PR #539, opened 2018-05-17
add owner role

opqdonut:


src/clj/rems/api/license.clj
```diff
@@ -15,5 +15,5 @@
       :summary "Get licenses"
       :return [License]
       (check-user)
-      (check-roles :approver) ;; TODO admin role needed?
+      (check-roles :owner)
```
Macroz: Maybe you could add the license tests in this PR or at least the auth one.

PR #540, opened 2018-05-17
rems.api.catalogue: add TODO

opqdonut:


src/clj/rems/api/catalogue.clj
```diff
@@ -11,6 +11,8 @@
 (def GetCatalogueResponse
   [CatalogueItem])
 
+;; TODO resid is misleading: it's the internal id, not the string id
```
Macroz: This is also confusing in the database. It requires some thought to change as the migrations probably need fixing too.

PR #546, opened 2018-05-18
Clean up

tjalkane:


src/clj/rems/context.clj
```diff
@@ -43,8 +43,6 @@
 
 (def ^:dynamic ^{:doc "User's preferred language."} *lang*)
 
-(def ^:dynamic ^{:doc "Contents of the cart."} *cart*)
-
 (def ^:dynamic ^{:doc "Flash session."} *flash*)
```
opqdonut: do we really still use flash?

src/clj/rems/events.clj
```diff
@@ -45,113 +42,12 @@
 (defn close-modal [app]
```
opqdonut: these modals can be removed also, just remove the call sites too

PR #563, opened 2018-05-29
fixes #562

### LINKED ISSUE #562: When logged in bogus api key should fail the request

opqdonut:


src/clj/rems/middleware.clj
```diff
@@ -35,17 +36,29 @@
...
+(defn get-api-key [request]
+  (get-in request [:headers "x-rems-api-key"]))
+
 (defn valid-api-key? [request]
-  (api-key/valid? (get-in request [:headers "x-rems-api-key"])))
+  (when-let [key (get-api-key request)]
+    (api-key/valid? key)))
+
+(defn wrap-csrf-token-in-headers
+  "For testing ease, publish csrf token in response headers."
```
tjalkane: if this is used only in testing maybe disable this in production mode?

PR #564, opened 2018-05-29
fix: show localization value instead of key in status change emails

### LINKED ISSUE #550: fix: status localization keyword shown in status change email instead of localized text

tjalkane:
closes #550

### DIFF
--- a/src/clj/rems/email.clj
+++ b/src/clj/rems/email.clj
@@ -38,7 +38,7 @@
                           (get-username recipient-attrs)
                           app-id
                           (format-items items)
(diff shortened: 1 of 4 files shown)


src/clj/rems/email.clj
```diff
@@ -8,6 +8,7 @@
                                get-username]]))
 
 (defn send-mail [to subject msg]
+  (clojure.pprint/pprint (str "subject: " subject "msg: " msg))
```
opqdonut: remove this

Issue #263, opened 2017-08-10
proper redirection for unauthenticated users following a link


opqdonut:
- reimplement for SPA
- [x] Store the entry url in session(?) before redirecting to login page.
- [x] Protected from redirect attacks.
- [x] check that it also works on Chrome

PR #568, opened 2018-06-05
Redirection

Macroz:
Implements proper redirection from #263.

This is difficult to test without browser-based tests. I will add a separate ticket for that.


src/cljs/rems/handlers.cljs
```diff
@@ -36,3 +37,18 @@
  :set-current-language
  (fn [db [_ language]]
    (assoc db :language language)))
+
+(reg-event-fx
```
opqdonut: qould it be clearer if this was just a function (in spa.cljs), after all it has all sorts of side effects like `dispatch!` and `removeItem`

src/cljs/rems/handlers.cljs
```diff
@@ -36,3 +37,18 @@
...
+     (let [roles (set (get-in db [:identity :roles]))]
+       (println "Selecting landing page based on roles" roles)
+       (.removeItem js/window.sessionStorage "rems-redirect-url")
+       (cond
+         (contains? roles :owner) (dispatch! "/#/administration")
+         (contains? roles :approver) (dispatch! "/#/actions")
+         (contains? roles :reviewer) (dispatch! "/#/actions")
+         :else (dispatch! "/#/catalogue"))
+       {})
+     {:dispatch [:landing-page-redirect!]})))
```
opqdonut: are you rerunning the same event here?

src/cljs/rems/spa.cljs
```diff
@@ -31,7 +32,8 @@
 
 (defn home-page []
   (if @(rf/subscribe [:user])
-    [:p "Logged in."]
+    (do (rf/dispatch [:landing-page-redirect!]) ; don't show the empty page
+          [:div])
```
opqdonut: couldn't this just `dispatch!` to `/landing_page`? or actually, could we get rid of landing pages altogether and have the logic in `/`?

src/clj/rems/landing_page.clj
```diff
@@ -2,20 +2,12 @@
...
-      redirect-to (assoc (redirect redirect-to)
-                         :session (dissoc (:session req) :redirect-to))
-      (has-roles? :owner) (redirect "/#/administration")
-      (has-roles? :approver) (redirect "/#/actions")
-      (has-roles? :reviewer) (redirect "/#/actions")
-      :else (redirect "/#/catalogue"))))
+  (redirect "/#/landing_page"))
 
 (defroutes landing-page-routes
   (GET "/landing_page" req (landing-page req)))
```
opqdonut: couldn't you just redirect directly to `/#/landing_page` from the various auth methods?

src/cljs/rems/handlers.cljs
Macroz: This event requires set-identity to have been processed first and since events may be processed in random order (esp. Firefox), so we'll dispatch the same event for the next "round" if there is no roles set yet.

src/cljs/rems/handlers.cljs
Macroz: I think we should dismantle the handlers.cljs altogether. Perhaps now is the time? Same for subs?

opqdonut: fine by me :)

src/cljs/rems/handlers.cljs
opqdonut: maybe a comment to this effect would be good

PR #586, opened 2018-06-19
Login options

tjalkane:
add alternative login options link to login component.

### DIFF
--- a/src/clj/rems/api/public.clj
+++ b/src/clj/rems/api/public.clj
@@ -17,6 +17,7 @@
 
 (def GetConfigResponse
   {:authentication s/Keyword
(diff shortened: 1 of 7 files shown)


src/clj/rems/api/public.clj
```diff
@@ -17,6 +17,7 @@
 
 (def GetConfigResponse
   {:authentication s/Keyword
+   (s/optional-key :login-options) s/Str
```
Macroz: Should this be named `:alternative-login-url` or something? I find that there is now login-options and alternative used as name. Plural also refers to there being multiple other options.

tjalkane: good point, will change it

PR #598, opened 2018-07-02
Api rename

Macroz:
Rename API to be more consistent 
- use plurals
- special /draft endpoint
- split /catalogue and /catalogue-items
- test renaming and refactoring


src/cljs/rems/administration.cljs
```diff
@@ -42,11 +42,6 @@
  (fn [db [_ forms]]
    (assoc db ::forms forms)))
 
-(rf/reg-event-db
```
tjalkane: good find

PR #599, opened 2018-07-04
Actions perf

### LINKED ISSUE #499: Handled applications are loaded only after 'Show more' has been clicked
Loading handled applications, although normally hidden, slows down the normal use case.
- [x] load handled applications when clicking show more
- [x] show spinner while loading

tjalkane:
closes #499

### DIFF
--- a/src/cljs/rems/actions.cljs
+++ b/src/cljs/rems/actions.cljs
@@ -50,8 +50,9 @@
 (rf/reg-event-db
  ::fetch-handled-actions-result
  (fn [db [_ result]]
-   (assoc db ::handled-actions result
-             ::loading-handled-actions? false)))
+   (-> db
+    (assoc ::handled-actions result)
+    (dissoc ::loading-handled-actions?))))
 
 (rf/reg-event-db
  ::loading-handled-actions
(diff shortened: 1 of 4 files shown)


src/cljs/rems/actions.cljs
```diff
@@ -15,26 +15,63 @@
...
  ::fetch-actions-result
  (fn [db [_ result]]
-   (assoc db ::actions result) ))
+   (assoc db ::actions result)))
+
+(rf/reg-event-db
+ ::fetch-handled-actions-result
+ (fn [db [_ result]]
+   (assoc db ::handled-actions result
+             ::loading-handled-actions? false)))
```
Macroz: Could `dissoc` here?

src/cljs/rems/collapsible.cljs
```diff
@@ -41,15 +44,16 @@
   `:id` unique id required
   `:class` optional class for wrapper div
   `:open?` should the collapsible be open? Default false
+  `:dispatch-on-open` triggers the reframe event dispatch given as an argument when load-more is clicked
```
Macroz: Perhaps a regular callback function here (that in this case dispatches a re-frame event) would be a better API (not tied fo re-frame)?

PR #600, opened 2018-07-06
feat: make default language configurable

### LINKED ISSUE #584: Make default language configurable

tjalkane:
closes #584

### DIFF
--- a/src/clj/rems/middleware.clj
+++ b/src/clj/rems/middleware.clj
@@ -22,8 +22,7 @@
             [taoensso.tempura :as tempura])
   (:import (javax.servlet ServletContext)))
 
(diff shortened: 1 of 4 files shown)


src/clj/rems/middleware.clj
```diff
@@ -22,7 +22,8 @@
             [taoensso.tempura :as tempura])
   (:import (javax.servlet ServletContext)))
 
-(def +default-language+ :en)
+(def +default-language+ (or (:default-language env)
```
Macroz: Can also write this `(:default-language env :en)`

tjalkane: Nice suggestion, didn't know that was a thing.

PR #602, opened 2018-07-06
Catalogue-item not link to metadata, instead have separate "more info" button

orfjackal:
for #480


src/cljs/rems/catalogue.cljs
```diff
@@ -33,24 +33,27 @@
...
-   :cart {:value (fn [i] [cart/add-to-cart-button i])
-          :sortable? false
-          :class "commands"}})
+  {:name     {:header     #(text :t.catalogue/header)
+              :value      (fn [item] [catalogue-item-title item lang])
+              :sort-value #(get-catalogue-item-title % lang)}
+   :commands {:values    (fn [item] [[catalogue-item-more-info item config]
+                                     [cart/add-to-cart-button item]])
+              :sortable? false
+              :class     "commands"}})
```
Macroz: You can probably remove `:class` as the key of the column goes into the class automatically?

PR #605, opened 2018-07-11
Scripting hooks

### LINKED ISSUE #591: configurable option to add scripts (such as google analytics)
- [x] possible to inject requires script tag to every page
- [x] possible to configure a custom callback fn for each supported event type
- [x] page transition events (router changes of SPA)
- [x] action events (i.e. apply, approve, ...)

Macroz:
Implement hooks for custom scripts and documentation for it.

- injecting script files is done through configuration at the server side through `:extra-scripts`
- refactors all ajax `GET` and `PUT` calls into `fetch` and `put!` utility functions that DRYs commonly used parameters
- `window.rems.hooks` is checked for potential callback hooks
- `window.rems.hooks.get` is for data loading requests via `fetch`
- `window.rems.hooks.put` is for command requests via `put!`
- `window.rems.hooks.navigate` is for SPA navigation events

Closes #591

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -22,8 +22,6 @@ function logIt(e) {
   console.log(e);
 }
 
-window.rems = {};
-window.rems.hooks = {};
 window.rems.hooks.get = logIt;
 window.rems.hooks.put = logIt;
 window.rems.hooks.navigate = logIt;
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -1,0 +1,28 @@
+# Hooks
+
+## Extra scripts
+
+You can configure REMS to include extra script files by defining configuration
+
+```clj
+ :extra-scripts {:root "/tmp" :files ["/scripts/foo.js"]}
+```
+
+There you can for example initialize your analytics scripts.
+
+## Event hooks
+
+If you want to hook code into additional events you can include a script file like the following. REMS will attempt to call hooks for `get`, `put` and `navigate` events, if they exist.
+
+```js
+function logIt(e) {
+  // e.g. call your analytics solution here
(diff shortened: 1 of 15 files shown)


docs/hooks.md
```diff
@@ -0,0 +1,28 @@
+# Hooks
+
+## Extra scripts
+
+You can configure REMS to include extra script files by defining configuration
+
+```clj
+ :extra-scripts {:root "/tmp" :files ["/scripts/foo.js"]}
```
orfjackal: The documentation should also say that the file needs to be located at /tmp/scripts/foo.js

docs/hooks.md
```diff
@@ -0,0 +1,28 @@
...
+If you want to hook code into additional events you can include a script file like the following. REMS will attempt to call hooks for `get`, `put` and `navigate` events, if they exist.
+
+```js
+function logIt(e) {
+  // e.g. call your analytics solution here
+  console.log(e);
+}
+
+window.rems = {};
+window.rems.hooks = {};
```
orfjackal: If the system would initialize those empty objects before the user script is executed, then the user would not need to do it.

src/cljs/rems/spa.cljs
```diff
@@ -255,6 +254,8 @@
     (events/listen
      HistoryEventType/NAVIGATE
      (fn [event]
+       (when (and js/window.rems js/window.rems.hooks js/window.rems.hooks.navigate)
+         (js/window.rems.hooks.navigate (.-token event)))
```
orfjackal: The conditionals here and with other hooks could be avoided by always having hooks; the default hooks would be just no-ops.

PR #607, opened 2018-07-13
Resource editor

tjalkane:
add a button for creating new resources to the administration page.

### DIFF
--- a/src/clj/rems/db/licenses.clj
+++ b/src/clj/rems/db/licenses.clj
@@ -30,11 +30,14 @@
        (format-licenses)
        (localize-licenses)))
 
-(defn get-all-licenses []
-  (->> (db/get-all-licenses)
-       (format-licenses)
-       (map #(dissoc % :start :end)) ;; HACK
-       (localize-licenses)))
+(defn get-all-licenses [filters]
+  (let [filters (or filters {})]
+    (->> (db/get-all-licenses)
+         (map db/assoc-active)
+         (filter #(db/contains-all-kv-pairs? % filters))
+         (format-licenses)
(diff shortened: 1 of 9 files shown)


src/cljs/rems/administration.cljs
```diff
@@ -236,9 +329,51 @@
               [[spinner/big]]
               [[:div
                 [:div.col.commands
+                 [to-create-resource-button]
                  [to-create-catalogue-item-button]]
                 [catalogue-list @catalogue @language]]])))))
 
+(defn create-resource-page []
```
opqdonut: would it be neater if this was in its own ns? perhaps rems.administration.resource? I think we have too many event and subscription definitions in this ns

src/clj/rems/db/licenses.clj
```diff
@@ -30,11 +30,14 @@
        (format-licenses)
        (localize-licenses)))
 
-(defn get-all-licenses []
-  (->> (db/get-all-licenses)
-       (format-licenses)
-       (map #(dissoc % :start :end)) ;; HACK
-       (localize-licenses)))
+(defn get-all-licenses [filters]
```
opqdonut: please document filters

src/clj/rems/db/licenses.clj
```diff
@@ -30,11 +30,14 @@
...
+  (let [filters (or filters {})]
+    (->> (db/get-all-licenses)
+         (map db/assoc-active)
+         (filter #(db/contains-all-kv-pairs? % filters))
+         (format-licenses)
+         (map #(dissoc % :start :end)) ;; HACK
```
opqdonut: a) do we still need this b) should assoc-active do this?

Issue #612, opened 2018-07-13
add support for date field on forms


tjalkane:

opqdonut:
- [ ] add a date field to test-data and demo-data

PR #631, opened 2018-08-08
Date fields

luontola:
Issue #612

The date input field is not supported by IE and desktop Safari, but the polyfill should handle it. https://caniuse.com/#feat=input-datetime


src/cljs/rems/spa.cljs
```diff
@@ -102,10 +102,16 @@
...
+(reg-event-fx
+  :set-current-language
+  (fn [{:keys [db]} [_ language]]
+    {:db (assoc db :language language)
+     :update-document-language (name language)}))
+
+(reg-fx
+  :update-document-language
+  (fn [language]
+    (set! (.. js/document -documentElement -lang) language)))
```
opqdonut: good stuff

Issue #582, opened 2018-06-19
creating new workflows in administration page


tjalkane:
As a "resource holder", I can add a new workflow to a new catalogue item.
- [x] api for creating workflows
- [x] workflow editor with possibility to add following information: approve round, review round, provide a name for the workflow

PR #633, opened 2018-08-20
Create workflow UI

luontola:
Issue #582


src/clj/rems/api/workflows.clj
```diff
@@ -47,6 +50,21 @@
       (assoc (format-workflow wf)
         :actors (db/get-workflow-actors {:wfid (:id wf)})))))
 
+(defn create-workflow [{:keys [prefix title rounds]}]
```
tjalkane: This should allow to create workflows that are automatically approved after applying for a resource

PR #662, opened 2018-09-04
Refactor admin pages

luontola:


src/cljs/rems/administration/resource.cljs
```diff
@@ -1,97 +1,133 @@
...
+
+(defn- build-request [form]
+  (let [request {:prefix (:prefix form)
+                 :resid (:resid form)
+                 :licenses (map :id (:licenses form))}]
+    (when (valid-request? request)
+      request)))
+
+(defn- create-resource [form]
+  (post! "/api/resources/create" {:params (build-request form)
```
opqdonut: what if build-request returns nil?

src/cljs/rems/spa.cljs
```diff
@@ -263,7 +263,10 @@
   (rf/dispatch [:set-active-page :administration]))
 
 (secretary/defroute "/create-catalogue-item" []
-  (rf/dispatch [:rems.administration.catalogue/reset-create-catalogue-item])
+  (rf/dispatch [:rems.administration.catalogue-item/start-fetch-workflows])
+  (rf/dispatch [:rems.administration.catalogue-item/start-fetch-resources])
+  (rf/dispatch [:rems.administration.catalogue-item/start-fetch-forms])
+  (rf/dispatch [:rems.administration.catalogue-item/reset-create-catalogue-item])
```
opqdonut: I'm not sure I like this pattern (even though I've used it myself). Is there some way we could keep the knowledge of which stuff needs to be started/resetted within the rems.administration.catalogue-item namespace?

How about having some sort of `:rems.administration.catalogue-item/enter-page` event that gets dispatched here, and which then causes these other events?

luontola: I was thinking about the same thing; an enter-page event. Shall I change all pages to use that style? I’ll make it in another PR because it affects so many files.

src/cljs/rems/administration/resource.cljs
luontola: The check for nil happens already when the save button is pressed. Though maybe it would be cleaner to send the request object instead of the form object to create-resource.

src/cljs/rems/administration/form.cljs
```diff
@@ -179,10 +179,11 @@
...
 
 (defn- save-form-button []
-  (let [form @(rf/subscribe [::form])]
+  (let [form @(rf/subscribe [::form])
+        request (build-request form)]
     [:button.btn.btn-primary
-     {:on-click #(rf/dispatch [::create-form form])
-      :disabled (not (build-request form))}
+     {:on-click #(rf/dispatch [::create-form request])
+      :disabled (not request)}
```
opqdonut: aka `(nil? request)`

Issue #610, opened 2018-07-13
creating new forms on administration page


ghost:
Form editor should ask for "title" and form items.

- [x] api for creating forms
- [ ] ~~editor for creating forms from existing form items~~
- [x] possibility to create a new form item with localizations in form editor

PR #666, opened 2018-09-11
Form editor, backend

luontola:
#610


src/clj/rems/api/forms.clj
```diff
@@ -29,6 +30,38 @@
...
+                               :langcode (name lang)
+                               :title (get title lang)
+                               :inputprompt (get input-prompt lang)}))))
+
+(defn- create-form [{:keys [prefix title items]}]
+  (let [form (db/create-form! {:prefix prefix
+                               :title title
+                               :user (get-user-id)})]
+    (doseq [[index item] (map-indexed vector items)]
+      (create-form-item form index item))
```
opqdonut: aka

```
(doall (map-indexed (partial create-form-item form) items))
```

or

```
(doall (map-indexed (fn [index item] create-form-item form index item) items))
```

or

```
(doall (map (partial create-form-item) (range) items))
```

but your version is fine too

Issue #609, opened 2018-07-13
vTHL2: Adding members to an application by applicant


ghost:
Note: application progress is not dependent on additional members' actions. Only entitlements are.
See also #653 and #870.

- [x] field for inviting members to application by email address and name
- ~if the member doesn't have a REMS user account, send an email invitation for joining~ (moved to #652)
- [x] when coming through the join link (with appropriate secret challenge code) allows your new user to be linked to the application

opqdonut:
split this into #652 and #653

opqdonut:
backend done

opqdonut:
postponed until after dynamic workflow

PR #679, opened 2018-09-25
add-member application event

opqdonut:
for #609


src/clj/rems/api/schema.clj
```diff
@@ -43,7 +43,8 @@
    :round s/Num
    :event s/Str
    :comment (s/maybe s/Str)
-   :time DateTime})
+   :time DateTime
+   :eventdata s/Any})
```
opqdonut: I'm not so sure about this...

luontola: `{s/Keyword s/Any}` perhaps?

opqdonut:
will rebase on top of master

src/clj/rems/api/schema.clj
opqdonut: I think an opaque s/Any is pretty clear. The contents of eventdata is up to the event implementations. I'm not sure whether we should even publish it via the api.

PR #681, opened 2018-09-25
redirect to new application page by resource ID

### LINKED ISSUE #634: Provide API to access items with their resource ID (eg URN)
I was sure I had reported this already here, but it has been burried in a mail. so here it is again:

>> "Resource IDs need to be part of Apply link"
>> 
>> LBR3 (the present lbr.csc.fi) does not use the resource IDs used in LBR2 (the
>> previous incarnation). This is a massive problem and needs to be fixed. The
>> resource IDs are URN PIDs for a reason, so that the resources can be identified
>> accross services. Take for example: http://urn.fi/urn:nbn:fi:lb-2016112317 If
>> you dont't have access rights you are encouraged to apply via REMS, the link
>> refers to the old version:
>> https://lbr.csc.fi/web/guest/catalogue?domain=LBR&resource=urn:nbn:fi:lb-201403268&target=application
>> 
>> It is very important that we can use a parameter like
>> "resource=urn:nbn:fi:lb-201403268" which contains the resource ID, here a URN.

luontola:
closes #634

### DIFF
--- a/src/clj/rems/api/forms.clj
+++ b/src/clj/rems/api/forms.clj
@@ -1,9 +1,7 @@
 (ns rems.api.forms
   (:require [compojure.api.sweet :refer :all]
             [rems.api.util :refer [check-roles check-user]]
-            [rems.db.core :as db]
             [rems.db.form :as form]
-            [rems.util :refer [get-user-id]]
             [ring.util.http-response :refer :all]
             [schema.core :as s])
   (:import [org.joda.time DateTime]))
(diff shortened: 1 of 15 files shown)


src/clj/rems/api/forms.clj
```diff
@@ -79,4 +53,4 @@
       :body [command CreateFormCommand]
       (check-user)
       (check-roles :owner)
-      (ok (create-form command)))))
+      (ok (form/create-form! command)))))
```
Koura: nice find

src/clj/rems/db/licenses.clj
```diff
@@ -6,24 +6,24 @@
...
                            (map #(update-in % [:langcode] keyword))
                            (group-by :licid))]
     (doall
-      (for [lic licenses]
-        (assoc lic :localizations
-                   (into {} (for [{:keys [langcode title textcontent]} (get localizations (:id lic))]
-                              [langcode {:title title :textcontent textcontent}])))))))
+     (for [lic licenses]
+       (assoc lic :localizations
+             (into {} (for [{:keys [langcode title textcontent]} (get localizations (:id lic))]
```
Koura: indent starting from the line with `into` seems wrong

test/clj/rems/test/redirects.clj
```diff
@@ -0,0 +1,52 @@
...
+          catid (dummy-catalogue-item resid)
+          response (-> (request :get "/apply-for?resource=urn:one-matching-resource")
+                       app)]
+      (is (= 302 (:status response)))
+      (is (= (str "http://localhost/#/application?items=" catid) (get-in response [:headers "Location"])))))
+
+  (testing "fails if no catalogue item is found"
+    (let [response (-> (request :get "/apply-for?resource=urn:no-such-resource")
+                       app)]
+      (is (= 404 (:status response)))
```
Koura: I would add a test to check that the redirecting still works if there are two matching catalogue items but the other one is disabled.

src/clj/rems/home.clj
```diff
@@ -4,11 +4,17 @@
...
+            [rems.db.catalogue :as catalogue]
             [rems.layout :as layout]
             [rems.util :refer [index-by]]
-            [ring.util.response :refer [content-type
-                                        redirect
-                                        response]]))
+            [ring.util.response :refer [content-type not-found redirect response]]))
+
+(defn- apply-for-resource [resource]
+  (let [items (catalogue/get-localized-catalogue-items {:resource resource})]
```
Koura: Disabled catalogue-items should be filtered out here.

PR #688, opened 2018-10-09
Attachment field

Koura:
Add api endpoints for saving/fetching attachments and expand form editor to include attachment field.


src/clj/rems/api/applications.clj
```diff
@@ -207,4 +223,14 @@
       ;; TODO: provide a nicer error message when user doesn't exist?
       (applications/add-member (:application-id request)
                                (:member request))
+      (ok {:success true}))
+
+    (POST "/add_attachment" []
+      :summary "Add an attachment file related to an application field"
+      :multipart-params [file :- upload/TempFileUpload]
```
opqdonut: is there a size limit?

PR #692, opened 2018-10-17
Document which roles are needed for api calls

### LINKED ISSUE #689: add roles needed to swagger descriptions

Koura:
closes #689

### DIFF
--- a/src/clj/rems/api/actions.clj
+++ b/src/clj/rems/api/actions.clj
@@ -1,7 +1,7 @@
 (ns rems.api.actions
   (:require [compojure.api.sweet :refer :all]
             [rems.api.schema :refer :all]
-            [rems.api.util :refer [check-user]]
+            [rems.api.util]
             [rems.db.applications :as applications]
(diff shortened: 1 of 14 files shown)


luontola:
Hand-written documentation is a bit fragile; there are no guarantees that it's in sync with the implementation. Would it be worth the effort to generate the documentation? For example add the allowed roles as metadata to the API operations, and use the same metadata for runtime checks and documentation.

luontola:
Currently it's not obvious from the code that what the allowed roles are. For example `/applications/judge`, `/applications/review_request` and `/applications/add_member` are all documented to require different roles, but the code only checks with `(check-user)` that the user is logged in. The actual role checks are somewhere deep in the code. Some of those deep checks might still be needed, but a route level permission check (based on same metadata as documentation) would make it more obvious and serve as a safety net.

opqdonut:
I wonder how easy it would be to extend compojure-api so that we could give a `:roles [:approver :reviewer]` next to the `:summary`, and it would generate docs and a check-roles call...

PR #694, opened 2018-10-19
Fix #454 add sample browser test

### LINKED ISSUE #454: First ui tests
- [ ] unit tests for e.g. cart re-frame functionality (see old `rems.test.cart`) (doo)
- [ ] browser tests for happy paths (selenium etc.)
- [ ] tests use semantic classes for e.g. buttons

hukka:


luontola:
CI fails with:

```
Testing rems.test.browser
Exception in thread "main" java.io.IOException: Cannot run program "chromedriver": error=2, No such file or directory, compiling:(/tmp/form-init7324751374938809052.clj:1:73)
```

.travis.yml
```diff
@@ -3,6 +3,8 @@ lein: 2.8.1
 dist: trusty
 addons:
   chrome: stable
+  packages:
```
opqdonut: should be:
```
addons:
  apt:
    packages:
      - chromium-chromedriver

**Comment by hukka** [comment]
```
2018-10-23 14:22:36,596 [qtp283705665-30] INFO  rems.middleware - > :get /js/app.js lang: :en user: nil roles: nil 
2018-10-23 14:22:36,653 [qtp283705665-30] INFO  rems.middleware - < :get /js/app.js 404  
```
For some reason locally the test doesn't even try to access that file, yet manages to go through the tests. selenium/chromedriver should work in "incognito" mode by default without any caches, but just in case I deleted all chromium caches. Still couldn't make it get the app.js locally. Have to investigate more.

opqdonut:
Just a guess: you have an `app.js` lying around locally but on Jenkins nobody has built it. You need to generate the js file with something like `lein cljsbuild once`. Another option is to move the headless test after the uberwar build.

You can test this theory by taking a fresh checkout of the repo and running the headless test.

hukka:
Superseded by #740 which is inside the main repo to test CircleCI too.

PR #715, opened 2018-10-30
Add example for creating a catalogue item with the API

### LINKED ISSUE #705: Document in an example how to to create a new catalogue through the API

Koura:
Closes #705

### DIFF
--- a/docs/usingtheapi.md
+++ b/docs/usingtheapi.md
@@ -43,7 +43,7 @@ Some API endpoints also require `X-REMS-User-Id` header that is the REMS user id
 
 See other methods in the [Swagger API documentation](https://rems2demo.csc.fi/swagger-ui).
 
-## Example: Applying & Approving
+## Example 1: Applying & Approving
 
 The following example contains instructions on how to send an application and to approve the requested access using the API. These instructions assume you are using curl, have the backend running locally and have populated your local database with test data. Please refer to the README.md on [github](https://github.com/CSCfi/rems) for the setup instructions.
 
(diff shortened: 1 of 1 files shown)


docs/usingtheapi.md
```diff
@@ -109,3 +109,75 @@ Now the application sent by Alice has been approved and an entitlement should ha
...
+```
+
+Store this id somewhere as we will needed later when we create the catalogue item.
+
+Next let's create a new resource. The resource API works in a similar way as the workflow API. Keep in mind though that REMS doesn't handle the actual distribution of the resource so the `resid` used here should point to the unique identifier of the resource that you wish to share. The API will return the internal resource id used by REMS as an output just like it did with the successful creation of a workflow. We will not attach any terms of use for our resource so we will not add any license ids to the `licenses` array.
+
+```sh
+curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' --header 'x-rems-api-key: 42' --header 'x-rems-user-id: owner' -d '{ \
+   "resid": "string", \
+   "organization": "string", \
```
luontola: Use a more realistic example value instead of "string".

docs/usingtheapi.md
```diff
@@ -109,3 +109,75 @@ Now the application sent by Alice has been approved and an entitlement should ha
...
+   ] \
+ }' 'http://localhost:3000/api/resources/create'
+```
+
+Then we only need to create an application form to finally be able to create a catalogue item. Let's create a form with only one mandatory text field with English localization. The rest should already be familiar to you from the two previous calls.
+
+```sh
+curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' --header 'x-rems-api-key: 42' --header 'x-rems-user-id: owner' -d '{ \
+   "organization": "string", \
+   "title": "string", \
```
luontola: Use a more realistic example value instead of "string".

luontola:
Is it worth the effort to describe how to create workflows and forms through the API? People can create them through the UI manually and likely new ones are created so seldom that it doesn't need to be automated.

PR #716, opened 2018-10-30
Fix localize-time function when dealing with objects instead of strings

Koura:
Fix bug where sometimes current time was shown instead of the actual time the event had occured.

### DIFF
--- a/src/cljs/rems/text.cljs
+++ b/src/cljs/rems/text.cljs
@@ -54,7 +54,10 @@
 
 (defn localize-time [time]
   (when time
-    (format/unparse-local time-format (time/to-default-time-zone (format/parse time)))))
+    (let [time (or (when (string? time)
+                     (format/parse time))
+                   time)]
+      (format/unparse-local time-format (time/to-default-time-zone time)))))
 
 (defn localize-item
   ([item]
--- a/src/cljs/rems/text.cljs
+++ b/src/cljs/rems/text.cljs
@@ -54,9 +54,7 @@
 
 (defn localize-time [time]
   (when time
-    (let [time (or (when (string? time)
-                     (format/parse time))
-                   time)]
+    (let [time (if (string? time) (format/parse time) time)]
       (format/unparse-local time-format (time/to-default-time-zone time)))))
 
 (defn localize-item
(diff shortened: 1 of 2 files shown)


src/cljs/rems/text.cljs
```diff
@@ -54,7 +54,10 @@
 
 (defn localize-time [time]
   (when time
-    (format/unparse-local time-format (time/to-default-time-zone (format/parse time)))))
+    (let [time (or (when (string? time)
+                     (format/parse time))
+                   time)]
```
opqdonut: err this is weird, why not just
```
(if (string? time) (format/parse time) time)
```
also, please add a test.

test/cljs/rems/test/text.cljs
```diff
@@ -0,0 +1,15 @@
...
+            [rems.text :refer [time-format localize-time]]))
+
+(def test-time #inst "1980-01-02T13:45:00.000Z")
+
+(defn expected-time  [time]
+  (format/unparse-local time-format (time/to-default-time-zone time)))
+
+(deftest localize-time-test
+  (is (= (expected-time test-time) (localize-time "1980-01-02T13:45:00.000Z")))
+  (is (= (expected-time (time/now)) (localize-time "")))
```
opqdonut: is this really what we want?

luontola: This is a fragile test; (time/now) may return a different time than (localize-time).

test/cljs/rems/test/text.cljs
```diff
@@ -0,0 +1,15 @@
...
+            [cljs-time.format :as format]
+            [rems.text :refer [time-format localize-time]]))
+
+(def test-time #inst "1980-01-02T13:45:00.000Z")
+
+(defn expected-time  [time]
+  (format/unparse-local time-format (time/to-default-time-zone time)))
+
+(deftest localize-time-test
+  (is (= (expected-time test-time) (localize-time "1980-01-02T13:45:00.000Z")))
```
opqdonut: hmm, why can't you just have the literal output string here?

luontola: It would be better to avoid the complexity of expected-time and instead assert the exact value returned from localize-time. Now it's not obvious that what the expected value should be.

Issue #683, opened 2018-09-27
short name for application


opqdonut:
- [x] new field type "short description"
- [x] show this short description in various application tables

PR #721, opened 2018-10-31
Expose the application description through APIs

luontola:
#683


src/clj/rems/db/applications.clj
```diff
@@ -206,9 +223,10 @@
                               :when (= (:id app) (:appid e))]
                           ;; :appid needed only for batching
                           (dissoc e :appid))]
-         (assoc (get-application-state app app-events)
-                :formid (:formid (first catalogue-items))
-                :catalogue-items catalogue-items))))))
+         (enrich-application-description
```
Koura: This is not good, because enrich-application-description makes way too many db calls.

hukka:
Looks good, but don't have time right this moment to look at the merge conflict. Will look into that later, if someone else won't.

PR #750, opened 2018-11-20
Refactor request review

Macroz:
NB: Do after PR #749 

- third party review request now without modal
- shows status
- error handler
- style changes


src/cljs/rems/application.cljs
```diff
@@ -695,25 +695,32 @@
...
+(defn- send-third-party-review-request [reviewers user application-id round comment description]
   (post! "/api/applications/review_request"
          {:params {:application-id application-id
                    :round round
                    :comment comment
                    :recipients (map :userid reviewers)}
           :handler (fn [resp]
+                     (rf/dispatch [::set-status {:status :saved
+                                                 :description description}])
                      (rf/dispatch [::send-third-party-review-request-success true])
```
luontola: It should be possible to remove `::send-third-party-review-request-success` and `::send-third-party-review-request-message` because ::set-status does the same job.

P.S. `::send-third-party-review-request-success false` in `reset-state` should be `::send-third-party-review-request-message false`, so it'll fix that bug as well.

luontola: Or actually the success modal only shows up momentarily, because `(rf/dispatch [::enter-application-page application-id])` resets the page state. There are multiple bugs at play in here:

One is that `reset-state` resets the wrong key, which results in the status message being shown when navigating to another application.

But once that is fixed, the message is no more shown because the handlers call `(rf/dispatch [::enter-application-page application-id])` which resets the page state, which also hides the modal.

If the `::enter-application-page` dispatch is removed, then the modal success message shows up ok, but the action form state is no more reset.

luontola: Maybe the status should be decoupled from the application state. A similar status display is anyways needed on many other pages, such as the admin pages, and everywhere else where an API call is made and an error message could be shown.

luontola: PoC:
```diff
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -36,7 +36,7 @@
          ::edit-application nil
          ::judge-comment ""
          ::review-comment ""
-         ::send-third-party-review-request-success false))
+         ::send-third-party-review-request-message false))
 
 (rf/reg-sub
  ::application
@@ -136,16 +136,16 @@
  ::set-status
  (fn [db [_ {:keys [status description validation error]}]]
    (-> db
-       (assoc-in [::edit-application :status] {:open? (not (nil? status))
-                                               :status status
-                                               :description description
-                                               :error error})
+       (assoc-in [:foo/status] {:open? (not (nil? status))
+                                :status status
+                                :description description
+                                :error error})
        (assoc-in [::edit-application :validation] validation))))
 
 (rf/reg-sub
  ::status
  (fn [db _]
-   (get-in db [::edit-application :status])))
+   (get-in db [:foo/status])))
 
 ;;; saving application
 
```

Macroz: I think it doesn't make sense to fix this here since the next PR will be about rewriting the messy event handling chains. There are many old bugs that especially break the attachments. The flash and other status messages should be fixed there as well.

luontola: Yes. Another PR is better.

Issue #699, opened 2018-10-25
Initial application v2 API


Macroz:
- [x] return workflow info in application API
- [x] POST /v2/applications/command
- [x] command: submit
- [x] command: approve
- [x] command: reject
- [x] command: close
- [x] command: request-decision
- [x] command: decision
- [x] command: request-comment
- [x] command: comment

luontola:
How should we document the commands?

opqdonut:
maybe `rems.workflow.dynamic` could export a schema for the commands

PR #753, opened 2018-11-23
fetching dynamic applications via application api

opqdonut:
with a couple of hacks

for #699

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -40,25 +40,24 @@
    :value (s/maybe s/Str)})
 
 (s/defschema Event
-  s/Any
-  ;; HACK for dynamic applications
-  #_{:userid (s/maybe s/Str)
-     :round s/Num
-     :event s/Str
-     :comment (s/maybe s/Str)
-     :time DateTime
-     :eventdata s/Any})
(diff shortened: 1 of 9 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -40,17 +40,19 @@
...
 
 (s/defschema Event
-  {:userid (s/maybe s/Str)
-   :round s/Num
-   :event s/Str
-   :comment (s/maybe s/Str)
-   :time DateTime
-   :eventdata s/Any})
+  s/Any
+  ;; HACK for dynamic applications
```
Macroz: Maybe these new kinds of events could be returned under a new key?

src/clj/rems/api/schema.clj
```diff
@@ -40,17 +40,19 @@
...
+     :event s/Str
+     :comment (s/maybe s/Str)
+     :time DateTime
+     :eventdata s/Any})
 
 (s/defschema Application
   {:id (s/maybe s/Num) ;; does not exist for unsaved draft
    :formid s/Num
-   :state s/Str
+   :state s/Any ;; HACK for dynamic applications
```
luontola: Could use s/conditional or s/cond-pre

PR #769, opened 2018-11-29
test for creating & submitting a new dynamic application

opqdonut:


test/clj/rems/test/api/applications.clj
```diff
@@ -869,3 +869,29 @@
                   "event/decided"
                   "event/approved"]
                  (map :event (get-in data [:application :dynamic-events])))))))))
+
+(deftest dynamic-application-create-test
```
Macroz: Btw. should we change to the `test-` naming convention sometime?

test/clj/rems/test/api/applications.clj
```diff
@@ -911,8 +911,8 @@
                                          :application-id application-id
                                          :items {}})
                              app)]
-          (is (not= 200 (:status try-submit)))
-          (is (= "unauthorized" (read-body try-submit)))))
+          (is (= 400 (:status try-submit)))
```
Macroz: Hmm should this be 200 OK but some `:success false` and errors by the logic of our recent discussions?

test/clj/rems/test/api/applications.clj
opqdonut: I think @luontola changed all of these to `-test` a while ago. I prefer `test-` but followed the convention in this file.

test/clj/rems/test/api/applications.clj
opqdonut: I think this is different, you're actually violating an API invariant (old-style request for new-style application)

test/clj/rems/test/api/applications.clj
Macroz: So maybe a separate PR sometime later.

PR #773, opened 2018-11-30
circleci: store uberwar artifact

opqdonut:


opqdonut:
See https://circleci.com/gh/CSCfi/rems/794#artifacts/containers/0

PR #778, opened 2018-12-04
Show dynamic

Macroz:
Show dynamic applications.

### DIFF
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -22,6 +22,11 @@
             [rems.workflow.dynamic :as dynamic])
   (:import [java.io ByteArrayOutputStream FileInputStream]))
 
+;; TODO could use schemas for these coercions
+(defn- fix-workflow-from-db [wf]
+  (update (cheshire/parse-string wf keyword)
+          :type keyword))
+
 (defn draft?
   "Is the given `application-id` for an unsaved draft application?"
   [application-id]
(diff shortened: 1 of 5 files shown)


src/clj/rems/db/applications.clj
```diff
@@ -81,18 +86,27 @@
   ([app userid round]
    (reviewed? (update app :events (fn [events] (filter #(= round (:round %)) events))) userid)))
 
-(defn- can-act-as?
+(defn- is-handler? [application-id user-id]
+  (let [workflow (fix-workflow-from-db (:workflow (first (db/get-applications {:id application-id}))))]
+    (prn workflow)
```
opqdonut: forgotten prn

src/clj/rems/db/applications.clj
```diff
@@ -413,6 +440,9 @@
    (let [form (db/get-form-for-application {:application application-id})
          _ (assert form)
          application (get-application-state application-id)
+         application (if (= :workflow/dynamic (get-in application [:workflow :type]))
+                       (dynamic/assoc-possible-commands (getx-user-id) application) ; TODO move even higher?
```
opqdonut: it's a good place for this since this is where we invoke `can-review?` etc.

src/clj/rems/db/applications.clj
```diff
@@ -413,6 +440,9 @@
    (let [form (db/get-form-for-application {:application application-id})
          _ (assert form)
          application (get-application-state application-id)
+         application (if (= :workflow/dynamic (get-in application [:workflow :type]))
```
opqdonut: how about a `dynamic-application?` function?

test/clj/rems/test/db/applications.clj
```diff
@@ -1,6 +1,34 @@
...
+
+(deftest can-act-as?-test
+  (binding [context/*user* {"eppn" "developer"}]
+    (is (can-act-as? (first (db/get-applications {:id 12}))
+                     "approver"))
+    (is (not (can-act-as? (first (db/get-applications {:id 12}))
+                          "reviewer"))))
+  (binding [context/*user* {"eppn" "alice"}]
+    (is (not (can-act-as? (first (db/get-applications {:id 12}))
+                          "approver")))))
```
opqdonut: the other integration tests for applications are in rems.test.db

src/clj/rems/db/applications.clj
```diff
@@ -22,6 +22,11 @@
             [rems.workflow.dynamic :as dynamic])
   (:import [java.io ByteArrayOutputStream FileInputStream]))
 
+;; TODO could use schemas for these coercions
+(defn- fix-workflow-from-db [wf]
```
opqdonut: I'd maybe keep this and the is-handler? function on the bottom of the file with the rest of the dynamic stuff, and use a `(declare is-handler?)`

or maybe it's time to split the dynamic application db stuff to its own ns?

test/clj/rems/test/db/applications.clj
Macroz: Should they rather be here?

src/clj/rems/db/applications.clj
Macroz: Yes, but not going to split it in this PR.

test/clj/rems/test/db/applications.clj
opqdonut: that's fine, but we also need a place for unit tests of pure logic. maybe the implementation file?

... not that the tests that were already in this ns are that useful

Macroz: Implementation file is my preference.

src/clj/rems/db/applications.clj
Macroz: The reasoning would be that it uses the `context/*user*` here implicitly. At least `user-id` could be passed here explicitly.

PR #787, opened 2018-12-12
777 missing component ids

### LINKED ISSUE #777: Guide shows warning about same id being used in many places
- [ ] make sure guide page doesn't show error

Kauko:
Fixes #777 

Also fixed a minor bug where an approved form didn't show the licenses as checked.

Use :default-value instead of :value, to get around warning about controlled vs uncontrolled components. This _seems_ to be the way to go, though to be honest I'm not a 100% sure.

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -664,7 +664,10 @@
                   [field (assoc (localize-item license)
                                 :validation (get-in validation-by-field-id [:license (:id license)])
                                 :readonly readonly?
-                                :approved (get licenses (:id license)))]))])]}]))
+                                :approved (let [licenses (if-not readonly?
+                                                           licenses
+                                                           (group-by :id (:licenses form)))]
+                                            (get licenses (:id license))))]))])]}]))
 
 (defn- info-field
   "A component that shows a readonly field with title and value.
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -401,6 +401,7 @@
   [msgs language]
   (into [:ul]
         (for [m msgs]
+          ^{:key (str "message_list_item_" m)}
           [:li (text-format (:key m) (get-in m [:title language]))])))
 
 (defn- pdf-button [id]
@@ -416,7 +417,9 @@
     (rf/dispatch [::set-field id (.. event -target -value)])))
 
 (defn- id-to-name [id]
-  (str "field" id))
+  (if-not id
+    (str "id_missing")
+    (str "field" id)))
 
 (defn- set-attachment
   [id description]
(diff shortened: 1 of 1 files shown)


src/cljs/rems/application.cljs
```diff
@@ -416,7 +417,9 @@
     (rf/dispatch [::set-field id (.. event -target -value)])))
 
 (defn- id-to-name [id]
-  (str "field" id))
+  (if-not id
+    (str "id_missing")
+    (str "field" id)))
```
luontola: ID should be a required attribute, so instead of returning a dummy value (that will likely conflict with others), it's safer to throw an exception.

src/cljs/rems/application.cljs
```diff
@@ -919,39 +929,50 @@
...
+               [review-action-button]])
             (when (= :third-party (:review-type app))
-              [[third-party-review-action-button]])
+              [^{:key "third-party-review_action_button"}
+               [third-party-review-action-button]])
             (when (and (:is-applicant? app) editable?)
-              [[save-button]
+              [^{:key "save_action_button"}
+               [save-button]
+               ^{:key "submit_action_button"}
```
luontola: All of these keys can be removed as unnecessary. There is already a fix in #785

src/cljs/rems/application.cljs
```diff
@@ -989,6 +1010,7 @@
        (text :t.form/alert-disabled-items)
        (into [:ul]
              (for [item items]
+               ^{:key (str "disabled_item_" item)}
```
luontola: This key is unnecessary, since it already uses `into`

src/cljs/rems/application.cljs
```diff
@@ -999,7 +1021,7 @@
       :always [:div.form-items.form-group
                (into [:ul]
                      (for [item catalogue-items]
-                       ^{:key (:id item)}
+                       ^{:key (id-to-name (:id item))}
```
luontola: Same here and everywhere else.

src/cljs/rems/application.cljs
```diff
@@ -656,10 +660,14 @@
...
                 (for [license form-licenses]
+                  ^{:key (id-to-name (:id license))}
                   [field (assoc (localize-item license)
                                 :validation (get-in validation-by-field-id [:license (:id license)])
                                 :readonly readonly?
-                                :approved (get licenses (:id license)))]))])]}]))
+                                :approved (let [licenses (if-not readonly?
+                                                           licenses
+                                                           (group-by :id (:licenses form)))]
+                                            (get licenses (:id license))))]))])]}]))
```
luontola: Why is this done?

PR #788, opened 2018-12-12
fix: don't remove validation results on modal close

Macroz:


luontola:
After correcting the validation errors and submitting again, the validation errors show up for a split second in the Send Application modal. The old validation errors should be cleared before opening that modal.

PR #789, opened 2018-12-12
Fix duplicate element IDs and remove action form duplication

luontola:
P.S. Handy plugin for finding duplicate IDs: https://chrome.google.com/webstore/detail/dup-id-scans-html-for-dup/nggpgolddgjmkjioagggmnmddbgedice

### DIFF
--- a/src/cljs/rems/actions/action.cljs
+++ b/src/cljs/rems/actions/action.cljs
@@ -30,6 +30,16 @@
                 :value comment
                 :on-change on-comment}]]))
 
+(defn action-comment [{:keys [id label comment on-comment]}]
+  (let [id (str "comment-" id)]
+    [:div.form-group
+     [:label {:for id} label]
+     [textarea {:id id
+                :name id
+                :placeholder (text :t.actions/comment-placeholder)
+                :value comment
+                :on-change on-comment}]]))
+
 (defn action-form-view [id title comment-title buttons content comment on-comment]
   [:div.collapse {:id (action-collapse-id id) :data-parent "#actions-forms"}
    [:h4.mt-5 title]
--- a/src/cljs/rems/actions/action.cljs
+++ b/src/cljs/rems/actions/action.cljs
@@ -36,13 +36,12 @@
    content
    (into [:div.col.commands.mr-3 [cancel-action-button id]] buttons)])
 
-(defn action-button [id content on-click]
+(defn action-button [{:keys [id text class on-click]}]
   [:button.btn.mr-3
    {:id (str id "-action-button")
-    ;; TODO: decouple from action names
-    :class (if (contains? #{"approve" "approve-reject"} id) "btn-primary" "btn-secondary")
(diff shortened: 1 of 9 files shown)


src/cljs/rems/application.cljs
```diff
@@ -758,24 +758,30 @@
                    :on-click #(rf/dispatch [::judge-application "review" (text :t.actions/review)])}])
 
 (defn- third-party-review-button []
-  [button-wrapper {:id "request-review"
+  [button-wrapper {:id "third-party-review"
```
Macroz: `do-third-party-review`?

src/cljs/rems/actions/action.cljs
```diff
@@ -20,10 +20,10 @@
     :data-target (str "#" (action-collapse-id id))}
    (text :t.actions/cancel)])
 
-(defn- action-judge-comment [id label-title comment on-comment]
+(defn action-judge-comment [{:keys [id label comment on-comment]}]
```
Macroz: Kind of  shame that the "old legacy" static workflow component goes into the new namespace.

src/cljs/rems/application.cljs
```diff
@@ -925,17 +928,17 @@
...
-            :add-member nil ; TODO implement
-            :return [[return-action-button]]
-            :request-decision [[request-decision-action-button]]
-            :decide [[decide-action-button]]
-            :request-comment [[request-comment-action-button]]
-            :comment [[comment-action-button]]
-            :approve [[approve-reject-action-button]]
-            :reject [[approve-reject-action-button]]
-            :close [[close-action-button]]}
+               {:submit [[save-button]
```
Macroz: What is this whitespace change?

src/cljs/rems/application.cljs
Macroz: I would think about what logical convention we want to have for naming buttons since that affects UI tests especially.

src/cljs/rems/application.cljs
luontola: It's how Cursive formats that code and there are no options for configuring it. There doesn't seem to be an established convention for formatting maps with namespaced keys - the syntax is not even mentioned at https://github.com/bbatsov/clojure-style-guide

src/cljs/rems/application.cljs
luontola: I've now removed the "do-" prefixes, since they're not anymore needed to avoid ID conflicts.

src/cljs/rems/actions/action.cljs
Macroz: I refer to the term judge here but I like the extracted comment component so perhaps we could combine these but you can do what you like in the next PR.

src/cljs/rems/application.cljs
Macroz: Looks quite strange option from cursive. I can reformat this later.

Issue #728, opened 2018-11-02
"option" field type


Macroz:
- [x] admin UI for adding option fields
- [x] allow selection of options as answer
- [x] remove feature flag in rems/administration/form.cljs

Longer option lists and customizable options are not required yet. Longer lists could be radio group or dropdown or maybe even autocomplete multiselect like the 3rd party reviewer?

PR #813, opened 2018-12-28
UI for configuring option fields

luontola:
Issue #728, admin UI only.


resources/migrations/20181231124500-option-form-item.up.sql
```diff
@@ -0,0 +1,15 @@
+-- :disable-transaction
+ALTER TYPE itemtype
+  ADD VALUE 'option';
+--;;
+CREATE TABLE application_form_item_options
```
Macroz: I wonder if it would actually make more sense to start moving these towards document orientation as well? I.e. store the heterogenous options in the item rows.

src/clj/rems/db/applications.clj
```diff
@@ -410,6 +410,26 @@
       (:filename (db/get-attachment query-params))
       (:value (db/get-field-value query-params)))))
 
+(defn- process-item-options [options]
+  (->> options
+       (map (fn [{:keys [key langcode label displayorder]}]
+              {:key key
+               :label {(keyword langcode) label}
+               :displayorder displayorder}))
```
Macroz: Isn't displayorder unnecessary at least outside SQL layer if you have a sequence always?

test/clj/rems/test/api/forms.clj
```diff
@@ -86,6 +86,53 @@
...
+                                         :label {:en "Yes"
+                                                 :fi "Kyllä"}}
+                                        {:key "no"
+                                         :label {:en "No"
+                                                 :fi "Ei"}}]}]}
+            response (-> (request :post "/api/forms/create")
+                         (authenticate api-key user-id)
+                         (json-body command)
+                         app)]
+        (assert-response-is-ok response)
```
Macroz: You can now move this row to the `(-> ... assert-response-is-ok)` since master contains the chainable version of the assert.

src/clj/rems/db/applications.clj
luontola: Yes. The displayorder is dropped by the select-keys operation a few rows below this.

resources/migrations/20181231124500-option-form-item.up.sql
luontola: That would be good. It was a pain to implement this relationally.

We'll need to discuss this refactoring more. We could add a JSON field to application_form, or as a smaller step add it to application_form_item.

src/clj/rems/db/applications.clj
luontola: Due to the group-by operation the displayorder is still needed in this function.

PR #814, opened 2018-12-31
Auth statuses

cscwooller:


src/clj/rems/auth/ForbiddenException.clj
```diff
@@ -0,0 +1,3 @@
+(ns ^{:doc "REMS custom exception thrown when user is not authenticated."}
```
Macroz: This docstring should say forbidden instead.

PR #817, opened 2019-01-03
Option field support in applications

### LINKED ISSUE #728: "option" field type

luontola:
Closes #728

### DIFF
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -117,6 +117,20 @@
                     :optional true
                     :type "attachment"}
 
+                   {:title {:en "Project team size"
+                            :fi "Projektitiimin koko"}
+                    :optional true
+                    :type "option"
+                    :options [{:key "1-5"
+                               :label {:en "1-5 persons"
+                                       :fi "1-5 henkilöä"}}
+                              {:key "6-20"
+                               :label {:en "6-20 persons"
+                                       :fi "6-20 henkilöä"}}
+                              {:key "20+"
+                               :label {:en "over 20 persons"
+                                       :fi "yli 20 henkilöä"}}]}
+
                    ;; fields which support maxlength
                    {:title {:en "Project acronym"
                             :fi "Projektin lyhenne"}
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -4,6 +4,7 @@
             [rems.context :as context]
             [rems.db.applications :as applications]
             [rems.db.core :as db]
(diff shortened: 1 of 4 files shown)


src/cljs/rems/application.cljs
```diff
@@ -564,13 +564,36 @@
...
 
+(defn- option-label [value language options]
+  (let [label (->> options
+                   (filter #(= value (:key %)))
+                   first
+                   :label)]
+    (get label language value)))
+
+(defn option-field [{:keys [id value options validation] :as opts}]
+  (let [language @(rf/subscribe [:language])]
```
Macroz: I think it would be nicer if you'd push this subscription one level higher so that `option-field` would be a pure functional component and the user would pass the language. It would make it a little better to show in the examples then too.

luontola:
Done

src/clj/rems/db/test_data.clj
```diff
@@ -87,43 +88,47 @@
...
-    (db/localize-form-item! {:item (:id start-date) :langcode "fi" :title "Projektin aloituspäivä" :inputprompt ""})
-    (db/localize-form-item! {:item (:id start-date) :langcode "en" :title "Start date of the project" :inputprompt ""})
-    (db/localize-form-item! {:item (:id expired) :langcode "en" :title "Expired form item" :inputprompt ""})
-    (db/localize-form-item! {:item (:id plan) :langcode "fi" :title "Projektisuunnitelma" :inputprompt ""})
-    (db/localize-form-item! {:item (:id plan) :langcode "en" :title "Project plan" :inputprompt ""})
-    (db/localize-form-item! {:item (:id maxlength-text) :langcode "fi" :title "Projektin lyhenne" :inputprompt ""})
-    (db/localize-form-item! {:item (:id maxlength-text) :langcode "en" :title "Project acronym" :inputprompt ""})
-    (db/localize-form-item! {:item (:id maxlength-texta) :langcode "fi" :title "Tutkimussuunnitelma" :inputprompt ""})
-    (db/localize-form-item! {:item (:id maxlength-texta) :langcode "en" :title "Research plan" :inputprompt ""})
+  (binding [context/*user* {"eppn" (users :owner)}]
```
Macroz: I think the user should be moved away from the context variable like in the application APIs also here. Maybe we can just add a task or TODO for it.

PR #818, opened 2019-01-03
Multiselect field

### LINKED ISSUE #682: multiselect form item type
- [ ] description
- [ ] multiple checkable checkboxes

luontola:
Closes #682

Builds on PR #817

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -594,6 +594,47 @@
              [:option {:value key}
               (get label language key)]))]))
 
+(defn- encode-multiple-values [values]
+  (->> values
+       sort
+       (str/join " ")))
+
+(defn- decode-multiple-values [value]
+  (-> value
+      (str/split #"\s+")
+      set
+      (disj "")))
+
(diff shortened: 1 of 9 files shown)


src/cljs/rems/application.cljs
```diff
@@ -594,6 +594,47 @@
...
               (get label language key)]))]))
 
+(defn- encode-multiple-values [values]
+  (->> values
+       sort
+       (str/join " ")))
+
+(defn- decode-multiple-values [value]
+  (-> value
+      (str/split #"\s+")
```
Macroz: Isn't this logic coupled with the normalizer? Should it be in one place and tested together?

src/cljs/rems/application.cljs
```diff
@@ -594,6 +594,47 @@
...
+      set
+      (disj "")))
```
Macroz: Maybe you could name this empty string step as the default or unselected?

luontola: That's not really what the empty string is. It's the garbage left over by the split operation when there is leading or trailing space or the set is empty.

luontola:
Done

PR #819, opened 2019-01-03
Prevent duplicate resource IDs

### LINKED ISSUE #807: In create resource check id is unique
Resource id must be unique, check needed to prevent duplicates when creating a new resource

luontola:
Closes #807


src/clj/rems/api/catalogue.clj
```diff
@@ -15,6 +15,6 @@
 
     (GET "/" []
       :summary "Get the catalogue of items for the UI (does not include disabled)"
+      :roles #{:applicant}
       :return GetCatalogueResponse
-      (check-user)
```
cscwooller: Why is check-user removed?

src/clj/rems/db/resource.clj
```diff
@@ -1,18 +1,34 @@
...
+                                        :organization organization
+                                        :owneruserid (getx-user-id)
+                                        :modifieruserid (getx-user-id)}))]
+      (doseq [licid licenses]
+        (db/create-resource-license! {:resid id
+                                      :licid licid}))
+      {:success true
+       :id id})
+    (catch Exception e
+      (if (duplicate-resid? e)
```
cscwooller: This is ok now, but if we start to look at more error cases a separate error-handler were exceptions were passed would make more sense rather than checking all possible error situations where the error happened?

src/cljs/rems/administration/resource.cljs
```diff
@@ -148,18 +156,34 @@
    (text :t.administration/cancel)])
 
 (defn create-resource-page []
```
cscwooller: This function does quite a lot of things, could there be a name that would describe it better? Or splitting it to smaller functions?

test/clj/rems/test/api/forms.clj
```diff
@@ -164,8 +164,8 @@
                                      :title "the title"
                                      :items []})
                          app)]
-        (is (response-is-forbidden? response))
-        (is (= "<h1>Invalid anti-forgery token</h1>" (read-body response))))))
+        (is (response-is-unauthorized? response))
```
cscwooller: Shouldn't the error be forbidden without authentication? Unauthorized is when we know who it is but they don't have rights?

test/clj/rems/test/api/licenses.clj
```diff
@@ -101,8 +101,8 @@
                                      :localizations {:en {:title "t"
                                                           :textcontent "t"}}})
                          app)]
-        (is (response-is-forbidden? response))
-        (is (= "<h1>Invalid anti-forgery token</h1>" (read-body response))))))
+        (is (response-is-unauthorized? response))
```
cscwooller: Same here than above?

test/clj/rems/test/api/resources.clj
```diff
@@ -38,7 +38,26 @@
...
+                             (authenticate api-key user-id)
+                             (json-body {:resid resid
+                                         :organization "TEST-ORGANIZATION"
+                                         :licenses [licid]})
+                             app)
+                body (read-body response)]
+            (is (= 200 (:status response)))
+            (is (= false (:success body)))
+            (is (= [{:key "t.administration.errors/duplicate-resid" :resid resid}] (:errors body)))))
+        (testing "duplicate resource ID is allowed between organizations"
```
cscwooller: Good test.

test/clj/rems/test/api/users.clj
```diff
@@ -31,8 +31,8 @@
       (let [response (-> (request :post (str "/api/users/create"))
                          (json-body new-user)
                          app)]
-        (is (response-is-forbidden? response))
-        (is (= "<h1>Invalid anti-forgery token</h1>" (read-body response))))))
+        (is (response-is-unauthorized? response))
```
cscwooller: Should this also be forbidden?

test/cljs/rems/test/administration/form.cljs
```diff
@@ -8,7 +8,6 @@
 (use-fixtures :each isolate-re-frame-state)
 
 (defn reset-form []
-  (stub-re-frame-effect ::f/fetch-form-items)
```
cscwooller: Why this away?

src/clj/rems/api/catalogue.clj
Macroz: refactor: use the declarative :roles support

src/clj/rems/db/resource.clj
Macroz: I would worry about that then. It's not possible to see yet what it would look like?

src/cljs/rems/administration/resource.cljs
Macroz: Hmm, to me it looks like a pretty simple container component that subcribes to some state and renders itself using a few other components. I don't see how this can be simplified. Do you have some ideas?

test/clj/rems/test/api/forms.clj
Macroz: CSRF error is typical when the user session is timed out and we wish to redirect to login in that case.

test/clj/rems/test/api/licenses.clj
Macroz: Same as above.

test/clj/rems/test/api/users.clj
Macroz: Same as above.

test/cljs/rems/test/administration/form.cljs
Macroz: There is no actual event called `fetch-form-items` in the `form.cljs` and it only creates unnecessary warnings when running test code.

src/cljs/rems/administration/resource.cljs
cscwooller: Maybe it is then just my unfamiliariness of cljs ui code.

PR #823, opened 2019-01-07
Draft state

cscwooller:

### DIFF
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -117,7 +117,7 @@
    {:href (str "#/application/" (:id app))}
    (text :t.applications/view)])
 
-(defn draft-list [drafts language]
+(defn draft-list [drafts]
   (when (seq drafts)
     [:div.drafts
      [:h4 (text :t.catalogue/continue-existing-application)]
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -1,5 +1,6 @@
 (ns rems.catalogue
-  (:require [re-frame.core :as rf]
+  (:require [clojure.string :as str]
+            [re-frame.core :as rf]
             [rems.atoms :refer [external-link]]
             [rems.cart :as cart]
             [rems.db.catalogue :refer [disabled-catalogue-item?
@@ -9,7 +10,7 @@
             [rems.guide-functions]
             [rems.spinner :as spinner]
             [rems.table :as table]
-            [rems.text :refer [text]]
+            [rems.text :refer [localize-time text]]
             [rems.util :refer [fetch]])
   (:require-macros [rems.guide-macros :refer [component-info example]]))
 
@@ -17,7 +18,8 @@
  ::enter-page
  (fn [{:keys [db]} _]
    {:db (assoc db ::loading-catalogue? true)
-    ::fetch-catalogue nil}))
+    ::fetch-catalogue nil
+    ::fetch-drafts nil}))
 
 ;;;; table sorting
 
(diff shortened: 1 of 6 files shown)


src/cljs/rems/table.cljs
```diff
@@ -18,16 +18,21 @@
...
   ((or (get-in column-definitions [col :sort-value])
        (get-in column-definitions [col :value])) item))
 
 (defn- row [column-definitions columns item]
   (into [:tr.action]
         (for [col columns]
-          (into [:td {:class (column-class column-definitions col)
-                      :data-th (column-header column-definitions col)}]
-                (column-values column-definitions col item)))))
+            (into [:td {:class (column-class-customized item column-definitions col)
```
Macroz: Indentation is wrong here?

src/cljs/rems/table.cljs
```diff
@@ -18,16 +18,21 @@
 (defn column-class [column-definitions col]
   (get-in column-definitions [col :class] (name col)))
 
+(defn column-class-customized [item column-definitions col]
+  (if (= "draft" (:state item))
+    "text-warning"
```
Macroz: Warning style is risky to use here because an existing draft is generally not something to warn about but to advertise in a positive sense. Maybe introduce a separate `text-highlight` style?

src/cljs/rems/table.cljs
```diff
@@ -18,16 +18,21 @@
 (defn column-class [column-definitions col]
   (get-in column-definitions [col :class] (name col)))
 
+(defn column-class-customized [item column-definitions col]
```
Macroz: How about introducing to the table component, an optional `:row-class` callback function that you give the item to that can then give back a class to be applied to the `tr` element?

src/cljs/rems/table.cljs
```diff
@@ -18,16 +18,21 @@
...
+  (if (= "draft" (:state item))
```
opqdonut: This sort of logic shouldn't be part of a "generic table widget" (see ns docstring)

I think this should be done either by
a) using a `[:span {:class "highlight-draft"} ...]` as the value of a column
b) adding a :class fn to the column definition (just like we have the :value fn)

src/cljs/rems/table.cljs
cscwooller: 👍

src/cljs/rems/table.cljs
```diff
@@ -18,19 +18,15 @@
...
-    "text-highlight"
-    (column-class column-definitions col)))
-
 (defn column-sort-value [column-definitions col item]
   ((or (get-in column-definitions [col :sort-value])
        (get-in column-definitions [col :value])) item))
 
-(defn- row [column-definitions columns item]
-  (into [:tr.action]
+(defn- row [{:keys [row-class]} column-definitions columns item]
```
opqdonut: why not just a normal argument? would simplify this and the calling code

cscwooller: It is a capability to give row options. There might be more in the future. :D

src/cljs/rems/catalogue.cljs
```diff
@@ -89,8 +109,30 @@
...
 
+(defn- format-catalogue-items [app]
+  (str/join ", " (map :title (:catalogue-items app))))
+
+(defn- view-button [app]
+  [:a.btn.btn-primary
+   {:href (str "#/application/" (:id app))}
+   (text :t.applications/view)])
+
+(defn draft-list [drafts language]
```
opqdonut: language not used

src/cljs/rems/table.cljs
opqdonut: hmm ok. please at least document these parameter changes in the `component` docstring

src/cljs/rems/table.cljs
```diff
@@ -117,7 +121,7 @@
...
                        :aria-hidden true}])])])))]
     (into [:tbody]
-          (map (fn [item] ^{:key (id-function item)} [row column-definitions visible-columns item])
-               (->> items
-                    (apply-filtering column-definitions filters)
-                    (apply-sorting column-definitions sort-column sort-order))))]])
+          (map (fn [item] ^{:key (id-function item)} [row (select-keys opts [:row-class]) column-definitions visible-columns item])
+               (cond->> items
+                        filters (apply-filtering column-definitions filters)
+                        sorting (apply-sorting column-definitions sort-column sort-order))))]])
```
opqdonut: please document that filters and sorting can be nil

src/cljs/rems/table.cljs
cscwooller: Ended up removing the row-class, as it was not used.

PR #831, opened 2019-01-14
Thl theme

cscwooller:


theme-thl/thl.edn
```diff
@@ -0,0 +1,22 @@
...
+ :warning-color "#6f572f"
+ :logo-bgcolor "#ffffff"
+ :logo-name "thl-logo-fi.png"
+ :logo-name-sm "thl-logo-fi.png"
+ :logo-content-origin "content-box"
+ :phase-color "#eee"
+ :phase-color-active "#cad2e6"
+ :phase-color-completed "#519b2f"
+ :table-bgcolor "#d1edc4"
+ :table-stripe-color "#d1edc4"
```
luontola: :table-bgcolor and :table-stripe-color are the same, so there are no stripes. Is this on purpose?

cscwooller: Yes. I think @Macroz said we should get rid of the double colored stripes, this was the easiest way not touching Language bank.

Macroz: I think the stripes are not how modern sites do it. We should instead add a feature where on hover we show the row in highlight. It's much easier for the eyes with similar result.

Issue #838, opened 2019-01-15
Review or approve request -> when reviewer or approver logs in, they should see that there is a request for them


cscwooller:

Macroz:
Unable to reproduce.

PR #850, opened 2019-01-16
Test data for reviews and fixes

Macroz:
Trying without success to reproduce #838 but ending up creating test data and fixing random things.


src/clj/rems/db/test_data.clj
```diff
@@ -409,10 +437,11 @@
...
-    (create-review-application! with-review (:with-review workflows) +demo-users+)
+    (create-review-applications! with-review (:with-review workflows) +demo-users+)
     (create-application-with-expired-resource-license! (:simple workflows) form +demo-users+)
     (create-application-before-new-resource-license! (:simple workflows) form +demo-users+)
     (create-expired-license!)
     (let [dynamic (create-catalogue-item! res1 (:dynamic workflows) form
                                           {"en" "Dynamic workflow" "fi" "Dynaaminen työvuo"})]
-      (create-dynamic-application! dynamic (:dynamic workflows) (+demo-users+ :applicant1)))))
+      (create-dynamic-application! dynamic (:dynamic workflows) (+demo-users+ :applicant1))
+      (create-dynamic-request-applications! dynamic (:dynamic workflows) +demo-users+))))
```
opqdonut: hmm why dynamic-request-application vs dynamic-application

also should these functions just be combined?

Macroz: They sure can be. The other tries to create just different kinds of requests but I think I'll combine them.

Issue #727, opened 2018-11-02
show diff for resubmitted application


Macroz:
- [x] diff support for fields
- [x] full application history in events: form values, submit, resubmit
- [ ] diff form values between the latest submit and the preceding submit

Macroz:
I think the diff for the dynamic workflows can be done now. Let's leave the old workflows as is.

PR #851, opened 2019-01-17
Save application form values as an event in dynamic workflows

luontola:
#727

### DIFF
--- a/src/clj/rems/workflow/dynamic.clj
+++ b/src/clj/rems/workflow/dynamic.clj
@@ -60,15 +60,17 @@
 
 (defmethod apply-event [:event/draft-saved :workflow/dynamic]
   [application _workflow event]
-  ;; TODO
-  application)
+  (assoc application :draft-application {:items (:items event)
+                                         :licenses (:licenses event)}))
 
 (defmethod apply-event [:event/submitted :workflow/dynamic]
   [application _workflow event]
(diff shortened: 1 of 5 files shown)


src/clj/rems/workflow/dynamic.clj
```diff
@@ -58,12 +58,19 @@
...
+                                         :licenses (:licenses event)}))
+
 (defmethod apply-event [:event/submitted :workflow/dynamic]
   [application _workflow event]
   (assoc application
          :state ::submitted
          :commenters #{}
-         :members [(:actor event)]))
+         :members [(:actor event)]
+         :previous-application (:submitted-application application)
```
opqdonut: should this be previous-submitted-application?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -58,12 +58,19 @@
 (deftest test-all-event-types-handled
   (is (= EventTypes (set (get-event-types)))))
 
+(defmethod apply-event [:event/draft-saved :workflow/dynamic]
+  [application _workflow event]
+  (assoc application :draft-application {:items (:items event)
```
opqdonut: application is so overloaded, how about just `:form` or `:form-contents`? do you need to mention the state in the key?

PR #860, opened 2019-01-21
don't use thl theme in config-defaults.edn

opqdonut:
- theme-thl is not bundled in uberjar
- dev-config.edn refers to theme-thl anyway


opqdonut:
(fixed by amending)

PR #875, opened 2019-01-22
Fix last modified time

### LINKED ISSUE #825: vTHL2: when modifying draft application, modified time is not updated
Should add to event table an event of draft modification. At the moment draft modified timestamp always stays at the time of creation.

luontola:
Closes #825


src/clj/rems/db/applications.clj
```diff
@@ -979,12 +979,14 @@
...
-                                 :state ::dynamic/draft
-                                 :dynamic-events events
-                                 :workflow (fix-workflow-from-db (:workflow application)))]
-    (assert (is-dynamic-application? fixed-application))
-    (dynamic/apply-events fixed-application events)))
+        application (assoc application
+                           :state ::dynamic/draft
+                           :dynamic-events events
+                           :workflow (fix-workflow-from-db (:workflow application))
+                           :last-modified (or (:time (last events))
```
Macroz: How about taking the time of the last event with `:time`?

src/cljs/rems/application.cljs
```diff
@@ -758,7 +758,7 @@
               [:th (text :t.form/comment)]
               [:th (text :t.form/date)]]]
             (into [:tbody]
-                  (for [e (sort-by :time > events)]
+                  (for [e (reverse events)]
```
Macroz: Shouldn't time be sent accurately and formatted in UI inaccurately?

src/clj/rems/db/applications.clj
```diff
@@ -667,6 +667,10 @@
   []
   (keys (methods apply-event)))
 
+(defmethod apply-event "save"
+  [application _event]
+  application)
```
cscwooller: Why is the event omitted? Bit confusing imho when the function name is apply-event.

src/cljs/rems/application.cljs
luontola: It is. The rounding to millisecond accuracy happens in rems.application/application-header and is used to avoid showing to the user duplicate notifications about comment requests, for which there are one event per receiver but the applicant doesn't see who the receiver is, so it will show up confusingly as duplicates in the event list.

And anyways sorting by timestamp is fundamentally incorrect. It's normal to have multiple events produced at the same millisecond, especially when a single command produces multiple events. The database currently sorts the events by ID and those IDs are produced using a `serial` column (i.e. `nextval` from a sequence), so there can be gaps in the ID values and they might not be committed in monotonically increasing order, but for our purposes it's probably good enough. We should however lock the application being updated, because currently if the same application is updated concurrently, it could produce an inconsistent sequence of events.

src/clj/rems/db/applications.clj
luontola: It *is* taking the time of the last event with `:time`. What do you mean?

Macroz: It takes the last event, and then its time. Instead of taking the time of the last event with time.

luontola: I don't understand. How would you write the code?

Macroz: `:last-modified (or (last (keep :time events))`

luontola: Ok, so if the newest event doesn't have `:time`, then it will skip that and use the next newest event which has `:time`.

luontola: All events should have a time, but we don't yet have validation for that, so we cannot be sure about that.

Macroz: That is why I suggest that you could take the time of the last event with time...

Issue #869, opened 2019-01-22
vTHL2: Dynamic role determination


Macroz:
Show catalogue, applications and actions based on roles in existing applications and not based on a database table. I.e. if someone is asked for review, then that makes the actions page show for him/her because effectively he/she is made a reviewer then.

- [x] everyone can be an applicant and sees the catalogue page
- [x] everyone who has made an application sees the applications page
- [x] everyone who has a request (to review, comment etc.) is a reviewer and sees the actions page
- [ ] ~~decide and possibly refactor if it makes sense to move approvers workflow based like handlers are~~
- [ ] ~~decide and possibly refactor if it makes sense to move reviewers workflow based like handlers are (third party reviews maybe are enough for everyone so reviewer role as it is in a rounds workflow may not be really needed)~~
- handler role in a workflow must be given explicitly
- admin role must be given explicitly
- domain owner role must be given explicitly (there is no domain owner role but it is a subset of admin)
- domain reporter role must be given explicitly #833

Issue #863, opened 2019-01-22
vTHL3: Performance testing


Macroz:
Check performance of static and dynamic applications in the new version.
- [x] check performance of especially actions page with some hundreds of applications closed and a few thousand archived.
- [x] check that the catalogue page can handle a thousand catalogue items
- [x] populate enough rows in test data so regular dev use hits the potential limits too
- [x] investigate how much data there is in production, how much new data is expected per year, and whether it all can be held in memory at once
- [x] consider whether `rems.api.applications-v2/all-applications-view` should be cached and built incrementally
  - the spike code which was removed in https://github.com/CSCfi/rems/pull/878/commits/2b85f467b1589b669cc3da9acec9f851288fc182 can serve as a base
  - even just memoizing the injections should improve the performance much
- [x] consider whether `rems.db.dynamic-roles/get-roles` should be cached and built incrementally

hukka:
Are we interested in basic web KPI like first paint and time to interactive, i.e. running lighthouse? Do we have any targets, other than "it is not unusable"?

In any case we would need some way to track performance, so that we can see if we would get better or worse.

Macroz:
History: the production installations have a lot more data than test dataset so there have been too slow page loads.

I think a suitable sweet spot could be a browser test to see that the initial page load and render is "instant" (< 100 ms) and the long content (lists) that is loading (i.e. with a spinner) should have been rendered soon after (< 1 second). Opening the archived / handled applications can take a bit longer. 

Sorting columns is also surprisingly slow but there is a separate issue for that.

Perhaps this could be a test fail or if it creates a lot of false positives because of variance then it'd be interesting to plot the times somehow so we see that the trend is not bad.

luontola:
Initial tests on a development machine with a production database dump with about 1000 applications gave the following results:

Every request (even `/api/config`) takes at least 1.2s because `rems.db.dynamic-roles/get-roles` loads all events on every request without caching.

Loading the actions page takes 30s. This is probably due to the lack of batching for dynamic applications.

luontola:
Now that the v2 APIs are in use, the actions page loads slightly faster, in about 10 seconds. Catalogue and applications pages also take as long to load, since they all are based on the same code for listing all applications.

mmatthiesencsc:
[...]

Issue #859, opened 2019-01-21
vTHL2: Event visibility for dynamic applications


opqdonut:
Should use the same logic as old-style applications, i.e.:

- show everything to handlers & commenters etc.
- don't show these to applicant:
  - comment-requested, commented (see also #845)
  - decision-requested, decided

hukka:
#843 might be easy to implement at the same time

Issue #853, opened 2019-01-17
Naming things


luontola:
Go through the core concepts in REMS and figure out how to name them. For example the meaning of "application" is overloaded when the get application API returns an application which has an application field.

opqdonut:
Some suggestions:

- form template: the set of fields that a user needs to fill to apply for a given resource
- form: the field values that the user filled
- application: the entirety of an application: applicant+form+resources+events+state

luontola:
Some more suggestions:

- application: the whole lifecycle of applying for a resource
- form: a filled in form template, possibly multiple draft forms and submitted forms per application
- form template: a set of empty fields
- field: previously called "form item"

Also the name "catalogue item" is clunky: the catalogue UI calls them resources, so maybe we should them the same way in code and rename resource to something else.

Macroz:
UI terminology is somewhat dependent on the user terminology i.e. LBR wants kielivara. I think it's possible to remove the resource from catalogue and use item there instead but still you are picking the resources you want to apply for so resource is just as valid there.

Macroz:
I don't like renaming form to form template. Form, resource and workflow are core concepts. Lets rather have (form) answers or something instead for the application.

Macroz:
`resid` in the `resource` table should be named something else because `resid` is used to refer to the id of the resource which is our internal id and not the external id. Maybe `extid`?

luontola:
There is a need for a common term to use about the handlers, commenters and deciders. After removing the round-based workflow, the word "reviewer" should be available for use. Got any better suggestions?

The word "actions" is currently overloaded. There is the actions page, but also the buttons in the application page are called actions. One suggestion is to rename the actions page to review page, where the reviewers will go through the review process of an application.

luontola:
The legacy data model (inherited from REMS1) uses the word "item" in lots of places, typically when one thing contains many things, instead of inventing a descriptive term for it. There are form items (i.e. fields), catalogue items, application items (no idea what it is) and at places the application itself is also called an item. Item is a similar empty word to "object", "data", "manager" etc. and should be replaced with a term in the users' domain. 

The hardest of these to name seems to be catalogue item. A mail-order catalogue contains mail-orderable products, a photo catalogue contains photos... an item catalogue contains catalogue items? 😕

[...]

PR #878, opened 2019-01-22
v2 application API spike

luontola:
Explores the route from legacy workflows to fully event-based workflows. Relates to multiple issues: #779 #852 #853 #859 #863 #869 #843 #202


src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,813 @@
...
+          (send-off projection-state update-projection))
+        (assoc state
+               :last-processed-event-id until-id
+               :applications (apply-events (:applications state) events))))))
+
+(defstate projection-updater
+  :start (future (while (not (Thread/interrupted))
+                   (send-off projection-state update-projection)
+                   (Thread/sleep (-> 60 time/seconds time/in-millis))))
+  :stop (future-cancel projection-updater))
```
opqdonut: I'd like a short comment about this architecture. I see agents + futures + mount. How does it all fit together and what is the goal?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,813 @@
...
+
+(defn get-user-applications-v2 [user-id]
+  (->> (vals (:applications @projection-state))
+       ;; TODO: not all events in test data have the created event, so they must be filtered out
```
opqdonut: please add them

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,813 @@
...
+            [rems.db.core :as db]
+            [rems.db.form :as form]
+            [rems.db.licenses :as licenses]
+            [rems.db.users :as users]
+            [rems.workflow.dynamic :as dynamic])
+  (:import (org.joda.time DateTime)))
+
+;;; user permissions
+
+(defn update-grants [application grants]
```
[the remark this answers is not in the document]
Macroz: There is also now multiple terminology as discussed before.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,813 @@
...
+
+(defn- get-user [user-id]
+  (users/get-user-attributes user-id))
+
+(def externals {:get-form get-form
+                :get-catalogue-item get-catalogue-item
+                :get-license get-license
+                :get-user get-user})
+
+(defn apply-user-permissions [application user-id]
```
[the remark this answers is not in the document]
luontola: I renamed update-grants to set-permissions

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,813 @@
...
+            [rems.workflow.dynamic :as dynamic])
+  (:import (org.joda.time DateTime)))
+
+;;; user permissions
+
+(defn update-grants [application grants]
+  (reduce (fn [application [subject permissions]]
+            (let [category (cond
+                             (keyword? subject) :permissions/by-role
+                             (string? subject) :permissions/by-user)]
```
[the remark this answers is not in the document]
luontola: Do you have suggestions for what kind of function API would be nicer?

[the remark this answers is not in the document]
luontola: I figured out a way. https://github.com/CSCfi/rems/pull/878/commits/4560703ee63074baf329fa6e7a04f0d7b90fbc05

src/clj/rems/form.clj
```diff
@@ -87,13 +88,19 @@
...
+                                                        :catalogue-item-ids catalogue-items
+                                                        :time (:start application)
+                                                        :actor actor}))
+        (if-let [error (applications/dynamic-command! {:type :rems.workflow.dynamic/save-draft
+                                                       :actor actor
+                                                       :application-id application-id
+                                                       :time (time/now)
+                                                       :items items
+                                                       :licenses licenses})]
+          (throw (RuntimeException. (str "error in save-draft command: " error)))))
```
Macroz: Why throw `RuntimeException`? This should be `bad-request` made by the user?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -75,6 +76,17 @@
...
          :application/comment s/Str))
+(s/defschema CreatedEvent
+  (assoc EventBase
+         :event/type (s/eq :application.event/created)
+         :application/resources [{:catalogue-item/id s/Int
+                                  :resource/ext-id s/Str}]
+         :application/licenses [{:license/id s/Int}]
+         :form/id s/Int
+         :workflow/id s/Int
+         ;; TODO: separate workflow specific data to a new event or make this an open schema?
```
Macroz: Why are this values at the top level? Could it be just `:workflow s/Any`  or alternatively something more specific under the `:workflow`?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -174,6 +187,11 @@
   (is (= (set (keys event-schemas))
          (set (get-event-types)))))
 
+(defmethod apply-event [:application.event/created :workflow/dynamic]
+  [application _workflow event]
+  ;; TODO
```
Macroz: What is to do here?

src/clj/rems/db/applications.clj
```diff
@@ -1037,6 +1044,37 @@
...
+                                              :applicantuserid actor
+                                              :start time
+                                              :wfid workflow-id}
+                                             catalogue-item-ids)]
+      (assert (= :workflow/dynamic (:type workflow))
+              (str "workflow type was " (:type workflow))) ; TODO: support other workflows
+      (add-dynamic-event! {:event/type :application.event/created
+                           :event/time time
+                           :event/actor actor
+                           :application/id application-id
```
Macroz: Shouldn't there be a composite value under `:application` key and in it the `:id`, `:resources`, `licenses` etc.? Why is everything top-level and not hierarchical?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+          (pr-str catalogue-item-id))
+  (first (applications/get-catalogue-items [catalogue-item-id])))
+
+(defn- get-license [license-id]
+  (licenses/get-license license-id))
+
+(defn- get-user [user-id]
+  (users/get-user-attributes user-id))
+
+(def ^:private externals {:get-form get-form
```
Macroz: This is called injections in `dynamic.clj`?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+                                content-key (assoc (localization-for :textcontent license)
+                                                   :default (:textcontent license))})))
+                      (sort-by :license/id))]
+    (assoc application :application/licenses (merge-lists-by :license/id
+                                                             licenses
+                                                             (:application/licenses application)))))
+
+(defn- assoc-externals [application {:keys [get-form get-catalogue-item get-license get-user]}]
+  (-> application
+      (assoc-form (get-form (:form/id application)))
```
Macroz: I think it would be better to use plain assoc here and have the helper function calculate the value. It's unintuitive to have `assoc-resources` take `catalogue-items` as parameter. Same applies to the other "assocs".

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+           (set (keys (methods application-view-specific)))))))
+
+
+(defn- log-event [events event]
+  ;; only include events which are about processing the application
+  (if (contains? #{:application.event/created
+                   :application.event/draft-saved}
+                 (:event/type event))
+    events
+    ;; choose only keys which will be shown in the UI's event log
```
Macroz: This creates a dependency to the UI where the UI should be the one to decide what to show?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+  [application event]
+  application)
+
+(deftest test-application-view-specific
+  (testing "supports all event types"
+    (is (= (set (keys dynamic/event-schemas))
+           (set (keys (methods application-view-specific)))))))
+
+
+(defn- log-event [events event]
```
Macroz: Why is this called `log-event`? There could be a better name or docstring?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+;;       (it will couple the write and read models, but it's probably okay
+;;        because they both are about a single application and are logically coupled)
+(defn- application-view
+  "Projection for the current state of a single application.
+  Pure function; must use `assoc-externals` to enrich the model with
+  data from other entities."
+  [application event]
+  (-> application
+      (application-view-specific event)
+      (application-view-generic event)))
```
Macroz: I think these functions could really be named better. How about `event-application-view` for the multimethod and just more specific fns for the generic parts?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+
+(defmethod application-view-specific :application.event/member-added
+  [application event]
+  application)
+
+(defmethod application-view-specific :application.event/submitted
+  [application event]
+  (-> application
+      (assoc :workflow/state ::dynamic/submitted)
+      (set-role-permissions {:applicant #{::dynamic/add-member}
```
Macroz: These are not up to date anymore.

`add-member` is only for the handler in specific states and `invite-member` is both for the applicant and handler in specific states.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+;;;; v2 API, pure application state based on application events
+
+(defmulti ^:private application-view-specific
+  "See `application-view`"
+  (fn [_application event] (:event/type event)))
+
+(defmethod application-view-specific :application.event/created
+  [application event]
+  (-> application
+      (assoc :application/id (:application/id event)
```
Macroz: Why are all the keys top-level? Why not have a hierarchy?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+(defn- reset-applications-state [_]
+  {:last-processed-event-id 0
+   :applications {}})
+
+(def ^:private applications-state
+  "Stateful projection tracking the current state of all applications.
+  Will be updated periodically by `applications-update-scheduler`,
+  but an update may also be triggered with `trigger-applications-update!`
+  e.g. after committing some new events."
+  (agent (reset-applications-state nil)
```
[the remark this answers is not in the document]
luontola: Because `update-applications!` does I/O. An agent ensures that only one update is running in parallel. Atom would allow multiple (slow) updates running in parallel and then it would retry all but one of them. Agents are not optimal either, because they won't discard extra `trigger-applications-update!` invocations. Something like https://github.com/luontola/cqrs-hotel/blob/master/src/main/java/fi/luontola/cqrshotel/framework/util/SingleThreadedTriggerableWorker.java would be best, but AFAIK there is no such built-in concurrency construct so one would need to be created.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+
+(comment
+  (send applications-state reset-applications-state)
+  (trigger-applications-update!)
+  @applications-state)
+
+(defstate applications-update-scheduler
+  :start (future (while (not (Thread/interrupted))
+                   (trigger-applications-update!)
+                   (Thread/sleep (-> 60 time/seconds time/in-millis))))
```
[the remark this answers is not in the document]
luontola: As said in `applications-state` docs, the idea is that `trigger-applications-update!` will be called after new events have been created.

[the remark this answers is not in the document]
luontola: There's also the option of checking for updates before each read operation, if extra consistency is needed.

src/clj/rems/db/test_data.clj
```diff
@@ -457,23 +457,43 @@
...
+               (applications/get-form-for user-id app-id))
+        dynamic-workflow? (= :workflow/dynamic (get-in form [:application :workflow :type]))
+        save-draft-command (atom {:type :rems.workflow.dynamic/save-draft
+                                  :actor user-id
+                                  :application-id app-id
+                                  :time (get-in form [:application :start])
+                                  :items {}
+                                  :licenses {}})]
+    (when dynamic-workflow?
+      (applications/add-application-created-event! {:application-id app-id
```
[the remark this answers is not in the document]
luontola: `create-new-draft-at-time` is only called from `create-draft!` and it doesn't get enough parameters for the created event. `create-draft!` is the lowest level operation that has enough context.

`create-draft!` and the rest of test data creation is also a big mess. After #852 it will be possible to delete much of it and simplify the code dramatically.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+  e.g. after committing some new events."
+  (agent (reset-applications-state nil)
+         :error-handler (fn [_agent exception]
+                          (log/error exception "Updating projection failed"))))
+
+(declare trigger-applications-update!)
+
+(defn- update-applications! [state]
+  (let [from-id (:last-processed-event-id state)
+        batch-size 1000
```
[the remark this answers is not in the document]
luontola: I'm assuming conman returns all rows eagerly instead of using cursors and returning them lazily, so batching is there to protect from returning 1M rows from database. Though probably all the data fits in memory and batching won't be needed. #863 should investigate that how much data there is to be expected production.

src/clj/rems/workflow/dynamic.clj
luontola: Populating the application from the event instead of relying the legacy application model. The second step in #852 will take care of that.

Trying to do that is outside the scope of this PR, because it would require modifying existing code and create too much risk of a merge conflict.

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
Macroz: We should expect a few thousand applications with a dozen events each so YAGNI?

src/clj/rems/db/applications.clj
luontola: Because https://github.com/CSCfi/rems/blob/master/docs/architecture/003-namespaced-keywords.md

src/clj/rems/db/test_data.clj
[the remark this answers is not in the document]
Macroz: What I object to is that this is logic in the test code. Shouldn't test code be able to call something to create a draft that adds the created event too?

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
luontola: Ok. Removing batching.

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
Macroz: Or only mark the view as dirty and have the future do the update as soon as required.

src/clj/rems/api/applications_v2.clj
luontola: Got name suggestions? `append-to-event-history`?

src/clj/rems/workflow/dynamic.clj
Macroz: TODO comment should say so then?

src/clj/rems/api/applications_v2.clj
luontola: Any incompatibility will be fixed by the second step in #852 by running the tests for the old implementation against this new implementation. That's outside the scope of this PR, because it will require changes to the existing code; in particular the dynamic workflow tests need to be refactored to be purely event-based.

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
Macroz: if the view is not always up to date, I think there'll be tricky to debug consistency related problems in our future. Wouldn't it be possible to always eagerly update the view so that every time you do something you can read a new up to date view that at least has what you updated? Eager updates could happen in the thread that causes them.

src/clj/rems/api/applications_v2.clj
luontola: Because it's not needed when we use namespaced keys: https://github.com/CSCfi/rems/blob/master/docs/architecture/003-namespaced-keywords.md

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
luontola: Yes, that's also an option. Trigger the update after committing new events and block until all projections have been updated. Currently the commit happens in `rems.api/transaction-middleware`, so the update would need to be done after that.

When the system starts, it also needs to do the update (starting from blank state). Should that also block so that the application starts only after the projections are up to date?

src/clj/rems/db/test_data.clj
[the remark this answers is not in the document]
luontola: Absolutely. All the following functions should be changed to use the high-level APIs instead of using the low-level APIs from `rems.db.core`.

rems.db.test-data/create-expired-form!
rems.db.test-data/create-expired-license!
rems.db.test-data/create-workflows!
rems.db.test-data/create-resource-license!
rems.db.test-data/create-catalogue-item!
rems.db.test-data/trim-value-if-longer-than-fields-maxlength
rems.db.test-data/create-draft!
rems.db.test-data/create-application-with-expired-resource-license!
rems.db.test-data/create-application-before-new-resource-license!
rems.db.test-data/create-test-data!
rems.db.test-data/create-demo-data!

But that's outside this PR. I added it as a bullet point to #849.

src/clj/rems/workflow/dynamic.clj
luontola: With [namespaced keys](https://github.com/CSCfi/rems/blob/master/docs/architecture/003-namespaced-keywords.md) they don't need be nested, but each workflow can add its own workflow-specific stuff (in this case only `:workflow.dynamic/handlers`) using its own namespace. The `:workflow.dynamic/handlers` key should probably be removed from this event schema and replaced with `s/Keyword s/Any`.

luontola: Or actually, removing `:workflow.dynamic/handlers` would prevent using coercion when deserializing the events. Making the key optional seems better.

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
Macroz: Yes at startup I think it would be best to block until the projection is up to date. Otherwise there can be surprising behavior. I guess most of our business logic happens with no need to use the updated projection so middleware is ok approach? But then keeping it synchronous makes no sense but it could happen eagerly.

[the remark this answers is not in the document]
Macroz: If we want to write business logic that uses the projection then we should keep the projection up to date always because otherwise there will just be unnecessarily hairy bugs IMHO.

src/clj/rems/api/applications_v2.clj
Macroz: I think that disagrees with my idea of how Clojure data structures are used.

src/clj/rems/api/applications_v2.clj
Macroz: Then there should at least be a TODO or maybe just assert that always throws that the problem is obvious later?

src/clj/rems/api/applications_v2.clj
Macroz: I don't understand why this function exists so I can't really think of a better name.

src/clj/rems/api/applications_v2.clj
luontola: What do you propose? Something like this?

```clj
{:application/id 1
 :application/created (DateTime. 1000)
 :application/modified (DateTime. 1000)
 :application/last-activity (DateTime. 1000)
 :application/applicant "applicant"
 :application/applicant-attributes {"eppn" "applicant"
                                    "mail" "applicant@example.com"
                                    "commonName" "Applicant"}
 :application/resources [{:catalogue-item/id 10
                          :resource/id 11
                          :resource/ext-id "urn:11"
                          :catalogue-item/title {:en "en title"
                                                 :fi "fi title"
                                                 :default "non-localized title"}
                          :catalogue-item/start (DateTime. 100)
                          :catalogue-item/state :enabled}
                         {:catalogue-item/id 20
                          :resource/id 21
                          :resource/ext-id "urn:21"
                          :catalogue-item/title {:en "en title"
                                                 :fi "fi title"
                                                 :default "non-localized title"}
                          :catalogue-item/start (DateTime. 100)
                          :catalogue-item/state :enabled}]
 :application/licenses [{:license/id 30
                         :license/accepted false
                         :license/type :link
                         :license/start (DateTime. 100)
                         :license/end nil
                         :license/title {:en "en title"
                                         :fi "fi title"
                                         :default "non-localized title"}
                         :license/link {:en "http://en-license-link"
                                        :fi "http://fi-license-link"
                                        :default "http://non-localized-license-link"}}
                        {:license/id 31
                         :license/accepted false
                         :license/type :text
                         :license/start (DateTime. 100)
                         :license/end nil
                         :license/title {:en "en title"
                                         :fi "fi title"
                                         :default "non-localized title"}
                         :license/text {:en "en license text"
                                        :fi "fi license text"
                                        :default "non-localized license text"}}]
 :application/events []
 :application/description ""
 :application/form {:form/id 40
                    :form/title "form title"
                    :form/fields [{:field/id 41
                                   :field/value ""
                                   :field/type :description
                                   :field/title {:en "en title" :fi "fi title"}
                                   :field/placeholder {:en "en placeholder" :fi "fi placeholder"}
                                   :field/optional false
                                   :field/options []
                                   :field/max-length 100}
                                  {:field/id 42
                                   :field/value ""
                                   :field/type :text
                                   :field/title {:en "en title" :fi "fi title"}
                                   :field/placeholder {:en "en placeholder" :fi "fi placeholder"}
                                   :field/optional false
                                   :field/options []
                                   :field/max-length 100}]}
 :application/workflow {:workflow/id 50
                        :workflow/type :dynamic
                        :workflow/state :rems.workflow.dynamic/draft
                        :workflow.dynamic/handlers #{"handler"}}
 :application/allowed-commands #{::dynamic/save-draft
                                 ::dynamic/submit}

```

src/clj/rems/api/applications_v2.clj
luontola: ::dynamic/add-member has now been removed

luontola: The TODO note is in 852

src/clj/rems/db/applications.clj
Macroz: I guess we need to have another discussion about this because this is not exactly what I intended.

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
luontola: `trigger-applications-update!` is now called after commit.

src/clj/rems/api/applications_v2.clj
luontola: Splitting the one generic handler for all events to multiple handlers did not improve things:

```clj

(defn- application-id-check-application-view
  [application event]
  (assert (= (:application/id application)
             (:application/id event))
          (str "event for wrong application "
               "(not= " (:application/id application) " " (:application/id event) ")"))
  application)

(defn- event-log-application-view
  [application event]
  (update application :application/events append-to-event-log event))

(defn- last-activity-application-view
  [application event]
  (assoc application :application/last-activity (:event/time event)))

(defn- application-view
  "Projection for the current state of a single application.
  Pure function; must use `assoc-injections` to enrich the model with
  data from other entities."
  [application event]
  (-> application
      (event-application-view event)
      (application-id-check-application-view event)
      (event-log-application-view event)
      (last-activity-application-view event)))
```

luontola: The name `event-application-view` is also confusing, because *all* the functions are applying events to the application model. The only difference that the multimethod is specific to some event types, whereas the other function contains generic functionality which applies to all events.

luontola: Another option:

```clj
(defn- application-view
  "Projection for the current state of a single application.
  Pure function; must use `assoc-injections` to enrich the model with
  data from other entities."
  [application event]
  (assert (or (nil? (:application/id application))
              (= (:application/id application)
                 (:application/id event)))
          (str "event for wrong application "
               "(not= " (:application/id application) " " (:application/id event) ")"))
  (-> application
      (event-specific-application-view event)
      (assoc :application/last-activity (:event/time event))
      (update :application/events append-to-event-log event)))
```

src/clj/rems/api/applications_v2.clj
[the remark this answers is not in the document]
Macroz: As discussed lets run the projection every time for now and come back to this later.

src/clj/rems/api/applications_v2.clj
Macroz: I like this `event-specific-application-view`. Could also be `event-type-specific-application-view`.

And the other "generic" code looks fine now, it's not hidden under the functions. When it gets more complicated, we can add named functions to do the specific common parts.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,869 @@
...
+(defn trigger-applications-update! []
+  (send-off applications-state update-applications!))
+
+(comment
+  (send applications-state reset-applications-state)
+  (trigger-applications-update!)
+  @applications-state)
+
+(defstate applications-update-scheduler
+  :start (future (while (not (Thread/interrupted))
```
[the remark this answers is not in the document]
Macroz: As discussed lets run the projection every time for now and come back to this later.

src/clj/rems/api/applications_v2.clj
luontola: Form and workflow attributes are now nested.

src/clj/rems/api/applications_v2.clj
luontola: Sounds good. I'll do it like in the last version.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+(defn- assoc-injections [application {:keys [get-form get-catalogue-item get-license get-user]}]
+  (-> application
+      (enrich-form get-form)
+      (update :application/resources enrich-resources get-catalogue-item)
+      (update :application/licenses enrich-licenses get-license)
+      (assoc :application/applicant-attributes (get-user (:application/applicant application)))))
+
+(defn- build-application-view [events injections]
+  (-> (reduce application-view nil events)
+      (assoc-injections injections)))
```
Macroz: Maybe @opqdonut can comment as well but I don't think this `assoc-injections` name clarifies at all what is happening here but only hides details. Do you like this pattern? I'd name this like `join-externals` or just have each of the "injections" lines of code here by themselves or name each one like `join-resources` or `enrich-resources`...

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+             :application/licenses (map (fn [license]
+                                          {:license/id (:license/id license)
+                                           :license/accepted false})
+                                        (:application/licenses event))
+             :application/events []
+             :application/form {:form/id (:form/id event)
+                                :form/fields []}
+             :application/workflow {:workflow/id (:workflow/id event)
+                                    :workflow/type (:workflow/type event)
+                                    ;; TODO: or would :workflow.dynamic/state be more appropriate?
```
Macroz: I agree with this TODO :+1:

src/clj/rems/api/applications_v2.clj
opqdonut: I think it's fine. The name could be `enrich-with-injections` or something. These are private functions and grouping them a bit instead of having one huge `->` is nice.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+                                        ::dynamic/reject
+                                        ::dynamic/return
+                                        ::dynamic/request-decision
+                                        ::dynamic/request-comment}})))
+
+(defmethod event-type-specific-application-view :application.event/returned
+  [application event]
+  application)
+
+(defmethod event-type-specific-application-view :application.event/comment-requested
```
Macroz: Would this be the place where you add the commenter permissions? Isn't there significant overlap with possible commands still?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+          (str "event for wrong application "
+               "(not= " (:application/id application) " " (:application/id event) ")"))
+  application)
+
+;; TODO: replace rems.workflow.dynamic/apply-event with this
+;;       (it will couple the write and read models, but it's probably okay
+;;        because they both are about a single application and are logically coupled)
+(defn- application-view
+  "Projection for the current state of a single application.
+  Pure function; must use `assoc-injections` to enrich the model with
```
Macroz: This docstring isn't up to date anymore?

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+                                     :license/title (assoc (localization-for :title license)
+                                                           :default (:title license))
+                                     content-key (assoc (localization-for :textcontent license)
+                                                        :default (:textcontent license))})))
+                           (sort-by :license/id))]
+    (merge-lists-by :license/id rich-licenses app-licenses)))
+
+(defn- assoc-injections [application {:keys [get-form get-catalogue-item get-license get-user]}]
+  (-> application
+      (enrich-form get-form)
```
Macroz: I'd prefer `enrich-form` and `enrich-resources`, ... to have the same shape.

src/clj/rems/api/applications_v2.clj
```diff
@@ -0,0 +1,810 @@
...
+                                                :resource/ext-id "urn:21"}]
+                       :application/licenses [{:license/id 30}
+                                              {:license/id 31}]
+                       :form/id 40
+                       :workflow/id 50
+                       :workflow/type :dynamic
+                       :workflow.dynamic/handlers #{"handler"}}
+
+        ;; expected values
+        new-application {:application/id 1
```
Macroz: Could name this `expected-new-application` and ditch the comment?

src/clj/rems/api/applications_v2.clj
Macroz: I'd prefer inlining that function and the "huge" `->` is exactly what I prefer here where each line in the expression is "at the same level of abstraction" and the same "shape".

luontola: Renamed to `enrich-with-injections`. Inlining it is not an option, because it is called from multiple places: the listing for all applications (`get-user-applications-v2`) needs to do the same things as `build-application-view` but in a bit different order.

src/clj/rems/api/applications_v2.clj
luontola: Yes, that would be the place. The idea is to derive the possible commands from the permissions.

src/clj/rems/api/applications_v2.clj
luontola: Done. `set-application-description` is the exception, because it depends on the result of `enrich-form` but updates a key outside `:application/form`.

PR #889, opened 2019-01-29
remove themes

opqdonut:
they've moved to the private rems-deploy repository


docs/configuration.md
```diff
@@ -61,7 +61,7 @@ For example, to add German localization, create a file `my-translations/de.edn`
 
 ## Themes
 
-Custom themes can be used by creating a file, for example `my-custom-theme.edn`, and specifying its location in the `:theme-path` configuration parameter. The theme file can override some or all of the theme attributes (see `:theme` in [config-defaults.edn](https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn)). Static resources can be placed in a `public` directory next to the theme configuration file. See [theme-lbr](https://github.com/CSCfi/rems/tree/master/theme-lbr) for an example theme.
+Custom themes can be used by creating a file, for example `my-custom-theme.edn`, and specifying its location in the `:theme-path` configuration parameter. The theme file can override some or all of the theme attributes (see `:theme` in [config-defaults.edn](https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn)). Static resources can be placed in a `public` directory next to the theme configuration file. See [resources/example-theme/theme.edn](resources/example-theme/theme.edn) for an example.
```
luontola: Should link to https://github.com/CSCfi/rems/blob/master/example-theme/theme.edn and the link text should be "example-theme/theme.edn"

example-theme/theme.edn
```diff
@@ -0,0 +1,24 @@
+{:color1 "#aa0000"
+ :color2 "#00aa00"
+ :color3 "#0000aa"
+ :color4 "#aa00aa"
+ ;; these are just the defaults:
```
luontola: The defaults don't need to be repeated here, because they will be inherited from config-defaults.edn

opqdonut: yes, but now it's a better example (for docs)

luontola: 👌

Issue #866, opened 2019-01-22
View details in admin pages


Macroz:
View details page for each concept in admin pages.

Check that all these actually make sense before implementation.

- [ ] view catalogue item
- [ ] view resource
- [ ] view workflow
- [ ] view form
- [ ] view license

PR #892, opened 2019-01-31
Resource detail view

Macroz:
Detail view for resources. See #866

### DIFF
--- a/src/cljs/rems/administration/resource.cljs
+++ b/src/cljs/rems/administration/resource.cljs
@@ -1,0 +1,103 @@
+(ns rems.administration.resource
+  (:require [clojure.string :as str]
+            [re-frame.core :as rf]
+            [rems.administration.administration :refer [administration-navigator-container]]
+            [rems.atoms :refer [external-link info-field readonly-checkbox]]
+            [rems.collapsible :as collapsible]
+            [rems.spinner :as spinner]
+            [rems.text :refer [localize-time text text-format]]
+            [rems.util :refer [andstr dispatch! fetch put!]]))
+
+(rf/reg-event-fx
+ ::enter-page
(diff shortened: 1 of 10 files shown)


src/cljs/rems/util.cljs
```diff
@@ -3,6 +3,16 @@
             [re-frame.core :as rf]
             [secretary.core :as secretary]))
 
+(defn andstr
+  "Like `apply str` but only produces something if all values are truthy like `and`.
+
+  (andstr nil 1 2 3) => nil
+  (andstr true 1 false 3) => nil
+  (andstr (:foo {:foo 2}) \"/\") => \"2/\""
```
luontola: Instead of having the examples only as documentation, move them to a unit test for some executable documentation.

Macroz: Done in another PR.

src/cljs/rems/administration/resource.cljs
```diff
@@ -0,0 +1,103 @@
...
+(defn licenses-view [licenses language]
+  [collapsible/component
+   {:id "licenses"
+    :title (text :t.administration/licenses)
+    :top-less-button? true
+    :collapse (if (seq licenses)
+                (into [:div]
+                      (for [license licenses]
+                        [license-view license language]))
+                [:p (text :t.administration/no-licenses)])}])
```
luontola: Instead of showing the "no licenses" text only after pressing "show more", it would be more user friendly to show the "no licenses" text right away and remove the "show more" link in that case.

luontola: And in case there are licenses, they could be shown right away. At least the license title and link (in current user language) could be shown without the need to press "show more".

Macroz: Already implemented in another PR

PR #893, opened 2019-01-31
Licenses with attachments

### LINKED ISSUE #808: vTHL2: Upload license in license editor
License can be a uploaded file (for example PDF). This is needed in addition to "external link" and "inline text" licence types.

Store license in db for now.

- [ ] attaching file in license editor
- [ ] store license file in db

Kauko:
Closes #808


src/clj/rems/api/licenses.clj
```diff
@@ -39,4 +61,34 @@
...
+      :summary "Add an attachment file that will be used in a license"
+      :roles #{:owner}
+      :multipart-params [file :- upload/TempFileUpload]
+      :middleware [upload/wrap-multipart-params]
+      :return AttachmentMetadata
+      (check-attachment-content-type (:content-type file))
+      (ok (licenses/create-license-attachment! file (util/getx-user-id))))
+
+    (POST "/remove_attachment" []
+      :summary "Remove an attachment file related to an application field"
```
opqdonut: application field?

src/clj/rems/api/licenses.clj
```diff
@@ -39,4 +61,34 @@
...
+      :roles #{:owner}
+      :query-params [attachment-id :- (describe s/Int "attachment id")]
+      :return SuccessResponse
+      (if (some? (licenses/remove-license-attachment! attachment-id))
+        (ok {:success true})
+        (ok {:success false})))
+
+    (GET "/attachments/:attachment-id" []
```
opqdonut: why attachment-id and not license-id?

oh right, because you first create the attachment and then the license.

test/clj/rems/test/api/licenses.clj
```diff
@@ -50,10 +60,80 @@
...
+                     :textcontent "license text"
+                     :attachment-id attachment-id
                      :localizations {:en {:title "en title"
-                                          :textcontent "en text"}
+                                          :textcontent "en text"
+                                          :attachment-id attachment-id}
                                      :fi {:title "fi title"
-                                          :textcontent "fi text"}}}]
+                                          :textcontent "fi text"
+                                          :attachment-id attachment-id}}}]
```
opqdonut: but what happens if you delete the attachment after creating the license?

src/clj/rems/db/test_data.clj
```diff
@@ -397,22 +397,26 @@
     ;; attach both kinds of licenses to all workflows
     (let [link (:id (db/create-license!
                      {:modifieruserid owner :owneruserid owner :title "non-localized link license"
-                      :type "link" :textcontent "http://invalid"}))
+                      :type "link" :textcontent "http://invalid" :attachmentId nil}))
```
opqdonut: does attachmentId need to be a mandatory field?

src/clj/rems/api/licenses.clj
```diff
@@ -39,4 +61,34 @@
...
+      :summary "Get a license's attachment"
+      :roles #{:owner}
+      :path-params [attachment-id :- (describe s/Int "attachment id")]
+      (if-let [attachment (db/get-license-attachment {:attachmentId attachment-id})]
+        (do (check-attachment-content-type (:type attachment))
+            (-> (:data attachment)
+                (java.io.ByteArrayInputStream.)
```
cscwooller: Not sure but can the stream stay open if something unexpected happens?

src/cljs/rems/administration/create_license.cljs
```diff
@@ -70,6 +74,35 @@
    (create-license request)
    {}))
 
+(defn- save-attachment [language form-data]
+  (post! (str "api/licenses/add_attachment")
+         {:body form-data
+          :handler (fn [{:keys [id] :as response}]
+                     (js/console.log (pr-str response))
```
Macroz: Remove console logging?

src/clj/rems/api/licenses.clj
```diff
@@ -1,17 +1,39 @@
...
+   .pdf, .doc, .docx, .ppt, .pptx, .txt, image/*"
+  [content-type]
+  (when-not (or (#{"application/pdf"
+                   "application/msword"
+                   "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
+                   "application/vnd.ms-powerpoint"
+                   "application/vnd.openxmlformats-officedocument.presentationml.presentation"
+                   "text/plain"}
+                 content-type)
+                (.startsWith content-type "image/"))
```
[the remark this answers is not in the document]
Macroz: What do you mean? 
- The UI does not currently check the content type and these checks are only in the backend. 
- The UI can at best offer some extensions that may be allowed but can't prevent the user from trying anything anyway.
- The application attachment and license attachment types may be different so they don't really make sense to share the same definitions.

src/clj/rems/api/licenses.clj
Macroz: The attachment is entirely read into memory and this stream is to that memory buffer so there is nothing special that can leak in the case of an exception. Only regular memory handled by the Garbage Collector.

src/clj/rems/api/licenses.clj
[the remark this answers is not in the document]
cscwooller: Ok, then no need.

src/cljs/rems/administration/workflow.cljs
```diff
@@ -65,6 +66,15 @@
...
                                             (str/upper-case (:langcode localization))) [:a {:target :_blank :href (:textcontent localization)} (:textcontent localization) " " [external-link]]]))
+                (when (= "attachment" (:licensetype license))
+                  (for [localization (:localizations license)]
+                    (when (:attachment-id localization)
+                      [inline-info-field
+                       (str (text :t.create-license/license-attachment)
+                            " "
+                            (str/upper-case (:langcode localization)))
+                       [attachment-link (:attachment-id localization) (:title localization)]
+                       {:no-box? true}])))
```
cscwooller: Could all the new code above be unified with resource licenses if you gave the text as a parameter to a function that both would use?

Macroz: Yeah that's what the TODO is there for. It's unclear at the moment if resource, workflow or form licenses all are needed, and if they are all equal. Also the structures that our API returns are currently not identical so they should be unified. This however is quite a lot of work that I'd rather do later when the first questions are resolved.

cscwooller: I was thinking just something like this:

```
(defn create-license-fields [info-text entity]
  (for [localization (:localizations entity)]
                    (when (:attachment-id localization)
                      [inline-info-field
                       (str info-text " " (str/upper-case (:langcode localization)))
                       [attachment-link (:attachment-id localization) (:title localization)]
                       {:no-box? true}])))

....
(when (= "attachment" (:licensetype license))
    (create-license-fields (text :t.create-license/license-attachment) license)
...

```

PR #897, opened 2019-02-01
Catalogue item detail view

Macroz:
Detail view for catalogue items. See #866 

First merge #894

### DIFF
--- a/src/cljs/rems/administration/workflow.cljs
+++ b/src/cljs/rems/administration/workflow.cljs
@@ -2,6 +2,7 @@
   (:require [clojure.string :as str]
             [re-frame.core :as rf]
             [rems.administration.administration :refer [administration-navigator-container]]
+            [rems.administration.components :refer [inline-info-field]]
             [rems.atoms :refer [external-link info-field readonly-checkbox]]
             [rems.collapsible :as collapsible]
             [rems.common-util :refer [andstr]]
@@ -40,9 +41,6 @@
   [:a.btn.btn-primary
    {:href "/#/administration/create-workflow"}
    (text :t.administration/create-workflow)])
-
-(defn inline-info-field [text value]
-  [info-field text value {:inline? true}])
 
 (defn get-localized-value [field key language]
   (key (first (filter (comp #{(name language)} :langcode)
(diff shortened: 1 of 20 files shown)


src/cljs/rems/administration/workflow.cljs
```diff
@@ -0,0 +1,138 @@
...
+                  [inline-info-field (str (text :t.administration/title)
+                                          " "
+                                          (str/upper-case (:langcode localization))) (:title localization)])
+                [[inline-info-field (text :t.administration/type) (:type license)]
+                 (when (:textcontent license)
+                   (case (:type license)
+                     "link" [inline-info-field (text :t.create-license/external-link) [:a {:target :_blank :href (:textcontent license)} (:textcontent license) " " [external-link]]]
+                     "text" [inline-info-field (text :t.create-license/license-text) (:textcontent license)]
+                     nil))]
+                (when (= "link" (:type license))
```
cscwooller: I don't get it why there is twice check if license type is link?

src/cljs/rems/administration/workflow.cljs
```diff
@@ -0,0 +1,138 @@
...
+        :open? (<= (count rounds 5))
+        :collapse (into [:div]
+                        (for [round rounds]
+                          [round-view round]))}])))
+
+(defn licenses-view [licenses language]
+  [collapsible/component
+   {:id "licenses"
+    :title (text :t.administration/licenses)
+    :top-less-button? (> (count licenses) 5)
```
cscwooller: The limit of 5 is now hard coded in quite many places, should it be a def somewhere?

src/cljs/rems/administration/workflow.cljs
Macroz: I thought it's more readable this day when there is one check per field.  I.e. there is a field for the type (that must do the typecase) and there is another field that is only shown for link licenses. Therefore two checks.

src/cljs/rems/administration/workflow.cljs
Macroz: I don't think it's useful to put it anywhere yet.

There might be a case for a reusable collapse with this kind of functionality later. The difficulty is in that the regular collapse takes any component as child and does not know of any counts. Also the counts might actually be use case specific.

PR #904, opened 2019-02-05
Drop unused tables catalogue_item_application_free_comment_values, *m…

cscwooller:
…embers, *_metadata, *_predecessor


resources/migrations/20190205094615-catalogue-item-application-extra-tables.down.sql
```diff
@@ -0,0 +1,46 @@
+CREATE TABLE public.catalogue_item_application_free_comment_values (
```
luontola: Remove the schema ("public." prefix). We cannot assume that what the user's default schema is.

PR #906, opened 2019-02-05
Improve browser test resilience

luontola:
- retry once if webdriver fails to start
- smoke test that app.js exists


opqdonut:
travis fails with
```
Failed to download https://github.com/technomancy/leiningen/releases/download/2.8.1/leiningen-2.8.1-standalone.zip (exit code 56)
It's possible your HTTP client's certificate store does not have the
correct certificate authority needed. This is often caused by an
out-of-date version of libssl. It's also possible that you're behind a
firewall and haven't set HTTP_PROXY and HTTPS_PROXY.
```
let's just disable travis

opqdonut:
merging despite failed travis

PR #909, opened 2019-02-08
project.clj: use new official rrb-vector, bump lein-doo version

opqdonut:


opqdonut:
thanks for the pointer @miikka

PR #911, opened 2019-02-08
Create catalogue item disabled

cscwooller:


src/clj/rems/db/test_data.clj
```diff
@@ -445,7 +445,7 @@
 
 (defn- create-catalogue-item! [resource workflow form localizations]
   (let [id (:id (db/create-catalogue-item!
-                 {:title "non-localized title" :resid resource :wfid workflow :form form}))]
+                 {:title "non-localized title" :resid resource :wfid workflow :form form :state "enabled"}))]
```
Macroz: Isn't `enabled` now the default?

resources/sql/queries.sql
```diff
@@ -45,8 +45,14 @@ WHERE ci.id = :item
...
 -- :doc Create a single catalogue item
 INSERT INTO catalogue_item
-(title, formid, resid, wfid)
-VALUES (:title, :form, :resid, :wfid)
+(title, formid, resid, wfid, state)
+VALUES (:title, :form, :resid, :wfid,
+/*~ (if (:state params) */
+CAST(:state as item_state)
+/*~*/
+CAST('enabled' as item_state)
```
Macroz: Shouldn't need a `CAST` here?

src/clj/rems/db/test_data.clj
cscwooller: Yes, but it is on lower levet, in queries.sql

resources/sql/queries.sql
cscwooller: True.

src/clj/rems/db/test_data.clj
Macroz: But doesn't `db/create-catalogue-item!` here refer to the db fn?

Issue #870, opened 2019-01-22
vTHL2: Adding members to an application after sending


Macroz:
Handler should be able to always add applicants to an application. Also applicant and handler should always be allowed to remove members (members are not allowed to make changes).

- [x] adding applicants with the editors like #609 
- ~should send emails to the new member~ (moved to #652)
- applicant can add a member after sending by making #832
- [x] handler can add a member "anytime"
- [x] applicant and handler can remove a member anytime
- ~send mail to member when member is removed~ (moved to #652)

PR #912, opened 2019-02-08
Refactor members and implement viewing

Macroz:
See #609 and #870.

- Removes add member support for old applications
- Splits add member events to add member (REMS user, handler action only) and invite member (not rems user, handler and applicant depending on state)
- Implements a view of the additional members of an application as well as invited members.

### DIFF
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -125,9 +125,13 @@
                    :view "View"
                    :workflow "Workflow"
                    :workflows "Workflows"}
-  :applicant-info {:applicants "Applicants"
+  :applicant-info {:applicant "Applicant"
+                   :applicants "Applicants"
                    :email "Email"
-                   :username "Name"}
+                   :invited-member "Invited member"
+                   :member "Member"
+                   :name "Name"
+                   :username "Username"}
   :applications {:application "Application"
                  :applications "Applications"
                  :created "Created"
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -125,7 +125,7 @@
                    :view "View"
                    :workflow "Workflow"
                    :workflows "Workflows"}
-  :applicant-info {:applicant "Applicant"
+  :applicant-info {:applicants "Applicants"
                    :email "Email"
                    :username "Name"}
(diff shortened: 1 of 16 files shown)


resources/translations/en.edn
```diff
@@ -139,6 +143,7 @@
                                   :decision-requested "Decision requested"
                                   :draft-saved "Modified"
                                   :member-added "Member added"
+                                  :member-invited "Invited member"
```
cscwooller: This is quite splitting hair, but would prefer "Member invited" for consistency.

resources/translations/fi.edn
```diff
@@ -139,6 +143,7 @@
                                   :decision-requested "Pyydetty päätöstä"
                                   :draft-saved "Muokattu"
                                   :member-added "Lisätty jäsen"
+                                  :member-invited "Kutsuttu jäsen"
```
cscwooller: Same here, as an event "Jäsen kutsuttu"? Although here the member-added is not in same form.

src/clj/rems/db/applications.clj
```diff
@@ -946,17 +940,6 @@
...
-  (let [application (get-application-state application-id)]
-    (when-not (= user-id (:applicantuserid application))
-      (throw-forbidden))
-    (when-not (#{"draft" "returned" "withdrawn"} (:state application))
-      (throw-forbidden))
-    (assert (users/get-user-attributes member) (str "User '" member "' must exist"))
-    (db/add-application-event! {:application application-id :user user-id :round 0
-                                :comment nil
-                                :event "add-member" :eventdata (cheshire/generate-string {"uid" member})})))
-
```
cscwooller: Good that we are getting code away from applications. :)

resources/translations/en.edn
Macroz: Yes it's better like that. Matches the others.

resources/translations/fi.edn
Macroz: In this case the other Finnish translation matches this form and I think both ways are fine so I'll keep it like this.

PR #920, opened 2019-02-12
refactor: map as error with :key like in field validations

Macroz:
Some dynamic workflow errors had the new `:key` field and some did not. This modifies the format to always have the `:key` but does not try to localize them all yet.

### DIFF
--- a/src/clj/rems/workflow/dynamic.clj
+++ b/src/clj/rems/workflow/dynamic.clj
@@ -272,7 +272,7 @@
 (defn- applicant-error
   [application cmd]
   (when-not (= (:actor cmd) (:applicantuserid application))
-    {:errors [{:key :forbidden}]}))
+    {:errors [{:type :forbidden}]}))
 
 (defn- handler?
   [application user]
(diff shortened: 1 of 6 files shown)


src/clj/rems/workflow/dynamic.clj
```diff
@@ -272,7 +272,7 @@
 (defn- applicant-error
   [application cmd]
   (when-not (= (:actor cmd) (:applicantuserid application))
-    {:errors [:forbidden]}))
+    {:errors [{:key :forbidden}]}))
```
opqdonut: why `:key` and not `:dynamic.error/type` or something?

Macroz: It is fed directly to the localization. That's the format used in field errors already.

Macroz: I remembered wrong and it's `:type` elsewhere already so I'll just use that. If we want we can sometime refactor it to e.g. `:error/type`

PR #930, opened 2019-02-13
Dynamic roles and permissions: side by side with old possible commands

luontola:
#869 

This PR is to make sure the new dynamic role system works the same way as the old possible commands system. Following PRs will remove the old code.


src/clj/rems/workflow/permissions.clj
```diff
@@ -0,0 +1,145 @@
+(ns rems.workflow.permissions
+  (:require [clojure.set :as set]
+            [clojure.test :refer [deftest is testing]]))
+
+(def ^:private conj-set (fnil conj (hash-set)))
```
cscwooller: What is the purpose of this def?

luontola: To produce a set when conj'ing nil.

cscwooller: I would rather put it to utils or sth with more descriptive name.

luontola: I think the name is already as descriptive as it is, and it's such a rarely needed operation that moving it elsewhere only complicates reading the code.

Macroz: I think this is fine as it is now.

PR #931, opened 2019-02-14
Topic/localization

foxlynx:


src/clj/rems/locales.clj
```diff
@@ -2,27 +2,38 @@
...
-(defn load-translations [{:keys [languages translations-directory]}]
-  (->> languages
-       (map #(load-translation % translations-directory))
-       (apply merge)))
+(defn- load-translation [language translations-directory extra-translations-directory]
+  (let [filename (str (name language) ".edn")
+        base-translations (translations-from-file filename translations-directory)]
+    (if extra-translations-directory
+      (deep-merge {language (read-string (slurp base-translations))}
+                  {language (read-string (slurp (translations-from-file filename extra-translations-directory)))})
```
opqdonut: why not keep read-string & slurp in translations-from-file?

src/clj/rems/util.clj
```diff
@@ -60,6 +60,15 @@
   ([user]
    (get user "mail")))
 
+(defn deep-merge [a & maps]
+  (if (map? a)
```
opqdonut: I'm not sure why this case is needed. Why not just
```
(defn deep-merge [& maps]
  (apply merge-with deep-merge maps))

File: test/clj/rems/test/locales.clj [hunk 3]
```diff
@@ -39,12 +43,31 @@
...
+                                                   :translations-directory "translations/"
+                                                   :extra-translations-directory nil})))))
+
+  (testing "missing translations-directory in config is an error"
+    (is (thrown-with-msg? RuntimeException #"^\Q:translations-directory was not set in config\E$"
+                          (locales/load-translations {:languages [:xx]}))))
 
   (testing "missing translations is an error"
-    (is (thrown-with-msg? FileNotFoundException #"^\Qtranslations for :xx language could not be found in some-dir/xx.edn file or translations/xx.edn resource\E$"
+    (is (thrown-with-msg? FileNotFoundException #"^\Qtranslations could not be found in some-dir/xx.edn file or some-dirxx.edn resource\E$"
```
opqdonut: should be `some-dir/xx.edn` right?

dev-config.edn
```diff
@@ -1,4 +1,7 @@
 {:dev true
  :nrepl-port 7000
  :database-url "postgresql://localhost/rems?user=rems"
- :authentication :fake-shibboleth}
+ :authentication :fake-shibboleth
+ :translations-directory "translations/"
```
opqdonut: that's the default, no need to set

resources/translations/lbr/fi.edn
```diff
@@ -0,0 +1,14 @@
+{:t
```
opqdonut: These seem like part of the LBR deployment. If so, they should be in the rems-deploy repository IMO

src/clj/rems/util.clj
cscwooller: It is needed. Without it the following happens:

(deep-merge {:a {:b {:c 1 :d 2}}} {:a {:b {:c 100}}})

expected: {:a {:b {:c 100, :d 2}}}
  actual: ({:a {:b {:c 100}}})

resources/translations/lbr/fi.edn
cscwooller: Yes, let's put them to the deploy repo.

src/clj/rems/locales.clj
```diff
@@ -2,27 +2,37 @@
...
   (:import (java.io FileNotFoundException)))
 
-(defn- load-translation [language translations-directory]
-  (let [file (when translations-directory
-               (io/file translations-directory (str (name language) ".edn")))
-        resource-path (str "translations/" (name language) ".edn")
+(defn- translations-from-file [filename dir]
+  (let [file (when dir
+               (io/file dir filename))
+        resource-path (str dir filename)
```
luontola: This is fragile. It will break if `dir` does not end with `/`. Also it's questionable that why the resource path should be parameterizable, because it is known at build time.

cscwooller: Ok, changing it so that it is always taken from under theme.

PR #933, opened 2019-02-15
Dynamic roles and permissions: using roles for page visibility

luontola:
#869, continues PR #930

The roles of a user are a union of the user's roles over all applications, plus the `:logged-in` role which everybody has, plus whatever user-specific legacy roles (some of those will still need to be removed). Everybody can create applications. The `:applicant` role now means that the user is the applicant in some application.


src/clj/rems/api/applications.clj
```diff
@@ -163,7 +156,7 @@
       (-> application
```
Macroz: The handler and other user permissions are also sensitive information that should be hidden

src/clj/rems/workflow/permissions.clj
```diff
@@ -0,0 +1,166 @@
...
+(def ^:private conj-set (fnil conj (hash-set)))
+
+(defn give-role-to-user [application role user]
+  (assert (keyword? role)
+          (str "role must be a keyword: " (pr-str role)))
+  (assert (string? user)
+          (str "user must be a string: " (pr-str user)))
+  (update-in application [::user-roles user] conj-set role))
+
+(defn give-role-to-users [application role users]
```
Macroz: Use only this plural version, no need for the singular function?

src/clj/rems/workflow/permissions.clj
```diff
@@ -0,0 +1,166 @@
...
+               (set-role-permissions {:role-1 [:foo]
+                                      :role-2 [:foo]})
+               (remove-permission-from-all :foo)))))
+  (testing "leaves unrelated permissions unchanged"
+    (is (= {::role-permissions {:role #{:bar}}}
+           (-> {}
+               (set-role-permissions {:role [:foo :bar]})
+               (remove-permission-from-all :foo))))))
+
+(defn user-permissions
```
Macroz: This could be written for example

```
(defn user-permissions
  "Returns the specified user's permissions to this application.
   Union of all role specific permissions."
  [application user]
  (->> (user-roles application user)
       (mapcat (fn [role]
              (get-in application [::role-permissions role])))
       (set)))
```

src/clj/rems/workflow/dynamic.clj
```diff
@@ -196,8 +198,150 @@
                              :event/actor "foo"
                              :application/id 123}))))))
 
+;;; Roles and permissions
 
-;;; Events
+(defn hide-sensitive-dynamic-events [events]
```
Macroz: This could be written

```
(defn hide-sensitive-dynamic-events [events]
  (remove (comp #{:application.event/comment-requested
                  :application.event/commented
                  :application.event/decided
                  :application.event/decision-requested}
                :event/type)
          events))
```

src/cljs/rems/application.cljs
```diff
@@ -1035,22 +1034,22 @@
 
 
 (defn- dynamic-actions [app]
```
Macroz: This could be written like so

```
(defn- dynamic-actions [app]
  (let [command->action {:rems.workflow.dynamic/save-draft [save-button]
                         :rems.workflow.dynamic/submit [submit-button]
                         :rems.workflow.dynamic/return [return-action-button]
                         :rems.workflow.dynamic/request-decision [request-decision-action-button]
                         :rems.workflow.dynamic/decide [decide-action-button]
                         :rems.workflow.dynamic/request-comment [request-comment-action-button]
                         :rems.workflow.dynamic/comment [comment-action-button]
                         :rems.workflow.dynamic/approve [approve-reject-action-button]
                         :rems.workflow.dynamic/reject [approve-reject-action-button]
                         :rems.workflow.dynamic/close [close-action-button]}]
    (distinct (map command->action (:possible-commands app)))))
```

src/clj/rems/workflow/dynamic.clj
```diff
@@ -196,8 +198,150 @@
...
+  (filter (fn [event]
+            ((complement contains?) #{:application.event/comment-requested
+                                      :application.event/commented
+                                      :application.event/decided
+                                      :application.event/decision-requested}
+             (:event/type event)))
+          events))
+
+(defmulti calculate-permissions
```
Macroz: I feel like this function is not working so well because it doesn't actually calculate the permissions. You have to have a separate step to remember to remove "impossible" permissions.

src/cljs/rems/roles.cljs
```diff
@@ -0,0 +1,24 @@
...
+(defn- has-some? [expected-roles actual-roles]
+  (not (empty? (set/intersection expected-roles actual-roles))))
+
+(defn is-logged-in? [roles]
+  (has-some? #{:logged-in} roles))
+
+(defn is-applicant? [roles]
+  (has-some? #{:applicant} roles))
+
+(defn is-handler? [roles]
```
Macroz: This function is not named well where the `handler` word is now reserved for the superuser and this should be something totally different. Possibly you could use `is-actor?` because e.g. the page is named actions but that is also overloaded in the event terminology.

src/cljs/rems/application.cljs
luontola: But then the buttons will be in unpredictable order (the order of commands in the set).

src/clj/rems/workflow/dynamic.clj
luontola: The impossible permissions filtering is there just for backward compatibility with the old implementation, so that the list of possible commands would be the same. It's not so much about permissions, but about application state: you're allowed to remove members, but if there are no members then you can't remove anything, even though technically you still have full authority to remove members. Assuming the UI has a remove member button next to each member, then when there are no members, the button is never rendered regardless of whether the user is authorized to remove members or not. The commands also anyways validate when the user tries to remove a member that the member exists. Once the the old permission system based on `impossible-command?` is removed, then `remove-impossible-permissions` can also be removed without affecting functionality or security.

src/cljs/rems/roles.cljs
luontola: How about "reviewer"? After removing the old workflow, that word should be free. The handler and others are all reviewing the application. [Definitions of review](https://www.dictionary.com/browse/review) include "to go over (lessons, studies, work, etc.) in review", "to view, look at, or look over again", "to inspect, especially formally or officially". "Inspector" could be used as a synonym.

The name "actions" is also overloaded. The buttons in the application page are called actions and some of the actions are done by the applicant. Calling it the review page would be more descriptive.

src/clj/rems/workflow/permissions.clj
luontola: Most of the usages give a role to a single user, `(:event/actor event)`. There are only two special cases (handlers and commenters) where there are multiple users. Why complicate the default case? There's also the operation in the other direction, `remove-role-from-user`, which doesn't even have a plural version. Should that also be made plural-only, even though it's not needed? And anyways for testing and code readability the singular versions would still exists, even if they would be private.

src/cljs/rems/application.cljs
luontola: `static-actions` has the buttons in predictable order.

src/clj/rems/api/applications.clj
luontola: The permissions are removed by calling `rems.workflow.permissions/cleanup` before it is returned from the public API. The API schema should prevent them slipping through accidentally.

The list of handlers in the workflow should probably be hidden from the applicant.

src/clj/rems/workflow/permissions.clj
```diff
@@ -0,0 +1,166 @@
+(ns rems.workflow.permissions
```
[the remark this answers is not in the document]
luontola: Moved to rems.permissions

src/clj/rems/workflow/permissions.clj
luontola: I considered `give-role-to-users [application role & users]` but calling apply to pass in the list of users seemed too complicated.

src/clj/rems/api/applications.clj
luontola: Handlers are now hidden.

src/clj/rems/workflow/permissions.clj
Macroz: Simply because passing `[user]` is quite as simple as just `user`. And the resulting combined function is still only a few lines long.

src/cljs/rems/application.cljs
Macroz: Yeah, I guess the original is better.

src/cljs/rems/roles.cljs
Macroz: I think for our the users the review means something else so I would not be so eager to use it either. They are used to so-called review and third party review functionality. Have to think about this.

src/clj/rems/workflow/dynamic.clj
Macroz: If that becomes possible and all permissions are updated by the events in `calculate-permissions` then it's good.

Issue #852, opened 2019-01-17
vTHL3: Remove round-based workflow


opqdonut:
Background:

- dynamic workflow should be enough for everybody
- we want to get rid of old round-based workflow
  - db tables
  - `get-form-for`
- old application API conflates forms with applications
- we might need a new autoapprove workflow (if LBR needs it)

Steps:
- [x] spike for new application API (#878)
- [x] code for converting round-based applications to dynamic applications (for LBR production data) (TBD: how much granularity to preserve) (#948)
- [x] feature complete new application API (#1004, #1012)
- [x] decouple dynamic workflow from round-based workflow
  - [x] form validation (#1013)
  - [x] creating/saving the application (#1031)
- [x] replace `rems.workflow.dynamic/apply-event` multimethod with new `rems.api.applications-v2/application-view` (#1050 #1054 #1056 #1062 #1065)
  - the old tests must be run against the new implementation
  - or then follow CQRS: keep the old method as the write model and use the new method as the read model
- [x] switch to new application API in frontend (#1015, #1017)
- [x] applications page: use the new application API directly instead of through v2->v1 transformation (#1024)
- [x] remove old application API (#1020 #1021 #1022 #1033)
- [x] remove /api/v2/applications/:id/migration (#1063 #1066)
- [x] rename new v2 API to v1 or rename all the remaining ones to v2 (#1064 #1088)
- [x] remove dead code:
  - [x] remove `applicant`, `approver` and `reviewer` roles from database (#1036)
    - also cleanup `rems/roles.cljs` and lots of other code
    - consider renaming the `owner` role to `admin`
  - leftovers in #1101 
- [x] solve the cyclic dependency and `resolve` hack at https://github.com/CSCfi/rems/blob/39f75e287d7a33dc0f424e0c6bbc63a27b19fff1/src/clj/rems/db/applications.clj#L1033 (#1033)

luontola:
Currently any further cleanup is blocked by the code which migrates old applications to dynamic workflow. After releasing a version of the application which can do the migration, the migration code can be removed and lots of dead code with it.

opqdonut:
leftovers split out

PR #948, opened 2019-02-20
Convert round-based applications to dynamic

luontola:
#852

### DIFF
--- a/src/clj/rems/migrations/convert_to_dynamic_applications.clj
+++ b/src/clj/rems/migrations/convert_to_dynamic_applications.clj
@@ -1,6 +1,79 @@
 (ns rems.migrations.convert-to-dynamic-applications
   (:require [clojure.java.jdbc :as jdbc]
-            [clojure.test :refer :all]
-            [conman.core :as conman]
-            [rems.db.core :as db :refer [*db*]]
-            [rems.db.workflow :as workflow]))
+            [rems.db.applications :as applications]
+            [rems.db.core :refer [*db*]]
+            [rems.db.roles :as roles]
+            [rems.db.users :as users]
+            [rems.db.workflow :as workflow])
+  (:import [java.util UUID]))
+
+(defn migrate-catalogue-items! [workflow-id]
+  (jdbc/execute! *db* ["update catalogue_item set wfid = ?" workflow-id]))
+
+(defn migrate-application! [application-id workflow-id]
+  (let [read-user (->> (users/get-all-users)
+                       (map (fn [user] (assoc user :roles (roles/get-roles (:eppn user)))))
+                       (filter (fn [user] (contains? (:roles user) :approver)))
+                       first
+                       :eppn)
+        form (applications/get-form-for read-user application-id)
+        application (:application form)
+        workflow (workflow/get-workflow workflow-id)]
+    (assert (= "workflow/dynamic" (get-in workflow [:workflow :type])))
+
+    ;; use the dynamic workflow
+    (jdbc/execute! *db* ["update catalogue_item_application set wfid = ? where id = ?" (:id workflow) (:id application)])
+    ;; delete old events
+    (jdbc/execute! *db* ["delete from application_event where appid = ?" (:id application)])
+
+    (applications/add-application-created-event! {:application-id (:id application)
+                                                  :catalogue-item-ids (->> (:catalogue-items application)
+                                                                           (map :id))
+                                                  :time (:start application)
+                                                  :actor (:applicantuserid application)})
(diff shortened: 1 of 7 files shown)


src/clj/rems/migrations/convert_to_dynamic_applications.clj
```diff
@@ -0,0 +1,21 @@
...
+            [conman.core :as conman]
+            [rems.db.core :as db :refer [*db*]]
+            [rems.db.workflow :as workflow]))
+
+(defn- use-db [jdbc-url f]
+  (binding [*db* (conman/connect! {:jdbc-url jdbc-url})]
+    (try
+      (f)
+      (finally
+        (conman/disconnect! *db*)))))
```
opqdonut: hmm should this be somewhere else?

or should we just use mount as usual for the db connection?

test/clj/rems/test/migrations/convert_to_dynamic_applications.clj
```diff
@@ -0,0 +1,499 @@
...
+            [rems.test.api :refer [api-fixture]])
+  (:import [java.util UUID]))
+
+(use-fixtures
+  :once
+  api-fixture)
+
+(defn migrate-catalogue-items! [workflow-id]
+  (jdbc/with-db-connection [conn *db*]
+    (jdbc/execute! conn ["update catalogue_item set wfid = ?" workflow-id])))
```
opqdonut: transactions?

test/clj/rems/test/migrations/convert_to_dynamic_applications.clj
```diff
@@ -0,0 +1,499 @@
...
+    (migrate-application! 2 (:id new-workflow))
+    (migrate-application! 3 (:id new-workflow))
+    (migrate-application! 4 (:id new-workflow))
+    (migrate-application! 5 (:id new-workflow))
+    (migrate-application! 8 (:id new-workflow))
+    (migrate-application! 9 (:id new-workflow))
+    (migrate-application! (:id application) (:id new-workflow))
+    (println "--- after ---")
+    (pprint (applications/get-application-state (:id application))))
+
```
opqdonut: you could run `rems.validate/validate` here

you could perhaps also add some checks to the validation code if needed

src/clj/rems/migrations/convert_to_dynamic_applications.clj
luontola: The `migrate-up` method is called by migratus. Currently `-main` does not `mount/start *db*` when running migrations, but that could be changed.

Best would be to use the same database connection as migratus uses. The Connection instance is inside the config map that `migrate-up` receives, but it would need to be wrapped inside a mock DataSource that always returns the same connection, because `*db*` expects a DataSource instead of Connection.

test/clj/rems/test/migrations/convert_to_dynamic_applications.clj
luontola: `rems.validate/validate` only checks whether `applications/get-application-state` throws an exception. The actual validation happens every time an event is written or read from the database.

src/clj/rems/migrations/convert_to_dynamic_applications.clj
luontola: This file will actually be removed. It's better to run the data migration as a separate command from the command line instead of as part of the schema migrations.

src/clj/rems/migrations/example.clj
```diff
@@ -0,0 +1,9 @@
+(ns rems.migrations.example
```
opqdonut: a word somewhere about how this is an example of how to do clj migrations would be nice

test/clj/rems/test/migrations/convert_to_dynamic_applications.clj
```diff
@@ -0,0 +1,704 @@
...
+            [rems.validate :as validate])
+  (:import [java.util UUID]))
+
+(use-fixtures
+  :once
+  api-fixture)
+
+(defn add-more-test-data! []
+  (jdbc/execute! *db* ["
+INSERT INTO catalogue_item_application (id, applicantuserid, start, endt, modifieruserid, wfid, description) VALUES (20, 'alice', '2019-03-05 11:56:19.353103', null, null, 1, 'direct approval');
```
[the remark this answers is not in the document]
luontola: It's just for this one-off migration, so it can be removed in the next version. I think there is a separate task about improving the test data.

src/clj/rems/migrations/example.clj
luontola: Done.

PR #949, opened 2019-02-20
Fix :last-modified being nil and failing schema validation

luontola:
This situation happens when the application has no events, which is possible with old applications.

The last modified time is already calculated robustly in:
rems.db.applications/get-application-state
rems.db.applications/get-dynamic-application-state


opqdonut:
for #852, related to #948

PR #952, opened 2019-02-20
Allow members to see application

### LINKED ISSUE #950: Members can't see open applications
1. Log in as developer, create a dynamic workflow application
2. Submit the application
3. Open the application, add Alice as member (remember the id)
4. Log out, log in as Alice
-> Alice can't see the form, access is denied even with a direct url

luontola:
Fixes #950 

The old `rems.db.applications/get-user-applications` is unaware of members and needed a hack. Updating it to be aware of members is probably not worth it because #852 will anyways replace it with `rems.api.applications-v2/get-user-applications-v2`

### DIFF
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -201,6 +201,21 @@
   ;; schema could do these coercions for us...
   (update-present cmd :decision keyword))
 
+(defn- get-user-applications [user-id]
+  ;; XXX: the old API doesn't know about members, so rems.db.applications/get-user-applications doesn't return applications where the user is a member
+  ;; XXX: this code cannot be moved to rems.db.applications/get-user-applications because it would create cyclic dependencies
+  (let [old-applications (applications/get-user-applications user-id)
+        member-application-ids (->> (applications/get-dynamic-application-events-since 0)
+                                    (reduce dynamic-roles/permissions-of-all-applications nil)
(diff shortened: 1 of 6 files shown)


src/clj/rems/api/applications.clj
```diff
@@ -206,7 +209,20 @@
       :summary "Get current user's all applications"
       :roles #{:logged-in}
       :return GetApplicationsResponse
-      (ok (applications/get-user-applications (getx-user-id))))
+      ;; XXX: the old API doesn't know about members, so it doesn't return applications where the user is a member
+      ;; XXX: this code cannot be moved to rems.db.applications because it would create a cyclic dependency
```
Macroz: What cylic dependency? `rems.api.applications` depends on `rems.db.applications` but there shall be no other direction dependency.

luontola: There are actually two cycles that would be produced by both `rems.api.applications/api-get-application` and `rems.db.dynamic-roles/permissions-of-all-applications`.

I extracted `rems.api.applications/get-user-applications` so at least it isn't inlined inside the router.

luontola:
The purpose of :past-commenter and :past-decider is to allow the commenter/reviewer to still view the application, even if they have been removed from it. When a member is removed from an application, does he still need to view the application? And in what situation would a handler be removed from the application - currently it's not possible?

PR #953, opened 2019-02-20
 811 comment request grouping

### LINKED ISSUE #811: vTHL2: Show who a request was sent to and what they replied
NB: Concerns both static and dynamic application events.

Approve and review requests and responses:
- application history must include the information to whom the approval, review or comment request was sent to. We store the actor in userid column in `application_event` table and the receiver in the `eventdata` column where it should be shown from or create a new column for static applications?
- responses must be tied to requests, meaning that on history list it must be unambiguous that which response belongs to which request. 

See image for design.

hukka:
Fix #811 for dynamic workflows (since static will be deprecated)

Still open:
- [x] Migrations from round based system will be looked in the future, when it's time
- [x] Test errors after merging to master
- [x] Change latest comment to latest event group


src/clj/rems/workflow/dynamic.clj
```diff
@@ -636,12 +639,17 @@
...
-      {:success true
-       :result {:event/type :application.event/commented
-                :event/time (:time cmd)
-                :event/actor (:actor cmd)
-                :application/id (:application-id cmd)
-                :application/comment (:comment cmd)}}))
+      (let [last-request-for-actor
+            (last (filter #(contains? (set (:application/commenters %))
+                                      (:actor cmd))
+                           (:dynamic-events application)))]
```
luontola: Don't iterate the events here, but assoc the necessary state to application in the `apply-event [:application.event/comment-requested :workflow/dynamic]` method. (Because the application here has mixed responsibilites - it's both the read model and the write model instead of following CQRS - you'll probably need to dissoc the state before returning the application from the public API. See usages of  `rems.permissions/cleanup` to find the right places.)

luontola: Also it seems weird to dig the last request ID from the events when it could come as part of the command. Or would that complicate the client unnecessarily?

luontola: If the comments are always related to the latest comment request, then the use of a request ID is unnecessary because the information can be derived from the existing events.

test/clj/rems/test/api/applications.clj
```diff
@@ -924,9 +924,7 @@
           (is (= ["application.event/created"
                   "application.event/draft-saved"
                   "application.event/submitted"
-                  ; decision- and comment-requested, and decided should be missing here
-                  "application.event/commented"
-                  "application.event/commented"
+                  ; all decision and comment related events should be missing here
```
Macroz: Should this test be rewritten so that both events that can be seen and cannot be seen are explicitly listed? There have been bugs related to this functionality before.

src/cljs/rems/application.cljs
```diff
@@ -788,11 +781,11 @@
                (when state (str ": " (localize-state state))))]
       :always [:div
                [:div.mb-3 {:class (str "state-" (if (keyword? state) (name state) state))} (phases phases-data)]
-               (when latest-comment
-                 (info-field (text :t.applications/latest-comment)
-                             latest-comment))]
-      :collapse (when (seq event-groups)
-                  [events-view event-groups])}]))
+               [:h4 (text :t.form/events)]
+               (when-let [g [(first event-groups)]]
```
Macroz: Should be `(when-let [g (first event-groups)]`

src/clj/rems/api/schema.clj
```diff
@@ -98,6 +98,7 @@
    (s/optional-key :decider) s/Str
    (s/optional-key :decision) s/Keyword
    (s/optional-key :commenters) #{s/Str}
+   (s/optional-key :latest-comment-requests) {s/Str s/Uuid}
```
Macroz: Maybe `:latest-request-by-user`. However this is problematic in that you see other users comment requests so it should only be returned to the "handlers".

hukka:
> Does this only group comment requests? How about decision request?

Just comment requests. Decisions in a later PR when we get this merged.

src/clj/rems/api/schema.clj
hukka: Hm. I decided to reveal it in the API so that the UI could show the relevant request near the comment action, but yes, this leaks information to too many users if we don't want applicants to see that there are any comment requests.

src/cljs/rems/application.cljs
hukka: Well, then need to create the seq in the next line, but fine.

src/clj/rems/api/schema.clj
```diff
@@ -98,6 +98,7 @@
    (s/optional-key :decider) s/Str
    (s/optional-key :decision) s/Keyword
    (s/optional-key :commenters) #{s/Str}
+   (s/optional-key :latest-comment-request-by-user) {s/Str s/Uuid}
```
luontola: This is only used by the commands, so it should not be exposed to the UI. Please remove it from the public schema.

It might be good to also change it to a private namespaced keyword, i.e. `::latest-comment-request-by-user`, so that it would be tied to the `rems.workflow.dynamic` namespace and not accidentally be used elsewhere.

src/cljs/rems/application.cljs
```diff
@@ -735,40 +735,44 @@
                                 :readonly readonly?
                                 :approved (get licenses (:id license)))]))])]}]))
 
+
+;; FIXME Why do we have both this and dynamic-event->event?
```
luontola: `dynamic-event->event` converts dynamic events to the legacy event format used by the round-based workflow. After removing the round-based workflow we should be able to clean up this.

test/clj/rems/test/api/applications.clj
```diff
@@ -813,6 +814,73 @@
           "user should be forbidden to send command"))
 
     (testing "send commands with authorized user"
+      (testing "even handler cannot comment without request"
```
luontola: These tests seem to be mostly duplicating `rems.workflow.dynamic/test-comment`. Integration tests should only check that integration (e.g. HTTP transport layer) is working, but not all the detailed business rules, which are better checked using fast unit tests.

test/clj/rems/test/api/applications.clj
```diff
@@ -813,6 +814,73 @@
...
+                                                         {:type :rems.workflow.dynamic/request-comment
+                                                          :application-id application-id
+                                                          :commenters [decider-id commenter-id]
+                                                          :comment "What say you?"})))
+            (is (= #{commenter-id decider-id}
+                   (set (get-in (get-application handler-id application-id)
+                               [:application :commenters])))))
+          (testing "cannot see internal commenter state"
+            (is (= nil (get-in (get-application handler-id application-id)
+                               [:application :latest-comment-request-by-user]))))
```
luontola: Removing `:latest-comment-request-by-user` from the schema will avoid the need for this test.

test/clj/rems/test/api/applications.clj
```diff
@@ -839,28 +907,33 @@
...
+                    "application.event/comment-requested"
+                    "application.event/comment-requested"
+                    "application.event/commented"
+                    "application.event/commented"
+                    "application.event/decision-requested"
+                    "application.event/decided"
+                    "application.event/approved"]
+                   handler-event-types)))
+          (testing "applicant cannot see all events"
+            (is (= (filter ;; all decision and comment related events should be missing here
```
luontola: I'd prefer writing out the list of expected events explicitly. Now there is a possibility of bugs in in the filter call. It's a question of "who tests the tests". The code has no obvious mistakes, but I much prefer it to obviously have no mistakes.

Macroz: I wanted to avoid the comment and have an explicit list of events both ways. We have had regressions in this functionality before (that's why check with a blacklist as well rather than just comment about it) and since it's about who can see what it should be coded defensively (i.e. whitelist).

PR #954, opened 2019-02-21
refactor: move joda-time-writer to db ns

Macroz:
This helps running single db tests that don't require the api


opqdonut:
related to #1081

PR #955, opened 2019-02-21
Update date related deps

Macroz:


src/clj/rems/db/applications.clj
```diff
@@ -954,16 +955,25 @@
...
-  {DateTime string->datetime})
+  {DateTime longstring->datetime})
+
+(def ^:private datestring-coercion-matcher
+  {DateTime datestring->datetime})
 
 (defn- coercion-matcher [schema]
-  (or (datetime-coercion-matcher schema)
+  (or (datestring-coercion-matcher schema)
+      (datetime-coercion-matcher schema)
```
luontola: Maybe better names for the prefixes would be  "iso-datetime" and "unix-datetime"

Macroz:
Rebased over @Hukka PR

Issue #806, opened 2018-12-21
vTHL2: Archiving resources, licences, workflows and forms   


foxlynx:
There should be a way to archive resources, workflows, forms and catalogue items (enable / disable exists for catalogue item already)

- [x] add state to all if there is not a column yet (keep start and end times also for other use)
- catalogue items
  - [x] add archive/unarchive action
  - [x] add "show archived" checkbox
  - [x] hide expired from applicants
  - [x] hide archived from applicants
  - disabled catalogue items already hidden from applicant, warning shown for drafts with disabled
- resources
  - [x] add enable/disable action
  - [x] add archive/unarchive action
  - [x]  add "show archived" checkbox
- forms
  - [x] add enable/disable action
  - [x] add archive/unarchive action
  - [x]  add "show archived" checkbox
- workflows
  - [x] add enable/disable action
  - [x] add archive/unarchive action
  - [x]  add "show archived" checkbox
- licenses
  - [x] add enable/disable action
  - [x] add archive/unarchive action
  - [x]  add "show archived" checkbox
- [x] item cannot be archived if it's part of some non-archived catalogue item
  - [x] better error messages (the conflicting items are not included in the UI error at the moment)
- [x] archived item (resource, WF, form) must not be selectable during catalogue item creation (filter autocomplete)
- [x] expired items (after end date) should also not be available anymore like archived
- [x] expired items should be hidden unless archived checkbox is selected
- [x] in progress applications: show warning about disabled, archived and expired items
- catalogue item, resource or workflow can not be unarchived if its parts are archived (split out to issue #1094)

Macroz:
Blocked by #805

luontola:
It was discussed that the database tables will need four columns: start (timestamp), end (timestamp), enabled (boolean), archived (boolean). On the user pages will be shown those with start<now<end & enabled=true & archived=false. On the admin pages will be shown all except archived which are hidden by default and can be shown at the click of a button/checkbox.

opqdonut:
after discussion, decided to implement this so that:

- catalogue items, forms, resources, workflows and licenses can be enabled/disabled independently
- disabling a catalogue item means it can't be applied for
- disabling a form/resource/workflow/license means it can't be used when creating new catalogue items
- a form/resource/workflow/license can only be archived if the catalogue items using it are all archived

luontola:
How shall expiring fit into this? After the expiry date, should the thing behave similar to disabled or archived or still a third way?

[...]

PR #960, opened 2019-02-22
Archiving: expose status flags from API

luontola:
#806

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -65,7 +65,9 @@ SELECT
   organization,
   resid,
   start,
-  endt
+  endt,
+  enabled,
+  archived
 FROM resource
 
 -- :name get-resource :? :1
@@ -76,7 +78,9 @@ SELECT
   organization,
   resid,
   start,
-  endt
+  endt,
+  enabled,
+  archived
 FROM resource
 WHERE id = :id
 
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -538,12 +538,12 @@ WHERE textvalues.catAppId = :application
 -- - Gets application licenses by workflow and catalogue item ids
 -- - :wfid workflow id for workflow licenses
 -- - :items vector of catalogue item ids for resource licenses
-SELECT lic.id, lic.title, lic.type, lic.textcontent, wl.start, wl.endt
+SELECT lic.id, lic.title, lic.type, lic.textcontent, wl.start, wl.endt, lic.enabled, lic.archived
 FROM license lic
 INNER JOIN workflow_licenses wl ON lic.id = wl.licid
 WHERE wl.wfid = :wfid
 UNION
-SELECT lic.id, lic.title, lic.type, lic.textcontent, rl.start, rl.endt
+SELECT lic.id, lic.title, lic.type, lic.textcontent, rl.start, rl.endt, lic.enabled, lic.archived
 FROM license lic
 INNER JOIN resource_licenses rl ON lic.id = rl.licid
 INNER JOIN catalogue_item item ON (item.resid = rl.resid)
@@ -551,17 +551,17 @@ WHERE item.id IN (:v*:items)
 ORDER BY id
 
 -- :name get-resource-licenses :? :*
-SELECT lic.id, lic.title, lic.type, lic.textcontent, rl.start, rl.endt
+SELECT lic.id, lic.title, lic.type, lic.textcontent, rl.start, rl.endt, lic.enabled, lic.archived
 FROM license lic
 INNER JOIN resource_licenses rl ON lic.id = rl.licid
 WHERE rl.resid = :id
 
 -- :name get-all-licenses :? :*
-SELECT lic.id, lic.title, lic.type, lic.textcontent, lic.start, lic.endt, lic.attachmentid
+SELECT lic.id, lic.title, lic.type, lic.textcontent, lic.start, lic.endt, lic.enabled, lic.archived, lic.attachmentid
 FROM license lic
 
 -- :name get-license :? :1
-SELECT lic.id, lic.title, lic.type, lic.textcontent, lic.start, lic.endt, lic.attachmentid
+SELECT lic.id, lic.title, lic.type, lic.textcontent, lic.start, lic.endt, lic.enabled, lic.archived, lic.attachmentid
 , TRUE AS active -- TODO implement active and archiving
 FROM license lic
 WHERE lic.id = :id
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -25,7 +25,6 @@ WHERE 1=1
   AND res.resid = :resource
 /*~ ) ~*/
 
-
 -- :name get-catalogue-item :? :1
 SELECT ci.id, ci.title, res.resid, ci.wfid, ci.formid, ci.start, ci.endt as "end", ci.enabled, ci.archived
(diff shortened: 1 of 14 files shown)


test/clj/rems/test/api/licenses.clj
```diff
@@ -26,7 +26,7 @@
                      assert-response-is-ok
                      read-body)]
         (is (coll-is-not-empty? data))
-        (is (= #{:id :start :end :licensetype :title :textcontent :localizations :attachment-id} (set (keys (first data)))))))
+        (is (:id (first data)))))
```
Macroz: Why not actually test here what the API returns?

test/clj/rems/test/api/resources.clj
```diff
@@ -12,14 +12,13 @@
...
     (testing "get"
-      ;; just a basic smoke test for now
       (let [data (-> (request :get "/api/resources")
                      (authenticate api-key user-id)
                      app
                      assert-response-is-ok
                      read-body)]
         (is (coll-is-not-empty? data))
-        (is (= #{:id :owneruserid :modifieruserid :organization :resid :start :end :active :licenses} (set (keys (first data)))))))
+        (is (:id (first data)))))
```
Macroz: See above about testing the API.

src/clj/rems/db/catalogue.clj
```diff
@@ -22,7 +22,10 @@
   "Associates localisations into a catalogue item from
   the preloaded state."
   [item]
-  (assoc item :localizations ((cached :localizations) (:id item))))
+  (-> item
+      (assoc :localizations ((cached :localizations) (:id item))
+             :end (:endt item))
```
Macroz: We could rename the SQL column to `end` so we don't need this?

src/clj/rems/db/resource.clj
```diff
@@ -1,9 +1,7 @@
 (ns rems.db.resource
   (:require [rems.db.core :as db]
-            [rems.InvalidRequestException]
             [rems.util :refer [getx-user-id]])
-  (:import (org.postgresql.util PSQLException)
-           (rems InvalidRequestException)))
+  (:import (org.postgresql.util PSQLException)))
```
Macroz: I prefer using `[]` inside the import like in requires.

test/clj/rems/test/api/licenses.clj
luontola: It is already tested by the schema validation. This test only duplicates that (and does it less thoroughly). It’s enough to check the precense of one field to know that the API is not returning a schema validation error or other junk.

src/clj/rems/db/catalogue.clj
luontola: Should endt be renamed to end in all tables? It’s in quite many places.

luontola: End is a reserved word 
https://www.postgresql.org/docs/8.1/sql-keywords-appendix.html

src/clj/rems/db/resource.clj
luontola: It's not worth the effort to change. The IDE adds imports automatically and the established convention in this project is to use `()` - there are 23 instances of `(:import (` and only 3 instances of `(:import [`.

src/clj/rems/db/catalogue.clj
Macroz: Yes it is but there is no need to avoid it . You can write `"end"`

Macroz: The rename of endt can be postponed to the cleaning phase next week if you add a task for it.

src/clj/rems/db/resource.clj
Macroz: Then I will do the change to all of them myself. They are obviously someone else's results :)

test/clj/rems/test/api/licenses.clj
Macroz: In general there is value to overlapping test coverage as long as it's not too tedious to maintain. Especially having some through the API stack (most valuable happy paths) testing which is what the API tests are mostly about.

Also why not remove all the coll-`is-not-empty?` then?

luontola: Each test has a cost - the time it takes to run it, the time it takes to understand why the test failed, and the time it takes to fix the test when it fails for reasons unrelated to what is the focus of the test. That's why tests should be written with as much care as the production code.

Removing `coll-is-not-empty?`

src/clj/rems/api/forms.clj
```diff
@@ -35,13 +37,15 @@
   [Form])
 
 (defn- format-form
-  [{:keys [id organization title start endt active?]}]
+  [{:keys [id organization title start endt active? enabled archived]}]
```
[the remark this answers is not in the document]
luontola: Adding the question marks.

src/clj/rems/db/catalogue.clj
luontola: I added a note to the ideas.

src/clj/rems/api/forms.clj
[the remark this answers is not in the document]
luontola: On a second thought, removing the question mark from `active?` is much easier.

resources/sql/queries.sql
```diff
@@ -3,7 +3,7 @@
 -- - Get catalogue items
 -- - :items vector of item ids
 -- - :resource resource id to fetch items for
-SELECT ci.id, ci.title, res.resid, ci.wfid, ci.formid, ci.start, ci.endt, ci.enabled, ci.archived
+SELECT ci.id, ci.title, res.resid, ci.wfid, ci.formid, ci.start, ci.endt as "end", ci.enabled, ci.archived
```
Macroz: This does not clean up the database however? Was it your intention?

src/clj/rems/api/forms.clj
```diff
@@ -37,13 +37,13 @@
   [Form])
 
 (defn- format-form
-  [{:keys [id organization title start endt active? enabled archived]}]
+  [{:keys [id organization title start endt active enabled archived]}]
```
Macroz: There's still endt here?

resources/sql/queries.sql
luontola: It's an old field. Better to rename it physically in a separate PR. It smells like an invasive operation.

src/clj/rems/api/forms.clj
luontola: And in all other old code. It's a very invasive refactoring, so better do it in a separate PR. There's a note for it in the ideas backlog.

src/clj/rems/api/licenses.clj
```diff
@@ -48,7 +48,7 @@
       :roles #{:owner}
       :query-params [{active :- (describe s/Bool "filter active or inactive licenses") nil}]
       :return Licenses
-      (ok (licenses/get-all-licenses (when-not (nil? active) {:active? active}))))
+      (ok (licenses/get-all-licenses (when-not (nil? active) {:active active}))))
```
Macroz: You can simply write this?
```
(ok (licenses/get-all-licenses (when active {:active active}))))
```

resources/sql/queries.sql
Macroz: Yes that's why I suggested a separate task for it.

src/clj/rems/api/licenses.clj
luontola: Done.

PR #961, opened 2019-02-22
Improve logging

luontola:
Adds information about the current user and request to [Mapped Diagnostic Context](https://logback.qos.ch/manual/mdc.html). The log will then look like this:

```
2019-02-22 16:28:34,760 [main - - -] INFO  rems.xyz - not in a request
2019-02-22 16:29:03,204 [qtp374065596-26 - GET /Shibboleth.sso/Login] INFO  rems.xyz - non-authenticated user
2019-02-22 16:29:19,343 [qtp374065596-23 developer GET /api/applications/7] INFO  rems.xyz - authenticated user
```

### DIFF
--- a/resources/logback.xml
+++ b/resources/logback.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<configuration>
+<configuration scan="true" scanPeriod="15 seconds">
     <statusListener class="ch.qos.logback.core.status.NopStatusListener"/>
 
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
             <charset>UTF-8</charset>
-            <pattern>%date{ISO8601} [%thread] %-5level %logger{36} - %msg %n</pattern>
+            <pattern>%date{ISO8601} [%thread %mdc{user:--} %mdc{request-method:--} %mdc{request-uri:--}] %-5level %logger{36} - %msg%n</pattern>
         </encoder>
     </appender>
 
--- a/src/clj/rems/logging.clj
+++ b/src/clj/rems/logging.clj
@@ -11,13 +11,13 @@
                        (str v)]))
                (into {}))]
     (try
-      (doseq [[key val] m]
-        (when-not (empty? val)
-          (MDC/put key val)))
+      (doseq [[k v] m]
+        (when (seq v)
+          (MDC/put k v)))
       (f)
       (finally
-        (doseq [key (keys m)]
-          (MDC/remove key))))))
+        (doseq [k (keys m)]
+          (MDC/remove k))))))
 
 (defmacro with-mdc
   "Adds a map of values to SLF4J's MDC for logging context."
--- a/src/clj/rems/logging.clj
+++ b/src/clj/rems/logging.clj
@@ -13,7 +13,7 @@
     (try
       (doseq [[key val] m]
         (when-not (empty? val)
-          (MDC/put key (str val))))
+          (MDC/put key val)))
       (f)
       (finally
         (doseq [key (keys m)]
--- a/src/clj/rems/logging.clj
+++ b/src/clj/rems/logging.clj
@@ -4,11 +4,11 @@
 
 (defn with-mdc* [m f]
   (let [m (->> m
-               (map (fn [[key val]]
-                      [(if (keyword? key)
-                         (name key)
-                         key)
-                       (str val)]))
+               (map (fn [[k v]]
+                      [(if (keyword? k)
+                         (name k)
+                         k)
+                       (str v)]))
                (into {}))]
     (try
       (doseq [[key val] m]
--- a/src/clj/rems/logging.clj
+++ b/src/clj/rems/logging.clj
@@ -1,0 +1,58 @@
+(ns rems.logging
+  (:require [clojure.test :refer [deftest is testing]])
+  (:import (org.slf4j MDC)))
+
+(defn with-mdc* [m f]
+  (let [m (->> m
+               (map (fn [[key val]]
+                      [(if (keyword? key)
+                         (name key)
+                         key)
+                       (str val)]))
+               (into {}))]
+    (try
+      (doseq [[key val] m]
+        (when-not (empty? val)
+          (MDC/put key (str val))))
+      (f)
+      (finally
+        (doseq [key (keys m)]
+          (MDC/remove key))))))
+
+(defmacro with-mdc
+  "Adds a map of values to SLF4J's MDC for logging context."
+  [m & body]
+  `(with-mdc* ~m (fn [] ~@body)))
+
+(deftest test-with-mdc
+  (testing "adds keys to context"
+    (with-mdc {"foo" "bar"}
+      (is (= "bar" (MDC/get "foo")))))
+  (testing "removes keys from context afterwards"
+    (with-mdc {"foo" "bar"})
+    (is (nil? (MDC/get "foo"))))
+  (testing "can add multiple keys at once"
+    (with-mdc {"foo" "one"
+               "bar" "two"}
+      (is (= "one" (MDC/get "foo")))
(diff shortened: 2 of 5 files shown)


resources/logback.xml
```diff
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<configuration>
+<configuration scan="true" scanPeriod="15 seconds">
```
Macroz: Is this a good value for production? Does it need to be done so often?

src/clj/rems/logging.clj
```diff
@@ -0,0 +1,58 @@
...
+  (let [m (->> m
+               (map (fn [[key val]]
+                      [(if (keyword? key)
+                         (name key)
+                         key)
+                       (str val)]))
+               (into {}))]
+    (try
+      (doseq [[key val] m]
+        (when-not (empty? val)
```
Macroz: If `val` is always a collection then `(seq val)` is correct.

In Clojure `k v` are usually preferred (also because `key` and `val` are both useful functions).

src/clj/rems/logging.clj
```diff
@@ -0,0 +1,58 @@
+(ns rems.logging
+  (:require [clojure.test :refer [deftest is testing]])
+  (:import (org.slf4j MDC)))
+
+(defn with-mdc* [m f]
```
Macroz: I think this `ns` or this function could use more documentation.

src/clj/rems/logging.clj
```diff
@@ -0,0 +1,58 @@
...
+  (testing "can add multiple keys at once"
+    (with-mdc {"foo" "one"
+               "bar" "two"}
+      (is (= "one" (MDC/get "foo")))
+      (is (= "two" (MDC/get "bar"))))
+    (is (empty? (MDC/getCopyOfContextMap))))
+  (testing "converts keyword keys to string"
+    (with-mdc {:foo "bar"}
+      (is (= "bar" (MDC/get "foo")))))
+  (testing "converts non-string values to string"
```
Macroz: Here you could also use `are` for the multiple tests.

src/clj/rems/middleware.clj
```diff
@@ -81,7 +83,8 @@
               context/*roles* (when context/*user*
                                 (set/union (roles/get-roles (getx-user-id))
                                            (dynamic-roles/get-roles (getx-user-id))))]
-      (handler request))))
+      (with-mdc {:roles (str/join " " (sort context/*roles*))}
```
Macroz: Roles are not logged?

src/clj/rems/logging.clj
```diff
@@ -0,0 +1,58 @@
+(ns rems.logging
+  (:require [clojure.test :refer [deftest is testing]])
+  (:import (org.slf4j MDC)))
```
Macroz: I prefer using `[]` here just like in the require.

resources/logback.xml
luontola: It's not used just in production - it's also used when starting the app with `lein run`. Thus it needs to be low enough to not slow down the developer. 5 seconds would probably be better; 15 seconds is quite long time to wait when you're debugging something.

The scan only checks the `java.io.File#lastModified` of the configuration files (see `ch.qos.logback.core.joran.spi.ConfigurationWatchList#changeDetected`), so it's a cheap operation.

src/clj/rems/logging.clj
luontola: There's thorough documentation on lines 27-58; the tests.

src/clj/rems/middleware.clj
luontola: There are usually lots of them, so they are not logged in by default to avoid log clutter. If they are needed for debugging something, you can add them to logback config and thanks to the periodic scan, the config changes are applied without restarting.

src/clj/rems/logging.clj
luontola: Using a template expression for just two asserts seems very overkill. It took me over a minute to read and understand that what `are` does.

src/clj/rems/logging.clj
Macroz: It would be good to document the namespace still about the intention of the file.

src/clj/rems/logging.clj
Macroz: There are at least three here alone and I'm sure you can think of even more if you look at it.

Macroz: Honestly this test looks like word porridge to me now.

luontola: In this project there are 825 usages of `is` and only 1 usage of `are` (in `rems.test.db.applications/test-handling-event?`). That's a rather obscure operation. Complicating these simple asserts with that would reduce the test readability considerably.

Macroz: `are` is a standard Clojure test structure and not at all obscure. This is something to what I would write

```
(deftest test-with-mdc
  (with-mdc {"foo" "one"
             "bar" "two"}
    (is (= "one" (MDC/get "foo")))
    (is (= "two" (MDC/get "bar"))))

  (is (empty? (MDC/getCopyOfContextMap)) "empty outside sexp")
  (is (nil? (MDC/get "foo")) "key does not exist outside sexp")

  (testing "converts values to string"
    (are [expected x] (= expected (with-mdc {"foo" x} (MDC/get "foo")))
      ":bar" :bar
      "123" 123
      "true" true))

  (with-mdc {"foo" nil} (is (empty? (MDC/getCopyOfContextMap)) "nil should be ignored"))
  (with-mdc {"foo" ""} (is (empty? (MDC/getCopyOfContextMap)) "empty string should be ignored"))
  (with-mdc {"foo" false} (is (= "false" (MDC/get "foo")) "false should not be ignored")))
```

Macroz: Here's another variant

```
(deftest test-with-mdc
  (with-mdc {"foo" "one"
             "bar" "two"}
    (is (= "one" (MDC/get "foo")))
    (is (= "two" (MDC/get "bar")))
    (is (nil? (MDC/get "foobar")) "non-existant key does not exist"))

  (is (nil? (MDC/get "foo")) "key does not exist outside sexp")

  (are [expected x] (= expected (with-mdc {"foo" x} (MDC/get "foo")))
    ":bar" :bar
    "123" 123
    "false" false
    nil nil
    nil ""))
```

Macroz:
I think this PR is ready now but I'll let you refactor the tests before merge if you see what I mean.

src/clj/rems/logging.clj
foxlynx: Does one still get with one what row the fail was, eg. if there are 5 conditions in are-clause, does one get accurate information which one of them failed? If yes, then I think are is ok.

luontola: For the string conversion case `are` makes sense. I'll refactor that.

luontola: Actually, when there are multiple failures, `are` reports them all but it does not tell that which of the test cases had the problem. In the following example all the assertion failures refer to line 54, which is the line where `are` starts, but the test cases are on lines 56 and 57. 

Especially when there are many test cases, it becomes harder to know what caused the failure. The failure message shows only the expected and actual values, but not the input value, so when there are multiple input values which produce the same output, it's not possible to know which of them failed unless you comment out the other test cases and try them one by one.
```clj
FAIL in (test-with-mdc) (logging.clj:54)
converts non-string values to string
expected: ":bar1"
  actual: ":bar"
    diff: - ":bar1"
          + ":bar"

FAIL in (test-with-mdc) (logging.clj:54)
converts non-string values to string
expected: "123x"
  actual: "123"
    diff: - "123x"
          + "123"
```

foxlynx: That was my worry, so it would seem more practical to use if?

luontola: Yes, seems like it.

PR #963, opened 2019-02-25
799 Make it possible to theme alert-primary, secondary, light and dark

### LINKED ISSUE #799: Themeable alert colors
- [x] make it possible to theme all the Bootstrap alert colors
- [x] for example add warning colour to theme (#ffa64d for LBR)
- [x] check all other alert colors (danger, success, info)

Kauko:
Fixes #799


hukka:
Related to #799

src/clj/rems/css/styles.clj
```diff
@@ -272,21 +272,51 @@
...
+   [:.alert-warning {:color (util/get-theme-attribute :warning-color)
+                     :background-color (util/get-theme-attribute :warning-bgcolor)}]
+   [:.alert-info
+    (s/descendant :.state-info :.phases :.phase.completed)
+    {:color (util/get-theme-attribute :info-color)
+     :background-color (util/get-theme-attribute :info-bgcolor)}]
+   [:alert-light {:color (util/get-theme-attribute :alert-light-bgcolor)
+                  :background-color (util/get-theme-attribute :alert-light-bgcolor)}]
+   [:alert-dark {:color (util/get-theme-attribute :alert-dark-bgcolor)
+                 :background-color (util/get-theme-attribute :alert-dark-bgcolor)}]
```
luontola: The color and background-color are the same in most of these.

src/clj/rems/css/styles.clj
```diff
@@ -272,21 +272,51 @@
                      :color "#ccc"}
                     [:&:hover {:color (util/get-theme-attribute :color4)
                                :background-color "#eee"}]]]
-   [:.alert-info
-    (s/descendant :.state-info :.phases :.phase.completed)
-    {:color (util/get-theme-attribute :info-color)
-     :background-color (util/get-theme-attribute :info-bgcolor)}]
+
+   [:text-primary {:color (util/get-theme-attribute :text-primary :alert-primary-color)}]
```
luontola: `:text-primary` refers to the `<text-primary>` element which doesn't exists. `:.text-primary` would refer to the class.

Macroz:
Definitely should be visible in `guide_page.cljs` alerts.

Also what about the already existing (from `example-theme/theme.edn`):
```
 :info-bgcolor "#d9edf7"
 :danger-bgcolor "#f2dede"
 :success-bgcolor "#dff0d8"
 :warning-bgcolor "#fcf8e3"
 :info-color "#ff0000"
 :danger-color "#a94442"
 :success-color "#3c763d"
 :warning-color "#8a6d3b"
```

Kauko:
Updated and tested themes from rems-deploy https://github.com/CSCfi/rems-deploy/pull/15

src/clj/rems/css/styles.clj
```diff
@@ -272,21 +272,59 @@
...
+   [:.bg-danger {:background-color (util/get-theme-attribute :bg-danger)}]
+   [:.bg-warning {:background-color (util/get-theme-attribute :bg-warning)}]
+   [:.bg-info {:background-color (util/get-theme-attribute :bg-info)}]
+   [:.bg-light {:background-color (util/get-theme-attribute :bg-light)}]
+   [:.bg-dark {:background-color (util/get-theme-attribute :bg-dark)}]
+   [:.bg-white {:background-color (util/get-theme-attribute :bg-white)}]
+
+   [:.alert-primary {:color (util/get-theme-attribute :alert-primary-color)
+                     :background-color (util/get-theme-attribute :alert-primary-bgcolor)
+                     :border-color (util/get-theme-attribute :alert-primary-bordercolor :alert-primary-color)}]
```
luontola: If the theme attributes are not defined, this will produce CSS about which Chrome gives the error "Invalid property value":

devtools

Kauko: I noticed that too, but does it really matter? At least Chrome seems to handle that just fine.

We could wrap everything in `styles.clj` with lets and conditionals, but I'm not sure it's worth it?

luontola: A little function which removes all keys with nil values could do it:
```
   [:.alert-primary (remove-nils {:color (util/get-theme-attribute :alert-primary-color)
                                  :background-color (util/get-theme-attribute :alert-primary-bgcolor)
                                  :border-color (util/get-theme-attribute :alert-primary-bordercolor :alert-primary-color)})]
```

Macroz: Put a default for everything or don't put nil?

Kauko: Esko has a good point, I'll just remove nil values.

Kauko:
```
(when-let [style (remove-nil-vals {:color (util/get-theme-attribute :alert-dark-color)
                                      :background-color (util/get-theme-attribute :alert-dark-bgcolor)
                                      :border-color (util/get-theme-attribute :alert-dark-bordercolor :alert-dark-color)})]
     [:.alert-dark style])
```
I wrapped the new classes like this. Seems to work.

I actually implemented  a recursive function that walks through the whole list of styles and removes nil values. I got it working (I think), but in the end I decided that it's too complicated, and thus prone to bugs. While this approach is quite verbose, it's at least clear what is being done.

src/clj/rems/css/styles.clj
```diff
@@ -272,21 +292,60 @@
...
-     :background-color (util/get-theme-attribute :success-bgcolor)}]
-   [:.alert-warning {:color (util/get-theme-attribute :warning-color)
-                     :background-color (util/get-theme-attribute :warning-bgcolor)}]
-   [:.alert-danger
-    :.state-rejected
-    (s/descendant :.state-rejected :.phases :.phase.completed)
-    {:color (util/get-theme-attribute :danger-color)
-     :background-color (util/get-theme-attribute :danger-bgcolor)}]
+
+   (when-let [style (remove-nil-vals {:color (util/get-theme-attribute :text-primary)})] [:.text-primary style])
```
Macroz: This unfortunately complicates reading the styles a lot. I think mostly having a default value for the themeable values would've been better. Btw. It can be "default" or "inherit" too!

Macroz:
```
[:.text-primary {:color (util/get-theme-attribute :text-primary "default")}]
```

Macroz: Also you could extract the whole theme getter call to the beginning of the file and refer to it as `text-primary-color`.

Macroz: I can do this refactoring later if you want this merged soon.

luontola: The `when-let` stuff can be avoided by just returning `{}` from `remove-nil-vals` if all properties are removed.

luontola: Or actually it doesn't matter whether it returns `{}` or `nil`. The CSS minifier removes the declaration automatically when it contains no properties.

src/clj/rems/css/styles.clj
```diff
@@ -202,6 +204,65 @@
...
+    (record? obj) obj
+    (map? obj) (let [m obj
+                     m (->> m
+                            (map-vals remove-nil-vals)
+                            (remove-vals nil?))]
+                 (when (seq m)
+                   m))
+    (sequential? obj) (let [coll obj]
+                        (map remove-nil-vals coll))
+    :default obj))
```
Macroz: This should always be `:else`.

src/clj/rems/css/styles.clj
```diff
@@ -202,6 +204,65 @@
...
+  [obj]
+  (cond
+    (record? obj) obj
+    (map? obj) (let [m obj
+                     m (->> m
+                            (map-vals remove-nil-vals)
+                            (remove-vals nil?))]
+                 (when (seq m)
+                   m))
+    (sequential? obj) (let [coll obj]
```
Macroz: Doesn't seem to make sense to rename here?

src/clj/rems/css/styles.clj
```diff
@@ -202,6 +204,65 @@
                                   :margin-top 0
                                   :margin-bottom (u/px 16)})
 
+(defn- remove-nil-vals
+  "Recursively removes all keys with nil values from a map."
+  [obj]
+  (cond
+    (record? obj) obj
+    (map? obj) (let [m obj
```
Macroz: Doesn't seem to make sense to rename here?

src/clj/rems/css/styles.clj
```diff
@@ -202,6 +204,65 @@
...
+    (is (= nil
+           (remove-nil-vals {}))))
+  (testing "flat"
+    (is (= nil
+           (remove-nil-vals {:a nil})))
+    (is (= {:a 1}
+           (remove-nil-vals {:a 1})))
+    (is (= {:a false}
+           (remove-nil-vals {:a false})))
+    (is (= {:a ""}
```
Macroz: Maybe empty strings should be considered as nils because they make no sense in the resulting CSS?

luontola: Maybe. We shouldn't have any empty strings right now, so it's safer to not touch them. If some day empty strings appear somewhere, it might blow up and produce confusing bugs if they are removed from places where we don't want to remove them.

Macroz: Then make it blow up if it meets empty strings?

src/clj/rems/css/styles.clj
```diff
@@ -202,6 +204,69 @@
                                   :margin-top 0
                                   :margin-bottom (u/px 16)})
 
+(defn- remove-nil-vals
+  "Recursively removes all keys with nil values from a map."
+  [obj]
+  (assert (not= "" obj))
+  (cond
+    (record? obj) obj
+    (map? obj) (let [m (->> obj
```
Macroz: This could now be simplified further?
```
  (map? obj) (->> obj
                  (map-vals remove-nil-vals)
                  (remove-vals nil?)
                  not-empty)
```

PR #964, opened 2019-02-25
Add db reset to standalone

foxlynx:


src/clj/rems/standalone.clj
```diff
@@ -67,6 +68,14 @@
     (do
       (mount/start #'rems.config/env)
       (migrations/migrate args (select-keys env [:database-url])))
+    (= "reset" (first args))
+    (do
+      (println "\n\n*** Are you absolutely sure??? Reset empties the whole database and runs migrations to empty db.***\nType 'YES' to proceed")
+      (if (= "YES" (read-line))
```
Macroz: `when`?

src/clj/rems/standalone.clj
```diff
@@ -67,6 +68,14 @@
...
+        (do
+          (println "Running reset")
```
luontola: Also add an else branch with `(println "Cancelled")`

opqdonut:
for #989

PR #970, opened 2019-02-27
Form labels are always optional

### LINKED ISSUE #966: vTHL2: Label should never be required
The label element is not an input element so it should not be required to be filled.

image

Kauko:
Fixes #966

Previously it was possible to create a form with mandatory labels, but labels are not input elements.

Now `:optional` is set to `true`, and the checkbox is disabled, when the element is of type label.

### DIFF
--- a/src/cljs/rems/administration/create_form.cljs
+++ b/src/cljs/rems/administration/create_form.cljs
@@ -238,7 +238,7 @@
                                          {:value "multiselect", :label (text :t.create-form/type-multiselect)}
                                          {:value "date", :label (text :t.create-form/type-date)}
                                          {:value "attachment", :label (text :t.create-form/type-attachment)}
-                                         {:value "label" :label (text :t.create-form/type-label)}]}])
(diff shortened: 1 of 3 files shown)


src/cljs/rems/administration/create_form.cljs
```diff
@@ -235,11 +240,17 @@
...
+                                         {:value "label" :label (text :t.create-form/type-label)
+                                          :on-change #(rf/dispatch [::set-form-field [:items item-index :optional] true])}]}])
 
 (defn- form-item-optional-checkbox [item-index]
-  [checkbox context {:keys [:items item-index :optional]
-                     :label (text :t.create-form/optional)}])
+  (let [item @(rf/subscribe [::form-item item-index])]
+    (if (= "label" (:type item))
+      [checkbox context {:label (text :t.create-form/optional)
+                         :disabled? true}]
```
Kauko: I thought about showing something else here, but a disabled checkbox was really fast to implement, doesn't make the UI jump, and is pretty clear for the user (especially for adminds), so I decided to keep it.

PR #971, opened 2019-02-27
840 multiple decisions

### LINKED ISSUE #840: vTHL2: Request decision from multiple deciders
Request decision can't be sent to many receivers at the same time. Track approval by request, just like comments in #811 

- [ ] multiple deciders
- [ ] group decisions to requests like in #811 
- [ ] show clearly if it was approve or reject decision

hukka:
Fix #840 in same way as we handle multiple commenters

### DIFF
--- a/src/clj/rems/workflow/dynamic.clj
+++ b/src/clj/rems/workflow/dynamic.clj
@@ -87,12 +87,14 @@
 (s/defschema DecidedEvent
   (assoc EventBase
          :event/type (s/eq :application.event/decided)
+         :application/request-id s/Uuid
          :application/decision (s/enum :approved :rejected)
          :application/comment s/Str))
 (s/defschema DecisionRequestedEvent
   (assoc EventBase
          :event/type (s/eq :application.event/decision-requested)
-         :application/decider s/Str
+         :application/request-id s/Uuid
+         :application/deciders [s/Str]
          :application/comment s/Str))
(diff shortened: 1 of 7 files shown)


src/clj/rems/workflow/dynamic.clj
```diff
@@ -529,29 +528,33 @@
...
-           :application/decider (:decider cmd)
-           :application/comment (:comment cmd)})))
-
-(defmethod command-handler ::decide
-  [cmd _application _injections]
-  (or (when-not (contains? #{:approved :rejected} (:decision cmd))
-        {:errors [{:type :invalid-decision :decision (:decision cmd)}]})
-      (ok {:event/type :application.event/decided
-           :application/decision (:decision cmd)
+           :application/request-id (java.util.UUID/randomUUID)
```
opqdonut: so what was the rationale for request-id?

Macroz: There could be several comment requests going on at the same time?

hukka: Yes. While currently we let deciders only act on the latest decision request, this was just a guess and didn't come from users. It might just as well be that deciders can reply to multiple decision requests in any order they want to.

We'll have to see, but I thought that an explicit link would make sense, and that it's a bit lighter approach to put it in the JSON blob.

PR #972, opened 2019-02-27
Archiving: add archive/unarchive button for catalogue items

luontola:
#806

### DIFF
--- a/src/cljs/rems/administration/catalogue_items.cljs
+++ b/src/cljs/rems/administration/catalogue_items.cljs
@@ -11,14 +11,21 @@
 (rf/reg-event-fx
  ::enter-page
  (fn [{:keys [db]}]
+   {:db (assoc db ::display-archived? false)
+    :dispatch [::reload-page]}))
+
+(rf/reg-event-fx
+ ::reload-page
+ (fn [{:keys [db]}]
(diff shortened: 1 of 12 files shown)


src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -11,14 +11,21 @@
...
+   {:db (assoc db ::loading? true)
+    ::fetch-catalogue {:archived (::display-archived? db)}}))
 
-(rf/reg-fx ::fetch-catalogue (fn [_] (fetch-catalogue)))
+(rf/reg-fx
+ ::fetch-catalogue
+ (fn [{:keys [archived]}]
+   (fetch "/api/catalogue-items/" {:url-params {:expand :names
+                                                :archived archived}
+                                   :handler #(rf/dispatch [::fetch-catalogue-result %])})))
```
hukka: I think it would make more sense to make `::fetch-catalogue` an event (handler) instead of a effect handler, and skip the  `::reload-page` event altogether. So fetching would also set loading (since it's always assumed to be in db in the fetch handler), `::enter-page` (and all other places where `::reload-page` is dispatched to) would dispatch to fetch directly.

PR #982, opened 2019-03-01
Accepting invitation

Macroz:
Implements accept invitation coming through a join link from #609

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -656,4 +656,3 @@ WHERE evt.eventdata->>'invitation/token' IS NOT NULL
 /*~ (when (:appid params) */
 AND appid = :appid
 /*~ ) ~*/
->>>>>>> feat: accept invitations
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -637,3 +637,23 @@ VALUES (:apikey, :comment);
 -- :name get-api-key :? :1
 SELECT apiKey FROM api_key
 WHERE apiKey = :apikey;
+
+-- :name get-application-by-invitation-token :? :1
+SELECT app.id
+FROM catalogue_item_application app
+JOIN application_event evt ON (app.id = evt.appid)
+WHERE evt.eventdata->>'invitation/token' = :token
+
+-- :name get-invitation-tokens :? :*
+SELECT
+evt.eventdata->>'event/type' AS event,
+evt.eventdata->>'invitation/token' AS token,
+evt.eventdata->>'event/actor' AS actor,
+evt.eventdata->'application/member'->>'name' AS name,
+evt.eventdata->'application/member'->>'email' AS email
+FROM application_event evt
+WHERE evt.eventdata->>'invitation/token' IS NOT NULL
+/*~ (when (:appid params) */
+AND appid = :appid
+/*~ ) ~*/
+>>>>>>> feat: accept invitations
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -46,7 +46,10 @@
                   [:p "You agree not to introduce to REMS any information that relates to an identified or identifiable natural person."]
                   [:h3 "Acknowledgements"]
                   [:p "CSC’s work for REMS has been supported by the Ministry of Education and Culture of Finland and by Academy of Finland grants 271642 and 263164 to construct Biomedinfra, the Finnish consortium for ELIXIR, BBMRI and EATRIS ESFRI."]]]}
-  :actions {:add-member "Add member"
+  :actions {:accept-invitation "Accept invitation"
+            :accept-invitation-success [:div [:p "Joining the application was successful."] [:p "Next you will be taken to the application."]]
+            :accept-invitation-already-member [:div [:p "You are already a member."] [:p "Next you will be taken to the application."]]
+            :add-member "Add member"
             :applicant "Applicant"
             :application "Application"
             :approve "Approve"
@@ -61,6 +64,7 @@
             :decide "Decide"
             :description "Description"
             :empty "You don't have any applications to process."
+            :errors {:invalid-token [:div [:p "Joining the application failed."] [:p "Check the invitation token " [:code "%1"]]]}
             :export-entitlements "Export entitlements as csv"
             :handled-approvals "Handled applications"
             :handled-reviews "Handled reviews"
@@ -154,6 +158,7 @@
                                   :draft-saved "Modified"
                                   :member-added "Member added"
                                   :member-invited "Member invited"
+                                  :member-joined "Member joined"
                                   :member-removed "Member removed"
                                   :member-uninvited "Member uninvited"
                                   :rejected "Rejected"
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -332,9 +332,6 @@
                :administration "Administration"
                :applications "Applications"
                :catalogue "Catalogue"
-               :create-catalogue-item "Create catalogue item"
-               :create-license "Create license"
-               :create-resource "Create resource"
                :home "Home"
                :login "Login"
                :logout "Sign Out"}
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -110,6 +110,11 @@
    :application-id s/Num
    s/Keyword s/Any})
 
+(s/defschema AcceptInvitationResult
+  {:success s/Bool
+   (s/optional-key :application-id) s/Num
+   (s/optional-key :errors) [s/Any]})
+
 ;; Api implementation
 
 (defn- api-judge [{:keys [command application-id round comment actor]}]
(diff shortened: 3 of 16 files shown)


resources/sql/queries.sql
```diff
@@ -637,3 +637,23 @@ VALUES (:apikey, :comment);
...
+evt.eventdata->>'invitation/token' AS token,
+evt.eventdata->>'event/actor' AS actor,
+evt.eventdata->'application/member'->>'name' AS name,
+evt.eventdata->'application/member'->>'email' AS email
+FROM application_event evt
+WHERE evt.eventdata->>'invitation/token' IS NOT NULL
+/*~ (when (:appid params) */
+AND appid = :appid
+/*~ ) ~*/
+>>>>>>> feat: accept invitations
```
opqdonut: left-over conflict marker

resources/translations/en.edn
```diff
@@ -61,6 +64,7 @@
             :decide "Decide"
             :description "Description"
             :empty "You don't have any applications to process."
+            :errors {:invalid-token [:div [:p "Joining the application failed."] [:p "Check the invitation token " [:code "%1"]]]}
```
opqdonut: can the user really check the code somehow?

src/clj/rems/api/applications.clj
```diff
@@ -143,12 +148,13 @@
   (let [is-handler? (or (contains? (set (applications/get-handlers (:application application))) user) ; old form
                         (contains? (get-in application [:application :possible-commands]) :see-everything))] ; dynamic
     (if is-handler?
-      application
+      (update-in application [:application] dissoc :invitation-tokens)
```
opqdonut: couldn't this dissoc go outside the if?

src/clj/rems/api/applications_v2.clj
```diff
@@ -510,7 +514,8 @@
 (defn- hide-sensitive-information [application]
```
opqdonut: hmm why do we have this and the `rems.api.applicaitons/hide-sensitive-information` functions?

src/clj/rems/db/applications.clj
```diff
@@ -327,7 +327,8 @@
 
 (defn get-approvals [user-id]
   (->> (get-applications-impl-batch user-id {})
-       (filterv (partial can-approve? user-id))))
+       (filterv (partial can-approve? user-id))
+       (map #(dissoc % :invitation-tokens)))) ; TODO use same hiding in all applications lists
```
opqdonut: yes, please fix this TODO

src/clj/rems/workflow/dynamic.clj
```diff
@@ -658,12 +715,28 @@
...
+  "Like assert but throw the result with ex-info and not as string. "
+  ([x message]
+   `(when-not ~x
+      (throw (ex-info (str "Assert failed: " ~message "\n" (pr-str '~x))
+                      (merge ~message {:expression '~x}))))))
+
+
+(defmacro catch-ex
+  "Like `catch` but automatically unwraps possible exception to the result."
+  [& body]
```
opqdonut: it's not really like `catch`, it's like `try`, right?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -681,7 +754,8 @@
 (defn possible-commands
   "Returns the commands which the user is authorized to execute."
   [actor application-state]
-  (permissions/user-permissions application-state actor))
+  (-> (permissions/user-permissions application-state actor)
+      (conj ::accept-invitation)))
```
opqdonut: hmm why cant user-permissions include this?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -966,6 +1046,73 @@
...
+
+    (let [submitted (apply-events application
+                                  [{:event/type :application.event/submitted
+                                    :event/actor "applicant"}])]
+      (testing "invited member can join submitted application"
+        (is (= [{:userid "applicant"} {:userid "somebody"}]
+               (:members
+                (apply-commands application
+                                [{:type ::accept-invitation :actor "somebody" :token "very-secure"}]
+                                injections))))))))
```
opqdonut: how about an accepted application?

src/cljs/rems/actions/accept_invitation.cljs
```diff
@@ -0,0 +1,53 @@
...
+             (text type))])])
+
+(defn error-handler [response]
+  (status-modal/set-error!
+   (merge {:on-close #(dispatch! "#/catalogue")}
+          (if (:error response)
+            {:result {:error response}}
+            {:error-content (errors-to-content (:errors response))}))))
+
+(defn success-handler [response]
```
opqdonut: these functions should probably be private

src/cljs/rems/spa.cljs
```diff
@@ -237,6 +239,14 @@
      [:div.container.main-content [content]]
      [footer]]))
 
+(reg-event-fx
+ :after-translations-are-loaded
+ (fn [{:keys [db]} [_ on-loaded]]
+   (if (seq (:translations db))
+     (on-loaded)
+     (.setTimeout js/window #(dispatch [:after-translations-are-loaded on-loaded]) 100))
+   {}))
```
opqdonut: hmm seems like bit of a hack. is there really no way to wait for another event in re-frame?

src/clj/rems/api/applications_v2.clj
luontola: They are v1 and v2 versions of the API.

luontola: Though I'm confused that how `:invitation-tokens` finds its way to the v2 API so that it would need to be removed here. AFAIK, only v1 code is setting `:invitation-tokens`.

resources/translations/en.edn
Macroz: The code will be shown in the mail with the link to join. Some people like to type links by hand in the address bar. There's not much else that can be said?

src/clj/rems/api/applications.clj
Macroz: I think it does not make it any easier for the eyes like that if you try to combine the two cases. This is code about to be deleted or significantly refactored so I'd rather just leave it as it is for now.

src/clj/rems/api/applications_v2.clj
Macroz: I'm at a loss about what should work and where. Seems like v2 is full of placeholders in `event-type-specific-application-view`?

`:invitation-tokens` is set by the code in the `dynamic.clj` but if that's not used in the future and if someone will later write the correct specific views then this line is obviously not needed.

src/clj/rems/db/applications.clj
Macroz: Doesn't make sense to fix now as this should all be part of v2 including a nice `get-applications-impl-batch` replacement.

src/clj/rems/workflow/dynamic.clj
Macroz: Well it catches the exception so you don't have to. Maybe `try-catch-ex`?

src/clj/rems/workflow/dynamic.clj
Macroz: It doesn't really support a set of permissions that are always available. It can technically be moved there of course but involves rewriting what that is in the first place. Perhaps @luontola can explain how it should be. I think we agreed on this place earlier.

src/clj/rems/workflow/dynamic.clj
Macroz: I didn't intend to test all the possible states where it's possible to join. It's more significant at least that a submitted application can be joined since a draft is already checked. Should I test all the states here?

src/cljs/rems/spa.cljs
Macroz: There isn't. Similar problem is also met in the redirection (see `:landing-page-redirect!`).

resources/translations/en.edn
opqdonut: ok fair enough

src/clj/rems/api/applications.clj
opqdonut: ok

src/clj/rems/db/applications.clj
opqdonut: ok

src/clj/rems/workflow/dynamic.clj
opqdonut: ok

src/clj/rems/workflow/dynamic.clj
opqdonut: I'd like at least one negative case where the state of the application is the reason for the failure (instead of invalid code or missing invitation), but I'm not insisting.

src/cljs/rems/spa.cljs
opqdonut: ok

src/clj/rems/workflow/dynamic.clj
Macroz: Well accepting the invitation should always be possible but maybe not when the application has been closed already. That also creates a new situation that has to be shown to the joiner so maybe I'll add that one.

PR #983, opened 2019-03-01
Small fixes

Macroz:
Test by clicking and fix problems caused by recent PRs and improve consistency.

### DIFF
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -116,7 +116,7 @@
         items @(rf/subscribe [::catalogue])]
     [:div
      [:h2 (text :t.catalogue/catalogue)]
-     (if (empty? items)
+     (if (nil? items)
        [spinner/big]
        [:div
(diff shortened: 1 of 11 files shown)


src/cljs/rems/catalogue.cljs
```diff
@@ -115,17 +112,18 @@
        :items drafts}]]))
 
 (defn catalogue-page []
-  (let [language (rf/subscribe [:language])]
+  (let [language (rf/subscribe [:language])
+        items @(rf/subscribe [::catalogue])]
     [:div
      [:h2 (text :t.catalogue/catalogue)]
-     (if @(rf/subscribe [::loading-catalogue?])
+     (if (empty? items)
```
luontola: On a freshly installed system (or dev instance without test-data) this spinner will be spinning even after the catalogue has been loaded. It does not disambiguate between loading catalogue and an empty catalogue. Using `nil?` instead of `empty?` should fix that.

PR #985, opened 2019-03-05
Show latest modified in UI

### LINKED ISSUE #843: vTHL3: Show last event time for applications
At the moment there is no visibility to the applicant if something is happening to their application. Something like "latest activity" needed.

- [x] in the status area, always show a "last event" time that includes any possibly hidden events
- [x] make sure no details of events are sent to the applicant that should not be seen (could either calculate on serve or send all event times even and withholding any other details)

hukka:
Fix #843.

Looks like
image


src/cljs/rems/application.cljs
```diff
@@ -598,7 +598,8 @@
       :title [:span#application-state
               (str
                (text :t.applications/state)
-               (when state (str ": " (localize-state state))))]
+               (when state (str ": " (localize-state state)))
+               " — " (text :t.applications/last-modified) ": " (localize-time last-modified))]
```
Macroz: Why not include the whole text in the translation?
`(text-format :t.applications/last-modified (localize-time last-modified))`

resources/translations/en.edn
```diff
@@ -188,6 +188,7 @@
                           :returned "Returned"
                           :unknown "Unknown"
                           :withdrawn "Withdrawn"}
+                 :last-modified "last modified at"
```
Macroz: Maybe just `Last modified`?

Macroz:
Alternative placement for the last modified could be just showing the last event and its time. For hidden events just show the time and say "Processed" or something. This would avoid the disconnect between the last modified time and the last event being different.

luontola:
"Last modified" can mean two things: when the application's content was last modified (i.e. draft saved or submitted) or when there last was some meta activity such as commenting or adding a member. Using a different term for those two things would help to avoid the ambiguity.

PR #988, opened 2019-03-06
Reframe docs

hukka:


docs/architecture/004-reframe-style.md
```diff
@@ -0,0 +1,89 @@
...
+ (fn [{{::keys [deciders application-id comment on-finished]} :db} _]
+   (fetch "/api/applications/request-decision"
+          {:params {:deciders deciders
+                    :application-id application-id
+                    :comment comment}
+           :handler on-finished})
+   {}))
+```
+
+In particular do not mix side-effects and effect handlers in the same flow!
```
opqdonut: what does this mean?

docs/architecture/004-reframe-style.md
```diff
@@ -0,0 +1,89 @@
...
+
+```clojure
+(rf/reg-event-fx
+ ::send-request-decision
+ (fn [{{::keys [deciders application-id comment on-finished]} :db} _]
+   (fetch "/api/applications/request-decision"
+          {:params {:deciders deciders
+                    :application-id application-id
+                    :comment comment}
+           :handler on-finished})
```
opqdonut: is this really a typical example? I think a typical example would have a dispatch in the `on-finished`

docs/architecture/004-reframe-style.md
```diff
@@ -0,0 +1,89 @@
...
+   {}))
+```
+
+In particular do not mix side-effects and effect handlers in the same flow!
+
+## Handling user interaction in components
+
+We could also launch the fetch directly from the components, like buttons,
+but we will keep this level of indirection for consistency,
```
opqdonut: which level of indirection?

docs/architecture/004-reframe-style.md
```diff
@@ -0,0 +1,89 @@
...
+                    :comment comment}
+           :handler on-finished})
+   {}))
+```
+
+In particular do not mix side-effects and effect handlers in the same flow!
+
+## Handling user interaction in components
+
+We could also launch the fetch directly from the components, like buttons,
```
opqdonut: how about "directly from the components (e.g. on-click callbacks of buttons)" or something that mentions callbacks

Macroz:
I reworded some parts. @luontola can you have a look that this is clear for you also?

PR #992, opened 2019-03-08
WIP circle: use docker images

opqdonut:
... where possible

on top of #991


opqdonut:
the problem is that `dev-db.sh` uses docker to set up postgres, but I'll try to dockerize some of the other builds anyway

opqdonut:
added a comment about why docker doesn't really work for us in commit 182540d7 in pr #991

PR #997, opened 2019-03-12
Support livereload of css with Figwheel

Macroz:
Support (again) livereload of CSS with Figwheel. 

While we normally generate the `screen.css` on the fly and memoize it, in development it's useful to render a file `target/resources/public/css/screen.css` that Figwheel can serve and update. We use mount here to render the initial file and re-render each time the styles are changed and the style file reloaded. Figwheel picks up the change and delivers it to the browser.

We also render all the configured languages into `screen-fi.css` etc. files so this works for localized CSS files as well.


.gitignore
```diff
@@ -21,3 +21,4 @@
 /.rebel_readline_history
 /package-lock.json
 /browsertest-errors
+/env/dev/resources/public/css/screen.css
```
opqdonut: please put generated files under `target/`

src/clj/rems/css/styles.clj
```diff
@@ -641,9 +647,27 @@
...
-    (is (string? (rems.css.styles/screen)))))
+    (is (string? (screen-css)))))
+
+;; for development use and Figwheel updates
+(mount/defstate
+  rendered-css-file
+  :start
+  (when (env :render-css-file?)
+    (render-css-file
+     (binding [context/*lang* :en] ; default language to start Figwheel tracking
```
opqdonut: so does the css file change language when `context/*lang*` changes?

src/clj/rems/css/styles.clj
```diff
@@ -1,8 +1,11 @@
 (ns rems.css.styles
   "CSS stylesheets are generated by garden automatically when
   accessing the application on a browser. The garden styles can also
-  be manually compiled by calling the function
-  rems.css.styles/generate-css"
+  be manually compiled by calling the function `rems.css.styles/screen-css`.
+
+  For development purposes with Figwheel the styles are rendered to
+  `env/dev/resources/public/css/screen.css` whenever this ns is evaluated
+  so that Figwheel can autoreload them."
```
opqdonut: is there no other way to hook into figwheel autoreload than the filesystem?

.gitignore
Macroz: No this should not be under target. Generally Garden files are put under the resources-directory so that they are picked by Figwheel and friends. This follows the same practice but puts development files under the development resources instead.

src/clj/rems/css/styles.clj
Macroz: I think in the development flow it mostly goes so that you edit changes in `styles.clj` and reloading that file updates dynamically. I removed the rendering of the css file each time `screen.css` is requested because I thought it's not necessary but I could put it back. Now the language does not actually update.

src/clj/rems/css/styles.clj
Macroz: No

.gitignore
opqdonut: generated files should be under `target/` for e.g. cleanable builds. we have a resource directory there already `target/cljsbuild`. could easily add a `target/css` e.g.

src/clj/rems/home.clj
```diff
@@ -36,17 +38,32 @@
...
+  (GET "/favicon.ico" [] (redirect "/img/favicon.ico")))
+
+(defn- get-language-from-filename [filename]
+  (keyword (subs filename (inc (str/index-of filename "-")) (str/index-of filename "."))))
+
+(deftest test-get-language-from-filename
+ (is (= :fi (get-language-from-filename "/css/screen-fi.css"))))
+
+(defroutes css-routes
+  (GET "/css/:filename" [filename]
```
opqdonut: would it be easier if the files were e.g. `/css/fi/screen.css`?

src/clj/rems/home.clj
```diff
@@ -36,17 +38,32 @@
...
   (GET "/accept-invitation" {{:keys [token]} :params} (redirect (str "/#/application/accept-invitation/" token)))
   (GET "/apply-for" {{:keys [resource]} :params} (apply-for-resource resource))
   (GET "/landing_page" req (redirect "/#/redirect")) ; DEPRECATED: legacy url redirect
   (GET "/markdown/:filename" [filename] (markdown-page filename))
-  (GET "/favicon.ico" [] (redirect "/img/favicon.ico"))
-  (GET "/css/screen.css" [] (memoized-render-css context/*lang*)))
+  (GET "/favicon.ico" [] (redirect "/img/favicon.ico")))
+
+(defn- get-language-from-filename [filename]
+  (keyword (subs filename (inc (str/index-of filename "-")) (str/index-of filename "."))))
```
opqdonut: just a minor thing but I'd prefer a regex to two subs+index-of+index-of

Macroz: I don't for this simple case but fortunately this piece of code is not needed anymore.

PR #998, opened 2019-03-12
Archiving buttons for resources

luontola:
#806

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -44,7 +44,7 @@ LEFT OUTER JOIN application_form form ON (ci.formid = form.id)
 WHERE ci.id = :id;
 
 -- :name set-catalogue-item-state! :insert
-UPDATE catalogue_item ci
+UPDATE catalogue_item
 SET
 /*~ (when (boolean? (:enabled params)) */
   enabled = :enabled,
@@ -53,7 +53,7 @@ SET
   archived = :archived,
 /*~ ) ~*/
   id = id
-WHERE ci.id = :id;
+WHERE id = :id;
 
 -- :name create-catalogue-item! :insert
 -- :doc Create a single catalogue item
@@ -97,6 +97,18 @@ INSERT INTO resource
 VALUES (:resid, :organization, :owneruserid, :modifieruserid,
  /*~ (if (:endt params) */ :endt /*~*/ NULL /*~ ) ~*/
 );
+
+-- :name set-resource-state! :insert
+UPDATE resource
+SET
+  /*~ (when (boolean? (:enabled params)) */
+  enabled = :enabled,
+  /*~ ) ~*/
+  /*~ (when (boolean? (:archived params)) */
+  archived = :archived,
+  /*~ ) ~*/
+  id = id
+WHERE id = :id;
 
 -- :name get-database-name :? :1
 SELECT current_database();
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -101,12 +101,12 @@ VALUES (:resid, :organization, :owneruserid, :modifieruserid,
 -- :name set-resource-state! :insert
 UPDATE resource
 SET
-  /*~ (when (boolean? (:enabled params)) */
+/*~ (when (boolean? (:enabled params)) */
   enabled = :enabled,
-  /*~ ) ~*/
-  /*~ (when (boolean? (:archived params)) */
+/*~ ) ~*/
+/*~ (when (boolean? (:archived params)) */
   archived = :archived,
-  /*~ ) ~*/
+/*~ ) ~*/
   id = id
 WHERE id = :id;
 
--- a/src/clj/rems/api/resources.clj
+++ b/src/clj/rems/api/resources.clj
@@ -72,9 +72,11 @@
     (GET "/" []
       :summary "Get resources"
       :roles #{:owner}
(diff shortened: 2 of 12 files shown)


src/cljs/rems/administration/resources.cljs
```diff
@@ -45,6 +59,63 @@
    {:href (str "/#/administration/resources/" resource-id)}
    (text :t.administration/view)])
 
+
+;;; Archiving
+;; TODO: deduplicate
```
luontola: Do you have ideas that how to remove this duplication between resource and catalogue pages?

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -116,6 +120,7 @@
      [:label.form-check-label {:for "display-archived"}
       (text :t.administration/display-archived)]]))
 
+
```
Macroz: Extra newline?

src/cljs/rems/administration/resources.cljs
Macroz: re-frame is a framework that does not support this kind of reusable components well. However refactoring this into Reagent components does not really help.

One option is to extract a common enable/disable as well as an archived ns and then parametrize each handler,  subscription and button there with the local key as well. I.e. `::set-display-archived?` subscription in `administration/archived_item.clj` would accept as a parameter the item prefix (i.e. `::resource`, `::catalogue-item`) and store and retrieve in the db under the path `[::resource ::archived]` (or the reverse if it makes more sense). Same for handlers and buttons.

resources/sql/queries.sql
```diff
@@ -98,6 +98,18 @@ VALUES (:resid, :organization, :owneruserid, :modifieruserid,
  /*~ (if (:endt params) */ :endt /*~*/ NULL /*~ ) ~*/
 );
 
+-- :name set-resource-state! :insert
+UPDATE resource
+SET
+  /*~ (when (boolean? (:enabled params)) */
```
Macroz: Should be no need to check `boolean?` here?

test/clj/rems/test/db/resources.clj
```diff
@@ -0,0 +1,69 @@
...
+(defn- status-flags [res-id]
+  (-> (resources/get-resource res-id)
+      (select-keys [:enabled :archived])))
+
+(deftest test-update-catalogue-item!
+  (let [user-id "test-user"
+        res-id (:id (resources/create-resource! {:resid "res1" :organization "abc" :licenses []} user-id))
+        res-id2 (:id (resources/create-resource! {:resid "res2" :organization "abc" :licenses []} user-id))]
+
+    (testing "new resources are enabled and not archived"
```
Macroz: I would write this like
```
      (is (= {:enabled true :archived false} (status-flags res-id)) "new resources are enabled and not archived")
```

src/clj/rems/api/resources.clj
```diff
@@ -72,9 +72,11 @@
     (GET "/" []
       :summary "Get resources"
       :roles #{:owner}
-      :query-params [{active :- (describe s/Bool "filter active or inactive resources") nil}]
+      :query-params [{active :- (describe s/Bool "filter active or inactive resources") nil}
+                     {archived :- (describe s/Bool "whether to include archived resources") false}]
       :return Resources
-      (ok (get-resources (when active {:active active}))))
+      (ok (get-resources (merge (when active {:active active})
+                                (when-not archived {:archived false})))))
```
Macroz: Since you are giving archived a default value that is either `true` or `false` could you just not use the when tricks here?

src/cljs/rems/administration/catalogue_items.cljs
luontola: To separate the archiving related functions from the rest.

resources/sql/queries.sql
luontola: It's a duplicate of `set-catalogue-item-state!`, which is in tests/testdata used to change just one of the fields.

src/clj/rems/api/resources.clj
luontola: `{:archived true}` would return *only* archived resources, but we want to return *both* archived and non-archived resources.

src/cljs/rems/administration/status_flags.cljs
```diff
@@ -0,0 +1,50 @@
+(ns rems.administration.status-flags
+  (:require [re-frame.core :as rf]
+            [rems.text :refer [text]]))
+
+(defn- disable-button [update-item-cmd item]
+  [:button.btn.btn-secondary.button-min-width
+   {:type "button"
+    :on-click #(rf/dispatch [update-item-cmd (assoc item :enabled false)])}
```
Macroz: I think I'd prefer to have full blown callback functions here so and not only passing effect names.

luontola: Would it be better for the callback to receive as parameter the updated item or just the enabled boolean?

luontola: The `(if (:enabled item)` already makes `enabled-toggle` aware of the existence of `:enabled`, so I think I'll make the callback receive the updated item. Otherwise for full decoupling it would be necessary to parameterize also the current value of `:enabled`.

luontola: Done.

PR #999, opened 2019-03-12
Only test disabled-items-warnings if application is in editable state.

foxlynx:
New, updated story with more description: https://github.com/CSCfi/rems/issues/1002

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -773,7 +773,8 @@
                        (map dynamic-event->event (:dynamic-events app)))
         applicant-attributes (:applicant-attributes application)
         messages (remove nil?
-                         [(when (editable? application)(disabled-items-warning (:catalogue-items application))) ; NB: eval this here so we get nil or a warning
+                         [(when (editable? application)
+                            (disabled-items-warning (:catalogue-items application))) ; NB: eval this here so we get nil or a warning
                           (when (:validation edit-application)
                             [flash-message
                              {:status :danger
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -765,8 +765,8 @@
               (:application/comment event))})
 
 (defn- show-items-warning? [application]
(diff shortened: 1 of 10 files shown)


src/cljs/rems/application.cljs
```diff
@@ -765,7 +765,7 @@
                        (map dynamic-event->event (:dynamic-events app)))
         applicant-attributes (:applicant-attributes application)
         messages (remove nil?
-                         [(disabled-items-warning (:catalogue-items application)) ; NB: eval this here so we get nil or a warning
+                         [(when (editable? application)(disabled-items-warning (:catalogue-items application))) ; NB: eval this here so we get nil or a warning
```
opqdonut: I'd prefer a line break at `)(`, but at least add a space :)

opqdonut: should the condition actually be "when application is a draft"?

Macroz:
Why is this a solution to the problem found in migration tests?

src/cljs/rems/application.cljs
foxlynx: I thought one can somehow modify returned application still, or does it create a new draft from the returned?

opqdonut: oh right, I guess editable is the right condition then

foxlynx: We are gonna dig a bit deeper tomorrow irl how this feature should be overall.

src/cljs/rems/application.cljs
```diff
@@ -773,7 +773,8 @@
                        (map dynamic-event->event (:dynamic-events app)))
         applicant-attributes (:applicant-attributes application)
         messages (remove nil?
-                         [(disabled-items-warning (:catalogue-items application)) ; NB: eval this here so we get nil or a warning
+                         [(when (editable? application)
```
Macroz: This check should be moved inside `disabled-items-warning`. However, there are also other needs as discussed so let's specify this task a bit better together.

Macroz:
Closes #1002

PR #1000, opened 2019-03-12
873 full text search

hukka:
- [x] resolve nulls being passed into props
- [x] style tweaking
- [ ] maybe we don't want to search from all translations?
- [x] fix tests


hukka:
Works for example by searching "developer draft 2019 elfa"

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
-  (every? (fn [[col filter-value]] ()
-            (matches-filter column-definitions col filter-value item))
-          filters))
+(defn match-filter? [column-definitions filter-word item]
+  (some (fn [col]
+          (str/includes? (str/lower-case (column-filter-value column-definitions col item))
+                         (str/lower-case filter-word)))
+        (keys column-definitions)))
+
+(deftest test-match-filter
```
Macroz: This should be named `test-match-filter?`?

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
+                          :description "bar"}))
+     "doesn't match with value if filter-value is provided"))
+
+(defn match-filters? [column-definitions filters item]
+  (every? (fn [filter-word]
+            (match-filter? column-definitions filter-word item))
+          (str/split filters #"\s+")))
+
+(deftest test-match-filters
+ (is (true? (match-filters? {:id {:value :id}
```
Macroz: `true?` also not required here.

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
+                          :its-a-trap "willnotmatch"
+                          :description "bar"}))
+     "doesn't match with value if filter-value is provided"))
+
+(defn match-filters? [column-definitions filters item]
+  (every? (fn [filter-word]
+            (match-filter? column-definitions filter-word item))
+          (str/split filters #"\s+")))
+
+(deftest test-match-filters
```
Macroz: This should be named `test-match-filters?`

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
+                             :description "bar"}))
+     "match with filter-value")
+ (is (not (match-filters? {:id {:value :id}
+                           :description {:value :its-a-trap
+                                         :filter-value :description}}
+                          "foo bar"
+                          {:id "foo"
+                           :its-a-trap "bar"
+                           :description "shouldnotmatch"}))
+     "doesn't match with value if filter-value is provided")
```
Macroz: This duplicates a lot of tests with `test-match-filter` so maybe just have the tests there and test some combination logic here.

Macroz: If you move the string splitting and lower-casing outside then this just tests `every?` that is not very useful.

src/cljs/rems/table.cljs
```diff
@@ -94,20 +140,38 @@
...
+(defn- filter-view [filtering]
+  (let [{:keys [filters set-filtering]} filtering]
+    [:div
+      [:input
+       {:type "text"
+        :name "table-search"
+        :value filters
+        :placeholder ""
+        :on-input (fn [event]
+                    (set-filtering (assoc filtering :filters (-> event .-target .-value))))}]
```
Macroz: Maybe this would be a good place for preparing the filters with a helper fn that parses the input value? You could also test that separately.

src/cljs/rems/table.cljs
```diff
@@ -94,20 +140,38 @@
...
+      (when (seq filters)
+        [:div.reset-button.icon-link.fa.fa-backspace
```
Macroz: Like discussed I would just use a regular button here because now we have space. On the other hand Google does not have a specific clear button so maybe it's not required at all.

src/cljs/rems/table.cljs
```diff
@@ -94,20 +140,38 @@
...
                                 (s/optional-key :show-filters) s/Bool
                                 (s/optional-key :set-filtering) Applicable}
    :id-function Applicable
    :items [s/Any]
    (s/optional-key :class) s/Str})
 
+(defn- filter-view [filtering]
+  (let [{:keys [filters set-filtering]} filtering]
+    [:div
+      [:input
```
Macroz: This input should be made full width. Also it doesn't align well with the search button.

src/cljs/rems/table.cljs
Macroz: The state should be reset when closing the search with the search button.

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
+                         "willnotmatch"
+                         {:id "foo"
+                          :its-a-trap "willnotmatch"
+                          :description "bar"}))
+     "doesn't match with value if filter-value is provided"))
+
+(defn match-filters? [column-definitions filters item]
+  (every? (fn [filter-word]
+            (match-filter? column-definitions filter-word item))
+          (str/split filters #"\s+")))
```
Macroz: I don't think this is a good place for the `str/split`. it should be done earlier so that filters can be a seq already.

src/cljs/rems/table.cljs
```diff
@@ -63,18 +67,60 @@
...
-                   (str/lower-case filter-value))))
-
-(defn matches-filters [column-definitions filters item]
-  (every? (fn [[col filter-value]] ()
-            (matches-filter column-definitions col filter-value item))
-          filters))
+(defn match-filter? [column-definitions filter-word item]
+  (some (fn [col]
+          (str/includes? (str/lower-case (column-filter-value column-definitions col item))
+                         (str/lower-case filter-word)))
```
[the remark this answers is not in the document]
hukka: I disagree. I think this is the functions that determines the search logic. If we ever want to have a case sensitive search instead, then only this function should be modified.

src/cljs/rems/table.cljs
```diff
@@ -29,11 +31,13 @@
...
-    (if filter-value-fn
-      (filter-value-fn item)
-      (throw (js/Error. (str "No `:filter-value` or `:value` defined for column \"" col "\""))))))
+  (if (get-in column-definitions [col :filterable?] true)
+    (let [filter-value-fn (or (get-in column-definitions [col :filter-value])
+                              (get-in column-definitions [col :value]))]
+      (if filter-value-fn
+        (str (filter-value-fn item))
+        (throw (js/Error. (str "No `:filter-value` or `:value` defined for column \"" col "\"")))))
+    ""))
```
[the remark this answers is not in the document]
hukka: Disagree, it makes callers implementations simpler when they can always assume that the result is a string.

src/cljs/rems/table.cljs
hukka: Button removed. Disagree about the reset; it's useful to quickly toggle filters on and off.

src/cljs/rems/table.cljs
hukka: At most it could be moved one level up to apply-filters, which doesn't make much of a difference. Any higher and we are in representational logic that shouldn't really care how filter logic wants to interpret the filter commands.

src/cljs/rems/table.cljs
hukka: Disagree, if we save a list instead of string to the db, we would have to then generate the string again (the input is a controlled component).

src/cljs/rems/table.cljs
Macroz: I think `str/split` belongs to the representational layer since it's parsing user input and not in the logic layer here. That's why I suggest that you move the parsing up to the border where the input callback is. Internally you can store the filters in a more optimal format for search and do less processing every call.

Macroz: I think it might make sense to refactor this later anyway since there is very little value in having `:filterable?` and `:sortable?` separately compared to only specifying `:filter-value` and `:sort-value` where necessary.

luontola: `:filter-value` and `:sort-value` fall back to `:value`, so `:filterable?` and `:sortable?` exist to disable the default behavior.

Macroz: Yes but that is also suspect.

PR #1004, opened 2019-03-13
Make v2 application API feature complete (support all events)

luontola:
#852 

Now the v2 application API handles all events. Additional testing is still needed to ensure that the v1 and v2 APIs are functionally equivalent.


src/clj/rems/api/applications_v2.clj
```diff
@@ -32,13 +33,16 @@
...
              :application/workflow {:workflow/id (:workflow/id event)
                                     :workflow/type (:workflow/type event)
                                     ;; TODO: other workflows
+                                    ;; TODO: extract an event handler for dynamic workflow specific stuff
                                     :workflow.dynamic/state :rems.workflow.dynamic/draft
-                                    :workflow.dynamic/handlers (:workflow.dynamic/handlers event)})))
+                                    :workflow.dynamic/handlers (:workflow.dynamic/handlers event)
+                                    :workflow.dynamic/awaiting-commenters #{}
+                                    :workflow.dynamic/awaiting-deciders #{}
+                                    :workflow.dynamic/invitations {}})))
```
luontola: What do you think, are `awaiting-commenters`, `awaiting-deciders` and `invitations` specific to the dynamic workflow or to the application? At least the first two seem quite workflow specific to me. Invitations seem more like a generic feature, not workflow specific.

Issue #835, opened 2019-01-15
vTHL3: Preview form


foxlynx:
When creating or editing a form, there should be a possibility to preview what it looks like to the applicant. This could open a new window, modal or show beside the editor on large screens.

- [x] preview action while creating a new form #1110 
- ~preview action while editing an existing form~ no form editing is planned
- [x] preview for existing form #1113 
- [x] style adjustments: editor and preview side-by-side

The form does not need to be functioning but it should look like what it will look like in a real application.

Plan of attack:
- [x] push current form creation model as is into db (still keeping old save functionality)
- [x] make an API for returning that model
- [x] UI preview with that model
- leftover refactoring opportunities in #913

opqdonut:
`/forms/v2/:id` returns the new form template

PR #1011, opened 2019-03-18
Store form template as jsonb on the side of old relational way

hukka:


resources/sql/queries.sql
```diff
@@ -207,6 +207,36 @@ WHERE 1=1
...
+SELECT
+  id,
+  organization,
+  title,
+  start,
+  endt as "end",
+  fields::TEXT,
+  enabled,
+  archived,
+  TRUE AS active -- TODO implement active and archiving
```
luontola: `rems.db.core/assoc-active` can be used to calculate active.

resources/sql/queries.sql
```diff
@@ -207,6 +207,36 @@ WHERE 1=1
...
+FROM form_template
+WHERE id = :id;
+
+-- :name get-form-templates
+select id from form_template;
```
luontola: Dead code?

test/clj/rems/api/test_forms.clj
```diff
@@ -78,10 +78,20 @@
...
+                                  assert-response-is-ok
+                                  read-body)]
             (is (= 1 (count forms))
                 "only one form got created")
             ;; TODO: create an API for reading full forms (will be needed latest for editing forms)
+            (testing "form template matches command"
+              (is (= (select-keys command [:title :organization])
+                     (select-keys form-template [:title :organization])))
+              (is (= (:items command)
+                     (:fields form-template))))
```
luontola: Also the command's `:items` should be renamed to `:fields`. It can also be done in a different PR if it seems too big a change.

resources/sql/queries.sql
hukka: Hm, I copy pasted from other queries. I have to see if I will fix the other parts here or in other PR.

resources/sql/queries.sql
hukka: Yeah… We need a proper query to replace the old plural query eventually, but this is not it. I'll remove it.

test/clj/rems/api/test_forms.clj
hukka: Didn't want to touch the UI code yet. Will do later.

opqdonut:
for #835

PR #1013, opened 2019-03-18
Break dependency from form validation to legacy code

luontola:
#852

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -99,9 +99,7 @@
    (s/optional-key :members) [AddedMember]
    (s/optional-key :description) (s/maybe s/Str)
    (s/optional-key :workflow) s/Any
-   (s/optional-key :possible-commands) #{s/Keyword}
-   (s/optional-key :deciders) #{s/Str}
-   (s/optional-key :commenters) #{s/Str}})
+   (s/optional-key :possible-commands) #{s/Keyword}})
 
 (s/defschema Entitlement
   {:resource s/Str
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -101,10 +101,7 @@
    (s/optional-key :workflow) s/Any
    (s/optional-key :possible-commands) #{s/Keyword}
    (s/optional-key :deciders) #{s/Str}
-   (s/optional-key :commenters) #{s/Str}
-   (s/optional-key :form-contents) s/Any
-   (s/optional-key :submitted-form-contents) s/Any
-   (s/optional-key :previous-submitted-form-contents) s/Any})
+   (s/optional-key :commenters) #{s/Str}})
 
 (s/defschema Entitlement
   {:resource s/Str
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -329,7 +329,9 @@
           :application/licenses
           :form-contents
           :submitted-form-contents
-          :previous-submitted-form-contents))
+          :previous-submitted-form-contents
+          :deciders
+          :commenters))
 
 (defn get-approvals [user-id]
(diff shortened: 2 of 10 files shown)


src/clj/rems/db/applications.clj
```diff
@@ -1018,15 +1022,15 @@
 (defn- valid-user? [userid]
   (not (nil? (users/get-user-attributes userid))))
 
-(defn- validate-form [application-id]
-  (let [user-id (get-applicant-of-application application-id)
-        validation (form-validation/validate (get-form-for user-id application-id))]
-    (when-not (= :valid validation)
-      validation)))
+(defn- get-form [form-id]
+  (-> ((resolve 'rems.db.form/get-form) form-id) ;; XXX: avoid cyclic dependency
```
opqdonut: this is bad. do you have a plan for getting rid of it right after this PR?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -544,16 +547,24 @@
...
+  (let [form (get-form (:form/id application))
+        _ (assert form)
+        answers (:form-contents application)
+        legacy-form {:items (for [field (:items form)]
+                              (assoc field :value (get-in answers [:items (:id field)])))
+                     :licenses (for [license (:application/licenses application)]
+                                 {:id (:license/id license)
+                                  :approved (= "approved"
+                                               (get-in answers [:licenses (:license/id license)]))})}
+        result (form-validation/validate legacy-form)]
```
opqdonut: hmm, it still might be better design to inject the whole validate-form function, but have this as its implementation

src/clj/rems/api/schema.clj
```diff
@@ -101,10 +101,7 @@
    (s/optional-key :workflow) s/Any
    (s/optional-key :possible-commands) #{s/Keyword}
    (s/optional-key :deciders) #{s/Str}
-   (s/optional-key :commenters) #{s/Str}
-   (s/optional-key :form-contents) s/Any
-   (s/optional-key :submitted-form-contents) s/Any
-   (s/optional-key :previous-submitted-form-contents) s/Any})
+   (s/optional-key :commenters) #{s/Str}})
```
opqdonut: so where will the form contents come from? is this refactor here to make room for the new `:answers` api?

luontola: They come from here: https://github.com/CSCfi/rems/blob/4d46dc7de71651b120da4a3de31ca42d428cd68b/src/clj/rems/api/schema.clj#L56-L57

The `:form-contents` et al. are never used in the UI. They are only used to calculate the field's `:value` and `:previous-value`.

src/clj/rems/workflow/dynamic.clj
luontola: The form validation is a pure function, so there is little need to inject it.

src/clj/rems/api/schema.clj
opqdonut: oh right, they're one level up on the application api

src/clj/rems/db/applications.clj
luontola: `rems.db.form` depends on `rems.db.applications`. `rems.db.applications/process-item` is applications ns even though it deals mostly with form fields. After removing the round-based workflow this will be easier to fix.

luontola: The new form API might also fix this.

src/clj/rems/workflow/dynamic.clj
luontola: I'll extract it to make workflows unaware of form details.

src/clj/rems/db/applications.clj
opqdonut: I'd just move proces-item to rems.db.form

luontola: 👌

Issue #653, opened 2018-08-28
vTHL2: Separate license acceptance for members out of application


opqdonut:
See also #609 

When an application has multiple members, the other members must each separately accept the licenses before being granted an entitlement. Also they must then be REMS users so they have come through a join link.

- [x] application can be submitted and processed without license acceptance of additional members
- [x] entitlements are granted only after member has accepted licences 
- [x] separate variant of application page where the member can accept the licenses, but cannot modify anything else
- [x] show members that have accepted licenses in handler's view
- [x] email to member with a link to join REMS where there is a challenge token so that the invited member can join
- [x] show user attrs of joined member in the application after user was created and logged in

PR #1016, opened 2019-03-20
Member acceptance

Macroz:
Small fixes then prepare for #653 by tracking accepted licenses per member.

### DIFF
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -155,7 +155,7 @@
                                   :created "Luotu"
                                   :decided "Päätetty"
                                   :decision-requested "Pyydetty päätöstä"
-                                  :draft-saved "Talletettu"
+                                  :draft-saved "Tallennettu"
                                   :member-added "Lisätty jäsen"
                                   :member-invited "Kutsuttu jäsen"
                                   :member-joined "Jäsen liittyi"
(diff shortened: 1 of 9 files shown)


resources/translations/fi.edn
```diff
@@ -155,7 +155,7 @@
                                   :created "Luotu"
                                   :decided "Päätetty"
                                   :decision-requested "Pyydetty päätöstä"
-                                  :draft-saved "Muokattu"
+                                  :draft-saved "Talletettu"
```
opqdonut: Tallennettu?

src/clj/rems/api/applications_v2.clj
```diff
@@ -54,7 +55,8 @@
   (-> application
       (assoc :application/modified (:event/time event))
       (assoc ::draft-answers (:application/field-values event))
-      (update :application/licenses set-accepted-licences (:application/accepted-licenses event))))
+      (update :application/licenses set-accepted-licences (:application/accepted-licenses event))
+      (update :application/accepted-licenses into (map :license/id (:application/accepted-licenses event)))))
```
luontola: This line apparently does nothing, because the expectations in test_applications_v2.clj have not been changed except for the initial state.

opqdonut: good catch! fixed in PR #1019 commit 1919135

PR #1018, opened 2019-03-20
Refactor tests

luontola:


src/cljc/rems/common_util.cljc
```diff
@@ -55,7 +55,8 @@
       (is (= "(2)" (andstr "(" (:foo x) ")"))))))
 
 (defn deep-merge [a b]
-  (cond (and (sequential? a)
+  (cond (:replace (meta b)) (merge-with deep-merge b)
```
foxlynx: I would rather that we would have a new function deep-replace or sth. Don't quite get why use a function called merge to replace. Imo unnecessary obfuscation.

test/clj/rems/api/test_applications_v2.clj
```diff
@@ -487,11 +487,11 @@
...
-                            expected-application (-> (deep-merge expected-application
-                                                                 {:application/last-activity (DateTime. 5000)
-                                                                  :application/events events
-                                                                  :application/members #{{:userid "member"}}})
-                                                     (assoc :application/invitation-tokens {}))]
+                            expected-application (deep-merge expected-application
+                                                             {:application/last-activity (DateTime. 5000)
+                                                              :application/events events
+                                                              :application/members #{{:userid "member"}}
+                                                              :application/invitation-tokens ^:replace {}})]
```
opqdonut: I'd just `merge` the other keys and `update` or `deep-merge` the member

src/cljc/rems/common_util.cljc
luontola: There are more than one way to merge things. For example most deep merge implementations you can find online concatenate lists, whereas this deep merge combines them by index because it’s more useful for the application tests.

test/clj/rems/api/test_applications_v2.clj
luontola: I’m not sure what you mean. Do you want :application/members done differently? Is the new version of :application/invitation-tokens okay? A code example would be helpful.

opqdonut:
(related to #852)

PR #1030, opened 2019-03-25
fix removing an attachment

### LINKED ISSUE #1028: Removing attachments is broken
1. Add an attachment to an application
2. Remove the attachment
3. Reload the page

Expected: there is no attachment

Actual: the attachment's name is still there, but the link is broken

opqdonut:
fixes #1028


opqdonut:
(rebased on top of master)

PR #1034, opened 2019-03-26
fix: img-path default

Macroz:
Fix the `:img-path` default. We now have the language subdirectory in the path that must be bypassed.


opqdonut:
for #1078

Issue #862, opened 2019-01-22
vTHL3: Custom numbering of applications


Macroz:
Support custom labeling of applications. For example a running number within the year + the year + the description i.e. `68/2019 Title of my application`. This can be done dynamically when showing applications in lists instead of separate id + description columns.

- [x] compute a human-readable id for every application
- [x] show the human-readable id in the same field as the description (in tables in the ui)
- [x] don't show the internal id in the ui
- ~human-readable id configurable (fallback to internal id?)~

PR #1038, opened 2019-03-26
external id (like "2019/11") for applications

opqdonut:
for #862


src/clj/rems/db/applications.clj
```diff
@@ -924,6 +935,9 @@
        :event/time time
        :event/actor actor
        :application/id application-id
+       :application/external-id (when allocate-external-id? ;; TODO parameterize id allocation?
+                                  (let [id-prefix (str (.getYear time))]
+                                    (format-external-id (allocate-external-id! id-prefix))))
```
foxlynx: I would like rows 939 and 940 to be in format-external-id function (rename maybe the fn?). Unnecessary details here what the external id consists of imho.

test/clj/rems/db/test_applications.clj
```diff
@@ -319,3 +324,13 @@
...
+
+(deftest test-allocate-external-id!
+  (is (= [] (db/get-external-ids {:prefix "a"})))
+  (is (= [] (db/get-external-ids {:prefix "b"})))
+  (is (= {:prefix "a" :suffix "1"} (allocate-external-id! "a")))
+  (is (= {:prefix "a" :suffix "2"} (allocate-external-id! "a")))
+  (is (= {:prefix "a" :suffix "3"} (allocate-external-id! "a")))
+  (is (= {:prefix "b" :suffix "1"} (allocate-external-id! "b")))
+  (is (= {:prefix "b" :suffix "2"} (allocate-external-id! "b")))
+  (is (= {:prefix "a" :suffix "4"} (allocate-external-id! "a"))))
```
foxlynx: If we are going to use year as the prefix at the moment, would maybe be consistent to give prefix in the same example format?

PR #1046, opened 2019-03-27
Refactor: rename keywords

luontola:


src/clj/rems/workflow/dynamic.clj
```diff
@@ -19,33 +19,33 @@
...
+    :application.command/decide
+    :application.command/reject
+    :application.command/request-comment
+    :application.command/request-decision
+    :application.command/remove-member
+    :application.command/return
+    :application.command/save-draft
+    :application.command/submit
+    :application.command/uninvite-member
+    #_:application.command/withdraw})
```
opqdonut: hmm, states and command types _could_ be workflow specific. are we sure they're not?

opqdonut: ok I guess it's a valid design

opqdonut:
for #852

PR #1047, opened 2019-03-27
command schemas

opqdonut:


src/clj/rems/workflow/dynamic.clj
```diff
@@ -694,7 +789,7 @@
       (invitation-token-error application (:token cmd))
       {:success true
        :result {:event/type :application.event/member-joined
-                :event/time (:time cmd)
+                :event/time (:time cmd) ;; TODO these are redundant
                 :event/actor (:actor cmd)
                 :application/id (:application-id cmd)
```
luontola: The `ok` function can be used to get rid of the `:success true` and the time, actor and application-id can be removed because `handle-command` will already fill them in by calling `enrich-result`.

src/clj/rems/workflow/dynamic.clj
```diff
@@ -744,7 +840,10 @@
                                         :event/actor "applicant"
                                         :workflow/type :workflow/dynamic
                                         :workflow.dynamic/handlers #{"assistant"}}])
-        command {:type :application.command/save-draft
+        command {:application-id 123 :time 456
+                 :type :application.command/save-draft
+                 :field-values []
+                 :accepted-licenses []
                  :actor "applicant"}]
     (testing "executes command when user is authorized"
```
luontola: Also add a test "fails when command fails schema validation"

src/clj/rems/workflow/dynamic.clj
```diff
@@ -27,25 +27,120 @@
...
-    :application.command/remove-member
-    :application.command/return
-    :application.command/save-draft
-    :application.command/submit
-    :application.command/uninvite-member
-    #_:application.command/withdraw})
+(s/defschema CommandInternal
+  {:actor UserId
+   ;; TODO: tests use non-DateTime times
+   :time s/Any})
```
luontola: All the places use the same constant `:time 456`, so it should be quick to search and replace them to use `(DateTime. 1000)` or similar.

src/clj/rems/workflow/dynamic.clj
```diff
@@ -772,11 +871,15 @@
      (catch RuntimeException e#
        (ex-data e#))))
 
+;; TODO rename to be test-specific
```
luontola: Or move the high-level tests (i.e. tests for commands and calculate-permissions) to another file.

src/clj/rems/workflow/dynamic.clj
```diff
@@ -27,25 +27,120 @@
...
+(s/defschema RequestDecisionCommand
+  (assoc CommandBase
+         :type (s/eq :application.command/request-decision)
+         :deciders [UserId]
+         :comment s/Str))
+
+(s/defschema DecideCommand
+  (assoc CommandBase
+         :type (s/eq :application.command/decide)
+         :decision s/Keyword ;; TODO: (s/enum :approved :rejected), validated now in the command handler
```
luontola: Could it be done here instead and removed from the command handler? It should require only changing the test's assertion because the error message will be different when it comes from schema validation.

src/clj/rems/workflow/dynamic.clj
opqdonut: done

src/clj/rems/workflow/dynamic.clj
opqdonut: yeah I guess the time is ripe for that

opqdonut:
for #1075

PR #1049, opened 2019-03-28
use :application/external-id in email texts

opqdonut:
for #862


src/clj/rems/poller/email.clj
```diff
@@ -24,6 +24,9 @@
 (defn- invitation-link [token]
   (str (:public-url env) "accept-invitation?token=" token))
 
+(defn- id-of-application [application]
+  (or (:application/external-id application) (:id application)))
```
luontola: Would `humane-application-id` be a more descriptive name?

PR #1051, opened 2019-04-01
rems.application.commands: fix swagger for :type constants

opqdonut:
single-value enums are supported by swagger, unlike s/eq


opqdonut:
for #1075

PR #1054, opened 2019-04-01
Refactor: decouple command handlers from apply-event

luontola:
#852


luontola:
That’s right. After a couple more PRs apply-event can be removed.

PR #1055, opened 2019-04-02
archiving forms, workflows and licenses

opqdonut:
- enabling & archiving forms, workflows and licenses
- error if trying to archive something that's in use

for #806

### DIFF
--- a/src/clj/rems/db/catalogue.clj
+++ b/src/clj/rems/db/catalogue.clj
@@ -46,5 +46,6 @@
     return))
 
 (defn update-catalogue-item! [command]
+  ;; TODO disallow unarchiving catalogue item if its resource, form or licenses are archived
   (db/set-catalogue-item-state! (select-keys command [:id :enabled :archived]))
   {:success true})
--- a/src/clj/rems/db/form.clj
+++ b/src/clj/rems/db/form.clj
@@ -57,3 +57,13 @@
       (create-form-item! user-id form-id index item))
     {:success (not (nil? form-id))
      :id form-id}))
+
+(defn update-form! [command]
+  (let [catalogue-items (db/get-catalogue-items {:form (:id command)})]
+    (if (and (:archived command) (seq catalogue-items))
+      {:success false
+       :errors [{:type :t.administration.errors/form-in-use :catalogue-items (mapv :id catalogue-items)}]}
+      (do
+        (db/set-form-state! command)
+        (db/set-form-template-state! command)
+        {:success true}))))
--- a/src/clj/rems/db/form.clj
+++ b/src/clj/rems/db/form.clj
@@ -59,7 +59,7 @@
      :id form-id}))
 
 (defn update-form! [command]
-  (let [catalogue-items (db/get-catalogue-items {:form (:id command)})]
+  (let [catalogue-items (db/get-catalogue-items {:form (:id command) :archived false})]
     (if (and (:archived command) (seq catalogue-items))
       {:success false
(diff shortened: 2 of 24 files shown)


src/clj/rems/db/catalogue.clj
```diff
@@ -46,5 +46,6 @@
     return))
 
 (defn update-catalogue-item! [command]
+  ;; TODO disallow unarchiving catalogue item if its resource, form or licenses are archived
```
opqdonut: should I add a checkbox for this into #806 ?

luontola: Yes, add all the checkboxes that come to your mind. It helps in keeping track of the big epic.

opqdonut: done

src/clj/rems/db/form.clj
```diff
@@ -57,3 +57,13 @@
       (create-form-item! user-id form-id index item))
     {:success (not (nil? form-id))
      :id form-id}))
+
+(defn update-form! [command]
+  (let [catalogue-items (db/get-catalogue-items {:form (:id command)})]
```
luontola: By default `db/get-catalogue-items` doesn't return archived catalogue items. I wonder if that would be a problem here? To get all catalogue items, the `:archived true` parameter is needed. I wonder whether `db/get-catalogue-items` should default to returning all items and hiding the archived should be handled at a higher level (near the api).

opqdonut: Hmm interesting, the negative test passes anyway...

luontola: I guess this works correctly now. An explicit `:archived false` would improve readability.

src/clj/rems/db/licenses.clj
```diff
@@ -112,3 +112,29 @@
 (defn remove-license-attachment!
   [attachment-id]
   (db/remove-license-attachment! {:id attachment-id}))
+
+(defn- get-license-usage [id]
+  ;; these could be db joins
+  (let [resources (->> (db/get-resources-for-license {:id id})
+                       (map :resid)
+                       (map #(db/get-resource {:id %}))
```
luontola: Worst case, this will make as many DB queries as there are resources. I'm okay with it, since the update license command should be rarely used.

opqdonut: yeah that was kinda my thinking as well

src/cljs/rems/administration/create_resource.cljs
```diff
@@ -52,6 +52,7 @@
    (status-modal/common-pending-handler! (text :t.administration/save))
    (post! "/api/resources/create"
           {:params request
+           ;; TODO: render the catalogue items that use this resource in the error handler
```
opqdonut: added a checkbox to #806 for this

src/clj/rems/db/resource.clj
```diff
@@ -37,5 +37,10 @@
         (throw e)))))
 
 (defn update-resource! [command]
-  (db/set-resource-state! (select-keys command [:id :enabled :archived]))
-  {:success true})
+  (let [catalogue-items (db/get-catalogue-items {:resource-id (:id command)})]
```
luontola: `:archived false` also here

src/clj/rems/db/workflow.clj
```diff
@@ -49,3 +49,12 @@
     (merge
      result
      {:success (not (nil? (:id result)))})))
+
+(defn update-workflow! [command]
+  (let [catalogue-items (db/get-catalogue-items {:workflow (:id command)})]
```
luontola: `:archived false` also here

PR #1056, opened 2019-04-02
Refactor: decouple the tests of command handlers from apply-event 

luontola:
#852 

Many of the tests for command handlers were asserting things about the old application model. In preparation for removing the `rems.workflow.dynamic/apply-event` method, this PR decouples the tests from `apply-event` method by asserting the events produced by the command handlers. 

Some tests which are already covered in `rems.application.test-model` were removed. There is still some overlap with the rather declarative `rems.application.model/calculate-permissions`, but maybe that's ok.

### DIFF
--- a/test/clj/rems/workflow/test_dynamic.clj
+++ b/test/clj/rems/workflow/test_dynamic.clj
@@ -136,7 +136,7 @@
                             :field-values {1 "updated"}
                             :accepted-licenses #{3}})))))))
 
-(deftest test-submit-form-validation
+(deftest test-submit
   (let [injections {:validate-form-answers fake-validate-form-answers}
         created-event {:event/type :application.event/created
                        :event/time test-time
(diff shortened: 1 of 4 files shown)


test/clj/rems/workflow/test_dynamic.clj
```diff
@@ -191,532 +153,693 @@
...
-             (handle-command {:application-id 123 :time test-time
-                              :type :application.command/add-member :actor "applicant" :member {:userid "member1"}}
-                             application
-                             injections)
-             (handle-command {:application-id 123 :time test-time
-                              :type :application.command/add-member :actor "member1" :member {:userid "member2"}}
-                             application
-                             injections))))
+             (fail-command application
+                           {:type :application.command/add-member
```
opqdonut: hmm this is not a valid command without :time and :application-id, right?

luontola: The `fail-command` and `ok-command` helpers add them automatically. `rems.workflow.dynamic/handle-command` also validates all commands, so it would detect such issues.

luontola:
There should already be a bunch of tests which go through the API to check the integration.

test/clj/rems/workflow/test_dynamic.clj
opqdonut: right!

PR #1058, opened 2019-04-03
WIP/POC: Test improvements

Macroz:
- try to use [mount.lite](https://github.com/aroemers/mount-lite)
- try to make most tests run in parallel
- try to make all tests use a single initialized test database and not break it
- run each test in a transaction and rollback after each test


resources/migrations/20190301085026-example.edn
```diff
@@ -1,3 +0,0 @@
-{:ns rems.migrations.example
- :up-fn migrate-up
- :down-fn nil}
```
luontola: Why was this removed? Do code based migrations not work with mount-lite?

test/clj/rems/api/test_resources.clj
```diff
@@ -91,46 +90,62 @@
...
+        (@handler)
+        assert-response-is-ok)
+    (testing "duplicate resource ID is allowed between organizations"
+      (-> (request :post "/api/resources/create")
+          (authenticate api-key user-id)
+          (json-body {:resid resid
+                      :organization "TEST-ORGANIZATION2"
+                      :licenses [licid]})
+          (@handler)
+          assert-response-is-ok))
```
luontola: The API returns HTTP 200 also when there are errors, so this test is not asserting anything. It should check that the response has `:success true` (both here and the request right above it).

resources/migrations/20190301085026-example.edn
Macroz: Simply because it's not required either.

test/clj/rems/api/test_resources.clj
Macroz: Ok this is old code so you should fix it in another PR. This PR was only a proof of concept that doesn't make sense to merge but only to keep as discussion.

luontola: #1109 created

PR #1059, opened 2019-04-03
administration: better error messages when archiving fails

opqdonut:
for #806

### DIFF
--- a/src/clj/rems/db/licenses.clj
+++ b/src/clj/rems/db/licenses.clj
@@ -1,7 +1,9 @@
 (ns rems.db.licenses
   "querying localized licenses"
   (:require [rems.common-util :refer [distinct-by]]
-            [rems.db.core :as db])
+            [rems.db.core :as db]
+            [rems.db.resource :as resource]
+            [rems.db.workflow :as workflow])
   (:import (java.io FileInputStream ByteArrayOutputStream)))
 
 (defn- format-license [license]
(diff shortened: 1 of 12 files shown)


src/clj/rems/db/licenses.clj
```diff
@@ -113,18 +115,24 @@
...
                        (remove :archived)
-                       (map :id))
+                       (map resource-in-error))
         workflows (->> (db/get-workflows-for-license {:id id})
                        (map :wfid)
-                       (map #(db/get-workflow {:wfid %}))
+                       (map workflow/get-workflow)
                        (remove :archived)
-                       (map :id))]
+                       (map workflow-in-error))]
```
luontola: The name `{resource,workflow}-in-error` brings to my mind an operation that finds workflows/resources which are in an error state. But here the usage is related to selecting only those keys which the UI will show in an error message. I'd inline the functions `(map #(select-keys % [:id :resid]))` or use very long but descriptive function names.

src/clj/rems/db/licenses.clj
```diff
@@ -115,24 +115,18 @@
...
-  (select-keys resource [:id :resid]))
-
 (defn- get-license-usage [id]
   ;; these could be db joins
   (let [resources (->> (db/get-resources-for-license {:id id})
                        (map :resid)
                        (map resource/get-resource)
                        (remove :archived)
-                       (map resource-in-error))
+                       (map #(select-keys resource [:id :resid])))
```
luontola: `resource` -> `%`

PR #1060, opened 2019-04-04
hide expired catalogue items from applicants

opqdonut:
for #806

### DIFF
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -146,10 +146,6 @@
    (filter #(some #{(:id %)} ids)
            localized-items)))
 
-;; TODO remove this function
-(defn get-catalogue-item [id]
-  (catalogue/get-localized-catalogue-item id))
-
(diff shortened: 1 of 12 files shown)


opqdonut:
added some kinda-related cleanup

src/clj/rems/db/applications.clj
```diff
@@ -138,23 +138,13 @@
...
+  ([ids] (catalogue/get-localized-catalogue-items {:ids ids}))
   ([ids localized-items]
-   (mapv translate-catalogue-item
-         (filter #(some #{(:id %)} ids)
-                 localized-items))))
-
-(defn get-catalogue-item [id]
-  (translate-catalogue-item (catalogue/get-localized-catalogue-item id)))
+   (filter #(some #{(:id %)} ids)
+           localized-items)))
```
luontola: The 2-parameter version can be removed. It's only used from `get-catalogue-items-by-application-items` which is never used.

PR #1061, opened 2019-04-04
fix /api/licenses/<id>

opqdonut:
for #806


src/clj/rems/api/schema.clj
```diff
@@ -143,6 +144,7 @@
    :license/start DateTime
    :license/end (s/maybe DateTime)
    :license/enabled s/Bool
+   :license/active s/Bool
```
opqdonut: this missing was caught only by our browser tests. I'll follow up with another PR that uses the v2 application api more widely in tests

PR #1065, opened 2019-04-04
Refactor: remove rems.workflow.dynamic/apply-event

luontola:
#852


src/clj/rems/db/applications.clj
```diff
@@ -328,8 +326,7 @@
                               :is-applicant? (is-applicant? user-id application)
                               :review-type review-type
                               :description description)
-                       (permissions/cleanup)
-                       (dynamic/clean-internal-state))
```
opqdonut: why is this no longer needed?

luontola: It's no more exposed from a public API.

opqdonut: oh okay

PR #1068, opened 2019-04-05
endpoint per command type

opqdonut:
for better swagger user experience

### DIFF
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -16,8 +16,8 @@
 
 (s/defschema SaveDraftCommand
   (assoc CommandBase
-         :field-values s/Any
-         :accepted-licenses s/Any))
+         :field-values {s/Num s/Str}
+         :accepted-licenses #{s/Num}))
 
 (s/defschema SubmitCommand
   CommandBase)
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -7,7 +7,8 @@
 (def UserId s/Str)
 
 (s/defschema CommandInternal
-  {:actor UserId
+  {:type s/Keyword
+   :actor UserId
    :time DateTime})
 
 (s/defschema CommandBase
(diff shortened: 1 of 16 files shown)


src/cljs/rems/application.cljs
```diff
@@ -157,12 +156,10 @@
...
                                               (rf/dispatch [::set-validation-errors (:errors response)]))))
                                :error-handler status-modal/common-error-handler!
-                               :params {:type :application.command/submit
-                                        :application-id application-id}})
+                               :params {:application-id application-id}})
                        (status-modal/common-error-handler! response)))
           :error-handler status-modal/common-error-handler!
-          :params {:type :application.command/save-draft
-                   :application-id application-id
+          :params {:application-id application-id
```
luontola: This would be easier to read if `:params` was the first parameter, before the `:handler`, so that it would be right next to the command URL. The same reordering need applies to all three `post!` calls in this file.

src/clj/rems/application/commands.clj
```diff
@@ -7,90 +7,76 @@
...
    :time DateTime})
 
 (s/defschema CommandBase
   {:application-id s/Int})
 
 (s/defschema SaveDraftCommand
   (assoc CommandBase
-         :type (s/enum :application.command/save-draft) ;; single-value enums are supported by swagger, unlike s/eq
          :field-values s/Any
          :accepted-licenses s/Any))
```
luontola: Now would be a good time to add more specific schema than s/Any.

luontola:
We also have `/api/v2/applications/create`, `/api/applications/accept-invitation` which are like commands, though implemented as a custom function instead of command handler. To unify the API from the client's perspective, we could take the "/command" part away from the URL so that all of them are "/applications/command-name".

Also rename `add_attachment` -> `add-attachment` and `remove_attachment` ->  `remove-attachment`

opqdonut:
related to #852

opqdonut:
for #1075

PR #1069, opened 2019-04-05
db tests can now be run in the dev profile

opqdonut:


test/clj/rems/db/testing.clj
```diff
@@ -0,0 +1,27 @@
...
+                         #'rems.config/env
+                         #'rems.db.core/*db*)
+  (db/assert-test-database!)
+  (migrations/migrate ["reset"] {:database-url (:test-database-url env)})
+  (f)
+  (mount/stop))
+
+(defn test-data-fixture [f]
+  (test-data/create-test-data!)
+  ;; XXX no teardown!
```
luontola: Why no teardown? Does it rely on `test-db-fixture` doing a reset? Clarifying comment would be nice.

opqdonut: added a comment

PR #1070, opened 2019-04-05
Accept licenses command

Macroz:
Applicant and member is treated the same and can use the new accept-licenses command for accepting all the terms of use. The previous checkbox solution is removed.

Member attributes are shown if there are any (e.g. Malice in test users)

See https://github.com/CSCfi/rems/issues/653


src/cljs/rems/application.cljs
```diff
@@ -111,22 +112,36 @@
...
+
+(rf/reg-event-fx
+ ::reload-application-page
+ (fn [{:keys [db]} [_ id]]
+   {::reload-application id}))
+
+(rf/reg-fx
+ ::reload-application
+ (fn [id]
+   (fetch (str "/api/v2/applications/" id)
```
luontola: `/api/v2/applications` has been renamed to `/api/applications` in `master`

src/cljc/rems/text.cljc
```diff
@@ -60,7 +60,8 @@
   (text (get states state :t.applications.states/unknown)))
 
 (def ^:private event-types
-  {:application.event/approved :t.applications.events/approved
+  {:application.event/accepted-licenses :t.applications.events/accepted-licenses
```
luontola: `:application.event/licenses-accepted` would be more in line with how the other events are named.

Macroz: Indeed, but somewhere we probably made an error with this because now you have to remember in which order the words were ...

resources/translations/fi.edn
```diff
@@ -256,8 +258,10 @@
          :event "Tapahtuma"
          :events "Tapahtumat"
          :failed "Epäonnistui"
+         :has-accepted-licenses "Olet hyväksynyt käyttöehdot."
          :licenses "Käyttöehdot"
          :maxlength "(enintään %1 merkkiä)"
+         :must-accept-licenses "Saadaksesi oikeudet hakemiisi resursseihin, sinun pitää hyväksyä seuraavat käyttöehdot."
```
opqdonut: "hakemiisi" sounds like applicant but this is for members too? not insisting you fix this

src/clj/rems/application/events.clj
```diff
@@ -63,7 +63,10 @@
 (s/defschema DraftSavedEvent
   (assoc EventBase
          :event/type (s/enum :application.event/draft-saved)
-         :application/field-values {s/Int s/Str}
+         :application/field-values {s/Int s/Str}))
+(s/defschema LicensesAcceptedEvent
+  (assoc EventBase
+         :event/type (s/enum :application.event/licenses-accepted)
          :application/accepted-licenses #{s/Int}))
```
opqdonut: yeah it does look funny that we have both accepted-licenses and licenses-accepted

src/clj/rems/db/test_data.clj
```diff
@@ -29,7 +29,7 @@
 (def +fake-user-data+
   {"developer" {:eppn "developer" :mail "developer@example.com" :commonName "Developer"}
    "alice" {:eppn "alice" :mail "alice@example.com" :commonName "Alice Applicant"}
-   "malice" {:eppn "malice" :mail "malice@example.com" :commonName "Malice Applicant"}
+   "malice" {:eppn "malice" :mail "malice@example.com" :commonName "Malice Applicant" :twinOf "alice" :other "Attribute Value"}
```
opqdonut: twinOf :)

src/cljs/rems/application.cljs
```diff
@@ -602,23 +595,32 @@
...
           license-validations (index-by [:license-id] (:validation-errors edit-application))
           form-fields-editable? (form-fields-editable? application)
           readonly? (not form-fields-editable?)]
       [collapsible/component
        {:id "form"
         :title (text :t.form/licenses)
         :always
-        [:div.form-group.field
+        [:div
+         [:p (text :t.form/must-accept-licenses)]
```
opqdonut: should this be shown only when licenses haven't yet been accepted?

src/cljs/rems/application.cljs
```diff
@@ -602,23 +595,32 @@
...
                              :app-id (:application/id application))]))]}]))
 
-(defn- application-licenses [application edit-application]
+(defn- application-licenses [application edit-application userid]
   (when-let [licenses (not-empty (:application/licenses application))]
-    (let [accepted-licenses (:accepted-licenses edit-application)
+    (let [application-id (:application/id application)
+          accepted-licenses (get (:application/accepted-licenses application) userid)
+          possible-commands (:application/permissions application)
           license-validations (index-by [:license-id] (:validation-errors edit-application))
```
opqdonut: I thought you removed these from the backend?

test/clj/rems/workflow/test_dynamic.clj
```diff
@@ -180,13 +185,6 @@
                  (apply-events [(assoc-in draft-saved-event [:application/field-values 41] "")])
                  (fail-command submit-command injections)))))
 
-    (testing "cannot submit when not all licenses are accepted"
```
opqdonut: once this is merged, entitlements will be granted even without licenses? I guess you'll follow up with changes to the entitlement poller?

src/cljs/rems/application.cljs
Macroz: Now the wording is more neutral

test/clj/rems/workflow/test_dynamic.clj
Macroz: Yes it's the remaining checkbox

src/cljs/rems/application.cljs
Macroz: Now removed from here also.

src/cljc/rems/text.cljc
luontola: No error. It follows the naming conventions of event sourced systems: commands are verbs in imperative mood, events are verbs in past tense. Typically it ends up with the verb changing location, for example "create order" and "order created".

"Accepted licenses" is not a verb, but a noun (participle + noun), short for "licenses which have been accepted". It's used in that sense in `:application/accepted-licenses`. Past verb for an event would be an active "applicant accepted licenses" or passive "licenses accepted". The active form could be used in situations where the events need to be handled differently based on who made the event ("applicant accepted licenses" vs "member accepted licenses"), but usually that's not needed and anyways the event contains information that who made the event, so the passive form is usually enough.

PR #1085, opened 2019-04-08
administration: hide disabled things by default

opqdonut:
- unify disabled/archived/expired filtering in listing apis
- use expired instead of misleading active in API
- change "display archived" checkbox into "display old"

for #806

### DIFF
--- a/src/clj/rems/api/catalogue_items.clj
+++ b/src/clj/rems/api/catalogue_items.clj
@@ -40,11 +40,17 @@
       :roles #{:logged-in}
       :query-params [{resource :- (describe s/Str "resource id (optional)") nil}
                      {expand :- (describe s/Str "expanded additional attributes (optional), can be \"names\"") nil}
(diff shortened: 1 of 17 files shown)


src/clj/rems/api/catalogue_items.clj
```diff
@@ -40,11 +40,17 @@
...
-                     {archived :- (describe s/Bool "'true' to include archived items, defaults to 'false'") false}]
+                     {archived :- (describe s/Bool "whether to include archived items") false}
+                     {disabled :- (describe s/Bool "whether to include disabled items") false}
+                     {expired :- (describe s/Bool "whether to include expired items") false}]
       :return GetCatalogueItemsResponse
-      (ok (catalogue/get-localized-catalogue-items {:resource resource
-                                                    :expand-names? (str/includes? (or expand "") "names")
-                                                    :archived archived})))
+      (ok (db/apply-filters
+           (merge (when-not expired {:active true})
```
opqdonut: I'll make a separate PR out of using :expired instead of :active elsewhere

PR #1086, opened 2019-04-08
fix /api/licenses/<id> again

opqdonut:
- there were two conflicting License schemas
- the single license GET api wasn't tested

part of #852
caused by #1064
see also #1061

### DIFF
--- a/test/clj/rems/api/test_licenses.clj
+++ b/test/clj/rems/api/test_licenses.clj
@@ -37,21 +37,21 @@
                                           :attachment-id nil}
                                      :fi {:title "fi title"
                                           :textcontent "http://example.com/license/fi"
-                                          :attachment-id nil}}}]
-        (-> (request :post "/api/licenses/create")
-            (authenticate api-key user-id)
-            (json-body command)
-            handler
-            assert-response-is-ok)
+                                          :attachment-id nil}}}
+            id (-> (request :post "/api/licenses/create")
+                   (authenticate api-key user-id)
+                   (json-body command)
+                   handler
+                   assert-response-is-ok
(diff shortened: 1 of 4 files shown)


test/clj/rems/api/test_licenses.clj
```diff
@@ -37,21 +37,21 @@
...
-                         read-body)
-                license (->> body
-                             (filter #(= (:title %) (:title command)))
-                             first)]
+          (let [license (-> (request :get (str "/api/licenses/" id))
+                            (authenticate api-key user-id)
+                            handler
+                            assert-response-is-ok
+                            read-body)]
+            (is id)
```
luontola: `(is id)` is testing the create operation, so it should be move to line 47½.

test/clj/rems/api/test_licenses.clj
```diff
@@ -37,21 +37,21 @@
...
-            (json-body command)
-            handler
-            assert-response-is-ok)
+                                          :attachment-id nil}}}
+            id (-> (request :post "/api/licenses/create")
+                   (authenticate api-key user-id)
+                   (json-body command)
+                   handler
+                   assert-response-is-ok
+                   read-body
```
luontola: There is a helper `read-ok-body`

test/clj/rems/api/test_licenses.clj
```diff
@@ -65,21 +65,20 @@
...
+                   :id)]
```
luontola: Add `(is id)` here

opqdonut:
also added similar tests for the indivirual resource and license endpoints

PR #1088, opened 2019-04-08
Remove v2 prefix from API URLs

luontola:
#852

### DIFF
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -15,7 +15,6 @@
             [ring.middleware.multipart-params :as multipart]
             [ring.swagger.upload :as upload]
             [ring.util.http-response :refer :all]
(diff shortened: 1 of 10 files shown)


src/clj/rems/api/applications.clj
```diff
@@ -203,25 +197,9 @@
...
-
-    (POST "/create" []
-      :summary "Create a new application"
-      :roles #{:logged-in}
-      :body [request CreateApplicationCommand]
-      :return CreateApplicationResponse
-      (ok (applications/create-application! (getx-user-id) (:catalogue-item-ids request))))
+    (command-endpoint :application.command/uninvite-member commands/UninviteMemberCommand)
 
+    ;; this route must be last - otherwise the path parameter will conflict with the other routes
```
opqdonut: but it isn't last?

PR #1098, opened 2019-04-09
Member entitlements

Macroz:
Implements granting entitlements for members after licenses are accepted and application approved from #653. Entitlements are also removed when a member is removed or the application is closed. The entitlement tests are moved into their own proper places.


src/clj/rems/application/model.clj
```diff
@@ -116,12 +131,7 @@
 (defmethod calculate-permissions :application.event/approved
   [application _event]
   (-> application
-      (permissions/set-role-permissions (update closed-permissions
-                                                :handler set/union #{:application.command/add-member
-                                                                     :application.command/remove-member
-                                                                     :application.command/invite-member
-                                                                     :application.command/uninvite-member
-                                                                     :application.command/close}))))
+      (permissions/set-role-permissions approved-permissions)))
```
foxlynx: Good that the approved permissions are now in separate definition. 👍

src/clj/rems/application/model.clj
```diff
@@ -167,7 +178,7 @@
              :application/licenses (map (fn [license]
                                           {:license/id (:license/id license)})
                                         (:application/licenses event))
-             :application/accepted-licenses {"applicant" #{}}
+             :application/accepted-licenses {}
```
foxlynx: Why did this change?

src/clj/rems/db/entitlements.clj
```diff
@@ -62,30 +62,56 @@
           (log/warnf "Post failed: %s", response))
         (db/log-entitlement-post! {:target target :payload json-payload :status status})))))
 
+(defn- accepted-licenses? [application user-id]
+  (every? (or (get (:application/accepted-licenses application)
+                   user-id)
+              #{})
+          (map :license/id (:application/licenses application))))
```
foxlynx: Not sure if I understand this correctly, but isn't this true also for an empty hash-set?

test/clj/rems/api/test_entitlements.clj
```diff
@@ -14,3 +15,39 @@
...
+                     handler
+                     read-body)]
+        (is (= 2 (count data)))))
+
+    (testing "just for alice"
+      (let [data (-> (request :get "/api/entitlements?user=alice")
+                     (authenticate api-key "developer")
+                     handler
+                     read-body)]
+        (is (= 1 (count data)))))
```
foxlynx: Maybe do some assert for the content of the data as well?

src/clj/rems/application/model.clj
Macroz: This was completely wrong. There should not be accepted licenses for the fixed value called "applicant" for every application. There should only be the correct ones that come later.

src/clj/rems/db/entitlements.clj
Macroz: First part is geting the accepted application licenses for a particular user, or empty set if there is no user there can be no accepted licenses.

Then the second part is that we check that every license in the application is acceted. i.e. in the set of accepted application licenses of the user..
```
(every? #{} [1 2 3]) => false
```

test/clj/rems/api/test_entitlements.clj
Macroz: This is an old test that I just moved here. There is not really a good spec on what should be returned so anything goes here. The more interesting tests about content are in the `db/test_applications.clj` so far

Issue #1045, opened 2019-03-27
vTHL3: LBR03 environment configuration and migration


foxlynx:
- Configuring new LBR server through rems-deploy
- Converting existing applications to dynamic workflow

PR #1107, opened 2019-04-10
Migration fixes for LBR data

foxlynx:
Related to #1045


opqdonut:
we can't merge this since it will break normal use: the additional values are needed for the enums

PR #1110, opened 2019-04-11
administration: preview when creating new form

opqdonut:
for #835


opqdonut:
rebased now that #1108 is in

luontola:
In the editor when a new field is added, by default the field has no type selected. This results in a "Unsupported field" error in the preview. To avoid that it would be possible to use a default field type.

Issue #830, opened 2019-01-14
vTHL2: Dynamic workflow additional license request


Macroz:
The handler should be able to request a new license to be approved. I.e. create an agreement that is then signed by accepting the license. Old REMS abuses licenses for this. We could instead have a license request that the receiver must approve just like the other actions of the dynamic application. The request form can have a shortcut into creating a new license.

- [x] license request (new licenses visible, explicit approve licenses button)

Kauko:
#653 should maybe be done first?

hukka:
Also the question in #914 should be discussed first

Macroz:
At the moment creation of a new license is not possible from the handler. This is because now we have that right only for the owners, not for the handlers. I think it's useful at this point to ask the users if handlers and owners are the same or different group of users or should we expose (some of) the admin pages to handlers as well.

PR #1111, opened 2019-04-11
Add licenses

Macroz:
Implements #830 by the way of using the regular licenses.
- `add-licenses` command, `licenses-added` event and an action UI `add_licenses.cljs`.
- Regular accept licenses workflow for the acceptance part.

Open questions:
- Currently owner is able to add new licenses and a handler doesn't. Should it be possible for a handler to admin or how would an ad-hoc license for one application be created?
- Does not refactor adding licenses away from the created away. Perhaps it does not make sense after all.

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -205,6 +205,7 @@
 (defmethod event-type-specific-application-view :application.event/licenses-added
   [application event]
   (-> application
+      (assoc :application/modified (:event/time event))
       (update :application/licenses
               (fn [licenses]
                 (->> (:application/licenses event)
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -172,12 +172,9 @@
              :application/members #{}
              :application/past-members #{}
              :application/invitation-tokens {}
-             :application/resources (map (fn [resource]
-                                           {:catalogue-item/id (:catalogue-item/id resource)
-                                            :resource/ext-id (:resource/ext-id resource)})
+             :application/resources (map #(select-keys % [:catalogue-item/id :resource/ext-id])
                                          (:application/resources event))
-             :application/licenses (map (fn [license]
-                                          {:license/id (:license/id license)})
+             :application/licenses (map #(select-keys % [:license/id])
                                         (:application/licenses event))
              :application/accepted-licenses {}
              :application/events []
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -205,9 +205,8 @@
       (assoc :application/modified (:event/time event))
       (update :application/licenses
               (fn [licenses]
-                (->> (:application/licenses event)
-                     (map (fn [license-id] {:license/id license-id}))
-                     (into licenses)
+                (-> licenses
+                    (into (:application/licenses event))
                      distinct
                      vec)))))
 
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -210,8 +210,9 @@
               (fn [licenses]
                 (->> (:application/licenses event)
                      (map (fn [license-id] {:license/id license-id}))
-                     (concat licenses)
-                     distinct)))))
+                     (into licenses)
+                     distinct
+                     vec)))))
 
 (defmethod event-type-specific-application-view :application.event/member-invited
   [application event]
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -38,6 +38,7 @@
                                                   :application.command/accept-licenses]
                                       :member [:application.command/accept-licenses]
                                       :handler [:see-everything
+                                                :application.command/add-licenses
                                                 :application.command/add-member
                                                 :application.command/remove-member
                                                 :application.command/invite-member
@@ -200,6 +201,16 @@
   (-> application
       (assoc :application/modified (:event/time event))
       (assoc-in [:application/accepted-licenses (:event/actor event)] (:application/accepted-licenses event))))
+
+(defmethod event-type-specific-application-view :application.event/licenses-added
+  [application event]
+  (-> application
+      (update :application/licenses
+              (fn [licenses]
+                (->> (:application/licenses event)
+                     (map (fn [license-id] {:license/id license-id}))
+                     (concat licenses)
+                     distinct)))))
 
 (defmethod event-type-specific-application-view :application.event/member-invited
   [application event]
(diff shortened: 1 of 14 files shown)


src/clj/rems/application/model.clj
```diff
@@ -201,6 +202,17 @@
...
 
+(defmethod event-type-specific-application-view :application.event/licenses-added
+  [application event]
+  (-> application
+      (assoc :application/modified (:event/time event))
+      (update :application/licenses
+              (fn [licenses]
+                (->> (:application/licenses event)
+                     (map (fn [license-id] {:license/id license-id}))
+                     (concat licenses)
```
opqdonut: hmmmm, `(into licenses)` would work here as well

src/clj/rems/workflow/dynamic.clj
```diff
@@ -189,6 +189,13 @@
              :application/request-id last-request-for-actor
              :application/comment (:comment cmd)}))))
 
+(defmethod command-handler :application.command/add-licenses
+  [cmd _application injections]
+  (or (must-not-be-empty cmd :licenses)
+      (ok {:event/type :application.event/licenses-added
+           :application/licenses (set (:licenses cmd))
```
opqdonut: should these be in `{:license/id ...}` format since the key is `:application/licenses`?

src/clj/rems/workflow/dynamic.clj
```diff
@@ -204,8 +211,7 @@
 
 (defmethod command-handler :application.command/accept-invitation
   [cmd application injections]
-  (or (invalid-user-error (:actor cmd) injections)
```
opqdonut: right, the actor of a command can never be invalid since they're logged in?

src/cljs/rems/actions/add_licenses.cljs
```diff
@@ -0,0 +1,116 @@
...
+            [rems.status-modal :as status-modal]
+            [rems.text :refer [text]]
+            [rems.util :refer [fetch post!]]))
+
+(rf/reg-fx
+ ::fetch-licenses
+ (fn [[user on-success]]
+   (fetch "/api/licenses"
+          {:handler on-success
+           :headers {"x-rems-user-id" (:eppn user)}})))
```
opqdonut: we should have a `fetch-as-current-user` util that injects this header. perhaps in another PR

src/cljs/rems/actions/add_licenses.cljs
```diff
@@ -0,0 +1,116 @@
...
+               ::comment ""
+               ::potential-licenses #{}
+               ::selected-licenses #{})
+    ::fetch-licenses [(get-in db [:identity :user])
+                      #(rf/dispatch [::set-potential-licenses %])]}))
+
+(defn- assoc-search-term
+  "Prepopulate a search term with localized names and unlocalized title"
+  [license]
+  (assoc license :search-term (str/join "" (conj (mapcat :title (vals (:localizations license)))
```
opqdonut: I think the name `:search-term` is a bit misleading... how about `:all-titles`

test/clj/rems/application/test_model.clj
```diff
@@ -274,262 +299,348 @@
...
+                                                             :application/state :application.state/approved})]
+                            (is (= expected-application (apply-events events)))
+
+                            (testing "> licenses accepted"
+                              (let [events (conj events
+                                                 {:event/type :application.event/licenses-accepted
+                                                  :event/time (DateTime. 4500)
+                                                  :event/actor "applicant"
+                                                  :application/id 1
+                                                  :application/accepted-licenses #{30 31 32 33}})
```
opqdonut: could you have two different users accept different sets of licenses here please?

test/clj/rems/application/test_model.clj
```diff
@@ -274,262 +299,348 @@
...
+                                          :event/time (DateTime. 3500)
+                                          :event/actor "handler"
+                                          :application/id 1
+                                          :application/licenses #{33}
+                                          :application/comment "Please sign these terms also"})
+                            expected-application (merge expected-application
+                                                        {:application/last-activity (DateTime. 3500)
+                                                         :application/modified (DateTime. 3500)
+                                                         :application/events events
+                                                         :application/licenses [{:license/id 30
```
luontola: If nothing about the original three licenses changed, they should not be repeated here. Just append the new license:

```clj
(-> expected-application
    (merge {:application/last-activity (DateTime. 3500)
            :application/modified (DateTime. 3500)
            :application/events events})
    (update :application/licenses conj {:license/id 33
                                        :license/type :attachment
                                        :license/title {:en "en title"
                                                        :fi "fi title"
                                                        :default "non-localized title"}
                                        :license/attachment-id {:en 3301
                                                                :fi 3302
                                                                :default 3300}
                                        :license/attachment-filename {:en "en filename"
                                                                      :fi "fi filename"
                                                                      :default "non-localized filename"}
                                        :license/start (DateTime. 100)
                                        :license/end nil
                                        :license/expired false
                                        :license/enabled true
                                        :license/archived false}))

src/clj/rems/workflow/dynamic.clj
Macroz: Yes, some role and the `getx-user-id are required by the API already. This was the only handler that had an explicit check.

src/clj/rems/workflow/dynamic.clj
Macroz: Good idea, I did this refactoring.

test/clj/rems/application/test_model.clj
luontola: There's also the case that how to interpret accepting licenses multiple times. If a user first accepts licenses 30 and 31, and then later accepts 32 and 33, are now all four licenses accepted or only the last two?

src/cljs/rems/actions/add_licenses.cljs
Macroz: Renamed

src/cljs/rems/actions/add_licenses.cljs
Macroz: Good idea, I'll add a task for it

Macroz: Actually, I'll just remove it in another PR since it's not needed (comes from the session)

test/clj/rems/application/test_model.clj
Macroz: I added two users case and also one with overwriting the previous accepted licenses

test/clj/rems/application/test_model.clj
Macroz: Now it's much shorter.

Issue #913, opened 2019-02-08
Store forms as JSON documents instead of relations


opqdonut:
(moved leftovers here from #835)

Current status:
- when a form is created, it's stored in the db both as a document (`form_template`) and as relations (`application_form`, `application_form_item`, `application_form_item_map`)
  - this is the document the _write model_
- there is an API for fetching the write model document (`/api/forms/:id`)
- there is an alternate document format that is returned by the new application api (schema `rems.api.schema/Form`)
  - this is the document _read model_
  - the read model is used by the application rendering code on the frontend
- when creating or viewing a form, the write model is converted to the read model to render a preview

To do:

- [x] unify read and write models (probably by using the read model for writing):
  - [x] store the read model in `form_template`
  - [x] get rid of read-to-write-model conversion in preview
- [x] read `form_template` instead of `application_form_*` in application api implementation

opqdonut:
after #1299, the `application_form_*` tables are no longer used by the application workflow code

opqdonut:
the last remaining user of the `application_form_*` tables is `rems.db.applications.legacy`

opqdonut:
I think this has progressed far enough now to unblock #865

luontola:
Removing old db tables and related source code has been moved to be part of #1101.

PR #1113, opened 2019-04-11
administration: preview for existing form

### LINKED ISSUE #835: vTHL3: Preview form

opqdonut:
plus some form api cleanup

for #835
related to #913


src/cljs/rems/administration/form.cljs
```diff
@@ -37,33 +38,6 @@
...
-         (for [localization (:localizations field)]
-           [inline-info-field (str (text :t.administration/title)
-                                   " "
-                                   (str/upper-case (name (:langcode localization)))) (:title localization)])
-         [[inline-info-field (text :t.create-form/type) (text (keyword (str "t.create-form/type-" (:type field))))]
-          [inline-info-field (text :t.create-form/input-prompt) (get-localized-value field :inputprompt language)]
-          [inline-info-field (text :t.create-form/optional) [readonly-checkbox (:formitemoptional field)]]
-          [inline-info-field (text :t.create-form/maxlength) (:maxlength field)]])))
-
-(defn form-fields [fields language]
```
opqdonut: I felt like these didn't add any value on top of the preview so I removed them

src/cljs/rems/administration/form.cljs
```diff
@@ -75,8 +49,8 @@
               [inline-info-field (text :t.administration/start) (localize-time (:start form))]
               [inline-info-field (text :t.administration/end) (localize-time (:end form))]
               [inline-info-field (text :t.administration/active) [readonly-checkbox (not (:expired form))]]]}]
-   [form-fields (:fields form) language]
-   [:div.col.commands [back-button]]])
+   [:div.col.commands [back-button]]
```
opqdonut: back is between the info and the preview. could also be at the top I guess. I don't want to have it below the preview, and I don't think a collapsible preview would be any better

Macroz:
I think the preview should not be open all the time, but rather open into a tab or maybe just modal container when you want. Perhaps it's also possible to split the view horizontally and render both on a wider display to actually use the space well but I'd go with the modal first.

Also noticed errors with THL form.

src/clj/rems/api/forms.clj
```diff
@@ -26,13 +26,13 @@
    :optional s/Bool
    :type (s/enum "attachment" "date" "description" "label" "multiselect" "option" "text" "texta")
    (s/optional-key :maxlength) (s/maybe (s/constrained s/Int not-neg?))
-   (s/optional-key :options) [{:key s/Str
+   (s/optional-key :options) [{:key s/Any ;; s/Str for multiselect, s/Bool for option
```
luontola: Smells fishy.

opqdonut: I removed the fishes

opqdonut:
hmm okay it seems the bool keys used in the thl form are actually invalid. let me fix those instead of the schema

test/clj/rems/api/test_forms.clj
```diff
@@ -61,32 +61,75 @@
...
+                             :maxlength 300
+                             :input-prompt localized}
+                            {:type "description"
+                             :title localized
+                             :optional false}
+                            {:type "option"
+                             :title localized
+                             :optional true
+                             :options [{:key "true" :label localized}
+                                       {:key "false" :label localized}]}
```
luontola: Option fields can have more than two options. The difference between option and multiselect is that option lets the user choose only one value, but multiselect lets the user choose many values. To make that explicit, this example could be:
```clj
:options [{:key "a" :label localized}
          {:key "b" :label localized}
          {:key "c" :label localized}]

PR #1119, opened 2019-04-15
administration: show form editor and preview side-by-side

### LINKED ISSUE #835: vTHL3: Preview form

opqdonut:
first implementation, can be improved later

closes #835

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -20,6 +20,8 @@
             [rems.util :as util]
             [rems.context :as context]))
 
+(def content-width (u/px 1200))
+
 (defn resolve-image [path]
   (when path
     (let [url (if (str/starts-with? path "http")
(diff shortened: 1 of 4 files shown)


src/clj/rems/css/styles.clj
```diff
@@ -385,6 +386,7 @@
     {:color (util/get-theme-attribute :nav-color :link-color :color3)
      :border 0}] ; for button links
    [:.navbar
+    {:max-width (u/px 1200)}
```
Macroz: This could be extracted to be a constant.

PR #1127, opened 2019-04-15
Performance optimizations cleanup

luontola:
Follow up from #1114 code review.

### DIFF
--- a/src/clj/rems/api/applications_v2.clj
+++ b/src/clj/rems/api/applications_v2.clj
@@ -51,6 +51,7 @@
         build-app (fn [_] (model/build-application-view events injections))]
     (if (empty? events)
       nil ; application not found
+      ;; TODO: this caching could be removed by refactoring the pollers to build their own projection
       (if (atom? application-cache) ; guard against not started cache
         (-> (swap! application-cache cache/through-cache cache-key build-app)
             (getx cache-key))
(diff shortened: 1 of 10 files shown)


src/clj/rems/api/applications_v2.clj
```diff
@@ -129,13 +143,16 @@
        (filter own-application?)))
 
 (defn reload-cache! []
+  ;; TODO: Here is a small chance that a user will experience a cache miss. Consider rebuilding the cache asynchronously and then `reset!` the cache.
   (events-cache/empty! all-applications-cache)
   (refresh-all-applications-cache!))
 
 ;; empty the cache occasionally in case some of the injected entities are changed
 (mount/defstate all-applications-cache-reloader
+  ;; TODO: consider refactoring this mount-cron-thing which is duplicated in a few places
```
opqdonut: yeah please do that

src/clj/rems/api/applications_v2.clj
```diff
@@ -98,6 +111,7 @@
    (fn [state events]
      ;; Because enrich-with-injections is not idempotent,
      ;; it's necessary to hold on to the "raw" applications.
+     ;; TODO: consider making enrich-with-injections idempotent (move dissocs to hide-non-public-information and other small refactorings)
```
opqdonut: please try this out, perhaps in a follow-up PR

src/clj/rems/db/dynamic_roles.clj
```diff
@@ -42,15 +42,22 @@
...
 (defn get-roles [user]
   (->> (events-cache/refresh!
         dynamic-roles-cache
         (fn [state events]
-          (let [apps (reduce permissions-of-all-applications (::apps state) events)]
-            {::apps apps
-             ::roles-by-user (group-roles-by-user apps)})))
+          (let [permissions (reduce permissions-of-all-applications (::permissions state) events)]
+            {::permissions permissions
+             ::roles-by-user (group-roles-by-user permissions)})))
```
opqdonut: is `group-roles-by-user` really so slow that we need to memoize it?

luontola: There are benchmark measurements in the commit message 87a598bafa9d210b31d690e70d85f8932a370615. It used to be a linear search which calls `apply-user-permissions` for every application, but now it's just a hashmap lookup. Filtering a person's own applications and open/handled reviews is still a linear search, but most users can see only a few applications so it's less of an issue.

src/clj/rems/api/applications_v2.clj
luontola: Done in #1138

Issue #1103, opened 2019-04-10
vTHL3: Edit workflow


opqdonut:
via administration ui

- [x] rename
- [x] change handlers

(split out from #867)

PR #1136, opened 2019-04-16
api: updating workflow title and handlers

opqdonut:
for #1103

### DIFF
--- a/test/clj/rems/api/test_workflows.clj
+++ b/test/clj/rems/api/test_workflows.clj
@@ -80,21 +80,7 @@
                              :handlers ["bob" "carl"]}
                   :enabled true
                   :archived false}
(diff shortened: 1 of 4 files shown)


test/clj/rems/api/test_workflows.clj
```diff
@@ -96,6 +96,69 @@
...
+      (is (:success (update {:enabled false :archived true})))
+      (is (= (assoc expected
+                    :enabled false
+                    :archived true)
+             (fetch))))
+    (testing "re-enable"
+      (is (:success (update {:enabled true})))
+      (is (= (assoc expected
+                    :archived true)
+             (fetch))))
```
luontola: Could also add test for "unarchive". That would avoid the need for `(assoc :archived true)` from the following tests.

PR #1137, opened 2019-04-16
minor changes to parsing and validation of maxlength in form creation

### LINKED ISSUE #1122: Maximum length of a field cannot be unset for the preview
When in the form editor, you can set the maximum length of a text field but when you clear the maximum length field, you still get the maximum number of characters text in the form preview.

okahilak:
closes #1122

 - parse maxlength also when the form is being previewed, thus
   preventing empty string from being a valid value in the preview,
   which previously caused "(max  characters)"

 - instead of checking the emptiness of the string using str/blank?,
   let parseInt do that part as well, returning NaN if the string is
   empty or nil, which is then filtered out using js/NaN

 - change input field for maxlength from number field to text field,
   thus allowing parseInt do the whole parsing, which has the benefit
   that it handles trimming the strings, which the html element doesn't
   seem to do (entering "1 " in the field did not work previously)

 - add validation for maxlength field, preventing saving a form with
   negative maxlength values, which crashed the backend previously


src/cljs/rems/administration/create_form.cljs
```diff
@@ -86,13 +86,19 @@
...
+    (when-not (js/isNaN parsed) parsed)))
+
 (defn- valid-request-item? [item languages]
   (and (valid-required-localized-string? (:title item) languages)
        (boolean? (:optional item))
        (not (str/blank? (:type item)))
        (if (supports-input-prompt? item)
          (valid-optional-localized-string? (:input-prompt item) languages)
          (nil? (:input-prompt item)))
+       (if (supports-maxlength? item)
```
opqdonut: this should probably be something like
```
(if (supports-maxlength? item)
  (not (neg? (:maxlength item)))
  (nil? (:maxlength item)))

**Comment by opqdonut** [comment]
also, you should fix `build-request-test` in test/cljs/rems/administration/test_create_form.cljs

File: src/cljs/rems/administration/create_form.cljs [hunk 2]
```diff
@@ -87,13 +87,20 @@
...
 (defn- valid-request-field? [field languages]
   (and (valid-required-localized-string? (:title field) languages)
        (boolean? (:optional field))
        (not (str/blank? (:type field)))
        (if (supports-input-prompt? field)
          (valid-optional-localized-string? (:input-prompt field) languages)
          (nil? (:input-prompt field)))
+       (if (supports-maxlength? field)
+         (not (neg? (:maxlength field)))
+	 (nil? (:maxlength field)))
```
luontola: You should use only spaces for indentation. The indentation on this line was wonky.

PR #1150, opened 2019-04-18
Show all apps for domain reporter on applications page

### LINKED ISSUE #833: vTHL3: Domain reporter role
A user with rights to see all applications. Can only view applications, not do any actions unless given the rights as usual.

- [x] domain reporter role
- [x] user rights added to the database and/or API (no UI yet)
- [x] domain reporter has a reports page that shows all applications (no other reports yet)

luontola:
closes #833

### DIFF
--- a/src/cljs/rems/applications.cljs
+++ b/src/cljs/rems/applications.cljs
@@ -19,7 +19,7 @@
 (rf/reg-event-fx
  ::fetch-my-applications
  (fn [{:keys [db]} _]
-   (fetch "/api/applications"
+   (fetch "/api/my-applications"
           {:handler #(rf/dispatch [::fetch-my-applications-result %])})
    {:db (assoc db ::loading-my-applications? true)}))
 
@@ -45,7 +45,7 @@
 (rf/reg-event-fx
  ::fetch-all-applications
  (fn [{:keys [db]} _]
-   (fetch "/api/applications/all"
+   (fetch "/api/applications"
           {:handler #(rf/dispatch [::fetch-all-applications-result %])})
    {:db (assoc db ::loading-all-applications? true)}))
 
--- a/src/cljs/rems/applications.cljs
+++ b/src/cljs/rems/applications.cljs
@@ -1,15 +1,15 @@
 (ns rems.applications
   (:require [re-frame.core :as rf]
             [rems.application-list :as application-list]
+            [rems.roles :as roles]
             [rems.spinner :as spinner]
             [rems.text :refer [localize-state localize-time text]]
-            [rems.util :refer [fetch]]
-            [rems.roles :as roles]))
+            [rems.util :refer [fetch]]))
 
 (rf/reg-event-fx
  ::enter-page
  (fn [{:keys [db]} _]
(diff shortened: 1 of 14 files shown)


src/cljs/rems/applications.cljs
```diff
@@ -3,20 +3,23 @@
...
             [rems.spinner :as spinner]
             [rems.text :refer [localize-state localize-time text]]
-            [rems.util :refer [fetch]]))
+            [rems.util :refer [fetch]]
+            [rems.roles :as roles]))
 
 (rf/reg-event-fx
  ::enter-page
  (fn [{:keys [db]} _]
    {:db (dissoc db ::my-applications)
```
hukka: Current logic dissocs :my-applications, when fetch starts. Should same be done to :all-applications?

I think that either all applications should follow the same logic, or the discrepancy should be documented.

src/cljs/rems/applications.cljs
```diff
@@ -37,6 +40,32 @@
...
+(rf/reg-sub
+ ::all-applications
+ (fn [db _]
+   (::all-applications db)))
+
+(rf/reg-sub
+ ::loading-all-applications?
+ (fn [db _]
+   (::loading-all-applications? db)))
+
```
hukka: These are very similar to my-applications. Would it make sense to make a generic that takes an URL and a db path, and stores the data and loading status under a map in that path?

Or is that abusing re-frame or just premature abstraction, as this is just the second version and only about 20 lines?

src/cljs/rems/applications.cljs
```diff
@@ -54,20 +83,34 @@
 
 ;;;; UI
 
+;; XXX: the application lists share sorting and filtering state
```
hukka: If we change to a model where all application data fetch state is in under one path in db, we could put the filter state there too.

src/cljs/rems/roles.cljs
```diff
@@ -3,12 +3,14 @@
...
-  (some #{:handler :commenter :decider :past-commenter :past-decider :reporter} roles))
+(defn show-all-applications? [roles]
+  (some #{:reporter} roles))
 
-(defn is-admin? [roles]
-  ;; TODO: add admin role (owner is a business user; admin is a technical user)
+(defn show-reviews? [roles]
+  (some #{:handler :commenter :decider :past-commenter :past-decider} roles))
+
+(defn show-admin-pages? [roles]
```
hukka: Is this now rather permissions.cljs, not roles? Pretty nitpicky. I agree that `is-applicant-or-member` was not really the most descriptive name to the relevant action/permission.

src/cljs/rems/applications.cljs
luontola: Similar loading is done on also other pages, so a generic component for doing that would make sense.

src/cljs/rems/roles.cljs
luontola: At least they still take roles as parameters, so they are to some extent about roles. I'm feeling too lazy to change the namespace name. 😉

PR #1151, opened 2019-04-18
sort catalogue initially to ascending order by :name

### LINKED ISSUE #1130: The catalogue should be in alphabetical order.
Catalogue items should be in alphapetical order.

okahilak:
closes #1130


luontola:
Such a click helper sounds like a good idea. We can do it later.

PR #1153, opened 2019-04-23
Transaction stuff

luontola:
- Make also reads transactional
- Test that events are observed in monotonically increasing order

### DIFF
--- a/test/clj/rems/db/test_transactions.clj
+++ b/test/clj/rems/db/test_transactions.clj
@@ -82,18 +82,18 @@
         user-id (create-dummy-user)
         app-ids (vec (for [_ (range applications-count)]
                        (create-dummy-application user-id)))
-        observed-event-count-marker 999
-        mark-observed-event-count (fn [result _cmd application]
+        observed-app-version-marker 999
+        mark-observed-app-version (fn [result _cmd application]
                                     (if (and (:success result)
                                              (= :application.event/draft-saved (:event/type (:result result))))
-                                      (assoc-in result [:result :application/field-values observed-event-count-marker]
+                                      (assoc-in result [:result :application/field-values observed-app-version-marker]
                                                 (str (count (:application/events application))))
                                       result))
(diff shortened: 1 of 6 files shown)


test/clj/rems/db/test_transactions.clj
```diff
@@ -0,0 +1,177 @@
...
+        observed-app-version-marker 999
+        mark-observed-app-version (fn [result _cmd application]
+                                    (if (and (:success result)
+                                             (= :application.event/draft-saved (:event/type (:result result))))
+                                      (assoc-in result [:result :application/field-values observed-app-version-marker]
+                                                (str (count (:application/events application))))
+                                      result))
+        write-event (fn [app-id]
+                      (try
+                        ;; Note: currently these tests pass only with serializable isolation
```
opqdonut: and thus that's what we use in the middleware? please clarify

test/clj/rems/db/test_transactions.clj
```diff
@@ -0,0 +1,177 @@
...
+
+      (testing "commands are executed serially"
+        (doseq [[_ events] final-events-by-app-id]
+          (is (= (->> (range 1 (count events))
+                      (map str))
+                 (->> events
+                      (filter #(= :application.event/draft-saved (:event/type %)))
+                      (map #(get-in % [:application/field-values observed-app-version-marker])))))))
+
+      ;; TODO?
```
opqdonut: please clarify. this does not pass right now due to transaction rollbacks, right?

luontola: Yes, transaction rollbacks. To guarantee that there are no gaps, it would be possible to use `max(id) + 1`.

test/clj/rems/db/test_transactions.clj
luontola: I'll first try adding a `select for update` which should reduce transaction conflicts and also make this work at any isolation level.

PR #1160, opened 2019-04-24
feat: optimize table filtering and rendering

Macroz:
- use separate close symbol
- only update the search results (which is slow) on enter, blur or
  clicking the new search button
- use Form 3 component to optimize the table rendering. Unfortunately
  the anonymous functions are not equal.
- empty the filters when closing
- buttons are not disabled anymore because that forces a re-render of the whole table
- shopping cart prevents adding same item multiple times

### DIFF
--- a/src/cljs/rems/cart.cljs
+++ b/src/cljs/rems/cart.cljs
@@ -26,14 +26,10 @@
 (defn add-to-cart-button
   "Hiccup fragment that contains a button that adds the given item to the cart"
   [item]
-  (let [cart @(rf/subscribe [::cart])
-        disabled? (and cart (contains? (set (map :id cart)) (:id item)))]
     [:button.btn.btn-primary.add-to-cart
      {:type "submit"
-      :disabled disabled?
-      :class (if disabled? " disabled" "")
       :on-click #(rf/dispatch [::add-item item])}
-     (text :t.cart/add)]))
+   (text :t.cart/add)])
 
 (defn remove-from-cart-button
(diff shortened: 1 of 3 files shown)


src/cljs/rems/table.cljs
```diff
@@ -123,29 +125,44 @@
...
    :id-function Applicable
    :items [s/Any]
    (s/optional-key :class) s/Str})
 
 (defn- filter-view [filtering]
-  (let [{:keys [filters set-filtering]} filtering]
+  (let [{:keys [filters-new set-filtering]} filtering
+        update-current-filters (fn [event]
+                                 (set-filtering (assoc filtering :filters filters-new)))]
+    [:div.flex-grow-1.d-flex
```
luontola: The indentation doesn't match the parentheses in multiple places in this file and thus breaks when the file is opened with parinfer. To a code reformat for the whole file.

src/cljs/rems/cart.cljs
```diff
@@ -16,24 +16,22 @@
...
   [item]
-  (let [cart @(rf/subscribe [::cart])
-        disabled? (and cart (contains? (set (map :id cart)) (:id item)))]
     [:button.btn.btn-primary.add-to-cart
      {:type "submit"
-      :disabled disabled?
-      :class (if disabled? " disabled" "")
       :on-click #(rf/dispatch [::add-item item])}
-     (text :t.cart/add)]))
+   (text :t.cart/add)])
```
luontola: This file also needs a code reformat.

Issue #1159, opened 2019-04-24
vTHL3: You can add same catalogue item many times


Macroz:
You can add Dynamic workflow catalogue item multiple times. Same problem applies to the THL catalogue item. You should only be able to add them once.

Issue #1102, opened 2019-04-10
vTHL3: Adding and removing resources to submitted application


opqdonut:
(split out from #842)

- [x] when the application is being processed, the handler can add and remove resources
- [ ] only the handler can add resources where the form is different to the one in the application
- [ ] if the new resources use a different form, give a warning

Macroz:
Do #842 first

PR #1168, opened 2019-04-25
Change resources

### LINKED ISSUE #842: vTHL3: Adding and removing resources from (draft) application
Resources should be able to be added and removed from an application.
Applicants can forget to add all resources. Sometimes the approver needs to add a resource that provides the information the applicant has wanted. Sometimes approval is not given to all the resources.

- [x] add and remove functionality into the resources block. Maybe use the autocomplete?
- [x] when the application is editable, the applicant should be able to add and remove resources

(see also #1102 #1104)

Macroz:
I'm getting blinded by these changes so I'm looking for comments and pointing out obvious problems.

Closes #842 and #1159. Also implements some of #1102 but no checking for bundling is done yet.

I have an additional idea that as an applicant maybe from the catalogue you could also add resources to existing applications in addition to just adding to the cart. I think I'll leave this out for now.


src/clj/rems/application/events.clj
```diff
@@ -101,6 +101,12 @@
   (assoc EventBase
          :event/type (s/enum :application.event/rejected)
          :application/comment s/Str))
+(s/defschema ResourcesChangedEvent
+  (assoc EventBase
+         :event/type (s/enum :application.event/resources-changed)
+         (s/optional-key :application/comment) s/Str
```
hukka: Base already has a comment. Accident, or making it optional (usually it's just empty)?

src/cljs/rems/catalogue.cljs
```diff
@@ -49,7 +49,16 @@
        (assoc ::catalogue catalogue)
        (dissoc ::loading-catalogue?))))
 
-(rf/reg-sub ::catalogue (fn [db _] (::catalogue db)))
+(rf/reg-sub ::full-catalogue (fn [db _] (::catalogue db)))
```
hukka: Is this used anywhere? The ::catalogue sub could depend on this.

src/clj/rems/application/events.clj
luontola: Base doesn't have a comment.

luontola: In the other events which have a comment, it's a mandatory field, but here it's optional. Any reason for that?

Macroz: The applicant does not need to add a comment when he/she is just creating an application but it would be nice if the handler can add a comment why he/she changed the resources.

luontola: 👌

src/cljs/rems/catalogue.cljs
Macroz: This is used by the `change_resources.cljs` and since this is the main file where catalogue is used I thought it's better to have the subscription be here.

It's true that the subs could be chained but I think the case here is so trivial that maybe it's unnecessary.

src/cljs/rems/actions/change_resources.cljs
```diff
@@ -0,0 +1,121 @@
...
+      (not= -1)))
+
+(rf/reg-sub ::initial-resources (fn [db _] (::initial-resources db)))
+(rf/reg-sub ::selected-resources (fn [db _] (::selected-resources db)))
+(rf/reg-event-db ::set-selected-resources (fn [db [_ resources]] (assoc db ::selected-resources resources)))
+(rf/reg-event-db ::add-selected-resources (fn [db [_ resource]] (update db ::selected-resources conj (:id resource))))
+(rf/reg-event-db ::remove-selected-resource (fn [db [_ resource]] (update db ::selected-resources disj (:id resource))))
+
+(rf/reg-sub ::comment (fn [db _] (::comment db)))
+(rf/reg-event-db ::set-comment (fn [db [_ value]] (assoc db ::comment value)))
```
hukka: Sub usually after effect handler

Macroz: I think they are pretty random in our code.

src/cljs/rems/catalogue.cljs
Macroz: Now with chaining!

PR #1169, opened 2019-04-26
Enable diff notification for attachements

### LINKED ISSUE #978: vTHL3: Add to application diff if attachment has been updated
Diff needs to show if attachement has been uplodaded again or if the name has changed etc. No need to diff the attachment file contents.

hukka:
Closes #978. Note that old versions of files are not saved, this merely
alerts the reader that the file has changed. Date and time is appended
to the name so that users wouldn't accidentally mask this by using a new
file with same name (there are no API level checks for this, so anyone
with a curl could fake it, though).

image

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -304,7 +304,7 @@
                [fields/field (assoc fld
                                     :on-change #(rf/dispatch [::set-field-value (:field/id fld) %])
                                     :on-set-attachment #(rf/dispatch [::save-attachment (:field/id fld) %1 %2])
-                                    :on-remove-attachment #(rf/dispatch [::remove-attachment (:field/id fld) %1 %2])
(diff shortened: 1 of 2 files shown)


src/cljs/rems/application.cljs
```diff
@@ -304,7 +304,7 @@
                [fields/field (assoc fld
                                     :on-change #(rf/dispatch [::set-field-value (:field/id fld) %])
                                     :on-set-attachment #(rf/dispatch [::save-attachment (:field/id fld) %1 %2])
-                                    :on-remove-attachment #(rf/dispatch [::remove-attachment (:field/id fld) %1 %2])
+                                    :on-remove-attachment #(rf/dispatch [::remove-attachment (:field/id fld) %])
```
hukka: I think those %1 and %2 were a copy paste error from on-set-attachment, since ::remove-attachment handler does not use that many fields.

Issue #1176, opened 2019-04-29
Summary of review 26.4.


mmatthiesencsc:
Refers to staging version:
https://github.com/CSCfi/rems/commits/95cd1d311ad209aa5fce3f83eb4cd04fcbd98822

# Positive feedback
- Tero and I decided that the new colour scheme just fine for now (unlike reported by Tero earlier)
- the UI is fast and snappy.

# UI issues
- [x] #1128 capitalize menu (e.g. "CATALOGUE") and hilight active menu as in production lbr.csc.fi  
- [x] #1133 application numbering: still important to us (was not promised for this rc) 
- ~~"..." in most buttons should be removed~~
- [x] #472 #1231 About > Privacy Policy (see production lbr.csc.fi)
- [x] #1186 UI logic problem: User cannot retract saved application or get out of application if she does not want to accept terms.  
- [x] #1179 Catalog items: Rightmost buttons chopped in FF ( as reported by Tero, tested on latest ESR in Linux);  (works ok in Chromium). 
- [x] https://github.com/CSCfi/rems/issues/1275

# Bugs
- [x] #1199 Bug: Request decision without specified deciders:
Request decision
Failed
Missing translation [:deciders]Missing translation [:must-not-be-empty]

- [x] #1191 Bug: https://lbr.csc.fi/#/application/809
shows up under "Open applications" after approval
Expected: Under processed applications

luontola:
> - capitalize menu (e.g. "CATALOGUE") and hilight active menu as in production lbr.csc.fi

This can be done with custom CSS:
```css
.nav-item {
    text-transform: uppercase;
}
```

The missing highlighting is a bug. Another CSS rule overwrites the highlight color.

mmatthiesencsc:
Hi Esko, thanks for the quick feedback, good that some of the issues are easy to fix. Please comment also on hard to fix issues, we can then decide to live with them if necessary. I really would like to get this update into production. Martin

luontola:
Application numbering is currently under development: #1133

The feature which will enable the privacy policy page is under development: #472

Redacting the application could be done by allowing the applicant to close draft applications. Currently only the handler can do that (and only for approved applications). The UI might need some adjustment, e.g. removing the comment field.

Macroz:
[...]

PR #1177, opened 2019-04-29
"More options" ellipsis using CSS

luontola:
This changes the "..." in buttons to be produced using CSS, so that it could be hidden using theme specific CSS:

```css
.btn-opens-more::after {
    content: "";
}
```

See #1176

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -526,14 +526,7 @@
    [(s/> :.form-actions "*:not(:first-child)")
     (s/> :.commands "*:not(:first-child)")
     {:margin-left (u/em 0.5)}]
-   [".btn-opens-more::after" {:display "inline-block"
-                              :font-style "normal"
-                              :font-variant "normal"
-                              :text-rendering "auto"
-                              :-webkit-font-smoothing "antialiased"
-                              :font-family "'Font Awesome 5 Free'"
-                              :font-weight 900
-                              :content "'\\a0\\f141'"}] ; nbsp and ellipsis-h
+   [".btn-opens-more::after" {:content "'...'"}]
 
    ;; form inputs
    ["input[type=date].form-control" {:width (u/em 12)}]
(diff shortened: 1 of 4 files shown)


src/cljs/rems/atoms.cljs
```diff
@@ -7,6 +7,9 @@
 (defn external-link []
   [:i {:class "fa fa-external-link-alt"}])
 
+(defn more-options []
+  [:i {:class "fa fa-ellipsis-h"}])
```
luontola: These icons could also be replaced with a CSS class. What do you think?

Macroz: I think using CSS for these things makes it more magic than it needs to be.

src/cljs/rems/actions/action.cljs
```diff
@@ -49,9 +49,10 @@
 (defn action-button [{:keys [id text class on-click]}]
   [:button.btn.mr-3
    {:id (str id "-action-button")
-    :class (or class "btn-secondary")
+    :class (str (or class "btn-secondary")
+                " more-options")
```
Macroz: How about `.btn-opens-more`?

luontola: Sounds good.

src/clj/rems/css/styles.clj
```diff
@@ -526,6 +526,14 @@
...
     (s/> :.commands "*:not(:first-child)")
     {:margin-left (u/em 0.5)}]
+   [".btn-opens-more::after" {:display "inline-block"
+                              :font-style "normal"
+                              :font-variant "normal"
+                              :text-rendering "auto"
+                              :-webkit-font-smoothing "antialiased"
+                              :font-family "'Font Awesome 5 Free'"
+                              :font-weight 900
+                              :content "'\\a0\\f141'"}] ; nbsp and ellipsis-h
```
Macroz: Hmm, isn't `fa-ellipsis-h` wrong here? We want the `...` and not the middle dots. Also I would put the `...` to the other place as well because the middle dots don't really work there so well.

PR #1178, opened 2019-04-29
Bundling warnings

### LINKED ISSUE #1102: vTHL3: Adding and removing resources to submitted application

Macroz:
Implements bundling warnings from #1102 for both applicant and handler.

Closes #1102


hukka:
A bit oddly, this will give the bundling warning when no resources are selected

hukka:
Could we somehow show the user which resources are ok to add? Now the software is just yelling at the user for not knowing beforehand.

src/clj/rems/workflow/dynamic.clj
```diff
@@ -78,10 +78,22 @@
...
+  "Checks the given catalogue items for validity and merges the errors"
   [catalogue-item-ids injections]
   (apply merge-with into (keep #(invalid-catalogue-item-error % injections) catalogue-item-ids)))
 
+(defn- unbundlable-catalogue-items
+  "Checks that the given catalogue items are bundlable by the given actor."
+  [application catalogue-item-ids actor injections]
+  (let [catalogue-items (map (:get-catalogue-item injections) catalogue-item-ids)
+        handlers (get-in application [:application/workflow :workflow.dynamic/handlers])]
+    (prn :kakka handlers actor)
```
hukka: Dev debug print?

test/clj/rems/workflow/test_dynamic.clj
```diff
@@ -190,6 +191,13 @@
                            {:type :application.command/change-resources
                             :actor applicant-user-id
                             :catalogue-item-ids [1 2]}
+                           injections)))
+      (is (= {:errors [{:type :unbundlable-catalogue-items
+                        :catalogue-item-ids [1 2 3]}]}
+             (fail-command application
+                           {:type :application.command/change-resources
+                            :actor applicant-user-id
+                            :catalogue-item-ids [1 2 3]}
```
hukka: Would it make sense to have separate tests for different form and different workflow? Or is that overkill?

hukka:
While the handler can bundle unbundlable resources, the message in the yellow box is still "The resources can't be bundled into one application because they have different forms or workflows." which implies that it is not possible. Maybe just "Did you mean to bundle these resources? They don't have the same workflow or form" or something like that.

Macroz:
Ready for review again.

PR #1180, opened 2019-04-29
Fix Firefox not showing the last columns on admin catalogue items page

### LINKED ISSUE #1179: Firefox: the last columns on admin catalogue items page are not visible
From #1176

before

luontola:
Closes #1179

This PR makes them visible. Alternatively, using overflow: auto would add a horizontal scroll bar, though you could not see it when it's at the bottom of the page.

In the long term, adjusting the page max width or removing columns would produce a cleaner outcome.

after


luontola:
Chrome splits the words at differently. For example "urn:uuid:00ac3bee-6bef-4c7e-ab91-b9c3b371c3e8" is split into three lines, but Firefox keeps it on a single line.

PR #1187, opened 2019-04-30
rems.poller.email: send emails to handlers for submitted applications

### LINKED ISSUE #1184: Send email to handler on application submit

opqdonut:
closes #1184

### DIFF
--- a/src/clj/rems/poller/email.clj
+++ b/src/clj/rems/poller/email.clj
@@ -43,6 +43,7 @@
       :body (text-format :t.email.application-submitted/message
                          handler
                          (:application/applicant application)
+                         (application-id-for-email application)
                          (link-to-application (:application/id application)))})))
 
 (defn- applicant-and-members [application]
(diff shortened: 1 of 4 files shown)


src/clj/rems/poller/email.clj
```diff
@@ -35,9 +35,19 @@
...
+      :body (text-format :t.email.application-submitted/message
+                         handler
+                         (:application/applicant application)
+                         (application-id-for-email application)
+                         (link-to-application (:application/id application)))})))
+
 (defn- applicant-and-members [application]
-  ;; XXX: the tests depend on the order of members
-  (conj (reverse (sort-by :userid (:application/members application)))
+  (conj (:application/members application)
```
foxlynx: Good that the sorting is moved to test. Just wondering if there is anywhere else that can break because the sorting has been here before. 🙈

opqdonut: :crossed_fingers:

PR #1189, opened 2019-04-30
Upgrade React 16.3 -> 16.8

luontola:
This makes it possible to use React's profiling tools. And anyways the version was about one year old.


project.clj
```diff
@@ -9,6 +9,9 @@
                  [clj-pdf "2.3.1"]
                  [clj-time "0.15.1"]
                  [cljs-ajax "0.8.0"]
+                 [cljsjs/react "16.8.3-0"]
```
opqdonut: could add a comment here that these override the react inside reagent, and can be removed once we upgrade reagent

PR #1198, opened 2019-05-03
move round-based application code to rems.db.applications.legacy

opqdonut:
part 1 out of several


src/clj/rems/db/applications.clj
```diff
@@ -35,106 +34,15 @@
...
-        (and (is-applicant? user-id application)
-             (not= "closed" (:state application))))))
-
-(defn can-withdraw? [user-id application]
-  (assert user-id)
-  (assert application)
-  (and (is-applicant? user-id application)
-       (= (:state application) "applied")))
+;;; Form & field stuff
+;; TODO these should be in rems.db.form or so
```
opqdonut: I'll follow up with a PR for this

PR #1199, opened 2019-05-03
Empty deciders 1188

### LINKED ISSUE #1188: LBR: Deciders should never be allowed to be empty
Check that commenters and deciders is never empty. Otherwise there will be an untranslated error from backend.

- [ ] disable action button if commenters is empty
- [ ] disable action button if deciders is empty

hukka:
closes #1188

### DIFF
--- a/src/cljs/rems/actions/request_comment.cljs
+++ b/src/cljs/rems/actions/request_comment.cljs
@@ -63,12 +63,11 @@
   [{:keys [selected-commenters potential-commenters comment on-set-comment on-add-commenter on-remove-commenter on-send]}]
   [action-form-view action-form-id
    (text :t.actions/request-comment)
-   [[button-wrapper (merge {:id "request-comment"
-                            :text (text :t.actions/request-comment)
-                            :class "btn-primary"
-                            :on-click on-send}
-                           (when (empty? selected-commenters)
-                             {:disabled true}))]]
+   [[button-wrapper {:id "request-comment"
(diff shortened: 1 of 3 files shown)


src/cljs/rems/actions/request_comment.cljs
```diff
@@ -63,10 +63,12 @@
   [{:keys [selected-commenters potential-commenters comment on-set-comment on-add-commenter on-remove-commenter on-send]}]
   [action-form-view action-form-id
    (text :t.actions/request-comment)
-   [[button-wrapper {:id "request-comment"
-                     :text (text :t.actions/request-comment)
-                     :class "btn-primary"
-                     :on-click on-send}]]
+   [[button-wrapper (merge {:id "request-comment"
```
Macroz: Wouldn't it be simpler if you did not do the merge but just set `:disabled` always?

src/cljs/rems/actions/request_decision.cljs
```diff
@@ -62,10 +62,12 @@
   [{:keys [selected-deciders potential-deciders comment on-set-comment on-add-decider on-remove-decider on-send]}]
   [action-form-view action-form-id
    (text :t.actions/request-decision)
-   [[button-wrapper {:id "request-decision"
-                     :text (text :t.actions/request-decision)
-                     :class "btn-primary"
-                     :on-click on-send}]]
+   [[button-wrapper (merge {:id "request-decision"
```
Macroz: See above also here.

PR #1202, opened 2019-05-03
Allow applicant to close a draft application

### LINKED ISSUE #1186: Applicant should be able to close the draft or application.
- [ ] applicant can close the application

luontola:
Closes #1186


luontola:
The command's functionality is tested in `rems.workflow.test-dynamic/test-close` and who is allowed to run the command is declarative data in `rems.application.model/draft-permissions`, so any additional tests would only duplicate the declarative data.

Macroz:
Well, I think a feature like this should not exist without explicit tests. For example, you should have a test for checking that the permissions model contains the relevant permissions for this feature to be available at correct times, Or even better, you could test as an integration or acceptance test, that an applicant can indeed close a draft application without giving a comment e.g. in `test_model.clj` and/or `test_applications.cljs` or elsewhere through API.

luontola:
I tested it manually. The API's schema validation for `rems.application.commands/CloseCommand` and the event schema validation at `rems.db.applications/event->json` would prevent a `nil` comment. The thing that could possibly fail in this situation, is that the UI for some reason sends a `nil` comment when the comment field is not visible. Due to the high cost of end-to-end tests, I don't consider this minor feature worth writing and end-to-end test for it, and we don't really have a way to unit test the UI either.

src/cljs/rems/application.cljs
```diff
@@ -539,13 +539,14 @@
...
 (defn- actions-form [application]
   (let [app-id (:application/id application)
+        see-everything? (contains? (:application/permissions application) :see-everything)
         actions (action-buttons application)
         reload (partial reload! app-id)
         forms [[:div#actions-forms.mt-3
                 [request-comment-form app-id reload]
                 [request-decision-form app-id reload]
                 [comment-form app-id reload]
-                [close-form app-id reload]
```
opqdonut: I find it kinda weird that you use the the `:see-everything` permission as a proxy for showing the comment field. I think it would be cleaner to either a) look at whether the user has the applicant role or b) have a separate close-without-comment permission

luontola: It's the same permission which is used to determine whether the user is allowed to see the comment: https://github.com/CSCfi/rems/blob/bc6436368eba588b1b40b7a73687a9c007fa8be7/src/clj/rems/application/model.clj#L573-L577 Thus I think it logical that the same permission is used - otherwise the user could write a comment which he is not allowed to see.

opqdonut: Ok I'm fine with it if you add a comment. Please also consider renaming the parameter to `close-view` and `close-form` to `show-comment-box?` or something like that.

Issue #1175, opened 2019-04-29
Old versions of attachments are not saved


Macroz:
Old versions of attachments are not saved. To get even a manual diff ("open both and compare") we need to store the older versions (e.g. have rows for each)  then link to those blobs, perhaps from a new event.

PR #1205, opened 2019-05-06
refactor application attachment storage (BREAKING change)

opqdonut:
for #1175


src/clj/rems/db/attachments.clj
```diff
@@ -21,38 +21,40 @@
     (throw (InvalidRequestException. (str "Unsupported content-type: " content-type)))))
 
 (defn save-attachment!
-  [{:keys [tempfile filename content-type]} user-id application-id item-id]
+  [{:keys [tempfile filename content-type]} user-id application-id]
   (check-attachment-content-type content-type)
-  (let [application (applications/get-application user-id application-id)
+  (let [application ((resolve 'rems.db.applications/get-application) user-id application-id) ;; will throw if unauthorized?
```
opqdonut: oh right need to figure out what to do about this cyclic dependency

src/clj/rems/db/applications/legacy.clj
```diff
@@ -122,7 +122,7 @@
            :value
            (cond
              (nil? application-id) ""
-             (= "attachment" (:type field)) (or (:filename (db/get-attachment query-params)) "")
+             ;;(= "attachment" (:type field)) (assert false "attachments not supported")
```
opqdonut: need to figure out what to do about this case

src/clj/rems/api/applications.clj
```diff
@@ -197,8 +201,8 @@
       :summary "Get an attachment for a field in an application"
```
luontola: No more "for a field"

src/clj/rems/db/attachments.clj
opqdonut: I'll move the access control to the api layer

src/clj/rems/db/applications/legacy.clj
luontola: It's used only in application conversion and in test data. Since there is no migration for attachments, it should be OK to let the conversion crash if there are attachments.

src/clj/rems/db/attachments.clj
```diff
@@ -21,38 +22,40 @@
...
-                            :form (get-in application [:application/form :form/id])
-                            :item item-id})))
-
-(defn get-attachment [user-id application-id field-id]
-  (let [application (applications/get-application user-id application-id)
-        form-id (get-in application [:application/form :form/id])]
-    (when-let [attachment (db/get-attachment {:item field-id
-                                              :form form-id
-                                              :application application-id})]
+(defn get-attachment [user-id application-id attachment-id]
```
luontola: `get-attachment-data` or `download-attachment` might be a better name, to differentiate between `get-attachments-for-application` which only returns the meta data.

test/clj/rems/api/test_applications.clj
```diff
@@ -434,31 +434,37 @@
...
+          (let [response (-> (read-request id)
+                             (authenticate api-key user-id)
+                             handler
+                             assert-response-is-ok)]
+            (is (= (slurp testfile) (slurp (:body response))))))
+        (testing "and retrieving it as non-applicant"
+          (let [response (-> (read-request id)
+                             (authenticate api-key "carl")
+                             handler)]
+            (is (response-is-forbidden? response))))))
```
luontola: There doesn't seem to be a test for attachment not found (i.e. wrong attachment ID). That would be a good addition to make sure that the `if attachment` code works right.

src/cljs/rems/application.cljs
```diff
@@ -195,37 +195,19 @@
...
             :handler (fn [response]
-                       ;; if we just remove the attachment from the backend but
-                       ;; don't save the field, the application will be left in an
-                       ;; inconsistent state (referring to a nonexistant attachment)
-                       ;; TODO: save only the value for the attachment field
                        (if (:success response)
-                         (rf/dispatch [::save-application (text :t.form/save)])
+                         (do
+                           (rf/dispatch [::set-field-value field-id (str (:id response))])
+                           (rf/dispatch [::save-application (text :t.form/upload)]))
```
opqdonut: this is a race condition

opqdonut: no it isn't

PR #1208, opened 2019-05-07
diff for attachment fields: show link to old and new file

opqdonut:
for #1175 and #978


opqdonut:
attachment-diff

luontola:
It needs to handle the case of adding or removing an attachment:

liite

src/cljs/rems/fields.cljs
```diff
@@ -205,19 +205,19 @@
...
-        filename-field [:div.field
-                        [:a.btn.btn-secondary.mr-2
-                         {:href (str "/api/applications/attachment/" value)
-                          :target :_new}
-                         filename " " (external-link)]]
+        link (fn [attachment-id filename]
+               [:div.field
+                [:a.btn.btn-secondary.mr-2
+                 {:href (str "/api/applications/attachment/" attachment-id)
+                  :target :_new}
```
luontola: Should be _blank https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target

opqdonut:
attachment-diff-empty

PR #1220, opened 2019-05-08
fold rems.workflow.dynamic/write-model into rems.application.model

opqdonut:
to simplify code

### DIFF
--- a/src/clj/rems/workflow/dynamic.clj
+++ b/src/clj/rems/workflow/dynamic.clj
@@ -158,7 +158,7 @@
            :application/comment (:comment cmd)})))
 
 (defn- actor-is-not-decider-error [application cmd]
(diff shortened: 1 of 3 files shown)


src/clj/rems/workflow/dynamic.clj
```diff
@@ -11,43 +11,10 @@
...
-  (-> application
-      (update ::latest-comment-request-by-user merge (zipmap (:application/commenters event)
-                                                             (repeat (:application/request-id event))))))
-
-(defmethod write-model :application.event/commented
-  [application event]
-  (-> application
-      (update ::latest-comment-request-by-user dissoc (:event/actor event))))
-
 (defn apply-events [application events]
```
opqdonut: I'll follow up with a PR getting rid of `apply-events`

PR #1221, opened 2019-05-08
get rid of rems.workflow.dynamic/apply-events

opqdonut:
on top of #1220


src/clj/rems/workflow/dynamic.clj
```diff
@@ -9,49 +9,7 @@
...
-
-(defn apply-events [application events]
-  (reduce (fn [application event] (-> application
-                                      (model/application-view event)
-                                      (write-model event)))
-          application
-          events))
-
-;;; Command handlers
+;; TODO move all of this to rems.application.commands?
```
opqdonut: what do you think?

Issue #845, opened 2019-01-16
vTHL4: Free commenting for dynamic workflow applications


foxlynx:
THL wants sometimes to add comments to application that are not visible to applicant. They are wishing a checkbox or similar if the comment can be used to the applicant.

Internal comment: everyone in THL can see who is a handler, or who is part of the handling (i.e. commenter, decider)
General comment: also applicant can see

- [ ] new action to leave a free comment
- [ ] possibility of selecting if the comment is internal or public (internal being default)
- [ ] test that the applicant doesn't get the comment data from API for private comments
- [ ] test that unrelated users don't get anything for the application

hukka:
If these would be different event types, it would be trivial to implement the hiding in `api.applications/hide-sensitive-events` and `hide-sensitive-dynamic-events`

hukka:
This is separate from current commenting functionality. This is related to non-dynamic roles only, does not thread and is not related to comment requests.
Some roles (handler for now) can leave comment events, that are either visible or not to the applicant.

luontola:
Support for non-requested comments (i.e. make request-id optional) would also solve this kludge: https://github.com/CSCfi/rems/blob/ec6dac0/src/clj/rems/migrations/convert_to_dynamic_applications.clj#L77-L78

PR #1223, opened 2019-05-08
Add option to make comments public and free

hukka:
Backend part of #845


src/clj/rems/application/events.clj
```diff
@@ -29,7 +29,8 @@
 (s/defschema CommentedEvent
   (assoc EventBase
          :event/type (s/enum :application.event/commented)
-         :application/request-id s/Uuid
+         :application/request-id (s/maybe s/Uuid)
+         :application/public s/Bool
```
opqdonut: hmm optional in command but not in event. I would've assumed the opposite. Now event compatibility is broken for no reason.

src/clj/rems/workflow/dynamic.clj
```diff
@@ -229,6 +229,7 @@
              ;; In the future this might change so that commenters can freely continue to comment
              ;; on any request they have gotten.
              :application/request-id last-request-for-actor
+             :application/public (true? (:public cmd))
```
opqdonut: more like `:event/public`, right?

test/clj/rems/application/test_model.clj
```diff
@@ -739,7 +739,7 @@
 
 (deftest test-calculate-permissions
   ;; TODO: is this what we want? wouldn't it be useful to be able to write more than one comment?
```
luontola: outdated comment

test/clj/rems/api/test_applications.clj
```diff
@@ -269,7 +269,22 @@
                   request-event (get-in application [:application/events eventcount])
                   comment-event (get-in application [:application/events (inc eventcount)])]
               (is (= (:application/request-id request-event)
-                     (:application/request-id comment-event)))))))
+                     (:application/request-id comment-event)))))
+          (testing "applicant can see public comments only"
```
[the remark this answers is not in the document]
Macroz: Since it is critical that public comments are not shown to the applicant, at least that case should be tested in the API level acceptance tests too.

Macroz:
Would it not be much clearer that the request comment and comment are completely separate from the free comment functionality? There was no requirement to change the existing functionality but to be able to add notes for handlers themselves and possibly also to communicate with the applicants. Therefore it would be much clearer to me that the comment requests stays as completely separate feature.

PR #1227, opened 2019-05-09
Change approved resources

### LINKED ISSUE #1104: vTHL4: Adding and removing resources to approved application
(split out from #842)

- [ ] when the application has been approved, resources can be added or removed by the handler
- [ ] entitlements should update

Macroz:
Closes #1104 

I had to rewrite the entitlement giving logic, but fortunately we have some coverage in the acceptance tests and I added tests for changing resources there.

There is possibly some refactoring opportunity still left.

### DIFF
--- a/src/clj/rems/db/entitlements.clj
+++ b/src/clj/rems/db/entitlements.clj
@@ -1,6 +1,7 @@
 (ns rems.db.entitlements
   "Creating and fetching entitlements."
   (:require [clj-http.client :as http]
+            [clojure.set :refer [union]]
             [clojure.string :refer [join]]
             [clojure.tools.logging :as log]
             [rems.auth.util :refer [throw-forbidden]]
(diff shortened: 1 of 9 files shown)


src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+          past-members (set (map :userid (:application/past-members application)))
+          application-resources (->> application
+                                     :application/resources
+                                     (map :resource/id)
+                                     set)
+          application-entitlements (->> (db/get-entitlements {:application app-id :is-active? true})
+                                        (group-by :userid)
+                                        (map (fn [[userid rows]]
+                                               [userid (set (map :resourceid rows))]))
+                                        (into {}))
```
opqdonut: this could be a helper function `get-entitlements-by-user`

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+          user-entitlements (into {}
+                                  (for [userid (union members past-members)]
+                                    [userid (or (application-entitlements userid) #{})]))
+          entitlements-to-add (->> (for [[userid resource-ids] user-entitlements
```
opqdonut: you could get rid of `user-entitlements` by writing this as:
```
(for [userid (union members past-members)]
  (let [resource-ids (get application-entitlements userid #{})]
    ...
```

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+                                         :when (and (= :application.state/approved (:application/state application))
+                                                    (contains? members userid)
+                                                    (accepted-licenses? application userid))
+                                         resource-id application-resources
+                                         :when (not (contains? resource-ids resource-id))]
```
opqdonut: lines 91-15 could be written as:
```
resource-id application-resources
:when (is-entitled? userid application resource-id)

File: src/clj/rems/db/entitlements.clj [hunk 4]
```diff
@@ -63,52 +64,62 @@
...
+                                     {userid #{resource-id}})
+                                   (apply merge-with union))
+          entitlements-to-remove (->> (for [[userid resource-ids] user-entitlements
+                                            resource-id resource-ids
+                                            :when (or (= :application.state/closed (:application/state application))
+                                                      (not (contains? members userid))
+                                                      (not (accepted-licenses? application userid))
+                                                      (not (contains? application-resources resource-id)))]
```
opqdonut: similarly we could use `:when (not (is-entitled? ...))` here

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+                             (concat (keys entitlements-to-add)
+                                     (keys entitlements-to-remove)))]
       (when (seq members-to-update)
-        (doseq [user-id members-to-update]
-          (log/info "granting entitlements on application" app-id "to" user-id)
-          (doseq [[resource-id ext-id] (->> (:application/resources application)
-                                            (map (juxt :resource/id :resource/ext-id)))]
+        (log/info "updating entitlements on application" app-id)
+        (doseq [[userid resource-ids] entitlements-to-add]
+          (log/info "granting entitlements on application" app-id "to" userid "resources" resource-ids)
```
opqdonut: the body of this doseq could be a helper function

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+          (doseq [resource-id (sort resource-ids)]
             (db/add-entitlement! {:application app-id
-                                  :user user-id
+                                  :user userid
+                                  :resource resource-id})
+            (post-entitlements :add (db/get-entitlements {:application app-id :user userid :resource resource-id}))))
+        (doseq [[userid resource-ids] entitlements-to-remove]
+          (log/info "revoking entitlements on application" app-id "to" userid "resources" resource-ids)
```
opqdonut: likewise

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,62 @@
...
+          (doseq [resource-id (sort resource-ids)]
+            (db/end-entitlement! {:application app-id
+                                  :user userid
                                   :resource resource-id})
-            (post-entitlements :add (db/get-entitlements {:application app-id :user user-id :resource ext-id})))))
+            (post-entitlements :remove (db/get-entitlements {:application app-id :user userid :resource resource-id})))))
       members-to-update)))
```
opqdonut: is the return value still used?

test/clj/rems/db/test_entitlements.clj
```diff
@@ -195,36 +197,56 @@
...
+            (is (= 6 (count (stub/recorded-requests server))))
             (testing "db"
               (is (= [[uid "resource1"] [uid "resource2"]]
                      (map (juxt :userid :resid) (db/get-entitlements {:application app-id :is-active? true})))))
             (testing "POST"
-              (let [data (nth (stub/recorded-requests server) 4)
-                    target (:path data)
-                    body (cheshire/parse-string (get-in data [:body "postData"]) keyword)]
-                (is (= "/remove" target))
+              (let [data (take 2 (drop 4 (stub/recorded-requests server)))
```
opqdonut: this test has grown pretty big. could we split it into separate tests or perhaps make some helpers for checking the entitlements & POSTs? also, is there a way to reset the stub to get rid of the magic `(take 2 (drop 4 ...))`?

src/clj/rems/db/entitlements.clj
Macroz: No but I think it's now slightly better when it's explicitly set at something instead of being random return value

test/clj/rems/db/test_entitlements.clj
Macroz: I refactored the implementation to use a helper so this tracking is not required anymore.

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,71 @@
...
+          current-members (set (concat (map :userid (:application/members application))
+                                       [(:application/applicant application)]))
+          past-members (set (map :userid (:application/past-members application)))
+          application-state (:application/state application)
+          application-resources (->> application
+                                     :application/resources
+                                     (map :resource/id)
+                                     set)
+          application-entitlements (get-entitlements-by-user application-id)
+          is-entitled? (fn [userid application resource-id]
```
opqdonut: no need to pass in application

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,71 @@
...
+                                   (apply merge-with union))
+          entitlements-to-remove (->> (for [userid (union current-members past-members)
+                                            :let [resource-ids (entitlements-by-user userid)]
+                                            resource-id resource-ids
+                                            :when (not (is-entitled? userid application resource-id))]
+                                        {userid #{resource-id}})
+                                      (apply merge-with union))
+          members-to-update (set
+                             (concat (keys entitlements-to-add)
+                                     (keys entitlements-to-remove)))]
```
opqdonut: [golf] `(keys (merge entitlementsto-add entitlements-to-remove))`

src/clj/rems/db/entitlements.clj
```diff
@@ -63,52 +64,71 @@
...
+      (when (seq members-to-update)
```
opqdonut: hmm if you split out the computation of `entitlements-to-{add,remove}` you could write unit tests for it

src/clj/rems/poller/common.clj
```diff
@@ -25,7 +25,7 @@
         (throw (InterruptedException.)))
       (try
         ;; TODO: add proper monitoring for pollers and caches
-        (log/info name-kw "processing event" (:event/id event))
+        (log/info name-kw "processing event" (:event/id event) (:event/type event))
```
opqdonut: consider `(select-keys event [:event/id :event/type])`

src/clj/rems/db/entitlements.clj
Macroz: I can't see an easy way to split them out that would not duplicate the fetching logic. I'll leave this for later PRs.

src/clj/rems/poller/common.clj
Macroz: There is no guaranteed order in logs then?

opqdonut: ok good point

Issue #1230, opened 2019-05-09
vTHL4: Truncate long application description in table


opqdonut:
Application tables that show the description column should show long descriptions well.

- [x] figure out a suitable max length
- [x] truncate description and show "..." in the end (i.e. with CSS styling)
- [x] show full description in tooltip while hovering (and application page)

See also #1174

PR #1241, opened 2019-05-13
rems.application-list: truncate long descriptions

### LINKED ISSUE #1174: LBR: Applications table layout breaks when userid is too long
The View button is rendered outside the visible area.

Example:

application-table-wide

See also #1230

opqdonut:
fixes #1174

### DIFF
--- a/src/cljs/rems/application_list.cljs
+++ b/src/cljs/rems/application_list.cljs
@@ -43,7 +43,11 @@
                  :header #(text :t.actions/id)}
    :id {:value :application/id
         :header #(text :t.actions/id)}
-   :description {:value :application/description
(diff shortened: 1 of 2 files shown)


src/cljs/rems/application_list.cljs
```diff
@@ -17,6 +17,14 @@
        (map localized)
        (str/join ", ")))
 
+(defn- format-description [app]
+  [:div {:style {:overflow :hidden
```
Macroz: I'd rather put these in `styles.clj`.

opqdonut: done. also simplified the styling a bit

opqdonut:
we can factor out the overflow style later when we know what we need

opqdonut:
this was for #1230

PR #1242, opened 2019-05-13
no need to :require exception namespaces

opqdonut:


opqdonut:
turns out they're needed to get the compilation order right in AOT

PR #1243, opened 2019-05-13
Visible focus indicators

### LINKED ISSUE #1213: A11Y: Make focus clearly visible
- [x] language selection links are currently hiding the focus outline (the only focus indicator is the tiny underlining)
- [x] the focus outline of secondary buttons is very faint

Required by 2.4.7 Focus Visible (Level AA) https://www.w3.org/WAI/WCAG21/quickref/

Part of #1172

luontola:
Closes #1213

- Adds focus indicators to language switcher and modal close button
- Makes the focus indicators of secondary buttons more visible
- Makes the focus indicators be the same on all browsers
- Add aria label for close dialog button

### DIFF
--- a/src/cljs/rems/language_switcher.cljs
+++ b/src/cljs/rems/language_switcher.cljs
@@ -6,8 +6,8 @@
 
 (defn lang-link-classes [current-language language]
   (if (= current-language language)
-    "nav-link active"
-    "nav-link"))
+    "btn btn-link active"
+    "btn btn-link"))
 
 (defn language-switcher
   "Language switcher widget"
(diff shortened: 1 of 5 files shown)


src/cljs/rems/language_switcher.cljs
```diff
@@ -18,13 +18,9 @@
...
             (let [lang-str (str/upper-case (name language))]
               [:form.inline
-               [:a {:href "#"
-                    :on-click (fn [event]
-                                (.preventDefault event)
-                                (rf/dispatch [:set-current-language language]))
-                    :class (lang-link-classes current-language language)
-                    :style {:padding 0}}
-                lang-str]])))))
+               #_(anti-forgery-field)
```
opqdonut: where did this come from?

luontola: It has been there since the beginning, 34e440ff. I had removed it in this PR, but then I reverted my changes to that file, so the mystery line lives on.

PR #1244, opened 2019-05-13
Submitted column

### LINKED ISSUE #1229: vTHL4: Show submitted instead of created timestamp to handler
THL wants the handler to see the submitted timestamp in tables, instead of the created timestamp. It's not useful for the handler to see when the applicant started creating the application but when it was sent.

- [x] Show two timestamp columns on the actions page: submitted and last activity instead of created and last activity.

Macroz:
Closes #1229 

- Shows submitted for applicant, reporter and handler. For the applicant it's shown in addition to created.
- Submitted is tracked from the very first submit. 
- Refactors `application-list/component` to remove extra layers.
- Fix the heading to be h1. That's the future with A11Y

### DIFF
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -120,7 +120,7 @@
      [:h4 (text :t.catalogue/continue-existing-application)]
      [application-list/component
       ;; TODO: use application-list/visible-columns like on other pages?
-      {:visible-columns application-list/+draft-columns+
+      {:visible-columns [:resource :last-activity :view]
        :items drafts}]]))
 
 (defn catalogue-page []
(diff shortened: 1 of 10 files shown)


src/cljs/rems/catalogue.cljs
```diff
@@ -120,7 +120,7 @@
      [:h4 (text :t.catalogue/continue-existing-application)]
      [application-list/component
       ;; TODO: use application-list/visible-columns like on other pages?
```
opqdonut: unnecessary TODO?

Macroz: Yes I think I just made it redundancy

opqdonut:
sorry, conflicts with my change

PR #1250, opened 2019-05-15
Fix/entitlements api 1249

### LINKED ISSUE #1249: Entitlements API does not work for owner and reporter
Entitlements api should:
* [ ] return entitlements for a user when called as an owner (only works for handler now)
* [ ] return entitlement for a user when called as an reporter  (only works for handler now)

Macroz:
Closes #1249

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -65,7 +65,7 @@
   {:resource s/Str
    :application-id s/Num
    :start s/Str
-   :end s/Str
+   :end (s/maybe s/Str)
    :mail s/Str})
 
 (s/defschema UpdateStateCommand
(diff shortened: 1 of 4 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -65,6 +65,7 @@
   {:resource s/Str
    :application-id s/Num
    :start s/Str
+   :end s/Str
```
opqdonut: the tests don't test `:end` at all. what value do we get when the db doesn't have an end date?

Macroz: I think I'll add some date checking here because this is a very valuable API anyway. Even though it may duplicate something of the db level tests.

Macroz: I'll add improved acceptance tests in another PR because we want this fix to be deployed for tomorrow.

PR #1256, opened 2019-05-16
Kaocha

hukka:


tests.edn
```diff
@@ -0,0 +1,16 @@
...
+                     :test-paths ["src" "test"]
+                     :skip-meta [:browser :integration]}
+                    {:id :browser
+                     :ns-patterns ["rems"]
+                     :test-paths ["src" "test"]
+                     :focus-meta [:browser]}
+                    {:id :integration
+                     :ns-patterns ["rems"]
+                     :test-paths ["src" "test"]
+                     :focus-meta [:integration]}]
```
opqdonut: this is different from the previous behaviour of integration

hukka: It is, because by default (without any flags) kaocha runs all test suites, and thus we would end up running unit tests twice, in two suites.

We need to either then stop using kaocha without any flags and always specify just one suite, give multiple suites to kaocha or use lein aliases.

opqdonut: oh right, kaocha uses suites and not selectors so these kinda need to be mutually exclusive

opqdonut:
let's merge this, we can always revert back to eftest if we want

PR #1261, opened 2019-05-16
Performance improvement for changing language

### LINKED ISSUE #1260: Changing the language on catalogue page is very slow
It takes over 20 seconds to change the language, which is more than 10 times what it takes to load the page.

luontola:
Closes #1260

### DIFF
--- a/src/cljs/rems/spa.cljs
+++ b/src/cljs/rems/spa.cljs
@@ -114,15 +114,21 @@
  (fn [db [_ theme]]
    (assoc db :theme theme)))
 
+(declare render-empty-page mount-components)
 (reg-event-fx
  :set-current-language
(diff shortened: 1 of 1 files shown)


hukka:
This is quite weird. My initial guess that leaves get rendered many times (first their language changes, then parent changes and re-renders children) wasn't correct. But it did accidentally reveal other bugs!

@luontola will redo the table component to fix the bug of catalogue items not changing their title. Let's see what happens to perf after that.

If this is a pattern in re-frame, then we should raise it upstream and remember that cross cutting subscription changes are always better to be done by clearing dom first, which is quite against the react philosophy.

luontola:
Superseded by #1277

PR #1269, opened 2019-05-17
Change text-highlight color from lightened color4 to color3

okahilak:
The vague idea here is that color3 could be such that it has good
contrast with many different backgrounds, making it useful as a
text color in normal-sized text, whereas color4 would not need to
have that. Maybe color3/4 could be renamed in the future to reflect
those properties.


opqdonut:
Your comment here sounds like it should be a comment in the code :)

okahilak:
Done.

PR #1276, opened 2019-05-20
Define button type to avoid browser-specific defaults

luontola:
The default is type=submit, but we don't really use HTML forms which could be submitted. Also IIRC, IE had some different defaults.

### DIFF
--- a/src/cljs/rems/actions/action.cljs
+++ b/src/cljs/rems/actions/action.cljs
@@ -7,9 +7,9 @@
 
 (defn button-wrapper [{:keys [text class] :as props}]
   [:button.btn
-   (merge
-    {:class (or class :btn-secondary)}
-    (dissoc props :class :text))
+   (merge {:type "button"
+           :class (or class :btn-secondary)}
+          (dissoc props :class :text))
    text])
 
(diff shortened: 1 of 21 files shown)


src/cljs/rems/actions/action.cljs
```diff
@@ -7,17 +7,18 @@
 
 (defn button-wrapper [{:keys [text class] :as props}]
   [:button.btn
-   (merge
-    {:class (or class :btn-secondary)}
-    (dissoc props :class :text))
+   (merge {:type "button"
```
Macroz: I'd use keywords here not strings.

src/cljs/rems/actions/action.cljs
```diff
@@ -7,17 +7,18 @@
...
+          (dissoc props :class :text))
    text])
 
 (defn collapse-action-form [id]
   (.collapse (js/$ (str "#" (action-collapse-id id))) "hide"))
 
 (defn cancel-action-button [id]
   [:button.btn.btn-secondary
-   {:id (str "cancel-" id)
+   {:type "button"
```
Macroz: Keyword also here and elsewhere.

PR #1277, opened 2019-05-20
Improve catalogue performance with new table component

### LINKED ISSUE #1260: Changing the language on catalogue page is very slow

luontola:
Closes #1260 and makes filtering fast. Later PRs will start using it also on other pages.

### DIFF
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -138,7 +138,15 @@
   (let [language @(rf/subscribe [:language])
         loading-catalogue? @(rf/subscribe [::loading-catalogue?])
         drafts @(rf/subscribe [::draft-applications])
-        loading-drafts? @(rf/subscribe [::loading-drafts?])]
+        loading-drafts? @(rf/subscribe [::loading-drafts?])
+        table-spec {:id :catalogue
+                    :columns [{:key :name
+                               :title (text :t.catalogue/header)
+                               :sortable? true
+                               :filterable? true}
+                              {:key :commands}]
+                    :rows ::catalogue-table-data
+                    :default-sort-column :name}]
     [:div
      [document-title (text :t.catalogue/catalogue)]
      (if (or loading-catalogue? loading-drafts?)
@@ -147,12 +155,8 @@
         [draft-application-list drafts]
         [:h2 (text :t.catalogue/apply-resources)]
         [cart/cart-list-container language]
-        [table2/filter-field ::catalogue-table-data]
-        [table2/table ::catalogue-table-data [{:key :name
-                                               :title "Resource"
-                                               :sortable? true
-                                               :filterable? true}
-                                              {:key :commands}]]])]))
+        [table2/filter-field table-spec]
+        [table2/table table-spec]])]))
 
 (defn guide []
   [:div
--- a/src/cljs/rems/catalogue.cljs
+++ b/src/cljs/rems/catalogue.cljs
@@ -1,5 +1,6 @@
 (ns rems.catalogue
-  (:require [re-frame.core :as rf]
+  (:require [clojure.string :as str]
+            [re-frame.core :as rf]
             [rems.application-list :as application-list]
             [rems.application-util :refer [form-fields-editable?]]
             [rems.atoms :refer [external-link document-title document-title]]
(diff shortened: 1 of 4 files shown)


src/cljs/rems/table2.cljs
```diff
@@ -0,0 +1,134 @@
+(ns rems.table2
```
opqdonut: I guess the plan is to replace rems.table with this ns?

src/cljs/rems/table2.cljs
```diff
@@ -0,0 +1,134 @@
...
+    (rf/subscribe [::filtering spec])])
+ (fn [[rows filtering] [_ spec]]
+   (let [filters (str/lower-case (str (:filters filtering)))
+         columns (->> (:columns spec)
+                      (filter :filterable?))]
+     (->> rows
+          (map (fn [row]
+                 ;; performance optimization: hide DOM nodes instead of destroying them
+                 (assoc row ::display-row? (display-row? row columns filters))))))))
+
```
opqdonut: I think I like how all the logic is in the subscriptions and only the views are left below.

But now this component isn't reusable, right? You can't have multiple table2s on the same page?

src/cljs/rems/table2.cljs
```diff
@@ -0,0 +1,134 @@
...
+(defn table [spec]
+  (let [rows @(rf/subscribe [::sorted-and-filtered-rows spec])
+        language @(rf/subscribe [:language])]
+    [:div.table-border
+     [:table.rems-table {:class (:id spec)}
+      [:thead
+       [table-header spec]]
+      [:tbody {:key language} ; performance optimization: rebuild instead of update existing components
+       (for [row rows]
+         ^{:key (:key row)} [table-row row spec])]]]))
```
opqdonut: no guide?

Macroz: definitely needed

src/cljs/rems/table2.cljs
Macroz: I guess this is intended to be fixed in later PRs

src/cljs/rems/catalogue.cljs
```diff
@@ -83,36 +75,28 @@
...
-           :class "catalogue"}
-          (when sorting {:sorting sorting})
-          (when filtering {:filtering filtering}))])
+(rf/reg-sub
+ ::catalogue-table-rows
+ (fn [_ _]
+   [(rf/subscribe [::catalogue])
+    (rf/subscribe [:language])])
+ (fn [[catalogue language] _]
+   (->> catalogue
```
Macroz: Threading is unnecessary here?

src/cljs/rems/catalogue.cljs
```diff
@@ -124,10 +108,17 @@
...
   (let [language @(rf/subscribe [:language])
-        catalogue @(rf/subscribe [::catalogue])
         loading-catalogue? @(rf/subscribe [::loading-catalogue?])
         drafts @(rf/subscribe [::draft-applications])
-        loading-drafts? @(rf/subscribe [::loading-drafts?])]
+        loading-drafts? @(rf/subscribe [::loading-drafts?])
+        table-spec {:id :catalogue
+                    :columns [{:key :name
+                               :title (text :t.catalogue/header)
+                               :sortable? true
```
Macroz: Shouldn't the default be filterable and sortable?

src/cljs/rems/catalogue.cljs
```diff
@@ -124,10 +108,17 @@
 
 (defn catalogue-page []
   (let [language @(rf/subscribe [:language])
-        catalogue @(rf/subscribe [::catalogue])
         loading-catalogue? @(rf/subscribe [::loading-catalogue?])
         drafts @(rf/subscribe [::draft-applications])
-        loading-drafts? @(rf/subscribe [::loading-drafts?])]
+        loading-drafts? @(rf/subscribe [::loading-drafts?])
+        table-spec {:id :catalogue
```
Macroz: spec is very much overloaded with Clojure spec so I'd use definition as earlier or something else entirely.

luontola:
The plan is to replace table with table2 and then rename table2 to table.

src/cljs/rems/table2.cljs
luontola: Yes

src/cljs/rems/catalogue.cljs
luontola: It will be figured out later once this table is used in more places and the duplication is more obvious.

src/cljs/rems/table2.cljs
luontola: All the subcriptions are parameterized with a unique ID, so it should be reusable unless I have fundamentally misunderstood re-frame's subscriptions.

Issue #1255, opened 2019-05-16
LBR: Embed Shibboleth EDS to the login page


jaakkocsc:
Currently (in their production service) the login page only shows Haka-login and EDS is provided by a separate link below Haka. The EDS should be directly under Haka option. THL has this option already in their REMS1 production instance (https://thl-biobank.elixir-finland.org/login)

opqdonut:
These seem to be the relevant instructions: https://wiki.eduuni.fi/display/CSCHAKA/Embedded+DS

opqdonut:
#1282 adds the option for embedding an additional login widget as an iframe. This can be used together with the `:extra-static-resources` config option to deploy EDS. See CSCfi/rems-deploy#55 for an example of that.

foxlynx:
Does not work with iFrame, so we used link instead. There is an issue on EDS for this: https://issues.shibboleth.net/jira/browse/EDS-29

foxlynx:
Martin's feedback:

> Please change the title: "IDP select test bed" does not sound too ensuring 
> Could you add clarin.eu and eduuni (as configured in production)

Moving the issue back to backlog.

PR #1282, opened 2019-05-21
configuration option for embedding a login widget in an iframe

opqdonut:
for #1255


opqdonut:
I guess this is no longer needed since EDS isn't compatible with an iframe (see #1255)

PR #1286, opened 2019-05-22
Use rems.table2 for application lists

luontola:
After this PR the administration pages still use the old rems.table, so they will be updated next.

Now there is probably enough duplication to see how to refactor the column and row definitions to simplify the default cases.


luontola:
I'm thinking about refactoring the row definitions to have the following defaults:

- `:value` is used only as a default value for the other keys
- `:sort-value` default is `(:value %)`
- `:display-value` default is `(:value %)`
- `:filter-value` default is `(str/lower-case (str (:display-value %)))` at least until we have filtering for non-text values (e.g. date ranges)
- `:td` default is `[:td {:class column-key} (:display-value %)]`

opqdonut:
those defaults sound good to me

PR #1287, opened 2019-05-22
Minor refactorings

Macroz:
I had a couple notes in my scratch file to do minor refactorings which I now did

### DIFF
--- a/src/cljs/rems/administration/create_workflow.cljs
+++ b/src/cljs/rems/administration/create_workflow.cljs
@@ -58,7 +58,7 @@
        (not (str/blank? (:title request)))
        (case (:type request)
          :auto-approve true
-         :dynamic (not (empty? (:handlers request)))
+         :dynamic (seq (:handlers request))
          nil false)))
 
 (defn build-create-request [form]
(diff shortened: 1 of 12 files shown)


src/cljs/rems/administration/create_workflow.cljs
```diff
@@ -58,7 +58,7 @@
        (not (str/blank? (:title request)))
        (case (:type request)
          :auto-approve true
-         :dynamic (not (empty? (:handlers request)))
+         :dynamic (seq (:handlers request))
```
opqdonut: this should return a boolean. IMO `(seq ...)` is a valid replacement for `(not (empty? ...))` only when the return value is used immediately

test/clj/rems/api/testing.clj
```diff
@@ -53,7 +53,7 @@
 
 (defn coll-is-not-empty? [data]
   (and (coll? data)
-       (not (empty? data))))
+       (seq data)))
```
opqdonut: note that now this function returns `false`, `nil`, or a sequence. I'd prefer it to return a boolean.

src/cljs/rems/administration/create_workflow.cljs
Macroz: I've reordered the statements so that it is not returned.

test/clj/rems/api/testing.clj
Macroz: I refactored this so that the duplication is removed and therefore also this problem.

PR #1288, opened 2019-05-23
Minor refactorings 2

Macroz:
Continue making minor improvements (see also #1287)

### DIFF
--- a/test/clj/rems/api/testing.clj
+++ b/test/clj/rems/api/testing.clj
@@ -51,13 +51,12 @@
 (defn response-is-not-found? [response]
   (= 404 (:status response)))
 
-(defn coll-is-not-empty? [data]
-  (and (coll? data)
-       (seq data)))
-
 (defn coll-is-empty? [data]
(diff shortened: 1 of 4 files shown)


test/clj/rems/api/testing.clj
```diff
@@ -51,14 +51,13 @@
...
-(defn coll-is-not-empty? [data]
-  (and (coll? data)
-       (seq data)))
-
 (defn coll-is-empty? [data]
   (and (coll? data)
        (empty? data)))
 
+(defn coll-is-not-empty? [data]
+  (not (coll-is-empty? data)))
```
luontola: This makes the behaviour of non-coll arguments weird:

```
(coll-is-empty? [])
=> true
(coll-is-empty? nil)
=> false
(coll-is-empty? "")
=> false

(coll-is-not-empty? [])
=> false
(coll-is-not-empty? nil)
=> true
(coll-is-not-empty? "")
=> true
```

PR #1301, opened 2019-05-28
rems.db.test-data: always create forms with form/create-form!

opqdonut:
so that they get stored in both application_form and form_template
tables

follow-up for #1299


opqdonut:
related to #913

PR #1303, opened 2019-05-29
migrations: generate missing form_template rows

opqdonut:
for #913


opqdonut:
We have one on the API layer. Might make sense to use it "through the stack".

luontola:
Sounds good. Just use the same schema at the persistence layer.

PR #1305, opened 2019-05-29
misc form code cleanup

opqdonut:
for #913


opqdonut:
probably easiest to review commit-by-commit

PR #1306, opened 2019-05-29
Use HTML element roles

### LINKED ISSUE #1212: A11Y: Use HTML element roles
- [x] use `<h1>` as the top-level heading of each page; currently some pages use `<h2>` (#1247)
- [x] use HTML heading elements for all headings; currently form section headings are only divs (#1306)
- [x] use `<nav>` for the whole navigation menu; currently the sign out link is outside it (#1306)
  - `<nav role="navigation">` is redundant and can be removed
- [x] use `<main>` for the page content (#1259)

https://a11yproject.com/posts/aria-landmark-roles/

Required by 4.1.2 Name, Role, Value (Level A) https://www.w3.org/WAI/WCAG21/quickref/

Part of #1172

luontola:
Closes #1212

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -540,8 +540,7 @@
    [:.navbar-flex {:display "flex"
                    :flex-direction "row"
                    :justify-content "space-between"
-                   :min-width "100%"}
-    ["> .navbar:first-child" {:flex 1}]]
+                   :min-width "100%"}]
    [(s/> :.form-actions "*:not(:first-child)")
     (s/> :.commands "*:not(:first-child)")
     {:margin-left (u/em 0.5)}]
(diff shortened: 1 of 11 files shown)


src/clj/rems/css/styles.clj
```diff
@@ -541,7 +541,7 @@
                    :flex-direction "row"
                    :justify-content "space-between"
                    :min-width "100%"}
-    [:nav {:flex 1}]]
+    ["> .navbar:first-child" {:flex 1}]]
```
Macroz: Is it possible to avoid this complicated selector?

luontola: What do you propose? Some options are adding a new class to the div and using that, or moving the flex:1 to a style tag.

Macroz: Maybe the child component could have a class (e.g. `d-flex`), that would be clearer.

luontola: Done

PR #1316, opened 2019-06-04
Focus the search field when it's opened

### LINKED ISSUE #1311: Other: Search is not focused when opening
The search input field is not focused when opening search.

- [x] focus the search text input when after opening search

luontola:
Closes #1311

### DIFF
--- a/src/cljs/rems/focus.cljs
+++ b/src/cljs/rems/focus.cljs
@@ -1,15 +1,15 @@
 (ns rems.focus)
 
-(def ^:private autofocus (atom nil))
+(def ^:private element-atom-to-focus (atom nil))
 
 (defn- grab-focus []
-  (when-let [element-atom @autofocus]
+  (when-let [element-atom @element-atom-to-focus]
     (when-let [element @element-atom]
       (.focus element)
-      (reset! autofocus nil))))
+      (reset! element-atom-to-focus nil))))
 
 (defn set-focus [element-atom]
-  (reset! autofocus element-atom)
+  (reset! element-atom-to-focus element-atom)
   (grab-focus))
 
 (defn ref-changed [element-atom element]
--- a/src/cljs/rems/focus.cljs
+++ b/src/cljs/rems/focus.cljs
@@ -1,0 +1,17 @@
+(ns rems.focus)
+
+(def autofocus (atom nil))
+
+(defn grab-focus []
+  (when-let [element-atom @autofocus]
+    (when-let [element @element-atom]
+      (.focus element)
+      (reset! autofocus nil))))
+
+(defn set-focus [element-atom]
+  (reset! autofocus element-atom)
+  (grab-focus))
+
+(defn ref-changed [element-atom element]
+  (reset! element-atom element)
+  (grab-focus))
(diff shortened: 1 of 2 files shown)


luontola:
The problem with [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autofocus) is that there can be only one element with that attribute on the page (we have multiple tables/searches on some pages), and it would also grab the focus when the page is loaded for the first time (after closing the search, the open-search button needs to be focused, but the open-search button is visible already on page load).

I can investigate using [refs](https://purelyfunctional.tv/guide/reagent/#refs). It might be possible to use the refs as a kind of a callback for finding out when the element becomes visible, which would avoid the polling.

src/cljs/rems/table.cljs
```diff
@@ -171,48 +172,39 @@
...
-   (let [target (.querySelector parent selector)]
-     (cond
-       target (.focus target)
-       (pos? timeout) (js/setTimeout #(focus-async! parent selector (- timeout interval) interval)
-                                     interval)
-       :else (js/console.warn (str "Could not focus element " selector " under:") parent)))))
-
 (defn search [table]
-  (let [filtering @(rf/subscribe [::filtering table])
+  (let [search-params (atom nil)
```
Macroz: It's unfortunate that there needs to be local state here with so specific requirements. Some possible alternatives:
- atom with map from `:id` to element maintained in the focus `ns`
- re-frame atom to contain the element in a path that contains `::focus` and `:id`

src/cljs/rems/table.cljs
```diff
@@ -171,48 +172,39 @@
...
-       [:button.btn.btn-secondary.close-search {:type :button
-                                                :aria-label (text :t.search/close-search)
-                                                :on-click on-toggle}
-        [close-symbol]]]
-      [:div.rems-table-search-toggle.d-flex.flex-row-reverse
-       [:button.btn.btn-primary.open-search {:type :button
-                                             :aria-label (text :t.search/open-search)
-                                             :on-click on-toggle}
-        [search-symbol]]])))
+    (fn []
```
Macroz: These parameters should match the overall function parameters (even though at the moment they are unused).

src/cljs/rems/table.cljs
luontola: Using a reagent atom will produce infinite render loop, because setting the ref will re-render the component. It needs to be a `clojure.core/atom` as said in https://cljdoc.org/d/reagent/reagent/0.8.1/doc/frequently-asked-questions/how-do-i-use-react-s-refs-

src/cljs/rems/table.cljs
luontola: Done.

src/cljs/rems/table.cljs
luontola: Since one page can contain multiple search components, the key for such a global mapping would need to be e.g. `[(:id table) :search-params]` which I think is more complex than a couple local atoms. It would also be different from the recommended way of using refs in Reagent, so extra justification is needed.

src/cljs/rems/focus.cljs
```diff
@@ -0,0 +1,17 @@
+(ns rems.focus)
```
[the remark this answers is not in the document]
luontola: Done.

PR #1317, opened 2019-06-04
Update table data when language changes

### LINKED ISSUE #1313: Other: Changing language does not update table data
Changing the language does not update table data completely. For example in Applications page, the state is kept in the original localization and does not update properly.

luontola:
Closes #1313

### DIFF
--- a/src/cljs/rems/administration/licenses.cljs
+++ b/src/cljs/rems/administration/licenses.cljs
@@ -3,11 +3,10 @@
             [rems.administration.administration :refer [administration-navigator-container]]
             [rems.administration.status-flags :as status-flags]
             [rems.atoms :refer [readonly-checkbox document-title]]
-            [rems.catalogue-util :refer [get-catalogue-item-title]]
             [rems.spinner :as spinner]
             [rems.status-modal :as status-modal]
             [rems.table :as table]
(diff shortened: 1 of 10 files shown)


src/cljs/rems/administration/licenses.cljs
```diff
@@ -66,11 +67,12 @@
...
  (fn [_ _]
-   [(rf/subscribe [::licenses])])
- (fn [[licenses] _]
+   [(rf/subscribe [::licenses])
+    (rf/subscribe [:language])])
+ (fn [[licenses language] _]
    (map (fn [license]
           {:key (:id license)
-           :title {:value (:title license)}
+           :title {:value (get-catalogue-item-title license language)} ; XXX: not really catalogue item, but the structure is the same
```
Macroz: Could this function be renamed to something like `get-localized-title` and then it would be nicer to use in these different places? Ideally we would use the same localization structure with all these different concepts and then we could use the same fns too.

luontola: Done. For data from V2 APIs (forms and applications) it's possible to use `rems.text/localized` for all localized strings, but catalogue items and licenses still use the old style .

PR #1322, opened 2019-06-06
Changes to email translations and corresponding changes to code

okahilak:


UFOPOLI:
"Päätetty" sounds even more final but if that's what you prefer, I guess it's fine.

luontola:
The UI button reads "Pyydä päätöstä", so we should use the same terminology everywhere, whatever that term is. If the term is not understandable then change it everywhere. Here is my suggestion:

```
          :decided {:subject "Päätös annettu (%2: %3)"
                    :message "Hyvä %1,\n\nKäyttäjä %2 on antanut päätöksensä hakemuksesta %3.\n\nTarkastele hakemusta: %4"}
          :decision-requested {:subject "Päätöspyyntö (%2: %3)"
                               :message "Hyvä %1,\n\nKäyttäjä %2 on pyytänyt sinua antamaan päätöksen hakemuksesta %3.\n\nTarkastele hakemusta: %4"}
```

luontola:
"Application decided" could be changed to "Decision made". The application itself doesn't make decisions. 😉

UFOPOLI:
"Antaa päätös" sounds really strange. Let's use "tehdä" instead.

I did some research, and "application decided" is actually used in official language (in native English countries). I don't like it particularly myself, though, which is why I originally wanted to express it differently. "Decision made" sounds better but is it anymore in line with the other headings?

luontola:
"application decided" -> About 74 900 results
"decision made" -> About 7 970 000 results

luontola:
We do have "Comment requested" & "Comment added", which is similar to "Decision requested" & "Decision made".

resources/translations/en.edn
```diff
@@ -192,27 +192,27 @@
...
-          :member-added {:subject "You've been added as a member to an application"
-                         :message "Dear %1,\nYou've been added as a member to application %2.\nView the application: %3"}
+  :email {:application-approved {:subject "Application approved (%2)"
+                                 :message "Dear %1,\n\nYour application %2 has been approved.\n\nView application: %3"}
+          :application-rejected {:subject "Application rejected (%2)"
+                                 :message "Dear %1,\n\nYour application %2 has been rejected.\n\nView application: %3"}
+          :application-closed {:subject "Application closed (%2)"
+                               :message "Dear %1,\n\nYour application %2 has been closed.\n\nView application: %3"}
+          :application-submitted {:subject "New application (%2: %3)"
+                                  :message "Dear %1,\n\n%2 has submitted a new application (%3): %4.\n\nView application: %5"}
```
luontola: This isn't necessarily a new application. It could be an application which was returned and then resubmitted.

okahilak: Changed to "Application submitted", and removed "new" from :message part. Likewise in Finnish translations.

okahilak:
Thanks, nice tip.

PR #1333, opened 2019-06-11
Add edit form functionality

### LINKED ISSUE #865: vTHL4: Update form
Admin should be able to update a form.

Implement copying first. See #836 

- [ ] update action
- [ ] modifies old if there are no uses of the form yet or makes a copy of the form and disables to old one if the old one is in use already
- [ ] allow updating of information
- [ ] show in list view in #805

okahilak:
Closes #865

### DIFF
--- a/src/clj/rems/api/forms.clj
+++ b/src/clj/rems/api/forms.clj
@@ -57,9 +57,7 @@
      (format-form wf))))
 
 (s/defschema FormCommand
-  {:edit-form? s/Bool
-   (s/optional-key :form-id) s/Num
-   :organization s/Str
+  {:organization s/Str
    :title s/Str
    :fields [FormField]})
 
@@ -82,6 +80,13 @@
                                      (when-not disabled {:enabled true})
                                      (when-not archived {:archived false})))))
 
+    (POST "/create" []
+      :summary "Create form"
+      :roles #{:owner}
+      :body [command FormCommand]
+      :return FormResponse
+      (ok (form/create-form! (getx-user-id) command)))
+
     (GET "/:form-id" []
       :summary "Get form by id"
       :roles #{:owner}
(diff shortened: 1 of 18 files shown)


okahilak:
I sent this patch to review as somewhat incomplete: I haven't written any tests yet, or necessarily updated all old tests.

However, I'd like to have a review of the general approach before putting time into writing new tests.

luontola:
Calling the operation "create" when it's "create or update" is suspicious. How about using the "update" operation changing the form fields?

When the form fields are updated, the code should first check that the form is not in use and raise an error if it is. Similar to the check when archiving a form.

Macroz:
> When the form fields are updated, the code should first check that the form is not in use and raise an error if it is. Similar to the check when archiving a form.

Shouldn't it then notify that a new copy will be created (i.e. not an error)?

src/clj/rems/api/forms.clj
```diff
@@ -88,15 +90,22 @@
       (ok (form/get-form-template form-id)))
 
     (POST "/create" []
-      :summary "Create form"
+      :summary "Create or edit form"
```
Macroz: Why not have two separate endpoints? Create is when a completely new form is created and update is when one is, well, updated?

src/clj/rems/api/forms.clj
```diff
@@ -88,15 +90,22 @@
...
       :roles #{:owner}
-      :body [command CreateFormCommand]
-      :return CreateFormResponse
-      (ok (form/create-form! (getx-user-id) command)))
+      :body [command FormCommand]
+      :return FormResponse
+      (ok (form/create-or-edit-form! (getx-user-id) command)))
 
     (PUT "/update" []
```
Macroz: There is likely a need to make sure the different kinds of update endpoints are named well. This should probably be named `/update-form-state` or maybe even `/form/:form-id/update-state` as is common in REST APIs.

src/clj/rems/api/forms.clj
okahilak: You're right (as well as Esko who commented about the same): better to have the API as clean as possible. Done.

src/clj/rems/api/forms.clj
okahilak: I started doing this but found out that there are quite a lot of places that the change touches (considering that similar update-state logic with similarly-named API end points is also implemented for catalogue items, licenses, and resources), so opted to add a TODO comment at this point and to do it in a later patch.

okahilak:
> Calling the operation "create" when it's "create or update" is smelly, as is the use of a boolean flag. How about using the "update" operation to change the form fields?

Good comments. I added another endpoint, /:form-id/edit for editing existing forms.

Using the existing "update" operation to change the form fields would have been problematic, as an identically-named operation is also in use for catalogue items, licenses, and resources, and they share the common parts. It seems therefore better to have a separate, form-specific edit operation.

I also marked as a TODO to change "update" to "update-state" in the future to avoid confusion with "edit".

> When the form fields are updated, the code should first check that the form is not in use and raise an error if it is. Similar to the check when archiving a form.

Done.

okahilak:
> > When the form fields are updated, the code should first check that the form is not in use and raise an error if it is. Similar to the check when archiving a form.
> 
> Shouldn't it then notify that a new copy will be created (i.e. not an error)?

I opted to just show the error if the user attemps to edit a form that is in use, and let the user explicitly press "Copy as new" (located next to "Edit" button, after all) if that is what the user wants. I think that doing that automatically risks guessing incorrectly what the user wants to do -- for example, it could be that the user is not expecting the form to be in use and after noticing that that is the case, decides to do something else.

But this can be discussed or changed further -- I believe that having it work analogously to "Archive" button, i.e., just showing the error if the operation is non-permitted, is a decent first step.

Thanks for the good comments, I think the patch is already a lot better now after this first review round.

PR #1341, opened 2019-06-12
Create pull_request_template.md

Macroz:
Definition of Done / Review checklist for review.

I think we should all first add relevant ideas here in this PR for discussion and decide on the first set together in the autumn. I added only those that were on the top of my mind, so likely there are a lot of good ones still missing.


opqdonut:
needs a checkbox for changelog

PR #1344, opened 2019-06-14
Fix label fields validation, fix max length validation

### LINKED ISSUE #1342: Label fields must not require input
This seems to be a regression, so it definitely needs tests. Also the example form needs one of each type of field.

luontola:
Closes #1342

### DIFF
--- a/test/clj/rems/migrations/test_convert_to_dynamic_applications.clj
+++ b/test/clj/rems/migrations/test_convert_to_dynamic_applications.clj
@@ -3,13 +3,12 @@
             [clojure.pprint :refer [pprint]]
             [clojure.test :refer :all]
             [conman.core :as conman]
-            [rems.db.applications :as applications]
+            [rems.api.testing :refer [api-fixture]]
             [rems.db.applications.legacy :as legacy]
             [rems.db.core :refer [*db*]]
             [rems.db.test-data :as test-data]
             [rems.db.workflow :as workflow]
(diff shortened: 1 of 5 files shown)


tests.edn
```diff
@@ -9,7 +9,11 @@
                     {:id :integration
                      :ns-patterns ["rems"]
                      :test-paths ["src" "test"]
-                     :focus-meta [:integration]}]
+                     :focus-meta [:integration]}
+                    {:id :focused
```
foxlynx: Should this be in some other pull request? Can't see anywhere focused appended elsewhere in this PR?

test/clj/rems/migrations/test_convert_to_dynamic_applications.clj
```diff
@@ -3,13 +3,12 @@
             [clojure.pprint :refer [pprint]]
```
foxlynx: Also don't get how this file is relevant to field changes?

tests.edn
luontola: The :focused tag is not meant to be committed. It's for easily choosing which subset of tests to run.

test/clj/rems/migrations/test_convert_to_dynamic_applications.clj
luontola: The test expectations must be changed when the label field was added to the example form.

tests.edn
foxlynx: Still don't get why it is added in this pull request.

PR #1389, opened 2019-06-28
Free comments 845

### LINKED ISSUE #845: vTHL4: Free commenting for dynamic workflow applications

hukka:
Closes #845


okahilak:
After the change, the buttons flow over the edge of the page, see screenshot below:

image

They were slightly over the edge previously, as well, but it was less conspicuous:

image

Thus, maybe this would be time to do something about the layout.

Another thing that I noticed: When a remark is submitted, the text field and the submit button are not hidden, like they are, e.g., when returning the application back to the applicant. Moreover, the contents of the text field remain, which could give the impression that the remark was not submitted after all. (With other actions, the text field is also emptied.)

I noticed that the same behavior affected "Comment" action, as well (but not "Request comment" action).

resources/translations/fi.edn
```diff
@@ -228,6 +231,7 @@
          :actions "Toiminnot"
          :add-comments-not-shown-to-applicant "Kommentoi (ei näy hakijalle)"
          :add-comments-shown-to-applicant "Kommentoi (näkyy hakijalle)"
+         :add-remark "Lisää merkintä (näytetään tai piilotetaan hakijalta)"
```
okahilak: minor: In other Finnish translations, the term used is "huomio" instead of "merkintä".

src/cljs/rems/actions/remark.cljs
```diff
@@ -0,0 +1,68 @@
...
+                     :on-click on-send}]]
+   [:div [action-comment {:id action-form-id
+                          :label (text :t.form/add-remark)
+                          :comment comment
+                          :on-comment on-set-comment}]
+         [:input {:type "checkbox"
+                  :id "public"
+                  :name "public"
+                  :value public
+                  ;:checked (contains? selected-keys key)
```
okahilak: Hmm, is this here on purpose?

test/clj/rems/application/test_commands.clj
```diff
@@ -1011,3 +1011,86 @@
...
+                              injections)
+          application (apply-events application [event-1])
+          event-2 (ok-command application
+                              {:type :application.command/remark
+                               :actor "commenter"
+                               :comment "first remark"
+                               :public false}
+                              injections)
+          application (apply-events application [event-2])]
+      (testing "commenter can remark before"
```
okahilak: minor: I didn't understand this before I read further. How about making it self-contained, i.e., "commenter can remark before commenting"?

test/clj/rems/application/test_commands.clj
```diff
@@ -1011,3 +1011,86 @@
...
+                                  injections)
+              application (apply-events application [event-1])
+              event-2 (ok-command application
+                                  {:type :application.command/remark
+                                   :actor "commenter"
+                                   :comment "second remark"
+                                   :public false}
+                                  injections)
+              application (apply-events application [event-2])]
+          (testing "and after commenting"
```
okahilak: The same here, I believe this string is written to stdout or somewhere if the test fails, so it will be difficult to extract the needed information from this string alone.

okahilak: Never mind, the author told me that inner testing-strings are concatenated with the outer ones.

test/clj/rems/application/test_commands.clj
okahilak: This is ok, as well.

hukka:
image
with flexbox

PR #1391, opened 2019-07-09
Write custom exceptions in Java to avoid compilation issues

luontola:
Clojure's support for custom exception classes is fragile, so let's write the them in Java. Earlier for example Kaocha used to occasionally fail in watch mode when it couldn't find the exception namespace. This fixes that.

### DIFF
--- a/src/java/rems/InvalidRequestException.java
+++ b/src/java/rems/InvalidRequestException.java
@@ -19,8 +19,4 @@ public class InvalidRequestException extends Exception {
     public InvalidRequestException(Throwable cause) {
         super(cause);
     }
-
-    public InvalidRequestException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
-        super(message, cause, enableSuppression, writableStackTrace);
-    }
 }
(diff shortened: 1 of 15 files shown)


src/java/rems/InvalidRequestException.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public InvalidRequestException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public InvalidRequestException(Throwable cause) {
+        super(cause);
+    }
+
+    public InvalidRequestException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
```
okahilak: This seems to have `protected` modifier in Exception class, should it be protected here as well? See https://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html

luontola: Ok, in that case the constructor can be removed.

okahilak: Hm, yeah, true.

PR #1393, opened 2019-07-10
Focus on action button when action-related cancel button is clicked

okahilak:


src/cljs/rems/actions/action.cljs
```diff
@@ -20,7 +23,8 @@
    {:type :button
     :id (str "cancel-" id)
     :data-toggle "collapse"
-    :data-target (str "#" (action-collapse-id id))}
+    :data-target (str "#" (action-collapse-id id))
+    :on-click #(.focus (js/$ (str "#" (action-button-id id))))}
```
luontola: Using jQuery is overkill for this when native JavaScript can do the same thing with `document.querySelector`

okahilak: Hm hm, I could change it, but then there's a similar way to retrieve an element a couple of lines above in the same namespace:

```(defn collapse-action-form [id]
  (.collapse (js/$ (str "#" (action-collapse-id id))) "hide"))```

For consistency, I'd like to either change them both or keep them using jQuery.

luontola: `.collapse` is a method in the jQuery object (from Bootstrap), so there that is needed

okahilak: Thanks for the clarification. Done.

PR #1394, opened 2019-07-10
Change focus to modal content when opened

okahilak:


okahilak:
Switching the focus to content seemed to work correctly, but the screen reader I'm using (ChromeVox) did not always start reading the content. Not sure if it is a problem with the screen reader or the code. It probably needs to be tested with another screen reader eventually. But at least switching the focus to the content of the modal is a step to the right direction.

okahilak:
Okay, I need to re-think this, probably using the existing focus tools.

PR #1398, opened 2019-07-11
Field schema unification

luontola:
Part of #913. Unifies the read and write models for form fields.


src/cljs/rems/administration/status_flags.cljs
```diff
@@ -50,7 +50,8 @@
 
 (defn- format-update-error [{:keys [type catalogue-items resources workflows]}]
   (let [language @(rf/subscribe [:language])]
-    [:p (text type)
+    [:<>
```
hukka: Doesn't this produce a "<<>> </<>>" tag pair in html?

src/clj/rems/migrations/refactor_form_templates.clj
```diff
@@ -0,0 +1,36 @@
+(ns rems.migrations.refactor-form-templates
+  (:require [hugsql.core :as hugsql]
+            [medley.core :refer [map-keys]]
```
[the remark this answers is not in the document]
luontola: Good point. We use it already in lots of places.

src/cljs/rems/administration/status_flags.cljs
luontola: It produces a React fragment, i.e. no tag.

PR #1406, opened 2019-07-17
Support multiple remove member forms & tweak visible attributes

### LINKED ISSUE #1099: vTHL4: remove member forms share state but aren't mutually exclusive
the comment state is shared between all remove-member forms, but you can have multiple remove-member forms open at once

### LINKED ISSUE #1346: Show only name and accepted terms, rest behind show more
Check changes with LBR or make customizable

### LINKED ISSUE #1353: Do not show the "Accepted terms" as unclickable checkbox at the "Applicants" section
Applicant can't do anything about it when the application is still a draft. This is mostly useful later.

Perhaps easy to change with #1346

luontola:
Closes #1099. Closes #1346. Closes #1353.

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -17,7 +17,7 @@
             [rems.actions.request-comment :refer [request-comment-action-button request-comment-form]]
             [rems.actions.request-decision :refer [request-decision-action-button request-decision-form]]
             [rems.actions.return-action :refer [return-action-button return-form]]
-            [rems.application-util :refer [accepted-licenses? form-fields-editable? get-applicant-name]]
+            [rems.application-util :refer [accepted-licenses? form-fields-editable? get-member-name]]
             [rems.atoms :refer [external-link file-download flash-message info-field readonly-checkbox textarea document-title]]
             [rems.collapsible :as collapsible]
             [rems.common-util :refer [index-by]]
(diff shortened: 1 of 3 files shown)


src/cljs/rems/application.cljs
```diff
@@ -448,13 +448,11 @@
   [{:keys [element-id attributes application group? can-remove? accepted-licenses?]}]
   (let [application-id (:application/id application)
         user-id (or (:eppn attributes) (:userid attributes))
-        sanitized-user-id (-> (or user-id (:email attributes) "") ;; use email for invited members
-                              str/lower-case
-                              (str/replace #"[^a-z]" ""))
+        user-id-or-email (or user-id (:email attributes) "") ; use email for invited members
```
okahilak: I've been reading this for some time and testing using the UI, but I can't figure out what this change does and why it's needed. The commit message says "fix all forms being opened with a single button". Which forms, which button? I can't find a button on Create Application page that would open several forms at once. Can you clarify?

hukka: The old version removed everything except letters from the user-id, so that completely valid, distinct user ids and emails got the same id. When collapsibles have same id, bootstraps operates on them all.

Test by adding perftester1 and perftester2, for example.

okahilak: Ok, thanks. I was mislead by how the commit message was formulated (strictly speaking, "all forms" are not necessarily opened, several forms can in some cases be).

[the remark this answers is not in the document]
hukka: > ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").

At least these are safe, since that's HTML4's definition. Newer HTML doesn't have restrictions, but I don't know about browser compatibility.

PR #1410, opened 2019-07-17
Allow handler to use commands on application returned to applicant

### LINKED ISSUE #1358: Do not disable handler actions when application is returned to applicant
Handler often needs to close stale application, and spirit of REMS is that handler can do anything to applications anyway. Only drafts should be hidden (as those are just noise for handlers).

okahilak:
Closes #1358

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -19,19 +19,19 @@
   application)
 
 (def ^:private submittable-application-commands
-  [:application.command/save-draft
-   :application.command/submit
-   :application.command/close
-   :application.command/remove-member
-   :application.command/invite-member
-   :application.command/uninvite-member
-   :application.command/accept-licenses
-   :application.command/change-resources])
+  #{:application.command/save-draft
+    :application.command/submit
+    :application.command/close
+    :application.command/remove-member
+    :application.command/invite-member
+    :application.command/uninvite-member
+    :application.command/accept-licenses
+    :application.command/change-resources})
 
 (def ^:private non-submittable-application-commands
-  [:application.command/remove-member
-   :application.command/uninvite-member
-   :application.command/accept-licenses])
+  #{:application.command/remove-member
+    :application.command/uninvite-member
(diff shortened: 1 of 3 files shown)


src/clj/rems/application/model.clj
```diff
@@ -18,82 +18,76 @@
...
+   :application.command/invite-member
+   :application.command/uninvite-member
+   :application.command/request-comment
+   :application.command/request-decision
+   :application.command/return
+   :application.command/approve
+   :application.command/reject])
+
+(def ^:private handler-returned-commands
+  (vec (remove #(= % :application.command/return) handler-all-commands)))
```
luontola: handler-all-commands could be a set, which would simplify this code to just a disj

okahilak: A nice point, done. Although it adds some inconsistency between handler-all-commands and, e.g., submittable-application-commands.

src/clj/rems/application/model.clj
```diff
@@ -18,82 +18,74 @@
...
-             :application.command/remove-member
-             :application.command/invite-member
-             :application.command/uninvite-member
-             :application.command/request-comment
-             :application.command/request-decision
-             :application.command/return
-             :application.command/approve
-             :application.command/reject]
+  {:applicant non-submittable-application-commands
+   :handler (vec (conj handler-all-commands :see-everything))
```
luontola: These vec calls are not needed. update-role-permissions doesn't care what kind of a collection the parameter is.

okahilak: Ok, should I then change all these vectors to sets for consistency?

src/clj/rems/application/model.clj
luontola: For consistency you could convert them all to sets.

src/clj/rems/application/model.clj
luontola: Yes, sounds good.

okahilak: Done.

PR #1411, opened 2019-07-17
Send email to applicant when handler adds terms of use

### LINKED ISSUE #1363: Send email to applicants when a license has been added
because they need to consent.

okahilak:
Closes #1363

### DIFF
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -212,6 +212,8 @@
                                :message "Dear %1,\n\nYour application %2 has been closed.\n\nView application: %3"}
           :application-returned {:subject "Application returned (%2)"
                                  :message "Dear %1,\n\nYour application %2 has been returned.\n\nView application: %3"}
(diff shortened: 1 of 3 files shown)


resources/translations/en.edn
```diff
@@ -212,6 +212,8 @@
                                :message "Dear %1,\n\nYour application %2 has been closed.\n\nView application: %3"}
           :application-returned {:subject "Application returned (%2)"
                                  :message "Dear %1,\n\nYour application %2 has been returned.\n\nView application: %3"}
+          :application-licenses-added {:subject "Terms of use added to application (%2)"
+                                       :message "Dear %1,\n\nTerms of use have been added to your application %2.\n\nView application: %3"}
```
hukka: What do you think, should we be even more explicit that action is required. Something like "You need to accept these before gaining access to the resources"?

okahilak: Done.

PR #1418, opened 2019-07-19
Stricter schemas

luontola:


src/clj/rems/application/commands.clj
```diff
@@ -26,7 +26,7 @@
 (s/defschema SaveDraftCommand
   (assoc CommandBase
          ;; {s/Num s/Str} is what we want, but that isn't nicely representable as JSON
```
okahilak: minor: can you update this comment? (it also seems to be the last occurrence of "s/Num" in the code)

Issue #1101, opened 2019-04-10
Clean up rest of round-based code once LBR has been migrated


opqdonut:
(split out from #852, #913)

- [x] remove `rems.migrations.convert-to-dynamic-applications`
- [x] rewrite `rems.db.test-data/create-draft!` (and code which uses it) to not depend on legacy code
- [x] remove any other round-based stuff from `test-data`
- [x] stop updating `application_form` (use only `form_templates`)
  - WIP: https://github.com/CSCfi/rems/commits/remove-legacy-form
- [x] remove old form content saving logic
  - `application_text_values` and `application_license_approval_values` tables
- [x] get rid of `workflow_actors` table (only used for round-based workflows)
- [x] other tables?
- [x] other code?
- [x] schemas
- [x] UI
- [x] translations
- [x] docs

PR #1424, opened 2019-07-24
Remove legacy test data

luontola:
#1101


src/clj/rems/db/test_data.clj
```diff
@@ -441,93 +439,57 @@
...
-    value))
-
 (defn- run-and-check-dynamic-command! [cmd]
   (let [result (applications/command! cmd)]
     (assert (nil? result) {:actual result})
     result))
 
-(defn- create-draft! [user-id catids wfid field-value & [now]]
-  (let [app-id (legacy/create-new-draft-at-time user-id wfid (or now (time/now)))
-        catids (if (vector? catids) catids [catids])
```
okahilak: minor: There used to be the logic here that allowed passing to create-draft! function a single catid that was not in a vector. It seems that all uses of create-draft! pass a single catid, so the parameters of the function could be changed so that it takes a single catid instead of several catids, which cleans up the call-sites a bit.

src/clj/rems/db/test_data.clj
```diff
@@ -693,16 +655,16 @@
...
                             {"en" "Dynamic workflow with extra license" "fi" "Dynaaminen työvuo ylimääräisellä lisenssillä"})
     (let [thlform (create-thl-demo-form! +fake-users+)
           thl-catid (create-catalogue-item! res1 (:dynamic workflows) thlform {"en" "THL catalogue item" "fi" "THL katalogi-itemi"})]
-      (create-member-applications! thl-catid (:dynamic workflows) (+fake-users+ :applicant1) (+fake-users+ :approver1) [{:userid (+fake-users+ :applicant2)}]))
+      (create-member-applications! thl-catid (+fake-users+ :applicant1) (+fake-users+ :approver1) [{:userid (+fake-users+ :applicant2)}]))
     (let [dynamic-disabled (create-catalogue-item! res1 (:dynamic workflows) form
                                                    {"en" "Dynamic workflow (disabled)"
                                                     "fi" "Dynaaminen työvuo (pois käytöstä)"})]
-      (create-disabled-applications! dynamic-disabled (:dynamic workflows) (+fake-users+ :approver1) (+fake-users+ :approver1))
+      (create-disabled-applications! dynamic-disabled (+fake-users+ :approver1) (+fake-users+ :approver1))
```
okahilak: minor: not introduced by this patch, but shouldn't the second parameter be (+fake-users+ :applicant1)?

src/clj/rems/db/test_data.clj
```diff
@@ -720,12 +682,12 @@
...
+      (create-applications! dynamic +demo-users+))
     (let [thlform (create-thl-demo-form! +demo-users+)
           thl-catid (create-catalogue-item! res1 (:dynamic workflows) thlform {"en" "THL catalogue item" "fi" "THL katalogi-itemi"})]
-      (create-member-applications! thl-catid (:dynamic workflows) (+demo-users+ :applicant1) (+demo-users+ :approver1) [{:userid (+demo-users+ :applicant2)}]))
+      (create-member-applications! thl-catid (+demo-users+ :applicant1) (+demo-users+ :approver1) [{:userid (+demo-users+ :applicant2)}]))
     (let [dynamic-disabled (create-catalogue-item! res1 (:dynamic workflows) form
                                                    {"en" "Dynamic workflow (disabled)"
                                                     "fi" "Dynaaminen työvuo (pois käytöstä)"})]
-      (create-disabled-applications! dynamic-disabled (:dynamic workflows) (+demo-users+ :approver1) (+demo-users+ :approver1))
+      (create-disabled-applications! dynamic-disabled (+demo-users+ :approver1) (+demo-users+ :approver1))
```
okahilak: Here also, shouldn't it be (+demo-users+ :applicant1)?

src/clj/rems/db/test_data.clj
luontola: It probably should, but it breaks many (badly written) tests. Need to refactor those tests first to create their own test data.

PR #1425, opened 2019-07-24
Add user-specific settings to database, save language preference there

### LINKED ISSUE #1381: UI should remember the user's language preference (LBR)
From LBR: "It seems that the interface doesn't remember the user's language preference. At least I've had to switch each time I've logged in."

okahilak:
Closes #1381

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -670,6 +670,12 @@ VALUES (:user, :userattrs::jsonb)
 ON CONFLICT (userId)
 DO UPDATE SET userAttrs = :userattrs::jsonb;
 
+-- :name update-user-settings! :!
+INSERT INTO user_settings (userId, settings)
+VALUES (:user, :settings::jsonb)
+ON CONFLICT (userId)
+DO UPDATE SET settings = :settings::jsonb;
+
 -- :name get-users :? :*
 SELECT userId
 FROM users;
@@ -683,6 +689,11 @@ WHERE role = :role;
 SELECT userAttrs::TEXT
 FROM users
 WHERE userId = :user;
+
+-- :name get-user-settings :? :1
+SELECT settings::TEXT
+from user_settings
+where userId = :user;
 
 -- :name get-application-events :? :*
 SELECT id, appId, userId, round, event, comment, eventData::TEXT, time
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -1,0 +1,31 @@
+(ns rems.api.user-settings
+  (:require [compojure.api.sweet :refer :all]
+            [rems.api.schema :refer :all]
+            [rems.db.user-settings :as user-settings]
+            [rems.util :refer [getx-user-id]]
+            [ring.util.http-response :refer :all]
+            [schema.core :as s]))
+
+(s/defschema UpdateUserSettingsCommand
+  {:user-id s/Str
+   (s/optional-key :language) s/Keyword})
+
+(s/defschema UserSettings
+  {(s/optional-key :language) s/Keyword})
+
+(def user-settings-api
+  (context "/user-settings" []
+    :tags ["user-settings"]
+
+    (GET "/" []
+      :summary "Get user settings"
+      :roles #{:logged-in}
+      :return UserSettings
+      (ok (user-settings/get-user-settings (getx-user-id))))
+
+    (PUT "/update" []
+      :summary "Update user settings"
+      :roles #{:logged-in}
+      :body [command UpdateUserSettingsCommand]
+      :return SuccessResponse
+      (ok (user-settings/update-user-settings! (:user-id command) command)))))
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -6,11 +6,8 @@
             [ring.util.http-response :refer :all]
             [schema.core :as s]))
 
-(s/defschema UpdateUserSettingsCommand
-  {(s/optional-key :language) s/Keyword})
-
 (s/defschema UserSettings
-  {(s/optional-key :language) s/Keyword})
+  {:language s/Keyword})
 
 (def user-settings-api
   (context "/user-settings" []
@@ -25,6 +22,6 @@
     (PUT "/" []
       :summary "Update user settings"
       :roles #{:logged-in}
-      :body [command UpdateUserSettingsCommand]
+      :body [settings UserSettings]
       :return SuccessResponse
-      (ok (user-settings/update-user-settings! (getx-user-id) command)))))
+      (ok (user-settings/update-user-settings! (getx-user-id) settings)))))
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -7,8 +7,7 @@
             [schema.core :as s]))
 
 (s/defschema UpdateUserSettingsCommand
-  {:user-id s/Str
-   (s/optional-key :language) s/Keyword})
+  {(s/optional-key :language) s/Keyword})
 
 (s/defschema UserSettings
   {(s/optional-key :language) s/Keyword})
@@ -23,9 +22,9 @@
       :return UserSettings
       (ok (user-settings/get-user-settings (getx-user-id))))
 
-    (PUT "/update" []
+    (PUT "/" []
       :summary "Update user settings"
       :roles #{:logged-in}
       :body [command UpdateUserSettingsCommand]
       :return SuccessResponse
-      (ok (user-settings/update-user-settings! (:user-id command) command)))))
+      (ok (user-settings/update-user-settings! (getx-user-id) command)))))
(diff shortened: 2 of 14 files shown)


test/clj/rems/db/test_user_settings.clj
```diff
@@ -0,0 +1,19 @@
...
+(use-fixtures :each rollback-db-fixture)
+
+(deftest test-user-settings
+  (users/add-user! "user1" {:eppn "user1"})
+  (testing "user settings are empty for a new user"
+    (is (= (user-settings/get-user-settings "user1")
+           {})))
+
+  (testing "add a user setting"
+    (user-settings/update-user-settings! "user1" {:language :en})
```
luontola: This tests the INSERT, but another test is needed for the UPDATE.

okahilak: Done

src/clj/rems/api/user_settings.clj
```diff
@@ -0,0 +1,31 @@
...
+  (context "/user-settings" []
+    :tags ["user-settings"]
+
+    (GET "/" []
+      :summary "Get user settings"
+      :roles #{:logged-in}
+      :return UserSettings
+      (ok (user-settings/get-user-settings (getx-user-id))))
+
+    (PUT "/update" []
```
luontola: This too could be "/" for a more REST style.

src/clj/rems/api/user_settings.clj
```diff
@@ -0,0 +1,31 @@
...
+      :summary "Update user settings"
+      :roles #{:logged-in}
+      :body [command UpdateUserSettingsCommand]
+      :return SuccessResponse
+      (ok (user-settings/update-user-settings! (:user-id command) command)))))
```
luontola: This allows anybody updating any other user's settings. The user-id parameter should probably be removed from the command, so that it'll only change the current user's settings.

src/clj/rems/db/user_settings.clj
```diff
@@ -0,0 +1,21 @@
...
+(defn update-user-settings! [user settings]
+  (assert (and user settings) "User or user settings missing!")
+  (db/update-user-settings!
+   {:user user
+    :settings (json/generate-string {:language (:language settings)})})
+  {:success true})
+
+(defn- parse-settings [settings]
+  (let [{:keys [language]} (json/parse-string settings)]
+    {:language (keyword language)}))
```
luontola: Since :language is an optional key in UpdateUserSettingsCommand, how does this handle nil language? We probably need a test which sets empty settings and then reads them back.

src/cljs/rems/language.cljs
```diff
@@ -0,0 +1,42 @@
...
+(defn- update-css [language]
+  (let [localized-css (str "/css/" (name language) "/screen.css")]
+    ;; Figwheel replaces the linked stylesheet
+    ;; so we need to search dynamically
+    (doseq [element (array-seq (.getElementsByTagName js/document "link"))]
+      (when (str/includes? (.-href element) "screen.css")
+        (set! (.-href element) localized-css)))))
+
+(defn update-language [language]
+  (set! (.. js/document -documentElement -lang) language)
```
luontola: This should be `(name language)`. Currently it produces `<html lang=":en">`.

src/cljs/rems/language.cljs
```diff
@@ -0,0 +1,42 @@
...
+ (fn [db _]
+   (:default-language db)))
+
+(reg-event-fx
+ ::set-language
+ (fn [{:keys [db]} [_ language]]
+   (let [user-id (get-in db [:identity :user :eppn])]
+     (if user-id
+       (rf/dispatch [:rems.user-settings/update-user-settings user-id {:language language}])
+       {:db (assoc db :language language)}))))
```
luontola: `update-css` and `update-language` need to be called after updating the language here.

src/cljs/rems/user_settings.cljs
```diff
@@ -0,0 +1,22 @@
+(ns rems.user-settings
+  (:require [re-frame.core :as rf :refer [reg-event-fx reg-event-db]]
+            [rems.language :as language]
+            [rems.status-modal :as status-modal]
+            [rems.util :refer [fetch put!]]))
+
+(reg-event-fx
+ :loaded-user-settings
+ (fn [{:keys [db]} [_ user-settings]]
+   (language/update-language (:language user-settings))
```
luontola: Need to check for a nil language. Currently this produces `<html lang="null">` when the user has no settings.

luontola: Or then make it impossible to have a nil language. It there are no user settings, then some reasonable defaults should be returned.

test/clj/rems/db/test_user_settings.clj
```diff
@@ -0,0 +1,24 @@
...
+  (:require [clojure.test :refer :all]
+            [rems.db.testing :refer [rollback-db-fixture test-db-fixture]]
+            [rems.db.user-settings :as user-settings]
+            [rems.db.users :as users]))
+
+(use-fixtures :once test-db-fixture)
+(use-fixtures :each rollback-db-fixture)
+
+(deftest test-user-settings
+  (users/add-user! "user1" {:eppn "user1"})
```
luontola: These tests would not notice it if the SQL queries were missing WHERE conditions. For that you need a second user and then check that its settings are not changed when some other user's settings are changed.

src/clj/rems/api/user_settings.clj
okahilak: Done

src/clj/rems/api/user_settings.clj
okahilak: Good point, done.

src/clj/rems/db/user_settings.clj
okahilak: Done, robustified both get-user-settings and update-user-settings! and added a couple of tests.

src/cljs/rems/language.cljs
okahilak: Good catch, this was actually wrong in the old code as well.

src/cljs/rems/language.cljs
okahilak: Done.

test/clj/rems/db/test_user_settings.clj
okahilak: Done.

src/cljs/rems/user_settings.cljs
okahilak: I ended up returning default settings from get-user-settings if they are unset. It seemed like the most robust solution. Thanks for the suggestion.

okahilak:
Thanks for the review, lots of good points.

resources/sql/queries.sql
```diff
@@ -670,6 +670,12 @@ VALUES (:user, :userattrs::jsonb)
 ON CONFLICT (userId)
 DO UPDATE SET userAttrs = :userattrs::jsonb;
 
+-- :name update-user-settings! :!
+INSERT INTO user_settings (userId, settings)
+VALUES (:user, :settings::jsonb)
+ON CONFLICT (userId)
+DO UPDATE SET settings = :settings::jsonb;
```
luontola: If you change this to "DO NOTHING", none of the tests will detect it.

src/clj/rems/api/user_settings.clj
```diff
@@ -0,0 +1,30 @@
...
+            [rems.db.user-settings :as user-settings]
+            [rems.util :refer [getx-user-id]]
+            [ring.util.http-response :refer :all]
+            [schema.core :as s]))
+
+(s/defschema UpdateUserSettingsCommand
+  {(s/optional-key :language) s/Keyword})
+
+(s/defschema UserSettings
+  {(s/optional-key :language) s/Keyword})
```
luontola: This can be made non-optional. Also the update command can be made so. Simplifies many things.

src/clj/rems/db/user_settings.clj
```diff
@@ -0,0 +1,30 @@
...
+  (assert user "User missing!")
+  (let [settings (get-user-settings user)
+        update-language? (and language
+                              (contains? (set (env :languages)) language))]
+    (db/update-user-settings!
+     {:user user
+      :settings (json/generate-string
+                 (merge settings
+                        (when update-language?
+                          {:language language})))})
```
luontola: The validation of allowed languages could be extracted to its own method. Then this line would become very simple: `(merge old-settings (validate new-settings))`

okahilak: Done.

src/clj/rems/api/user_settings.clj
okahilak: I was probably thinking of some other settings that are potentially changed from somewhere else and kind of preparing for partial updating. But better to have it simple for now. Done.

resources/sql/queries.sql
okahilak: The test was there previously (due to your previous review comment) but got lost in the latest patch. Done.

PR #1426, opened 2019-07-24
Send emails in recipients language

### LINKED ISSUE #1371: Let user set the language of the emails they receive (LBR)

okahilak:
Closes #1371

### DIFF
--- a/src/clj/rems/poller/email.clj
+++ b/src/clj/rems/poller/email.clj
@@ -9,6 +9,7 @@
             [rems.context :as context]
             [rems.db.applications :as applications]
             [rems.db.events :as events]
+            [rems.db.user-settings :as user-settings]
             [rems.db.users :as users]
             [rems.poller.common :as common]
             [rems.scheduler :as scheduler]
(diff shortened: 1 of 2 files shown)


okahilak:
On top of https://github.com/CSCfi/rems/pull/1425

src/clj/rems/poller/email.clj
```diff
@@ -60,21 +61,24 @@
...
-                            (link-to-application (:application/id event)))
-      :body (text-format body-text
-                         recipient
-                         (:event/actor event)
-                         (application-id-for-email application)
-                         (:application/applicant application)
-                         (resources-for-email application)
-                         (link-to-application (:application/id event)))})))
+     (with-language (or (:language (user-settings/get-user-settings recipient))
+                        (:default-language env))
```
luontola: This could be simplified if the user settings would return defaults for missing users, as mentioned in the other PR.

luontola:
Look good. Need to just resolve the issues in the first PR first.

luontola:
Actually this PR's ancestry has diverged from #1425, so this needs to be rebased after the that PR is merged.

okahilak:
> Actually this PR's ancestry has diverged from #1425, so this needs to be rebased after the that PR is merged.

Done.

src/clj/rems/poller/email.clj
okahilak: Done.

mmatthiesencsc:
Hi, could you explain in layman's terms how this fix behaves? Is the language either actively chosen by the user or derived from the browser's settings?

okahilak:
The email is sent in the language that the recipient of the email has selected in the UI. For example, say that the applicant has been using the UI in Finnish and the handler has used the UI in English. When the applicant submits an application, a notification is sent to the handler in English. When the handler approves the application, a notification is sent to the applicant in Finnish. Changing the language in which the notifications are received is done using EN/FI buttons in the UI.

We could also have a more explicit user-configurable language setting in the UI, e.g., under a "Settings" tab. Maybe it can be implemented if we end up having more user-configurable settings in the UI...

mmatthiesencsc:
Thanks OP, now it is clear. I think selecting the language in the header and storing that setting in a cookie should be fine, REMS is not a tool one uses day in day out from different browsers.

okahilak:
(Just a technical sidenote: It is actually now implemented by storing the UI setting in the database and retrieving it from there. The problem with storing it in a cookie is that the back-end needs to know the settings for, say, the handler when the applicant is logged in and submits the application. So storing it locally for the user was not quite sufficient for implementing this.)

mmatthiesencsc:
(A sidenote back: Makes sense and I think this suffices.)

UFOPOLI:
Tying the mail language to the interface language is exactly what we wanted, thank you! Combined with having Finnish as the default language and remembering the user's choice, of course.

PR #1437, opened 2019-07-31
Docs cleanup

luontola:
#1101

### DIFF
--- a/docs/dictionary.md
+++ b/docs/dictionary.md
@@ -2,19 +2,23 @@
 
 The glossary below provides definitions for words used in REMS:
 
-**Application Form Item**: An application form item is an element (such as, a text box, checkbox or a place to upload an attachment) on an application form.
+**Application**: An instance of an applicant applying for entitlement to some resources. Includes a form for the applicant to fill and licenses to accept.
+
+**Application Form**: A form which is part of an application. Includes the values of fields filled in by the applicant.
 
 **Catalogue**: A catalogue lists the resources to which access can be applied for. Catalogue consists of catalogue items.
 
 **Catalogue Item**: A catalogue item is a resource whose owner has attached a workflow and an application form to it. To make it appear in the catalogue, the resource owner must set the catalogue item to an enabled state.
 
-**Close Period**: Defines the time (in days) after which an approved application moves automatically to the closed state, making the entitlement expire.
-
-**Dataset**: See resource in this dictionary.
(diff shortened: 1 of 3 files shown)


docs/dictionary.md
```diff
@@ -2,26 +2,28 @@
...
+**Entitlement**: Access right to a resource.
 
-**Dataset**: See resource in this dictionary.
+**Field**: A textarea, checkbox, attachment upload or some other place for user input.
 
-**Entitlement**: Access right to a resource.
+**Form**: A set of fields.
 
-**License**: A license is a building block for an application form or a workflow. The applicant (and optionally, application members) need to commit to the license term of the resource before receiving the entitlement.
+**Form Template**: A form which is not part of an application.
```
okahilak: I didn't understand this.

luontola: We have `form_template` table in the database which contains the structure of the forms. How would you describe that?

okahilak: Hmm. Is it visible to the end user anywhere? If not, I'm not sure if it needs to be listed in the dictionary.

docs/dictionary.md
```diff
@@ -2,26 +2,28 @@
...
 The glossary below provides definitions for words used in REMS:
 
-**Application Form Item**: An application form item is an element (such as, a text box, checkbox or a place to upload an attachment) on an application form.
+**Application**: An instance of an applicant applying for entitlement to some resources. Includes a form for the applicant to fill and licenses to accept.
+
+**Application Form**: A form which is part of an application. Includes the values of fields filled in by the applicant.
 
 **Catalogue**: A catalogue lists the resources to which access can be applied for. Catalogue consists of catalogue items.
 
 **Catalogue Item**: A catalogue item is a resource whose owner has attached a workflow and an application form to it. To make it appear in the catalogue, the resource owner must set the catalogue item to an enabled state.
```
opqdonut: should say form template instead of application form here

docs/dictionary.md
opqdonut: Perhaps something like "A set of fields for an applicant to fill, see also _application form_ and _catalogue item_".

I'm not sure we need a separate Form definition though?

luontola: Ok. Restricting the dictionary to user-facing concepts simplifies things. The glossary in turn goes into more technical details.

Issue #1435, opened 2019-07-30
Improve accessibility of form validation


luontola:
- [x] Add "required" attribute to required for inputs
- [ ] ~~Add aria-label "required" for the asterisk (?)~~
- [x] Use fieldsets and legends to associate groups of checkboxes and radio buttons
- [x] Add aria-invalid to fields with errors
- [x] Associate error messages with the field using aria-describedby
- [x] In the error summary on the top of the page, add links to the fields with errors
- [x] In the modal popups, already the title should say whether the operation was successful or failed. Maybe add "Success:"/"Failed:" to the front of the title.
  - [ ] ~~Need aria-live regions?~~
- [x] Validation message for required attachment field
- [x] Validation message for required checkbox group

Reference: https://webaim.org/techniques/formvalidation/

PR #1443, opened 2019-08-01
Form accessibility improvements

luontola:
#1435


src/cljs/rems/fields.cljs
```diff
@@ -179,13 +217,14 @@
...
         selected-keys (decode-option-keys value)]
-    ;; TODO: for accessibility these checkboxes would be best wrapped in a fieldset
     [field-wrapper
-     (assoc opts :readonly-component [readonly-field {:id (id-to-name id)
-                                                      :value (->> options
-                                                                  (filter #(contains? selected-keys (:key %)))
-                                                                  (map #(localized (:label %)))
-                                                                  (str/join ", "))}])
+     (assoc opts
+            :fieldset true
```
opqdonut: could fieldset be always true?

luontola: Fieldset is meant for grouping together multiple labels and their inputs. It doesn't make much sense for individual inputs - the fieldset legend and the field label would need to repeat the same text. The `field-wrapper` component needs to be refactored. Rich Hickey would say it's too complex.

PR #1453, opened 2019-08-05
rename request comment to request review

### LINKED ISSUE #1350: Rename comments into comment (related to action), request review and add note/remark (new feature).
Related to #845

opqdonut:
only in the UI, events unchanged

closes #1350


resources/translations/en.edn
```diff
@@ -17,7 +17,7 @@
             :cancel "Cancel"
             :close "Close application"
             :close-intro [:div.intro [:p "Note that if the application is closed, it cannot be opened again."]]
-            :comment "Comment"
+            :comment "Review"
```
okahilak: Hmm, maybe use "katselmoi" or "arvioi" or something similar in the Finnish translations.

opqdonut: Oops, forgot all about finnish...

okahilak:
docs/glossary.md and possibly docs/dictionary.md (depending on the interpretation) seem to contain remains of the old terminology, maybe change them as well.

opqdonut:
I kinda feel like I should rename the front end components even if I don't rename the events & commands right now. Opinions?

okahilak:
> I kinda feel like I should rename the front end components even if I don't rename the events & commands right now. Opinions?

Hmm, now that you mention it, I also have the vague feeling that it might be better that way. It could be that no-one has time to change the terms in the events & commands in the next two years or so, and changing it at least in the front-end would minimize the amount of code using the wrong terms during that indefinite amount of time.

src/cljs/rems/actions/review.cljs
```diff
@@ -24,18 +24,18 @@
...
                   :on-click #(rf/dispatch [::open-form])}])
 
-(defn comment-view
+(defn review-view
   [{:keys [comment on-set-comment on-send]}]
   [action-form-view action-form-id
    (text :t.actions/comment)
-   [[button-wrapper {:id "comment"
+   [[button-wrapper {:id "review-button"
                      :text (text :t.actions/comment)
```
okahilak: I guess you could also change the translations as part of the front-end changes.

PR #1455, opened 2019-08-05
Improve alignment of navbar with main-content

### LINKED ISSUE #1380: Check that alignment of navbar matches the page content (LBR)
From LBR: "The main navigation bar looks nice now. It would be further improved if the leftmost item (catalogue/kielivarat) was aligned with the left edge of the body (and the same with "sign out" on the right)."

okahilak:
Closes #1380

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -417,7 +417,7 @@
    [:.alert-dark {:color (util/get-theme-attribute :alert-dark-color)
                   :background-color (util/get-theme-attribute :alert-dark-bgcolor)
                   :border-color (util/get-theme-attribute :alert-dark-bordercolor :alert-dark-color)}]
-   [:#navbar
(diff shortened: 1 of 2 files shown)


okahilak:
Tested with different sized windows, seemed to work. Here is an example:

Old:

image


New:

image

src/clj/rems/css/styles.clj
```diff
@@ -417,9 +417,14 @@
    [:.alert-dark {:color (util/get-theme-attribute :alert-dark-color)
                   :background-color (util/get-theme-attribute :alert-dark-bgcolor)
                   :border-color (util/get-theme-attribute :alert-dark-bordercolor :alert-dark-color)}]
+   [:#navbar
```
luontola: The distinction between `#navbar` and `.navbar` seems arbitrary. Maybe rename `#navbar` to `.navbar-wrapper` or something more descriptive.

PR #1459, opened 2019-08-06
Fix duplicate key error in dropdown menus

okahilak:
- As a default, the value of an option is used as the key for the
  element in react-select, but the library does not work well with
  non-primitive keys -- instead it gives an error about a duplicate
  key.

- As a fix, use getOptionValue parameter to set a custom mapping
  from an option to the key.


luontola:
The guide page crashes because item-key null.

okahilak:
> The guide page crashes because item-key null.

Thanks. Done.

PR #1462, opened 2019-08-06
prototype of better form preview

opqdonut:
- preview floats when scrolling
- button for scrolling to the current field in the preview

for #1361


opqdonut:
next steps:
- instead of button, use focus in editor to control preview
- more widely supported scrolling logic

Issue #890, opened 2019-01-30
Make CSC theme great again


Macroz:
The CSC theme does not look after the recent theme style additions. It (or the defsults) should be tuned to make it look good again.

- [x] CSC theme including e.g. administration pages look good again
- [x] table header color is nice
- [x] table rows look nice without stripes
- [x] footer text is readable on footer color
- [x] example theme should have some logo

Macroz:
Changes implemented in #937

PR #1468, opened 2019-08-08
Remove custom spacing from headings

okahilak:
- I could not find the reason for using these in the first place,
  and LBR preferred default spacing in large headings (h1, h2, h3).
  The smaller headings are only used in one place (h4 in create form)
  or nowhere (h5, h6), so remove custom spacing from them, as well.


okahilak:
Screenshots for comparison:

Current (0.17 rem):

image

After this change:

image

Another option would be to use custom spacings but smaller, e.g., here is 0.05 rem for h1, h2, and h3:

image

opqdonut:
Note: spacing was introduced in PR #937 for issue #890

Issue #1352, opened 2019-06-17
Full-text search from all applicants/members' name


hukka:
Related to #873

Issue #873, opened 2019-01-22
vTHL4: Full text search


Macroz:
Improve the search functionality by offering a full text search based on all information in an application. One search field. This possibly replaces the column based filters or they should be improved to suit the improved functionality (create new tasks if need updating and not just removal).

- [x] search from all content of the application
- [x] one search field
- [x] try implementing full content search within backend first
- [ ] ~~search handled applications checkbox~~

hukka:
Closed by #1000

Macroz:
Reopening. Still missing MVP of searching from all the application fields.

luontola:
Could application fields using in-memory Lucene index: http://lucene.apache.org/core/8_0_0/memory/org/apache/lucene/index/memory/MemoryIndex.html

PR #1471, opened 2019-08-08
Search by applicant, API only

luontola:
POC of #1352 and #873

### DIFF
--- a/src/clj/rems/application/search.clj
+++ b/src/clj/rems/application/search.clj
@@ -70,5 +70,5 @@
      (let [results (.search searcher
                             (-> (StandardQueryParser. analyzer)
                                 (.parse query "applicant")) ; TODO: change defaultField to full text search
-                            10000)]
+                            Integer/MAX_VALUE)]
        (set (get-application-ids searcher results))))))
--- a/src/clj/rems/application/search.clj
+++ b/src/clj/rems/application/search.clj
@@ -43,6 +43,14 @@
       (.maybeRefresh searcher-manager)
       (swap! search-index assoc ::last-processed-event-id (:event/id (last events))))))
 
+(defn- with-searcher [f]
+  (let [searcher-manager ^SearcherManager (::searcher-manager @search-index)
+        searcher ^IndexSearcher (.acquire searcher-manager)]
+    (try
+      (f searcher)
+      (finally
+        (.release searcher-manager searcher)))))
+
 (defn- get-application-ids [^IndexSearcher searcher ^TopDocs results]
   (doall (for [^ScoreDoc hit (.-scoreDocs results)]
            (let [doc (.doc searcher (.-doc hit))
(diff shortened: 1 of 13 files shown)


src/clj/rems/standalone.clj
```diff
@@ -39,14 +41,21 @@
...
 (defn start-app [& args]
   (doseq [component (-> args
                         (parse-opts cli-options)
                         mount/start-with-args
                         :started)]
     (log/info component "started"))
   (.addShutdownHook (Runtime/getRuntime) (Thread. stop-app))
-  (validate/validate))
+  (validate/validate)
+  (refresh-caches))
```
opqdonut: excellent

src/clj/rems/application/search.clj
```diff
@@ -0,0 +1,69 @@
...
+           [org.apache.lucene.search IndexSearcher ScoreDoc TopDocs SearcherManager SearcherFactory]
+           [org.apache.lucene.store Directory NIOFSDirectory]))
+
+(def ^:private analyzer (StandardAnalyzer.))
+
+(mount/defstate ^Directory search-index
+  :start (let [directory (NIOFSDirectory. (.toPath (io/file (:search-index-path env))))]
+           (with-open [writer (IndexWriter. directory (-> (IndexWriterConfig. analyzer)
+                                                          (.setOpenMode IndexWriterConfig$OpenMode/CREATE)))]
+             (.deleteAll writer))
```
opqdonut: right so the indexes are cleared every restart. that's good.

src/clj/rems/application/search.clj
```diff
@@ -0,0 +1,69 @@
...
+           [org.apache.lucene.document Document StringField Field$Store TextField]
+           [org.apache.lucene.index IndexWriter IndexWriterConfig IndexWriterConfig$OpenMode]
+           [org.apache.lucene.queryparser.flexible.standard StandardQueryParser]
+           [org.apache.lucene.search IndexSearcher ScoreDoc TopDocs SearcherManager SearcherFactory]
+           [org.apache.lucene.store Directory NIOFSDirectory]))
+
+(def ^:private analyzer (StandardAnalyzer.))
+
+(mount/defstate ^Directory search-index
+  :start (let [directory (NIOFSDirectory. (.toPath (io/file (:search-index-path env))))]
```
opqdonut: should you guard against a a nil :search-index-path here?

src/clj/rems/application/search.clj
```diff
@@ -0,0 +1,69 @@
...
+             (Long/parseLong id)))))
+
+(defn find-applications [^String query]
+  (refresh!) ; TODO: call from a background thread asynchronously?
+  (with-searcher
+   (fn [^IndexSearcher searcher]
+     (let [results (.search searcher
+                            (-> (StandardQueryParser. analyzer)
+                                (.parse query "applicant")) ; TODO: change defaultField to full text search
+                            10000)]
```
opqdonut: nit: magic number

src/clj/rems/application/search.clj
luontola: There is a default for it in config-defaults.edn, so it should never be nil.

src/clj/rems/application/search.clj
luontola: Shall we use `Integer/MAX_VALUE` instead? Anyways we have no paging, but the UI shows all applications by default.

PR #1478, opened 2019-08-09
Scroll to top after closing a modal dialog

### LINKED ISSUE #1477: Scrolling to top of page after modal closes
was implemented originally in #1337 (see also #1435) but is now broken

luontola:
Closes #1477

### DIFF
--- a/src/cljs/rems/status_modal.cljs
+++ b/src/cljs/rems/status_modal.cljs
@@ -82,7 +82,8 @@
                            :content [:<> [status-text success? error-content] content]
                            :on-close (fn []
                                        (close)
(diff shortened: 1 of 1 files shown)


okahilak:
The patch itself seems to work nicely, but I'm not sure if it's advantageous to scroll to the top, e.g., in the following cases:

- Accepting terms of use
- Inviting a new member to an application

In both cases, I would expect the window to remain in the same position so I can see the action modifying the page.

Then again, I'm not sure why there's a success pop-up in these cases in the first place, as the page changing should already give enough feedback to the user that the action was successful.

luontola:
See my comment on #1477

PR #1483, opened 2019-08-13
Search all application data

luontola:
#1352 and #873

### DIFF
--- a/src/clj/rems/application/search.clj
+++ b/src/clj/rems/application/search.clj
@@ -6,7 +6,8 @@
             [rems.application-util :as application-util]
             [rems.config :refer [env]]
             [rems.db.applications :as applications]
-            [rems.db.events :as events])
+            [rems.db.events :as events]
+            [rems.text :as text])
   (:import [org.apache.lucene.analysis.standard StandardAnalyzer]
            [org.apache.lucene.document Document StringField Field$Store TextField]
            [org.apache.lucene.index IndexWriter IndexWriterConfig IndexWriterConfig$OpenMode]
@@ -50,7 +51,13 @@
                       (mapcat (fn [resource]
                                 (vals (:catalogue-item/title resource))))
                       (str/join " "))
-        all (str/join " " [id applicant member title resource])]
+        state (->> (:languages env)
+                   (map (fn [lang]
+                          (text/with-language lang
+                            (fn []
+                              (text/localize-state (:application/state app))))))
+                   (str/join " "))
+        all (str/join " " [id applicant member title resource state])]
     ;; metadata
     (.add doc (StringField. "app-id" (str (:application/id app)) Field$Store/YES))
     ;; searchable fields
@@ -59,6 +66,7 @@
     (.add doc (TextField. "member" member Field$Store/NO))
     (.add doc (TextField. "title" title Field$Store/NO))
     (.add doc (TextField. "resource" resource Field$Store/NO))
+    (.add doc (TextField. "state" state Field$Store/NO))
     (.add doc (TextField. "all" all Field$Store/NO))
     (.addDocument writer doc)))
 
--- a/src/clj/rems/application/search.clj
+++ b/src/clj/rems/application/search.clj
@@ -32,6 +32,9 @@
 (defn- index-application! [^IndexWriter writer app]
   (log/info "Indexing application" (:application/id app))
   (let [doc (Document.)
+        id (->> [(:application/id app)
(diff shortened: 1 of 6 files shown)


src/clj/rems/application/search.clj
```diff
@@ -29,9 +32,48 @@
 
 (defn- index-application! [^IndexWriter writer app]
   (log/info "Indexing application" (:application/id app))
-  (let [doc (Document.)]
-    (.add doc (StringField. "id" (str (:application/id app)) Field$Store/YES))
-    (.add doc (TextField. "applicant" (str (:application/applicant app)) Field$Store/NO))
+  (let [doc (Document.)
+        id (->> [(:application/id app)
```
opqdonut: minor: how about making a separate `index-terms-for-application` function that returns a map like
```
{:id "3 2019/3"
 :applicant "Alice Applicant alice@appli.cant"
 ...}
```
This would be slightly cleaner, and also you could implement the `"all"` field by catenating all values of this map

src/clj/rems/application/search.clj
```diff
@@ -29,9 +32,48 @@
...
+                    (str/join " "))
+        title (:application/description app)
+        resource (->> (:application/resources app)
+                      (mapcat (fn [resource]
+                                (vals (:catalogue-item/title resource))))
+                      (str/join " "))
+        state (->> (:languages env)
+                   (map (fn [lang]
+                          (text/with-language lang
+                            (fn []
```
opqdonut: nit: why not `#()` here?

src/clj/rems/application/search.clj
```diff
@@ -29,9 +32,48 @@
...
+                              [(:userid member)
+                               (application-util/get-member-name member)
+                               (:mail member)]))
+                    (str/join " "))
+        title (:application/description app)
+        resource (->> (:application/resources app)
+                      (mapcat (fn [resource]
+                                (vals (:catalogue-item/title resource))))
+                      (str/join " "))
+        state (->> (:languages env)
```
opqdonut: nit: how about a `for-all-languages` helper?

test/clj/rems/application/test_search.clj
```diff
@@ -1,30 +1,104 @@
...
             [rems.application.search :as search]
+            [rems.db.applications :as applications]
             [rems.db.test-data :as test-data]
-            [rems.db.testing :refer [test-db-fixture rollback-db-fixture]]))
+            [rems.db.testing :refer [test-db-fixture search-index-fixture]]))
 
-(use-fixtures :once test-db-fixture)
-(use-fixtures :each rollback-db-fixture)
+(use-fixtures :once (join-fixtures [test-db-fixture
+                                    search-index-fixture]))
```
opqdonut: rollback not needed?

opqdonut:
looks good to me, some minor comments

test/clj/rems/application/test_search.clj
luontola: The rollback doesn't affect the search index, so it's better to not pretend that there is any rollback. It would only create weird error conditions if a second deftest were added.

src/clj/rems/application/search.clj
luontola: Let's wait until the code is duplicated somewhere. This far there is only one.

src/clj/rems/application/search.clj
luontola: 👍

src/clj/rems/application/search.clj
```diff
@@ -30,50 +30,46 @@
...
-    (.add doc (TextField. "applicant" applicant Field$Store/NO))
-    (.add doc (TextField. "member" member Field$Store/NO))
-    (.add doc (TextField. "title" title Field$Store/NO))
-    (.add doc (TextField. "resource" resource Field$Store/NO))
-    (.add doc (TextField. "state" state Field$Store/NO))
-    (.add doc (TextField. "form" form Field$Store/NO))
-    (.add doc (TextField. "all" all Field$Store/NO))
+    (doseq [[k v] terms]
+      (.add doc (TextField. (name k) v Field$Store/NO)))
+    (.add doc (TextField. "all" (str/join " " (vals terms)) Field$Store/NO))
```
opqdonut: if ordering matters, you should probably do something like

```
(select-vals terms (sort (keys terms)))

luontola: It doesn't matter.

PR #1487, opened 2019-08-14
Add rems.api.services.licenses, providing API with licenses

okahilak:


src/clj/rems/api/services/licenses.clj
```diff
@@ -0,0 +1,79 @@
+(ns rems.api.services.licenses
+  "Serving licenses for API."
+  (:require [rems.common-util :refer [distinct-by]]
+            [rems.db.core :as db]
+            [rems.db.licenses :as licenses]
+            [rems.db.resource :as resource]
+            [rems.db.workflow :as workflow])
+  (:import (java.io FileInputStream ByteArrayOutputStream)))
+
+(defn create-license! [{:keys [title licensetype textcontent localizations attachment-id]} user-id]
```
opqdonut: this feels like it should be in rems.db.licences

src/clj/rems/api/services/licenses.clj
```diff
@@ -0,0 +1,79 @@
...
+    (doseq [[langcode localization] localizations]
+      (db/create-license-localization! {:licid licid
+                                        :langcode (name langcode)
+                                        :title (:title localization)
+                                        :textcontent (:textcontent localization)
+                                        :attachmentId (:attachment-id localization)}))
+    {:success (not (nil? licid))
+     :id licid}))
+
+(defn create-license-attachment! [{:keys [tempfile filename content-type]} user-id]
```
opqdonut: ditto

src/clj/rems/api/services/licenses.clj
```diff
@@ -0,0 +1,79 @@
...
+                     (clojure.java.io/copy input buffer)
+                     (.toByteArray buffer))]
+    (select-keys
+     (db/create-license-attachment! {:user user-id
+                                     :filename filename
+                                     :type content-type
+                                     :data byte-array})
+     [:id])))
+
+(defn remove-license-attachment!
```
opqdonut: ditto

src/clj/rems/api/services/licenses.clj
```diff
@@ -0,0 +1,79 @@
+(ns rems.api.services.licenses
```
opqdonut: consider naming this `rems.services.licenses`

src/clj/rems/api/services/licenses.clj
okahilak: Hmm, one problem is that in the current implementation, the functions in db.licenses have wrappers in api.services.licenses so that api.licenses can depend only on api.services.licenses and not on db.licenses directly.

I could change api.licenses to depend also on db.licenses, but I'm not sure what's the logic from api.licenses viewpoint behind figuring out which functions belongs where -- it feels arbitrary. Having wrappers here makes it at least consistent.

But moving (most of) these functions to db.licenses makes this intermediate step mostly just a collection of wrappers.

One criterion would be to have a minimal set of functions in db.licenses so that the other services have the functions they need, but not more. It is how it's done in the current implementation -- I'll need get-licenses in other services, but not create-license!.

src/clj/rems/api/services/licenses.clj
okahilak: Hmm, I find it slightly "non-informative" to have a directory structure where src/clj/rems/services, /api, and /db directories are all at the same level (together with, e.g., /css), while I feel that services is kind of subordinate to api -- I wouldn't understand from that directory structure alone what "services" is related to. Though I guess it could be rems.db.services, as well.

This is just from the directory structure viewpoint, but I've understood that the namespace naming needs to conform to that.

src/clj/rems/api/services/licenses.clj
opqdonut: Ok, that'll work I guess

src/clj/rems/api/services/licenses.clj
opqdonut: Fair enough

opqdonut:
Let's go with this and see where it takes us. Once you've got the structure figured out, a short ADR might make sense.

PR #1492, opened 2019-08-14
circleci: kill more apt-daily stuff

opqdonut:


opqdonut:
with this workaround 6/6 builds passed, so the issue seems fixed

PR #1504, opened 2019-08-15
Refactor fetching licenses for workflow

okahilak:
- Previously, /api/workflows/ did not work correctly, returning
  an empty vector of licenses for a workflow regardless of if it
  had licenses or not.

- Simplify SQL queries and move parts of the logic to clojure side,
  namely, joining workflow-licenses and the actual license data,
  which has the benefit of having to do fewer JSON round-trips.

- Previously, workflow had license localizations that were formatted
  differently compared to fetching only the license using API, which
  will cause potential trouble when using the license.

  To fix that, use rems.db.licenses for joining license data to workflow,
  ensuring that the license is always formatted in the same way.

- Consequently, remove license-related parts from rems.db.workflow.

- Unify License and WorkflowLicense schemas: use License schema for
  workflows, as well.

- Be more explicit about using workflow_license table's start and end times
  when joining a license to a workflow, previously it was slightly hidden
  in the SQL query returning start, start_2, end, and end_2, from which
  the correct keys were selected.

### DIFF
--- a/resources/sql/queries.sql
+++ b/resources/sql/queries.sql
@@ -331,17 +331,15 @@ UPDATE workflow_licenses wl
 SET start = :start, endt = :end
 WHERE wl.licid = :licid;
 
+-- :name get-workflow-licenses :? :*
+SELECT licid, start, endt as "end"
+FROM workflow_licenses
+WHERE wfid = :wfid
+
 -- :name get-workflow :? :1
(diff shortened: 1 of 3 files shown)


okahilak:
It would probably be appropriate to add regression tests for the changes of this patch, I'll do it tomorrow.

PR #1508, opened 2019-08-16
Refactor fetching licenses for workflow

okahilak:
- Previously, /api/workflows/ did not work correctly when fetching
  all workflows. It returned an empty vector of licenses for a
  workflow regardless of if it had licenses or not. (Do not add
  test for that because having workflows contain licenses is a
  legacy feature, so it is would be more difficult to test.)

- Simplify SQL queries and move parts of the logic to clojure side,
  namely, joining workflow-licenses and the actual license data,
  which has the benefit of doing fewer JSON round-trips.

- Previously, workflow had license localizations that were formatted
  differently compared to fetching only the license using API, which
  will cause potential trouble when using the license.

  To fix that, use rems.db.licenses for joining license data to workflow,
  ensuring that the license is always formatted in the same way.

- Consequently, remove license-related parts from rems.db.workflow.

- Unify License and WorkflowLicense schemas: use License schema for
  workflows, as well.

- Be more explicit about using workflow_license table's start and end times
  when joining a license to a workflow, previously it was slightly hidden
  in the SQL query returning start, start_2, end, and end_2, from which
  the correct keys were selected.

### DIFF
--- a/src/clj/rems/db/workflow.clj
+++ b/src/clj/rems/db/workflow.clj
@@ -1,67 +1,40 @@
 (ns rems.db.workflow
-  (:require [clj-time.core :as time-core]
-            [clojure.test :refer [deftest is]]
-            [rems.db.catalogue :as catalogue]
-            [rems.db.core :as db]
+  (:require [rems.db.core :as db]
+            [rems.db.licenses :as licenses]
             [rems.db.users :as users]
             [rems.json :as json]
-            [rems.util :refer [update-present]])
-  (:import [org.joda.time DateTime DateTimeZone]))
+            [rems.util :refer [update-present]]))
 
-(defn- parse-db-time [s]
-  (when s
-    (-> (DateTime/parse s)
-        (.withZone DateTimeZone/UTC))))
+;; XXX: Overwriting :start and :end from license table with :start and :end
+;;      from workflow_license table seems error-prone - they could at least
+;;      be named differently to avoid confusion.
+;;
+;;      See a related comment in rems.db.licenses regarding the use of
+;;      various start and end times.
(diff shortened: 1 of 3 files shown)


src/clj/rems/db/workflow.clj
```diff
@@ -1,67 +1,40 @@
...
   (-> wf
-      (update-present :workflow update :handlers #(mapv users/get-user %))
-      format-workflow))
-
-(defn- parse-workflow-body [json]
-  (json/parse-string json))
-
-(defn- parse-licenses [json]
-  (json/parse-string json))
+      (update :workflow #(json/parse-string %))
```
luontola: `#(json/parse-string %)` could be just `json/parse-string`

src/clj/rems/db/workflow.clj
```diff
@@ -1,67 +1,40 @@
...
+      (assoc :licenses (get-workflow-licenses (:id wf)))
+      db/assoc-expired
+      (update-present :workflow update :handlers #(mapv users/get-user %))))
```
luontola: Since the old workflows have been removed, `:workflow` is always present. This could be changed to `(update-in [:workflow :handlers] #(mapv users/get-user %))`. Though some day there might be a workflow without `:handlers`, in which case `(update :workflow update-present :handlers #(mapv users/get-user %))` is more future proof.

src/clj/rems/db/workflow.clj
okahilak: Done.

src/clj/rems/db/workflow.clj
okahilak: Done.

Issue #832, opened 2019-01-15
vTHL4: Update approved application


foxlynx:
Applicant needs to be able to update an existing, approved application, for example to continue the research and keeping access valid. A new application is created that refers to the old one. The new application will start with a copy of the information in the old one and will also show a diff compared to it. 
The handler needs to see that the new application is actually an update, not a new application. He/she also needs to see the original application and the diff. 

**Update 28.2.2019: From demo there was a requirement that the approved application needs to be seen as it was when it was approved. This should be kept in mind when implementing this story.** 

- [x] application action to update (UI action that triggers new command type?) 
     --> old application is closed
     --> new backend event would contain a link to the old application 
- [x] make a copy of the application with a link to the old one: link to old shown to the applicant in the application title?
- [x] handler can see that this is an update application (e.g. heading)
- [x] **handler** can diff changes compared to the old application
    * diff always to last submit
- [x] under the hood it is a new application
- [ ] ~~hide the old application from normal views if there is a linked new application~~
     * old application is closed, so it is not shown by default

**Entitlements**
- Old application entitlements are revoked when/if handler closes the old application
- When handler approves the new application, entitlements are granted

PR #1512, opened 2019-08-16
Copy as new application

luontola:
#832


src/clj/rems/application/commands.clj
```diff
@@ -392,9 +398,10 @@
   [cmd application injections]
   (or (already-member-error application (:actor cmd))
       (invitation-token-error application (:token cmd))
-      (ok {:event/type :application.event/member-joined
-           :application/id (:application-id cmd)
-           :invitation/token (:token cmd)})))
+      (ok-with-data {:application-id (:application-id cmd)}
```
opqdonut: why do you need to return the app id? it was specified in the cmd so the sender knows it.

src/clj/rems/application/commands.clj
```diff
@@ -417,14 +424,32 @@
...
+        new-app-id (:application-id result)]
+    (ok-with-data
+     {:application-id new-app-id}
+     {:event/type :application.event/draft-saved
+      :application/id new-app-id
+      :application/field-values (->> (:form/fields (:application/form application))
+                                     (map (fn [field]
+                                            [(:field/id field) (:field/value field)]))
+                                     (into {}))}
+     {:event/type :application.event/copied-from
```
opqdonut: I find it a bit disturbing that `create-application!` emits the created event and then we emit other events here. Couldn't you emit the created event explicitly?

src/clj/rems/application/model.clj
```diff
@@ -432,20 +441,19 @@
...
+               (let [item (get-catalogue-item (:catalogue-item/id resource))]
+                 {:resource/id (:resource-id item)
+                  :catalogue-item/title (assoc (localization-for :title item)
+                                               :default (:title item))
+                  ;; TODO: remove unused keys
+                  :catalogue-item/start (:start item)
+                  :catalogue-item/end (:end item)
+                  :catalogue-item/enabled (:enabled item)
+                  :catalogue-item/expired (:expired item)
+                  :catalogue-item/archived (:archived item)}))))
```
opqdonut: I preferred the old version, it was more explicit about which keys are emitted.

src/clj/rems/application/commands.clj
```diff
@@ -247,8 +251,10 @@
     {:errors [{:type :already-member :application-id (:id application)}]}))
 
 (defn- ok [& events]
-  {:success true
-   :events events})
+  {:events events})
+
+(defn- ok-with-data [data & events]
```
opqdonut: I'd prefer `[data events]`

test/clj/rems/test_locales.clj
```diff
@@ -10,9 +10,11 @@
             [taoensso.tempura.impl :refer [compile-dictionary]])
   (:import (java.io FileNotFoundException)))
 
-(def loc-en (read-string (slurp (io/resource "translations/en.edn"))))
+(defn loc-en []
+  (read-string (slurp (io/resource "translations/en.edn"))))
```
opqdonut: good catch

src/clj/rems/application/commands.clj
luontola: That simplifies the code in https://github.com/CSCfi/rems/pull/1512/files#diff-5045c91462d19bb329c2d32136b4f565R137-R140 by using common code instead of the old custom solution https://github.com/CSCfi/rems/pull/1512/files#diff-6abc0721f32efbf6b854ba6af49ae23fL134-L143

src/clj/rems/application/commands.clj
luontola: I can try refactor it. `rems.db.applications/create-application!` needs some refactoring to do that.

opqdonut: I'm fine with a TODO comment and refactoring in a follow-up PR

src/clj/rems/application/model.clj
luontola: The problem is that in the tests `item` is `nil`, which results in even the known IDs being nilled.

opqdonut: well how about still spelling out the keys instead of merging? that is,
```
{:catalogue-item/id (:catalogue-item/id resource)
 :resource/ext-id (:resource/ext-id resource)
 ...
 :catalogue-item/start (:start item)}

Comment (luontola) [reply src/clj/rems/application/model.clj]: 👍

File: src/clj/rems/application/commands.clj (hunk 2 above)
Comment (luontola) [reply src/clj/rems/application/commands.clj]: 👍

File: src/clj/rems/application/events.clj [hunk 6]
```diff
@@ -121,6 +121,11 @@
 (s/defschema SubmittedEvent
   (assoc EventBase
          :event/type (s/enum :application.event/submitted)))
+(s/defschema CopiedFromEvent
+  (assoc EventBase
+         :event/type (s/enum :application.event/copied-from)
+         :application/copied-from {:application/id s/Int
+                                   :application/external-id (s/maybe s/Str)}))
```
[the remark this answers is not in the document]
luontola: Yes. I'll do that in another PR.

PR #1519, opened 2019-08-20
Link the old application to the copies made from it

### LINKED ISSUE #832: vTHL4: Update approved application

luontola:
Closes #832

### DIFF
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -52,7 +52,8 @@
   (let [id-prefix (str (.getYear time))]
     (format-external-id (allocate-external-id! id-prefix))))
 
-(defn application-created-event [{:keys [application-id catalogue-item-ids time actor allocate-external-id?]}]
+(defn application-created-event [{:keys [application-id catalogue-item-ids time actor allocate-external-id?]
+                                  :or {allocate-external-id? true}}]
   (assert (seq catalogue-item-ids) "catalogue item not specified")
   (let [application-id (or application-id (:id (db/create-application!)))
         items (catalogue/get-localized-catalogue-items {:ids catalogue-item-ids})]
@@ -88,8 +89,7 @@
 (defn create-application2 [{:keys [catalogue-item-ids time actor]}]
   (application-created-event {:catalogue-item-ids catalogue-item-ids
                               :time time
-                              :actor actor
-                              :allocate-external-id? true}))
+                              :actor actor}))
 
 (defn create-application! [user-id catalogue-item-ids]
   (let [event (create-application2 {:catalogue-item-ids catalogue-item-ids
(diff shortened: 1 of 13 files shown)


src/cljs/rems/application.cljs
```diff
@@ -431,6 +434,23 @@
...
+    (when (or old-app new-apps)
+      [:<>
+       " ("
+       (when old-app
+         [:<> (text :t.applications/copied-from) " " [application-link old-app nil]])
+       (when (and old-app new-apps)
+         "; ")
+       (when new-apps
+         (into [:<> (text :t.applications/copied-to) " "]
+               (->> (for [new-app new-apps]
```
opqdonut: [nit] I'd prefer a direct call instead of threading here. If you want to keep threading, also thread the `into`

src/clj/rems/db/applications.clj
```diff
@@ -52,9 +52,11 @@
   (let [id-prefix (str (.getYear time))]
     (format-external-id (allocate-external-id! id-prefix))))
 
-(defn application-created-event [{:keys [application-id catalogue-item-ids time actor allocate-external-id?]}]
+(defn application-created-event [{:keys [application-id catalogue-item-ids time actor allocate-external-id?]
```
opqdonut: How about keeping this function pure and always passing in the application-id? That way it would suffice to inject `db/create-application!` and always use this pure function to generate the created event.

opqdonut: Hmm okay this function is fetching a lot of stuff from the db. I wonder if it could be turned into a function in `rems.application.commands` that takes the injections map as a parameter.

luontola: It's a complex beast.

Issue #1529, opened 2019-08-22
Remove the unlocalized text/title feature


opqdonut:
It's at least in the catalogue items. It doesn't make sense to localize everything that is not shown to applicants i.e. form name.

opqdonut:
licenses also seem to have an unlocalized title

PR #1534, opened 2019-08-22
Remove unlocalized title for catalogue item

okahilak:
Part of fixing #1529

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [x] all fields have a label
- [x] contrast is checked

## Follow-up


src/cljc/rems/text.cljc
```diff
@@ -49,8 +50,13 @@
         (first (vals m)))))
```
luontola: This `:default` can be removed.

src/clj/rems/api/schema.clj
```diff
@@ -6,21 +6,21 @@
...
-   (s/optional-key :langcode) s/Keyword
    :start DateTime
    :end (s/maybe DateTime)
    :enabled s/Bool
    :archived s/Bool
    :expired s/Bool
-   :localizations (s/maybe {s/Any s/Any})})
+   :localizations (s/maybe {s/Keyword {:id s/Int
+                                       :langcode s/Keyword
+                                       :title s/Str}})})
```
luontola: Could `s/maybe` be removed?

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -68,11 +68,12 @@
  ::catalogue-table-rows
  (fn [_ _]
    [(rf/subscribe [::catalogue])
-    (rf/subscribe [:language])])
- (fn [[catalogue language] _]
+    (rf/subscribe [:language])
+    (rf/subscribe [:default-language])])
```
luontola: The `:language` subscription already falls back to `:default-language`, so the fallback doesn't need to be done again.

src/cljc/rems/text.cljc
```diff
@@ -49,8 +50,13 @@
...
 ;; TODO: replace usages of `get-localized-title` with `localized`
-(defn get-localized-title [item language]
+;;
+;; TODO: remove fallback that fetches :title from item: localizations
+;;;  could be forced to always have the title in either the language of
+;;   choice or the default language.
+(defn get-localized-title [item language default-language]
   (or (get-in item [:localizations language :title])
+      (get-in item [:localizations default-language :title])
       (:title item)))
```
luontola: The admin UI should already force translations for all languages to be filled in, so all these fallbacks should be unnecessary.

okahilak: Yeah, I started thinking along the same lines after your previous comment. The problem is that at least LBR has some missing Finnish translations, but maybe I'll just copy the English translations to fill them in, in the migrations.

src/clj/rems/api/schema.clj
okahilak: I considered that, but the problem is that API allows adding a catalogue item without adding localizations, which would then cause an error when fetching the catalogue item. Maybe the API could be changed so that localizations would be added with a single API call, but that would probably be another patch...

src/cljc/rems/text.cljc
okahilak: I think not yet, because licenses also have unlocalized titles that are used as :default keys. I will add a TODO comment here to remove this once all unlocalized titles are removed.

src/clj/rems/api/schema.clj
luontola: An empty map should pass the schema without s/maybe

src/cljc/rems/text.cljc
luontola: 👌

src/cljc/rems/text.cljc
luontola: A more reliable fallback, in case the item is missing the requested translation, is to randomly pick whatever translation the item does have.

luontola: It's possible to have a situation that the default language is English, but the item has only Finnish translation.

src/clj/rems/api/schema.clj
okahilak: Ah, true. I'll do it that way.

src/cljc/rems/text.cljc
okahilak: True, I'll use that instead.

okahilak: I ended up using just the fallback to care for missing Finnish translations for catalogue items in the LBR database and not add extra logic to migrations. LBR can probably fill in the missing translations using the UI once the catalogue items are editable.

src/cljc/rems/text.cljc
```diff
@@ -41,23 +41,20 @@
...
+        ;;   fallback can be removed.
         (get m :default)
         (first (vals m)))))
 
 ;; TODO: replace usages of `get-localized-title` with `localized`
-;;
-;; TODO: remove fallback that fetches :title from item: localizations
-;;;  could be forced to always have the title in either the language of
-;;   choice or the default language.
 (defn get-localized-title [item language default-language]
```
luontola: default-language is an unused parameter.

luontola:
rems.db.test-data/create-catalogue-item! has still one `:title ""` which can be removed.

luontola:
This change could be mentioned in the changelog.

PR #1535, opened 2019-08-23
Document the search syntax

### LINKED ISSUE #1531: Document full-text search query syntax
It uses the Lucene query syntax: http://lucene.apache.org/core/8_2_0/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#package.description

Field names should also be documented.

luontola:
Closes #1531

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -1,0 +1,54 @@
+# Search
+
+## Application Search
+
+The application listings have a search which uses the [Lucene query parser syntax][query-syntax]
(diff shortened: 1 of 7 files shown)


opqdonut:
I was actually expecting something in the UI, but this is nice also. Consider adding some example queries to the UI?

luontola:
Maybe examples of prefix search and the + - modifiers? And a link to the wiki page.

opqdonut:
Yeah, something like that.

Issue #1469, opened 2019-08-08
Remove success pop-up windows


okahilak:
LBR preferred success pop-up windows to not be shown to the user.

opqdonut:
See also #1461 . Consider "flash" messages for success, perhaps accompanied with button animations.

opqdonut:
Actually, this is superseded by #1461

opqdonut:
(closed by accident)

luontola:
Possibly related to #1501

opqdonut:
Meeting on 2019-08-22: let's just do this.

PR #1537, opened 2019-08-26
floating save button

### LINKED ISSUE #1525: Make saving easier (e.g. with a floating save button)
This will mitigate some of the problems discussed in #979

opqdonut:
closes #1525

### DIFF
--- a/src/cljs/rems/flash_message.cljs
+++ b/src/cljs/rems/flash_message.cljs
@@ -23,7 +23,7 @@
 (rf/reg-event-fx
  ::show-flash-message
  (fn [{:keys [db]} [_ message]]
-   (focus/focus-element-async "#flash-message" {:scroll? true})
+   (focus/focus-element-async "#flash-message")
    ;; TODO: flash the message with CSS
    {:db (assoc db ::message (assoc message :expires (+ 500 (current-time-millis))))}))
 
--- a/src/cljs/rems/flash_message.cljs
+++ b/src/cljs/rems/flash_message.cljs
@@ -19,8 +19,7 @@
    ;; On Chrome, focusing the element scrolls it fully into view,
    ;; but on Firefox the element is hidden behind the navigation menu,
    ;; so explicit scrolling is needed.
-   (.scrollTo js/window 0 0)
-   (focus/focus-element-async "#flash-message")
+   (focus/focus-element-async "#flash-message" {:scroll? true})
    ;; TODO: flash the message with CSS
    {:db (assoc db ::message message)}))
 
--- a/src/cljs/rems/flash_message.cljs
+++ b/src/cljs/rems/flash_message.cljs
@@ -16,9 +16,6 @@
 (rf/reg-event-fx
  ::show-flash-message
  (fn [{:keys [db]} [_ message]]
-   ;; On Chrome, focusing the element scrolls it fully into view,
-   ;; but on Firefox the element is hidden behind the navigation menu,
-   ;; so explicit scrolling is needed.
    (focus/focus-element-async "#flash-message" {:scroll? true})
    ;; TODO: flash the message with CSS
    {:db (assoc db ::message message)}))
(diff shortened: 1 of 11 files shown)


opqdonut:
The two commits here present two options for a floating actions div. Which do you prefer?

## float bottom
float-bottom

## float right
float-right

hukka:
Float on right stands out more, and leaves more space for the actual form. Can we do both easily, with breakpoints, so that we use the horizontal space when it's available?

opqdonut:
We looked at this with @Macroz and moved the actions pane below the application title

Responsiveness works: in desktop & tablet sizes we get the floating div. In mobile sizes the actions stay at the bottom.

The remaining issue is that saving a draft reloads the application and scrolls to the top, which kinda defeats the purpose of a floating button. Waiting for #1469 to get implemented first.

opqdonut:
Rebased on top of fresh master.

luontola:
#1549 has been merged, so this can proceed.

opqdonut:
Flash messages now appear above the floating actions div and don't scroll the page.

As a side-effect, in phone layout the actions div is at the top.

opqdonut:
I think this is ready now

okahilak:
Hmm, so now it looks like this:

image

I might prefer the message element to be below the action panel, that somehow corresponds more to the causal direction of the action and its message.

src/cljs/rems/focus.cljs
```diff
@@ -2,13 +2,18 @@
...
-         (.setAttribute element "tabindex" "-1")
-         (.focus element))
-       (js/setTimeout #(focus-element-async selector (dec tries))
-                      10)))))
+  "Focus an element when it appears. Options can include:
+    :tries -- number of times to poll, defaults to 100
+    :scroll? -- whether to .scollIntoView the element"
+  [selector & [options]]
+  (let [tries (get options :tries 100)
+        scoll? (get options :scroll? false)]
```
okahilak: There's a typo here. It seems to work anyway, which is kind of funny & strange. However, I assume that having :scroll? as true wouldn't work.

src/cljs/rems/focus.cljs
```diff
@@ -2,13 +2,18 @@
...
-   (when (pos? tries)
-     (if-let [element (.querySelector js/document selector)]
-       (do
-         (.setAttribute element "tabindex" "-1")
-         (.focus element))
-       (js/setTimeout #(focus-element-async selector (dec tries))
-                      10)))))
+  "Focus an element when it appears. Options can include:
+    :tries -- number of times to poll, defaults to 100
+    :scroll? -- whether to .scollIntoView the element"
```
okahilak: a typo

okahilak:
> I might prefer the message element to be below the action panel, that somehow corresponds more to the causal direction of the action and its effects.

Ok, so this might be difficult to get to work together with the alternative position for the action panel on the top of the page, in which case it is nice to have the flash message element above the panel.

So, all in all, I'm not proposing this change in this patch after all. I think we can fine-tune the positioning later on, if need be.

src/cljs/rems/flash_message.cljs
```diff
@@ -19,8 +19,7 @@
    ;; On Chrome, focusing the element scrolls it fully into view,
    ;; but on Firefox the element is hidden behind the navigation menu,
    ;; so explicit scrolling is needed.
```
opqdonut: this bug now happens again in responsive mode. I'll have a look if I can fix it...

opqdonut: created #1578

opqdonut: created #1578

foxlynx:
Needs a bit more consistency in success announcements:

* better feedback for adding attachment
* better feedback when inviting/adding member

Now the two cases above appear to the UI without notification or highlight colors, which is ok if it is consistent I guess, but when accepting terms there is a green higlighted success box on spot, so it is confusing for user

opqdonut:
Some stuff that came up while working on this: #1580 #1581

PR #1538, opened 2019-08-26
disallow submitting disabled/archived/expired catalogue items

### LINKED ISSUE #1524: Shouldn't be able to submit application if catalogue item is disabled/archived/expired
... but submitted applications can still be handled.

opqdonut:
fixes #1524

### DIFF
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -100,6 +100,10 @@
 (mount/defstate application-cache
   :start (atom (cache/ttl-cache-factory {} :ttl 10000)))
 
+;; TODO combine with reload-cache!?
+(defn reset-application-cache! []
+  (swap! application-cache empty))
+
 (defn get-unrestricted-application
   "Returns the full application state without any user permission
    checks and filtering of sensitive information. Don't expose via APIs."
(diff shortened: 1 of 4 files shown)


src/clj/rems/db/applications.clj
```diff
@@ -100,6 +100,10 @@
 (mount/defstate application-cache
   :start (atom (cache/ttl-cache-factory {} :ttl 10000)))
 
+;; TODO combine with reload-cache!?
+(defn reset-application-cache! []
+  (swap! application-cache empty))
```
luontola: Yes, better combine with `reload-cache!`.

test/clj/rems/api/test_applications.clj
```diff
@@ -432,6 +434,38 @@
...
+(deftest test-application-submit
+  (let [user-id "alice"
+        form-id (test-data/create-form! {})
+        cat-id (test-data/create-catalogue-item! {:form-id form-id})
+        app-id (test-data/create-application! {:catalogue-item-ids [cat-id] :actor user-id})]
+    (testing "submit with disabled catalogue item fails"
+      (is (:success (catalogue/update-catalogue-item! {:id cat-id
+                                                       :enabled false
+                                                       :archived false})))
+      (rems.db.applications/reset-application-cache!)
```
luontola: For this test you could stop rems.db.applications/application-cache, in which case the reads would happen without a cache.

opqdonut: I think having an explicit cache reset makes the problem more visible so I'll leave it.

src/clj/rems/db/applications.clj
opqdonut: I'll do this

PR #1539, opened 2019-08-27
Remove start, end and expiration from all licenses

### LINKED ISSUE #1116: Only use license table expiration
We have start and end dates for resource and workflow licenses stored separately and in the common license table. We should have only one place for the, i.e. the common table.

Now a license can be expired in the resource licenses sense but not in the general license sense. And the API will happily return it when you get all licenses.

- [ ] only have start and end dates in licenses table
- [ ] migration for old db
- [ ] test for this case of expiration in the API

opqdonut:
Fixes #1116 

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up


test/clj/rems/api/test_workflows.clj
```diff
@@ -99,10 +99,9 @@
                              :handlers [{:userid "bob" :email "bob@example.com" :name "Bob Approver"}
                                         {:userid "carl" :email "carl@example.com" :name "Carl Reviewer"}]}
                   :enabled true
-                  :expired false
                   :archived false}
         fetch (fn []
-                (let [wfs (-> (request :get "/api/workflows" {:archived true :expired true :disabled true})
+                (let [wfs (-> (request :get "/api/workflows" {:archived true :disabled true})
```
luontola: The workflow API still has expired.

opqdonut: bah I botched a revert. I'll amend

test/clj/rems/db/test_licenses.clj
```diff
@@ -24,23 +24,15 @@
...
+                                            {:id :normal2 :enabled true :archived false}
+                                            {:id :disabled :enabled false :archived false}
+                                            {:id :archived :enabled true :archived true}])]
+    (testing "filters"
+      (is (= [:normal :normal2 :disabled :archived]
+             (map :id (get-all-licenses {}))))
+      (is (= [:archived]
+             (map :id (get-all-licenses {:archived true}))))
+      (is (= [:normal :normal2]
+             (map :id (get-all-licenses {:archived false :enabled true})))))))
```
luontola: This test is missing come cases: `{:archived false}`, `{:enabled true}`, `{:enabled false}`

Possibly also the various combinations of :archived and :enabled, if they are interesting.

opqdonut: I'll write more tests

PR #1547, opened 2019-08-28
Add option to rename catalogue item

### LINKED ISSUE #1507: Rename catalogue item
Spinoff from https://github.com/CSCfi/rems/issues/867

Invalidate cache for related items.

okahilak:
Closes #1507

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [x] all fields have a label
- [x] contrast is checked

## Follow-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Changes since v2.4
 - The pdf button and API have been removed. We recommend using "print
   to pdf" in your browser.
 - The `start`, `end` and `expired` fields have been removed from licenses
+- API for creating catalogue item and its localizations has been changed.
+  There is now a single API call that is used to create both a catalogue
+  item and the localizations, namely, /api/catalogue-items/create.
 
 ### Additions
 - New field types: description, option, multiselect
@@ -39,6 +42,7 @@ Changes since v2.4
 - Extra pages (#472)
 - Full-text search for all application content (#873)
 - Creating a new application as a copy from an older application (#832)
+- Re-naming a catalogue item (#1507)
 
 ### Enhancements
 - Improved version information in footer
--- a/src/clj/rems/api/catalogue_items.clj
+++ b/src/clj/rems/api/catalogue_items.clj
@@ -17,17 +17,17 @@
   {:form s/Int
    :resid s/Int
    :wfid s/Int
+   :localizations {s/Keyword {:title s/Str}}
    (s/optional-key :enabled) s/Bool
    (s/optional-key :archived) s/Bool})
+
+(s/defschema EditCatalogueItemCommand
+  {:id s/Int
+   :localizations {s/Keyword {:title s/Str}}})
 
 (s/defschema CreateCatalogueItemResponse
   {:success s/Bool
    :id s/Int})
-
-(s/defschema CreateCatalogueItemLocalizationCommand
-  {:id s/Int
-   :langcode s/Str
-   :title s/Str})
 
 ;; TODO use declarative roles everywhere
 (def catalogue-items-api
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Changes since v2.4
 - The pdf button and API have been removed. We recommend using "print
   to pdf" in your browser.
 - The `start`, `end` and `expired` fields have been removed from licenses
+- API for creating catalogue item and its localizations has been changed.
+  There is now a single API call that is used to create both a catalogue
+  item and the localizations, namely, /api/catalogue-items/create.
```
opqdonut: we seem to be gathering quite a lot of these breaking API changes. we'll soon need to be more careful with them

src/clj/rems/api/catalogue_items.clj
```diff
@@ -69,16 +69,16 @@
...
       :roles #{:owner}
-      :body [command UpdateStateCommand]
+      :body [command EditCatalogueItemCommand]
       :return SuccessResponse
-      (ok (catalogue/update-catalogue-item! command)))
+      (ok (catalogue/edit-catalogue-item! command)))
 
-    (POST "/create-localization" []
-      :summary "Create a new catalogue item localization"
+    (PUT "/update" []
```
opqdonut: I can't see any meaningful difference between `/edit` and `/update`. How about just adding new fields to UpdateStateCommand?

src/clj/rems/db/test_data.clj
```diff
@@ -128,18 +129,14 @@
...
-                                                                     :title text})]
-          (assert (:success result) {:command [id lang text] :result result}))))
-    id))
+  (let [localizations (map-vals (fn [title] {:title title}) title)
+        result (catalogue/create-catalogue-item!
+                {:resid (or resource-id (create-resource! {}))
+                 :form (or form-id (create-form! {}))
+                 :wfid (or workflow-id (create-dynamic-workflow! {}))
+                 :localizations (or localizations {})})
+        _ (assert (:success result) {:command command :result result})]
```
opqdonut: [nit] no need for the `_` trick here, the assert can just be in the body of the let

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -64,6 +64,11 @@
    (str "/#/administration/catalogue-items/" catalogue-item-id)
    (text :t.administration/view)])
 
+(defn- to-edit-catalogue-item [catalogue-item-id]
```
opqdonut: could this be shared with `catalogue-item/edit-button`?

src/clj/rems/api/catalogue_items.clj
okahilak: Hmm. So, this naming is analogous to how similar operations are named in forms. There the distinction is meaningful because updating form [state], i.e., changing archived/disabled status, is always doable, whereas editing form can only be done if the form is not in use. Also, editing form shares schema with creating a new form, so it made sense there to keep the operations distinct.

There's also UpdateStateCommand defined in Schemas, which consists of fields id, enabled, and archived, and which is in use for catalogue items, forms, licenses, and resources.

However, workflow has UpdateWorkflowCommand, which includes fields for editing the actual workflow (i.e., handlers and title).

Based on this, I'm not quite sure which route to take. My feeling is that I'd like to keep things modular and extract the common operations (archiving/disabling) to their own API call to have a more uniform API over different kinds of "elements" (resources, workflows, etc.) However, the naming is arguably confusing -- update should probably be update-state or similar.

opqdonut: Hmmm ok you're right. Maybe the operations need to be separate, but I'd like better names. We can go with update for now but could you create a technical debt ticket about renaming update to update-state (or whatever)?

src/cljs/rems/administration/catalogue_items.cljs
okahilak: Almost... It seems that one is btn-primary whereas the other is btn-secondary. It seems slightly messier to pass that information to the function than to have two separate functions.

Also, it would build a dependence between rems.administration.catalogue-items and .catalogue-item where there is currently none (except via page change).

Therefore, it would seem a tiny bit better to keep them separate for now.

src/clj/rems/db/test_data.clj
okahilak: Done.

src/clj/rems/api/catalogue_items.clj
okahilak: Done. https://github.com/CSCfi/rems/issues/1548

src/cljs/rems/administration/catalogue_items.cljs
opqdonut: fair enough

PR #1549, opened 2019-08-28
Use flash messages on application page

luontola:
#1469

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Accessibility
- [x] all fields have a label
- [x] contrast is checked

## Follow-up

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -18,7 +18,7 @@
             [rems.actions.return-action :refer [return-action-button return-form]]
             [rems.actions.review :refer [review-action-button review-form]]
             [rems.application-util :refer [accepted-licenses? form-fields-editable? get-member-name]]
-            [rems.atoms :as atoms :refer [external-link file-download info-field readonly-checkbox textarea document-title success-symbol]]
+            [rems.atoms :refer [external-link file-download info-field readonly-checkbox textarea document-title success-symbol]]
             [rems.collapsible :as collapsible]
             [rems.common-util :refer [index-by]]
             [rems.fields :as fields]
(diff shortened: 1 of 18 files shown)


src/cljs/rems/application.cljs
```diff
@@ -87,7 +87,7 @@
 (rf/reg-event-fx
  ::enter-application-page
  (fn [{:keys [db]} [_ id]]
-   {:db (dissoc db ::application ::edit-application)
+   {:db (dissoc db ::application ::edit-application ::flash-message/message)
```
opqdonut: hmm, clearing somebody else's "private" state feels a bit iffy

src/cljs/rems/flash_message.cljs
```diff
@@ -0,0 +1,47 @@
...
+  (:require [clojure.string :as str]
+            [re-frame.core :as rf]
+            [rems.atoms :as atoms]))
+
+(rf/reg-sub ::message (fn [db _] (::message db)))
+
+(rf/reg-event-fx
+ ::show-flash-message
+ (fn [{:keys [db]} [_ message]]
+   (.scrollTo js/window 0 0)
```
opqdonut: shouldn't we scroll to the flash message component?

okahilak:
Most of the flash messages seem to work nicely! I noticed one that had a slight visual glitch. When submitting an application that has an empty field, there is first a green "success" flash message, which is quickly replaced by the "error" flash message.

I also noticed that "Close Application" does not work properly -- it is left open after closing, which allows another "Close" operation, which causes an error. However, it turns out that this patch did not introduce the problem, but it exists already in master, so I created another issue for that: https://github.com/CSCfi/rems/issues/1562

luontola:
That fixes the glitch.

okahilak:
> That fixes the glitch.

Seems to work now, and the code change looks fine. Thanks.

PR #1551, opened 2019-08-28
Use case-insensitive sorting in tables

### LINKED ISSUE #1527: Make sorting in tables case-insensitive

okahilak:
Closes #1527

## Reviewability
- [ ] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up

### DIFF
--- a/src/cljs/rems/table.cljs
+++ b/src/cljs/rems/table.cljs
@@ -1,6 +1,5 @@
 (ns rems.table
-  (:require [cljs.test :refer-macros [deftest is testing]]
-            [clojure.string :as str]
+  (:require [clojure.string :as str]
             [re-frame.core :as rf]
             [rems.atoms :refer [close-symbol search-symbol sort-symbol]]
             [rems.search :as search]
@@ -68,14 +67,18 @@
     m
     (assoc m k (make-default m))))
 
-(defn- apply-row-defaults [row]
+(defn apply-row-defaults [row]
   (->> row
        (map (fn [[column opts]]
               [column
                (if (= :key column) ; not a column, but the row key
                  opts
                  (-> opts
-                     (default-if-missing :sort-value :value)
+                     (default-if-missing :sort-value (fn [opts]
+                                                       (let [val (:value opts)]
+                                                         (if (string? val)
+                                                           (str/lower-case val)
+                                                           val))))
                      (default-if-missing :display-value (comp str :value))
                      (default-if-missing :filter-value (fn [opts]
                                                          (if (string? (:display-value opts))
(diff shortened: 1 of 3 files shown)


src/cljs/rems/table.cljs
```diff
@@ -152,8 +152,8 @@
    (->> rows
         (sort-by #(get-in % [(:sort-column sorting) :sort-value])
```
opqdonut: [nit] you could put str/lower-case here in the keyfn, instead of below in the comparison

okahilak: Done.

src/cljs/rems/table.cljs
```diff
@@ -68,14 +67,18 @@
     m
     (assoc m k (make-default m))))
 
-(defn- apply-row-defaults [row]
+(defn apply-row-defaults [row]
```
okahilak: I had to change this function to public to test it outside the namespace -- otherwise `lein doo once` gives the warning `WARNING: var: rems.table/apply-row-defaults is not public at line 12 test/cljs/rems/test_table.cljs`.

Are there any better solutions?

okahilak:
The logic for lower-casing was in a slightly wrong place previously, now it fell inside a function that was already being tested. I moved the tests into their own testing namespace (maybe not necessary but I started doing it before realizing that I don't need another deftest but can use the existing ones) and added an explicit test for string and non-string values.

src/cljs/rems/table.cljs
opqdonut: The usual trick is to refer to privates like `@#'rems.table/apply-row-defaults`, but I'm not sure if that works in cljs.
Another option is to have the test in this file.
Making the function public is fine.

test/cljs/rems/test_table.cljs
```diff
@@ -0,0 +1,81 @@
...
+                                :foo {:td [:td.foo [:button "Button"]]}}))))
+  (testing ":sort-value is normalized to lowercase"
+    (is (= {:key 123
+            :foo {:value "FooBar"
+                  :sort-value "foobar"
+                  :display-value "FooBar"
+                  :filter-value ""
+                  :td [:td ""]}}
+           (apply-row-defaults {:key 123
+                                :foo {:value "FooBar"
```
opqdonut: But if you have `:sort-value "FooBar"`, it won't get lower-cased. Is that intentional?

okahilak: Nope, I'll fix..

okahilak: Hmm, even though that wasn't intentional, I think it could work that way. Currently, :filter-value works analogously, and it makes sense to use the same logic for both, i.e., if someone wants to filter (or sort) explicitly by a non-lower case string, it can be done, but the default is lower-cased.

I added test cases for that for both :filter-value and :sort-value, so that the behavior is more explicit and kind of documented in the test cases.

So, in short, I'd like to keep :filter-value and :sort-value analogous in this regard, and wouldn't want to change the existing behavior of :filter-value without strong enough reasons.

opqdonut: that sounds reasonable

Issue #1548, opened 2019-08-28
Clarify and unify API for updating and editing


okahilak:
- [x] Update (archive/disable) and edit (change content) are easily confused by name only. Improve naming, e.g., change "update" to "update-state" or similar.

- [x] Change API endpoint path for updating to be consistent with editing. Currently, e.g., form updating is /api/forms/update whereas editing is /api/forms/[form-id]/edit.

- [x] Workflow has currently one API endpoint for both updating state and editing content, whereas form has separate endpoints for both. Split workflow API endpoint to have two separate endpoints for increased consistency.

PR #1553, opened 2019-08-28
Split previous update workflow into archived, enabled, and edit

okahilak:
Split previous update workflow into archived, enabled, and edit
    
There used to be a single API endpoint for all these operations. Have separate API endpoints for each.

Part of #1548

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [x] all fields have a label
- [x] contrast is checked

## Follow-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ Changes since v2.4
 - API for creating catalogue item and its localizations has been changed.
   There is now a single API call that is used to create both a catalogue
   item and the localizations, namely, /api/catalogue-items/create.
+- API for editing workflow has been changed. The API endpoint for
+  editing name and handlers is now /api/workflows/edit, whereas the
+  endpoint for archiving or unarchiving workflow is /api/workflows/archived
+  and the endpoint for enabling or disabling workflow is
+  /api/workflows/enabled.
 
 ### Additions
 - New field types: description, option, multiselect
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -47,9 +47,19 @@
    :end (s/maybe DateTime)
    :mail s/Str})
 
+;; TODO: Remove once all admin APIs have separate endpoints for enabled
+;;       and archived.
 (s/defschema UpdateStateCommand
   {:id s/Int
    :enabled s/Bool
+   :archived s/Bool})
+
+(s/defschema EnabledCommand
+  {:id s/Int
+   :enabled s/Bool})
+
+(s/defschema ArchivedCommand
+  {:id s/Int
    :archived s/Bool})
 
 (s/defschema SuccessResponse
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Changes since v2.4
 - API for creating catalogue item and its localizations has been changed.
   There is now a single API call that is used to create both a catalogue
   item and the localizations, namely, /api/catalogue-items/create.
+- API for editing workflow content has been changed. The API endpoint for
+  editing is now /api/workflows/edit, whereas the API endpoint for changing
+  the state of a workflow remains /api/workflows/update.
```
opqdonut: Hmm,
a) if you're going to make a breaking change, why not rename /update to /state while you're at it
b) if you add new /edit and /state endpoints, you can retain the current /update for backwards compatibility (it'll just call edit-workflow and set-workflow-state)

src/clj/rems/api/workflows.clj
```diff
@@ -52,10 +50,17 @@
...
 
+    (PUT "/edit" []
+      :summary "Edit workflow"
+      :roles #{:owner}
+      :body [command EditWorkflowCommand]
+      :return SuccessResponse
+      (ok (workflow/edit-workflow! command)))
+
     (PUT "/update" []
       :summary "Update workflow"
```
opqdonut: These api docs need to be fixed. It was fine to have one endpoint that is described as "update workflow", but now that there is update and edit the difference needs to be explained. Consider something like "edit workflow title and settings" and "update workflow state (enabled, archived)"

okahilak: Done, the summaries are now more detailed.

CHANGELOG.md
okahilak: This new version now has /edit, /archived, and /enabled endpoints, as we seemed to have a reasonable consensus about that already, even though ADR is still in review.

src/clj/rems/api/schema.clj
```diff
@@ -47,11 +47,21 @@
...
+;; TODO: Remove once all admin APIs have separate endpoints for enabled
+;;       and archived.
 (s/defschema UpdateStateCommand
   {:id s/Int
    :enabled s/Bool
    :archived s/Bool})
 
+(s/defschema EnabledCommand
+  {:id s/Int
+   :enabled s/Bool})
```
okahilak: A comment about the design decision to have separate commands for enabling and archiving:

The UI components ('Enable'/'Disable' etc. buttons) currently rely on the item (workflow etc.) having :enabled and :archived keys. If the backend had one command with, e.g., :id and :value fields, the UI would need to map :enabled -> :value, and :archived -> :value, which backend or database query would then map back to :enabled and :archived before updating the db. Having two separate commands avoids that round trip, so the pipe between UI and the database stays cleaner.

PR #1566, opened 2019-08-30
Use flash messages on admin pages

### LINKED ISSUE #1469: Remove success pop-up windows

luontola:
Closes #1469

- [x] catalogue items
- [x] resources
- [x] forms
- [x] workflows
- [x] licenses

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Accessibility
- [x] all fields have a label
- [x] contrast is checked

## Follow-up

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -366,6 +366,7 @@
                      :color "#ccc"}
                     [:&:hover {:color (util/get-theme-attribute :color4)
                                :background-color "#eee"}]]]
+   [:.flash-message-title {:font-weight :bold}]
 
    [:.text-primary {:color (util/get-theme-attribute :text-primary)}]
    [:.text-secondary {:color (util/get-theme-attribute :text-secondary)}]
--- a/src/cljs/rems/administration/create_form.cljs
+++ b/src/cljs/rems/administration/create_form.cljs
@@ -10,7 +10,6 @@
             [rems.fields :as fields]
             [rems.flash-message :as flash-message]
             [rems.spinner :as spinner]
-            [rems.status-modal :as status-modal]
             [rems.text :refer [text text-format]]
             [rems.util :refer [dispatch! fetch put! post! normalize-option-key parse-int remove-empty-keys in-page-anchor-link]]))
 
--- a/src/cljs/rems/administration/create_form.cljs
+++ b/src/cljs/rems/administration/create_form.cljs
@@ -8,6 +8,7 @@
             [rems.atoms :as atoms :refer [document-title]]
             [rems.collapsible :as collapsible]
             [rems.fields :as fields]
+            [rems.flash-message :as flash-message]
             [rems.spinner :as spinner]
             [rems.status-modal :as status-modal]
             [rems.text :refer [text text-format]]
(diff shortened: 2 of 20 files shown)


okahilak:
I noticed that the active flash message is not updated when the translation is changed. But the problem is not introduced by this patch, so I created a new issue for that: https://github.com/CSCfi/rems/issues/1570

src/clj/rems/css/styles.clj
```diff
@@ -366,6 +366,7 @@
                      :color "#ccc"}
                     [:&:hover {:color (util/get-theme-attribute :color4)
                                :background-color "#eee"}]]]
+   [:.flash-message-title {:font-weight :bold}]
```
okahilak: I'm not sure how I feel about this change. The problem I have with it is that the boldface message is slightly distracting to me -- it puts too much emphasis on the message in a situation where I just want to note the message quickly and then move on to do other things on the page.

It also does not seem to be necessary for accessibility: the color contrast is high enough so that small text suffices. See screenshot below (similarly for the red text used for failure messages, not shown here):

image

It also introduces a small inconsistency with the warning box, which does not have boldface text:

image

Also, at the very least, the red failure message should have the boldface text, as well. This change only seems to affect the success message.

okahilak: Looking at this more closely, the problem that all flash messages do not have the boldface text seems to be caused by flash-message-title attribute only being set in show-default-success! and show-default-error! helper functions in rems.flash-messages, whereas some flash messages are created by calling show-success! and show-error! functions directly.

okahilak: After getting slightly more familiar with the patch, I now realize where the bold text comes from. In those cases where there is an actual error content in addition to the title, it is good to visually distinguish the title from the rest:

image

Without bolded text, it looks like this:

image

From these two, I prefer the former, with the bold text.

Overall, based on this, I'm fine with this change after all. Some messages might need fine-tuning afterwards regarding what is considered a title and what is content, but the idea is fine.

src/cljs/rems/administration/create_form.cljs
```diff
@@ -169,24 +169,25 @@
...
+         request (merge (build-request (db ::form) (db :languages))
+                        (when edit?
+                          {:form/id (db ::form-id)}))]
+     (when-not form-errors
+       (send-verb send-url
+                  {:params request
+                   :handler (flash-message/default-success-handler
+                             description
+                             (fn [response]
+                               (dispatch! (str "#/administration/forms/" (or (::form-id db)
```
okahilak: This seems to have changed back to how it had been previously before a recent fix: compare this line to L186-188 in the old version.

opqdonut: hmm maybe I botched the merge

opqdonut: fixed now

src/cljs/rems/administration/create_form.cljs
```diff
@@ -185,8 +185,10 @@
                    :handler (flash-message/default-success-handler
                              description
                              (fn [response]
-                               (dispatch! (str "#/administration/forms/" (or (::form-id db)
-                                                                             (:id response))))))
+                               (dispatch! (str "#/administration/forms/"
+                                               (if edit?
+                                                 (db ::form-id)
+                                                 (:id %))))))
```
okahilak: hmm, (:id response)?

okahilak: (oh, I think Esko changed the order of the arguments so that it would be (::form-id db) instead of (db ::form-id). )

opqdonut: fixed, tested and amended. I'll push a separate commit for the reorder.

okahilak: Thanks!

src/cljs/rems/status_modal.cljs
```diff
@@ -14,7 +14,7 @@
             [rems.text :refer [text]])
   (:require-macros [rems.guide-macros :refer [component-info example]]))
 
-(defn- format-errors [errors]
+(defn format-errors [errors]
```
okahilak: minor: This function is not used outside the namespace, so could keep it private.

src/cljs/rems/flash_message.cljs
```diff
@@ -40,7 +47,9 @@
 
     :component-will-unmount
     (fn [_this]
-      (rf/dispatch [::reset]))
+      (let [message @(rf/subscribe [::message])]
+        (when (expired? message)
```
okahilak: I tried to check when component-will-unmount is called, but couldn't easily think of a case in which this check is useful to have. Can someone clarify?

opqdonut: at least in dev mode when doing a live reload the components will be un- and remounted because there's a rerender

something similar might happen when an application is reloaded

okahilak: Hmm, ok. I guess it doesn't hurt to have the message expired and not re-shown in these kind of cases, so it's fine.

luontola: It happens when the page is changed as a response to a successful submit, so that the message will be shown on the next page since the user did not have time to see it on the original page. And when the user navigates manually to another page, the message will have expired and it won’t be shown anymore.

PR #1600, opened 2019-09-09
show "more info" links on application page

### LINKED ISSUE #1369: "More info" link to catalogue items
Similar to how LBR uses id links, but more general.

Could be edited in place.

opqdonut:
closes #1369

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -70,6 +70,9 @@
 (s/defschema LocalizedString
   {s/Keyword s/Str})
 
+(s/defschema OptionalLocalizedString
+  {s/Keyword (s/maybe s/Str)})
+
 (s/defschema LocalizedInt
   {s/Keyword s/Int})
 
@@ -78,6 +81,7 @@
    :resource/ext-id s/Str
    :catalogue-item/id s/Int
    :catalogue-item/title LocalizedString
+   :catalogue-item/infourl OptionalLocalizedString
    :catalogue-item/start DateTime
    :catalogue-item/end (s/maybe DateTime)
    :catalogue-item/enabled s/Bool
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -70,9 +70,6 @@
 (s/defschema LocalizedString
   {s/Keyword s/Str})
 
-(s/defschema OptionalLocalizedString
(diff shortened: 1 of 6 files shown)


src/cljs/rems/application.cljs
```diff
@@ -639,6 +640,19 @@
                                   actions)]
                       forms)}])))
 
+(defn- render-resource [resource]
+  ^{:key (:catalogue-item/id resource)}
+  [:div.application-resource
+   (localized (:catalogue-item/title resource))
+   ;; slight duplication with rems.catalogue/catalogue-item-more-info,
+   ;; but the data has a different schema here (V2Resource vs. CatalogueItem)
+   (when-let [url (or (localized (:catalogue-item/infourl resource))
```
okahilak: Hm, this is kind of subtle, but if I have a catalogue item with EN more info link and FI link defaulting to the urn, the localized function already does the fallback to the only available language (EN) and thus the urn link does not work when the language is set to FI.

opqdonut: good catch! I'll add a comment

src/clj/rems/api/schema.clj
```diff
@@ -78,6 +81,7 @@
    :resource/ext-id s/Str
    :catalogue-item/id s/Int
    :catalogue-item/title LocalizedString
+   :catalogue-item/infourl OptionalLocalizedString
```
okahilak: Is OptionalLocalizedString needed here? I checked that localization-for function works so that it only fills the map with keys that actually exist, so there should not be nil values there. Am I missing something?

opqdonut: Hmm okay maybe it only needs to be optional in the admin api. I'll check.

okahilak:
Hm, I'm not a fan of the capitalized letter in "More info" when listing the resources:

image

I think it's worse in the Finnish translation, maybe because of the tendency to capitalize less in Finnish:

image

Not sure what to do about it, though.

src/clj/rems/api/schema.clj
opqdonut: Yep, unneeded.

okahilak:
Hmm, how about using en dash there instead? I think it would make the capitalization slightly less conspicuous.

image

okahilak:
Or even em dash?

image

PR #1601, opened 2019-09-09
Show which actions an application is waiting for

### LINKED ISSUE #1596: Show which actions an application is waiting for
Make the Actions page more actionable by highlighting what is expected from the user

- [ ] remove the State column from the Actions page, it is always "Applied"
- [ ] instead, show a column with values like "New application", "Waiting for comment", "Waiting for your decision", "Waiting for a decision"
- [ ] searching the new column?

luontola:
Closes #1596

Actions_-_REMS

## Reviewability
- [x] link to issue

## API
- [X] Events are backwards compatible

## Documentation

## Testing

## Follow-up

### DIFF
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -149,6 +149,7 @@
                  :submitted "Lähetetty"
                  :todo "Toimia tarvitaan"
                  :todos {:new-application "Uusi hakemus"
+                         :no-pending-requests "Pyyntöihin on vastattu"
                          :resubmitted-application "Uudelleenlähetetty hakemus"
                          :unknown "Tuntematon"
                          :waiting-for-decision "Odottaa päätöstä"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -135,6 +135,7 @@
                  :id "Tunniste"
                  :intro [:div.intro [:p "Täyttämällä ja lähettämällä tämän hakemuksen kaikki hakijat voivat saada oikeuden haettuihin resursseihin. Kunkin hakijan on lisäksi hyväksyttävä käyttöehdot viimeistään hakemuksen käsittelyn jälkeen."]]
                  :last-activity "Muokattu"
+                 :latest-activity "Viimeisin toimenpide"
                  :my-applications "Omat hakemukset"
                  :resource "Resurssi"
                  :state "Tila"
@@ -146,7 +147,7 @@
                           :submitted "Haettu"
                           :unknown "Tuntematon"}
                  :submitted "Lähetetty"
-                 :latest-activity "Viimeisin toimenpide"
+                 :todo "Toimia tarvitaan"
                  :view "Näytä"}
   :cart {:add "Lisää koriin"
          :apply "Hae"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -151,10 +151,10 @@
                  :todos {:new-application "Uusi hakemus"
                          :resubmitted-application "Uudelleenlähetetty hakemus"
                          :unknown "Tuntematon"
-                         :waiting-for-comment "Odottaa kommenttia"
                          :waiting-for-decision "Odottaa päätöstä"
-                         :waiting-for-your-comment "Odottaa sinun kommenttiasi"
-                         :waiting-for-your-decision "Odottaa sinun päätöstäsi"}
+                         :waiting-for-review "Odottaa katselmointia"
+                         :waiting-for-your-decision "Odottaa sinun päätöstäsi"
+                         :waiting-for-your-review "Odottaa sinun katselmointiasi"}
                  :view "Näytä"}
   :cart {:add "Lisää koriin"
          :apply "Hae"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -149,7 +149,7 @@
                  :submitted "Lähetetty"
                  :todo "Toimia tarvitaan"
                  :todos {:new-application "Uusi hakemus"
-                         :no-pending-requests "Pyyntöihin on vastattu"
+                         :no-pending-requests "Kaikkiin pyyntöihin on vastattu"
                          :resubmitted-application "Uudelleenlähetetty hakemus"
                          :unknown "Tuntematon"
                          :waiting-for-decision "Odottaa päätöstä"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -148,6 +148,13 @@
                           :unknown "Tuntematon"}
                  :submitted "Lähetetty"
                  :todo "Toimia tarvitaan"
+                 :todos {:new-application "Uusi hakemus"
+                         :resubmitted-application "Uudelleenlähetetty hakemus"
+                         :unknown "Tuntematon"
+                         :waiting-for-comment "Odottaa kommenttia"
+                         :waiting-for-decision "Odottaa päätöstä"
+                         :waiting-for-your-comment "Odottaa sinun kommenttiasi"
+                         :waiting-for-your-decision "Odottaa sinun päätöstäsi"}
                  :view "Näytä"}
   :cart {:add "Lisää koriin"
          :apply "Hae"
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -264,9 +264,9 @@
       (update :application/first-submitted #(or % (:event/time event)))
       (dissoc ::draft-answers)
       (assoc :application/state :application.state/submitted)
-      (assoc :application/todo (if (nil? (:application/first-submitted application))
-                                 :new-application
-                                 :resubmitted-application))))
+      (assoc :application/todo (if (:application/first-submitted application)
+                                 :resubmitted-application
+                                 :new-application))))
 
 (defmethod event-type-specific-application-view :application.event/returned
   [application _event]
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -276,7 +276,7 @@
       (assoc :application/todo nil)))
 
 (defn- update-waiting-for-todo [application]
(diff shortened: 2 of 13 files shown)


src/cljs/rems/application_list.cljs
```diff
@@ -58,6 +58,12 @@
                           {:class (when (application-util/form-fields-editable? app)
                                     "text-highlight")}
                           value]})
+           :todo (let [value (localize-todo (:application/todo app))]
+                   {:value value
+                    :td [:td.state
+                         {:class (when (str/includes? (str (:application/todo app)) "-your-")
```
okahilak: Hmm, it would probably be cleaner if there was an explicit list of todo types that are highlighted.

src/cljs/rems/application_list.cljs
```diff
@@ -58,6 +58,12 @@
...
+                                   "text-highlight")}
```
okahilak: It might not be a problem of this patch (as a similar means for highlighting is used elsewhere), but I think the grey color alone doesn't work very well for highlighting the text. How about using bold in addition? See screenshots below.

image

and in a context:

image

src/cljs/rems/application_list.cljs
okahilak: Two possible semantics for highlight are: 1) highlight if the current user solely needs to act, 2) highlight if the current user or another user in a similar position needs to act.

Selecting 1) means that waiting-for-your-decision, waiting-for-your-review are highlighted (as it is currently implemented).

Selecting 2) would mean that new-application, resubmitted-application, waiting-for-your-decision, waiting-for-your-review are highlighted, as one of the handlers needs to act on an application, but if does not need to be any particular handler.

I think it would be safer to select 2), as the current "conservative" highlighting would allow every one of the handlers to dismiss a new or re-submitted application if they based their decision to act on highlighting only.

src/clj/rems/application/model.clj
```diff
@@ -265,49 +263,67 @@
...
-  [application event]
+  [application _event]
   (-> application
       (assoc ::draft-answers (::submitted-answers application)) ; guard against re-submit without saving a new draft
-      (assoc :application/state :application.state/returned)))
+      (assoc :application/state :application.state/returned)
+      (assoc :application/todo nil)))
+
+(defn- update-waiting-for-todo [application]
+  (cond-> (assoc application :application/todo nil)
```
okahilak: Hm, setting todo to nil here seems problematic, as "Action needed" is then rendered empty after another user has done commenting or deciding, even though some kind of action is still needed from the handler.

I'm not quite sure what to do here, though. I think the state of the application is formally close to "New application" -- it's just that the application is not new anymore. One problem here is that strictly speaking "New application" does not indicate an action, so just removing "New" from that does not make that much sense.

Maybe someone else has an idea of what to show to the user in these states.

(I'd also prefer setting an explicit value to todo instead of null value, but that depends a bit on how this turns out...)

okahilak: Another problem with the "empty action" is that it is difficult to determine what was the previous action that has happened (which affects the next action), e.g., has the application just been reviewed by another user and has returned to the handler, or has the application been long enough in the system that it does not count as a new application anymore, etc.

After thinking a bit more, I think one solution for this would be that "Action needed" is changed back to "State", which is, in some senses, a more general term -- it can consist of an action if the action is easy enough to describe, or just more general state in cases where the next action is not easily described or is ambiguous, which makes it more flexible.

After that change, the state could be, e.g., "Reviewed". In short, the state would describe, in decreasing order of priority:

1) The next action if it is unambiguous and easily described,
2) The previous action so that the user can infer what to do next,
3) The plain state of the application (e.g., "New application").

src/cljs/rems/application_list.cljs
luontola: Highlighting for the handler has been added.

src/clj/rems/application/model.clj
luontola: Added a "All requests have been responded to" message for those situations.

src/clj/rems/application/model.clj
```diff
@@ -265,49 +263,67 @@
       (assoc ::submitted-answers (::draft-answers application))
       (update :application/first-submitted #(or % (:event/time event)))
       (dissoc ::draft-answers)
-      (assoc :application/state :application.state/submitted)))
+      (assoc :application/state :application.state/submitted)
+      (assoc :application/todo (if (nil? (:application/first-submitted application))
```
okahilak: minor: nil? not needed here?

src/clj/rems/application/model.clj
```diff
@@ -265,49 +263,67 @@
...
-  [application event]
+  [application _event]
   (-> application
       (assoc ::draft-answers (::submitted-answers application)) ; guard against re-submit without saving a new draft
-      (assoc :application/state :application.state/returned)))
+      (assoc :application/state :application.state/returned)
+      (assoc :application/todo nil)))
+
+(defn- update-waiting-for-todo [application]
+  (cond-> (assoc application :application/todo :no-pending-requests)
```
okahilak: Thanks, I think this is a good idea.

src/cljc/rems/text.cljc
```diff
@@ -63,6 +63,20 @@
...
+  {:new-application :t.applications.todos/new-application
+   :no-pending-requests :t.applications.todos/no-pending-requests
+   :resubmitted-application :t.applications.todos/resubmitted-application
+   :waiting-for-decision :t.applications.todos/waiting-for-decision
+   :waiting-for-review :t.applications.todos/waiting-for-review
+   :waiting-for-your-decision :t.applications.todos/waiting-for-your-decision
+   :waiting-for-your-review :t.applications.todos/waiting-for-your-review})
+
+(defn localize-todo [todo]
+  (if (nil? todo)
```
okahilak: Hmm, is this check needed anymore? If for some reason a nil value ends up as far as here, the "unknown" text would probably suffice.

resources/translations/fi.edn
```diff
@@ -146,7 +147,15 @@
                           :submitted "Haettu"
                           :unknown "Tuntematon"}
                  :submitted "Lähetetty"
-                 :latest-activity "Viimeisin toimenpide"
+                 :todo "Toimia tarvitaan"
+                 :todos {:new-application "Uusi hakemus"
+                         :no-pending-requests "Pyyntöihin on vastattu"
```
okahilak: This could also be "Kaikkiin pyyntöihin on vastattu", it would resolve the ambiguity of "pyyntöihin" being in both definite and indefinite form.

okahilak:
I'm happy after these changes. Maybe someone else would like to have a look?

src/cljc/rems/text.cljc
luontola: It's nil before submitting and after approve/reject/close. Those states are not shown in the todo table, but they will be found by the application search in the other tables.

okahilak: Ok right. Is there any known case in which unknown todo text is showed? Maybe that could be replaced by an empty string to make the function more simple and get rid of the unneeded translations?

okahilak: Or actually, never mind, I think an explicit nil check is good to have here to make the logic clear.

luontola: The unknown text would be shown if the backend returns a keyword for which there is no translation, i.e. somebody forgot to update the frontend.

okahilak: Right, I think it's fine. Thanks.

resources/translations/fi.edn
```diff
@@ -146,7 +147,15 @@
                           :submitted "Haettu"
                           :unknown "Tuntematon"}
                  :submitted "Lähetetty"
-                 :latest-activity "Viimeisin toimenpide"
+                 :todo "Toimia tarvitaan"
+                 :todos {:new-application "Uusi hakemus"
+                         :no-pending-requests "Kaikkiin pyyntöihin on vastattu"
+                         :resubmitted-application "Uudelleenlähetetty hakemus"
+                         :unknown "Tuntematon"
```
opqdonut: "Tuntematon" sounds bad. Is it shown in any real case currently?

src/clj/rems/application/model.clj
```diff
@@ -265,49 +263,67 @@
...
+    (not (empty? (::latest-comment-request-by-user application)))
+    (assoc :application/todo :waiting-for-review)
+
+    (not (empty? (::latest-decision-request-by-user application)))
+    (assoc :application/todo :waiting-for-decision)))
```
opqdonut: AKA

```clojure
(assoc application :application/todo
       (cond (not (empty? (::latest-decision-request-by-user application)))
             :waiting-for-decision
             (not (empty? (::latest-comment-request-by-user application)))
             :waiting-for-review
             :else
             :no-pending-requests))

File: src/clj/rems/application/model.clj [hunk 10]
```diff
@@ -581,6 +606,7 @@
       (-> (if see-everything?
             application
             (hide-sensitive-information application))
+          (personalize user-id)
```
opqdonut: how about `personalize-todo`?

src/clj/rems/application/model.clj
```diff
@@ -265,49 +263,67 @@
...
 (defmethod event-type-specific-application-view :application.event/returned
-  [application event]
+  [application _event]
   (-> application
       (assoc ::draft-answers (::submitted-answers application)) ; guard against re-submit without saving a new draft
-      (assoc :application/state :application.state/returned)))
+      (assoc :application/state :application.state/returned)
+      (assoc :application/todo nil)))
+
+(defn- update-waiting-for-todo [application]
```
opqdonut: why not `update-todo`?

resources/translations/fi.edn
luontola: The same as with event types and states. If the backend returns some new value, it'll use that unknown.

PR #1602, opened 2019-09-09
Show accepted status separately for each license in application

### LINKED ISSUE #1591: When accepting licenses, show which licenses have already been accepted
When a handler adds a new license, the user is asked to reaccept all licenses and he might not spot the difference.

okahilak:
Closes #1591

## Reviewability
- [x] link to issue

## Documentation

## Accessibility
- [x] contrast is checked

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -1,5 +1,6 @@
 (ns rems.application
-  (:require [clojure.string :as str]
+  (:require [clojure.set :refer [union]]
+            [clojure.string :as str]
             [medley.core :refer [map-vals]]
             [re-frame.core :as rf]
             [rems.actions.accept-licenses :refer [accept-licenses-action-button]]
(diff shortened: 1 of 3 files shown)


okahilak:
Screenshots showing how "Terms of use" looks like at different stages:

A new application:

image

"Accept terms of use" pressed:

image

Handler views the application:

image

Handler adds terms of use:

image

Handler adds another member, the new member views the application:

image

Accepts terms of use:

image

The original applicant views the application:

image

Accepts terms of use:

image

The handler views the application:

image

okahilak:
I also experimented with sorting the licenses so that the accepted licenses are shown first in the list and only then the non-accepted licenses. The problem with that seemed to be that the order of the list changes when the remaining licenses are accepted, which was a bit confusing to me. Therefore, I ended up showing the licenses always in the same order regardless of the accepted status.

src/cljs/rems/application.cljs
```diff
@@ -346,6 +349,10 @@
 (defn- application-licenses [application edit-application userid]
   (when-let [licenses (not-empty (:application/licenses application))]
     (let [application-id (:application/id application)
+          show-accepted-licenses?
+          (or (= (:application/applicant application) userid)
+              (contains? (set (map :userid (:application/members application)))
+                         userid))
```
luontola: This could also be done, probably simpler, by checking whether `(:application/roles application)` contains `:applicant` or `:member`.

okahilak: Done and tested, seemed to work nicely..

PR #1608, opened 2019-09-10
Use readonly-fields in edit-catalogue-item for non-editable fields

### LINKED ISSUE #1598: When editing catalogue item, resource, workflow and form fields should not be editable
- don't use react-select (doesn't support disabled style) but render a regular readonly info-field or other similar static text instead

okahilak:
Closes #1598

## Reviewability
- [x] link to issue

## Accessibility
- [x] all fields have a label
- [x] contrast is checked


okahilak:
After this change the edit catalogue item page looks like this:

image

Issue #1580, opened 2019-09-04
misc. flash message improvements


opqdonut:
- don't use popup for failures. check at least:
  - [x] invite-member validation failure
  - [x] rems.actions.change-resources
  - [x] rems.actions.accept-licenses
  - [x] rems.administration.catalogue-items
  - [x] rems.administration.resources
- [ ] ~~fading / disappearing for flash message~~
- [x] mechanism for having multiple named locations for flash-messages on page. e.g. `:top`, `:application-members`, `[:attachment 13]`

opqdonut:
See also #1578

luontola:
A fading flash message may conflict with https://www.w3.org/WAI/WCAG21/Understanding/enough-time

PR #1619, opened 2019-09-12
Flash message improvements, remove modal

luontola:
#1580

## Reviewability
- [x] link to issue


src/clj/rems/application/commands.clj
```diff
@@ -268,7 +268,7 @@
 
 (defn already-member-error [application userid]
   (when (contains? (all-members application) userid)
-    {:errors [{:type :already-member :application-id (:id application)}]}))
+    {:errors [{:type :already-member :application-id (:application/id application)}]}))
```
okahilak: There seems to be a mismatch as it's :application/id here but the handler in ::accept-invitation accesses :application-id.

src/cljs/rems/actions/accept_invitation.cljs
```diff
@@ -1,52 +1,48 @@
...
+                        (let [error (first (:errors response))]
+                          (cond
+                            (:success response)
+                            (do
+                              (flash-message/show-success! :top (text :t.actions/accept-invitation-success))
+                              (dispatch! (str "#/application/" (:application-id response))))
+
+                            (= :already-member (:type error))
+                            (do
+                              (flash-message/show-success! :top (text :t.actions/accept-invitation-already-member))
```
okahilak: Not sure if this should be an error or a success. If it's success on purpose, it's fine, but if it is by accident, you may consider changing it to error.

src/clj/rems/application/commands.clj
luontola: I see only `command-handler :application.command/accept-invitation` accessing `(:application-id cmd)`, not the application model.

src/cljs/rems/actions/accept_invitation.cljs
luontola: It's on purpose.

PR #1633, opened 2019-09-16
Make comment textarea slightly smaller, do not autosize it

### LINKED ISSUE #1614: Parts of action bar do not always fit in the window
- [x] when submitting an invalid (non-complete) application, if the list of validation errors is long enough, the action bar scrolls out of the window (try THL form for example)

- [ ] when the handler adds terms of use, requests review, or requests decision, the dropdown menu, used for selecting the terms of use or the users to send the request to, is so low on the screen that the option list barely fits in the window with a full-sized browser window (with 1920x1080 resolution), most likely with a smaller resolution it does not fit properly.

okahilak:
Closes #1614 

## Reviewability
- [ ] link to issue

### DIFF
--- a/src/cljs/rems/actions/action.cljs
+++ b/src/cljs/rems/actions/action.cljs
@@ -32,6 +32,8 @@
     [:div.form-group
      [:label {:for id} label]
      [textarea {:id id
+                :min-rows 4
+                :max-rows 4
                 :name id
                 :placeholder (text :t.actions/comment-placeholder)
                 :value comment


okahilak:
Old:

image

New:

image

It could be even smaller (e.g., min-rows/max-rows 3), but I feel that when approving or rejecting an application, the text area is good to be slightly larger. The other actions, e.g., adding terms of use, probably would suffice with smaller, but then again they all use the same element, so going with the larger option.

PR #1635, opened 2019-09-17
OIDC first support

hukka:
Enables support for logging in using Auth0.

If we want to get completely rid of fake-shibboleth, we also need to modify test-data so that we can log in with users that have handler and owner rights. There are also some error conditions that shouldn't really happen, but technically might (like reusing authorization codes, or sending wrong API requests to Auth0, if they happen to change the API).


docs/architecture/006-authentication.md
```diff
@@ -0,0 +1,183 @@
...
+     Data can be transferred as unsecured, or in JWS or JWE payload
+     (note that OIDC allows only signed JWT, with or without encryption).
+
+OAuth2: RFC 6749; authorization protocol based on browser redirects and JSON, supersedes OAuth1
+OIDC: OpenID Connect; authentication protocol on top of OAuth2 and signed JWT, supersedes OpenId 2.0
+OIDC client: a system that requires authentication services — in this context REMS is a client, while the actual person is user and service doing the authentication is identity provider
+SAML: Security Assertion Markup Language; XML based protocol for authentication and authorization
+Shibboleth: Single Sign-On system based on SAML
+Elixir: EU level life sciences consortium
+Elixir AAI proxy: Autehntication and Authorisation Infrastructure; Elixir's authentication system that aggregates different ways for users to log in, and works as a SAML and OIDC identity provider for other systems, like REMS
```
luontola: I suppose each of these lines is meant to be its own paragraph. Now markdown clumps them all together in one paragraph.

luontola: Also those parts of the doc which are indented 4 spaces are rendered as a big code block (except for the first line which is clumped in the same paragraph with the previous line), which is probably not what was intended. Check the preview.

hukka:
Also this likely won't work as is with AAI proxy, but we have to see what kind of data is actually returned from there via OIDC.

resources/config-defaults.edn
```diff
@@ -14,13 +14,19 @@
...
  :ldap {:connection {:host "localhost:2636"
                      :ssl? true}
         :search-root "dc=Suivohtor,dc=local"}
  :alternative-login-url nil
 
+ ;; Open ID Connect settings
+ ; These allow login to http://localhost:3000/oidc-callback only
+ :oidc-domain "remsdev.eu.auth0.com"
+ :oidc-client-id "65TIN1Joo6EvsHSPCN4zyCaxxPh0WJF1"
+ :oidc-client-secret "e1yr9R_8WW3Xg03wm6wJ4UmNVVjiqdVhwFGvHhttyoRS3h04F4MGh0zRaV7vKm30"
```
luontola: These should be in `dev-config.edn` and not defaults. Otherwise it's possible to login to a production system which accidentally doesn't override the default values.

resources/translations/en.edn
```diff
@@ -300,6 +300,7 @@
   :link {:download-file "Download file"
          :opens-in-new-window "Opens in a new window"}
   :login {:alternative "Other login options"
+          :auth0-text "Login with Auth0"
```
luontola: I don't think the user should know anything about Auth0. From the user's point of view he logs in using his Google/Facebook/etc. account.

src/clj/rems/auth/oidc.clj
```diff
@@ -0,0 +1,58 @@
...
+        now (Instant/now)
+        oidc-data (jwt/validate jwt issuer audience now)]
+    ;; oidc-data has keys:
+    ; sub – unique ID
+    ; name - non-unique name
+    ; locale – could be used to set preferred lang on first login
+    ; email – non-unique (!) email
+    (assoc (redirect "/") ; TODO Could redirect with state param
+           :session (assoc (:session request)
+                           :identity
```
luontola: `assoc-in [:session :identity]`

resources/config-defaults.edn
hukka: Sure, can move them. But the IdP never allows callbacks to unconfigured domains, so there's a double check.

resources/translations/en.edn
hukka: Auth0 does provide direct credentials too, social media accounts are not mandatory. Also this UI is really meant for development and not end users, similarly how fake-shibboleth has a Haka login logo even though Haka is not used. Actual user would use the Elixir AAI once we have the necessary information.

If we do want to have a more polished Auth0 support for end users, what would be a sensible text for that? "Login with Google/Facebook/Github/Auth0"? Or we could use the Auth0 Universal Login widget (https://auth0.com/docs/libraries/when-to-use-lock for looks) or directly have buttons for Google, Facebook etc. (we would still always use Auth0, but each link would specify the actual auth provider).

src/clj/rems/auth/oidc.clj
hukka: I think
```
(assoc (redirect "/")
  :session
 (:session
   (assoc-in [:session :identity]
    request
 {data})))
```

is less clear. Did you have something else in mind?

hukka: Maybe destructure the request already in the function params?

`(defn oidc-callback [{session :session params :params}]`

luontola: (-> (redirect "/") ; TODO Could redirect with state param
        (assoc-in [:session :identity] {:eppn (:sub oidc-data)
                                        :commonName (:name oidc-data)
                                        :mail (:email oidc-data)}))))

or if there is some existing state in the session which needs to be kept:

    (-> (redirect "/") ; TODO Could redirect with state param
        (assoc :session (:session request))
        (assoc-in [:session :identity] {:eppn (:sub oidc-data)
                                        :commonName (:name oidc-data)
                                        :mail (:email oidc-data)}))))

resources/config-defaults.edn
```diff
@@ -21,12 +21,6 @@
         :search-root "dc=Suivohtor,dc=local"}
  :alternative-login-url nil
 
- ;; Open ID Connect settings
- ; These allow login to http://localhost:3000/oidc-callback only
- :oidc-domain "remsdev.eu.auth0.com"
- :oidc-client-id "65TIN1Joo6EvsHSPCN4zyCaxxPh0WJF1"
- :oidc-client-secret "e1yr9R_8WW3Xg03wm6wJ4UmNVVjiqdVhwFGvHhttyoRS3h04F4MGh0zRaV7vKm30"
-
```
luontola: The `config-defaults.edn` should have at least `nil` placeholders for all the configuration parameters. It's linked as a configuration reference in https://github.com/CSCfi/rems/blob/master/docs/configuration.md

luontola:
Logout doesn't seem to work.
```
2019-09-20 15:16:53,093 [qtp2093276055-26 - GET /logout] ERROR rems.middleware - Internal error 
java.lang.IllegalArgumentException: No matching clause: :oidc
	at rems.auth.auth$logout_url.invokeStatic(auth.clj:35)
	at rems.auth.auth$logout_url.invoke(auth.clj:34)
	at rems.auth.auth$auth_routes$fn__39260.invoke(auth.clj:42)

resources/translations/en.edn
luontola: "Login with OpenID" could be better, since it could be some other OpenID provider as well.

luontola: Or just one big Login button.

hukka:
> Logout doesn't seem to work.
> 
> ```
> 2019-09-20 15:16:53,093 [qtp2093276055-26 - GET /logout] ERROR rems.middleware - Internal error 
> java.lang.IllegalArgumentException: No matching clause: :oidc
> 	at rems.auth.auth$logout_url.invokeStatic(auth.clj:35)
> 	at rems.auth.auth$logout_url.invoke(auth.clj:34)
> 	at rems.auth.auth$auth_routes$fn__39260.invoke(auth.clj:42)
> ```

Ugh, good catch. Had so much trouble getting login to work that forgot about that.

PR #1641, opened 2019-09-18
validate extra-translations

### LINKED ISSUE #1639: Validate extra-translations content
- [ ] check that translation keys exist
- [ ] check that %-parameters match
- [ ] emit warnings to log if problems found

opqdonut:
- warn if unused keys are overridden in extra translations
- check that format parameters match between base and extra translations

fixes #1639 

## Reviewability
- [x] link to issue

## Testing

### DIFF
--- a/src/clj/rems/locales.clj
+++ b/src/clj/rems/locales.clj
@@ -9,15 +9,23 @@
             [rems.config :refer [env]])
   (:import (java.io FileNotFoundException)))
 
+(defn- file-or-resource
+  "Given a path (a String or File), resolve it to either an existing file or existing resource.
+  Returns a File or URL object, or nil if file nor resource can be found.
+  Prefers files over resources."
+  [path]
+  (let [file (io/file path)
+        resource (io/resource (str file))]
+    (cond
+      (and file (.exists file)) file
+      resource resource
+      :else nil)))
+
 (defn- translations-from-file [file]
-  (let [file (io/file file)
-        resource (io/resource (str file))
-        chosen (cond
-                 (and file (.exists file)) file
-                 resource resource
-                 :else (throw (FileNotFoundException.
-                               (str "translations could not be found in file or resource \"" file "\""))))]
-    (read-string (slurp chosen))))
+  (if-let [chosen (file-or-resource file)]
+    (read-string (slurp chosen))
+    (throw (FileNotFoundException.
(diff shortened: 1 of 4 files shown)


opqdonut:
example warning messages:
```
2019-09-18 13:03:26,010 [nREPL-worker-1 - - -] WARN  rems.locales - Unused translation keys defined in example-theme/extra-translations/ : ((:t :unused-key))
2019-09-18 13:03:26,016 [nREPL-worker-1 - - -] WARN  rems.locales - Nonmatching format parameters in example-theme/extra-translations/ : ({:key (:t :create-license :license-text), :translations #{}, :extra-translations #{"%1"}} {:key (:t :unused-key), :translations nil, :extra-translations #{}})

**Comment by opqdonut** [comment]
hmm I'll refactor a bit

File: src/clj/rems/locales.clj [hunk 1]
```diff
@@ -26,17 +28,42 @@
         translations-path (str path "/extra-translations/")]
     translations-path))
 
+(defn extract-format-parameters [string]
+  (when (string? string)
+    (set (re-seq #"%\d+" string))))
```
luontola: If this would always return a set instead of nil, it would avoid the unnecessary warning about unused keys.

opqdonut: Good point

src/clj/rems/locales.clj
```diff
@@ -1,42 +1,77 @@
...
+            (str "translations could not be found in file or resource \"" file "\"")))))
 
 (defn- extra-translations-path [theme-path]
-  (let [path (str/join "/" (butlast (str/split theme-path #"/"))) ;;Theme-path is of form /foo/bar/theme.edn
-        translations-path (str path "/extra-translations/")]
-    translations-path))
+  (-> theme-path
+      (io/file)
+      (.getParentFile)
+      (io/file "extra-translations")))
```
luontola: I think the following would communicate the meaning better, because the thing being passed through `->` changes type (theme file vs. parent directory vs. extra translations directory). See https://tonsky.me/blog/readable-clojure/#dont-spare-names 
```clj
(let [parent-dir (.getParentFile (io/file theme-path))]
  (io/file parent-dir "extra-translations"))

opqdonut: fair enough

Issue #1323, opened 2019-06-06
Language selection does not persist in login


Macroz:
When I change the language of REMS before logging in, I expect to see the changed language also after logging in.

PR #1647, opened 2019-09-19
Language cookie

### LINKED ISSUE #1570: Active flash message is not updated when language is changed
If there is an active flash message shown on the page and the language is changed, the other parts of the page are updated with the new language but the old text in the flash message remains. See screenshot below:

image

Macroz:
Fixes #1570 and #1323.

When changing language, we hide the flash message because updating the content is difficult as it's stored in app db elsewhere and we have no possibility to re-render it.

We also save the chosen language into a cookie so the UI would often start already in the correct language. Also it persists after login and possibly overrides and changes the user's preferred language.

## Reviewability
- [x] link to issue
- ~~[ ] note if PR is on top of other PR~~
- ~~[ ] note if related change in rems-deploy repo~~
- ~~[ ] consider adding screenshots for ease of review~~

## Documentation
- ~~[ ] add or update docstrings for namespaces and functions~~
- ~~[ ] components are added to guide page~~
- ~~[ ] documentation _at least_ for config options (i.e. docs folder)~~
- ~~[ ] ADR for major architectural decisions or experiments~~

## Different installations
- ~~[ ] new configuration options added to rems-deploy repository~~
- ~~[ ] instance specific translations (i.e. LBR kielivara)~~

## Testing
- ~~[ ] complex logic is unit tested~~

## Accessibility
- ~~[ ] all icons have the aria-label attribute~~
- ~~[ ] all fields have a label~~
- ~~[ ] errors are linked to fields with aria-describedby~~
- ~~[ ] contrast is checked~~
- ~~[ ] conscious decision about where to move focus after an action~~

## Follow-up
- ~~[ ] new tasks are created for pending or remaining tasks~~


CHANGELOG.md
```diff
@@ -74,10 +74,12 @@ Changes since v2.4
 - Email messages now use the application title and full names of users
 - Email message texts improved
 - Show three latest events as a default on the application page (#1632)
+- User's language setting is stored in a cookie so that it persists after login. Also the language is possibly correct already at login because the cookie is remembered.
```
opqdonut: I don't understand the second sentence.

src/cljs/rems/language.cljs
```diff
@@ -2,11 +2,22 @@
   (:require [clojure.string :as str]
             [re-frame.core :as rf :refer [reg-event-fx reg-sub]]))
 
+(def ^:private language-cookie-name "rems-user-preferred-language")
+
+(defn get-language-cookie []
+  (when-let [value (.. js/document -cookie)]
+    (keyword (second (str/split value #"=")))))
```
opqdonut: there's no cookie parser available?

src/cljs/rems/language.cljs
```diff
@@ -2,11 +2,22 @@
...
+
+(defn- set-language-cookie! [language]
+  (let [year-from-now (.setFullYear (js/Date.) (inc (.getFullYear (js/Date.))))]
+    (set! (.. js/document -cookie)
+          (str language-cookie-name "=" (name language) "; expires=" (.toString year-from-now) "; path=/"))))
```
opqdonut: there's no cookie generator available?

src/cljs/rems/language.cljs
```diff
@@ -2,11 +2,22 @@
...
             [re-frame.core :as rf :refer [reg-event-fx reg-sub]]))
 
+(def ^:private language-cookie-name "rems-user-preferred-language")
+
+(defn get-language-cookie []
+  (when-let [value (.. js/document -cookie)]
+    (keyword (second (str/split value #"=")))))
+
+(defn- set-language-cookie! [language]
+  (let [year-from-now (.setFullYear (js/Date.) (inc (.getFullYear (js/Date.))))]
```
opqdonut: cljs-time?

src/cljs/rems/user_settings.cljs
```diff
@@ -7,8 +7,16 @@
 (rf/reg-event-fx
  :loaded-user-settings
  (fn [{:keys [db]} [_ user-settings]]
-   (language/update-language (:language user-settings))
-   {:db (assoc db :user-settings user-settings)}))
+   (let [new-language (or (language/get-language-cookie)
+                          (:language user-settings))
```
opqdonut: hmm so we have the overriding here and in the `:language` subscription... could we get rid of the other?

src/cljs/rems/user_settings.cljs
```diff
@@ -17,9 +25,11 @@
 
 (rf/reg-event-fx
  ::update-user-settings
```
opqdonut: would it be clearer if this was called something like `store-user-settings`?

src/cljs/rems/language.cljs
```diff
@@ -31,13 +42,12 @@
...
  ::set-language
- (fn [{:keys [db]} [_ language]]
-   (let [user-id (get-in db [:identity :user :eppn])]
-     (if user-id
-       (rf/dispatch [:rems.user-settings/update-user-settings {:language language}])
-       (do (update-language language)
-           {:db (assoc db :non-user-language language)})))))
+ (fn [_ [_ language]]
+   (update-language language)
+   {:dispatch [:rems.user-settings/update-user-settings {:language language}]}))
```
opqdonut: if we merged with existing settings here, we wouldn't need to merge in update-user-settings

src/cljs/rems/user_settings.cljs
```diff
@@ -17,9 +25,11 @@
...
-          :error-handler (flash-message/default-error-handler :top "Update user settings")})
-   nil))
+ (fn [{:keys [db]} [_ settings]]
+   (let [user-id (get-in db [:identity :user :eppn])
+         new-settings (merge (:user-settings db) settings)]
+     (when user-id
+       (put! "/api/user-settings"
+             {:params new-settings
+              :error-handler (flash-message/default-error-handler :top "Update user settings")}))
+     {:db (assoc db :user-settings new-settings)})))
```
opqdonut: hmm, or should we have separate events for a) updating `:user-settings` in db b) sending the current user settings from the db to the server

src/cljs/rems/language.cljs
luontola: Also, doesn't this assume that there is only one cookie?

src/cljs/rems/language.cljs
luontola: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie says that the expiry time should be in GMT, so need to use toUTCString() instead of toString(). But either way, this code should not be written manually.

src/cljs/rems/language.cljs
luontola: Quick search for clojurescript cookie returns at least two libraries which both delegate to [goog.net.Cookies](https://google.github.io/closure-library/api/goog.net.Cookies.html).
https://github.com/reagent-project/reagent-utils/blob/master/src/reagent/cookies.cljs
https://github.com/Quantisan/cljs-cookies/blob/master/src/cljs_cookies/core.cljs

src/cljs/rems/user_settings.cljs
```diff
@@ -7,8 +7,16 @@
...
-   {:db (assoc db :user-settings user-settings)}))
+   (let [new-language (or (language/get-language-cookie)
+                          (:language user-settings))
+         new-settings (assoc user-settings :language new-language)]
+     (language/update-language new-language)
+     ;; if the user has changed language before login
+     ;; the cookie will not match the language setting
+     ;; and we should save the new language setting
+     (when-not (= (language/get-language-cookie) (:language user-settings))
+       (rf/dispatch [::update-user-settings new-settings]))
```
luontola: Isn't this called also when the user has no cookie?

test/clj/rems/test_browser.clj
```diff
@@ -236,3 +236,36 @@
...
+      (go *driver* +test-url+)
+      (wait-visible *driver* {:tag :h1 :fn/text "Welcome to REMS"})
+      (login-as "alice")
+      (wait-visible *driver* {:tag :h1, :fn/text "Catalogue"})
+      (wait-page-loaded))
+
+    (testing "changing language while logged out"
+      (click *driver* :logout)
+      (wait-visible *driver* {:tag :h1 :fn/text "Welcome to REMS"})
+      (click *driver* [{:fn/has-class :btn-link :fn/text "FI"}])
```
luontola: This should look for the button under `.language-switcher` to be explicit that it's a link for changing the language.

CHANGELOG.md
Macroz: I reworded these. Essentially the second sentence is that the language is possibly correct already at login because the language setting stored in the cookie is remembered :)

src/cljs/rems/language.cljs
Macroz: I prefer not bringing in a new dependency for this simple usage.

src/cljs/rems/language.cljs
Macroz: I refactored into UTC string though it's unlikely to be different. There is no need to complicate this feature with encoding here.

src/cljs/rems/language.cljs
Macroz: It would be many more characters because of the requires and formatting. This is as simple as it gets?

src/cljs/rems/language.cljs
Macroz: It is likely that the settings will be a complicated map, so it's more useful to have the caller do less work than to remember to set everything? There used to be no merge which would eventually lead to bugs.

Macroz: Maybe even a deep-merge here? I agree a plain setter is "simpler" in some sense.

src/cljs/rems/language.cljs
opqdonut: I would've preferred the `(plus (now) (years 1))` of cljs-time, but this is fine too.

src/cljs/rems/language.cljs
opqdonut: What about the one cookie assumption? We also have the session cookie.

src/cljs/rems/language.cljs
opqdonut: I think I would be fine with having deep-merge in `update-user-settings`, but splitting the PUT to a different event. See my comment on `update-user-settings`.

test/clj/rems/test_browser.clj
Macroz: Using `.language-switcher` is indeed good. It could be YAGNI for the extraction.

Macroz: I extracted `change-language` and `logout` just for slight improvement in readability.

src/cljs/rems/user_settings.cljs
Macroz: Yes it is. Now it's slightly different.

src/cljs/rems/language.cljs
Macroz: Btw. I fixed a problem from here but it still stands that cljs-time is not simpler in this.

src/cljs/rems/language.cljs
Macroz: If someone requires complicated cookie usage later, then they can do the refactoring and use an external dependency (or goog library directly)

hukka: session cookie is http only, so JS should only see this one.

luontola: `document.cookie` returns a `;`-separated list of all cookies, which this code does not parse. If somebody adds e.g. Google Analytics to their installation, then this code will break.

Issue #1643, opened 2019-09-18
Fix application print styling


opqdonut:
"Print page" should result in a nice archive-quality version of an application

- [x] firefox truncates the content to 1 page
- [x] buttons should be hidden
- [x] all expandable sections should be expanded
- [x] nicer page breaks (currently breaks inside terms of use)

opqdonut:
Let's call this done, but like I said in the PR, we might need to consider a different approach in the future if there are more requirements.

PR #1651, opened 2019-09-20
Print styles for application page

opqdonut:
With these styles, the application page prints nicely in firefox & chrome. Since the changes are small, I decided to go with the existing page and CSS. An alternative would have been a fresh print.css, or a separate printable application page with less styling and no interactions.

The main caveat is that large (>1 page) textareas don't get printed nicely. Luckily we don't use textarea for readonly applicaitons, so this problem affects only drafts.

For #1643.

## Reviewability
- [x] link to issue
- ~note if PR is on top of other PR~
- ~note if related change in rems-deploy repo~

## API
- ~API is documented and shows up in Swagger UI~
- ~API is backwards compatible or completely new~
- ~Events are backwards compatible~

## Documentation
- ~add or update docstrings for namespaces and functions~
- ~components are added to guide page~
- ~documentation _at least_ for config options (i.e. docs folder)~
- ~ADR for major architectural decisions or experiments~

## Different installations
- ~new configuration options added to rems-deploy repository~
- ~instance specific translations (i.e. LBR kielivara)~

## Testing
- ~complex logic is unit tested~

## Accessibility
- ~all icons have the aria-label attribute~
- ~all fields have a label~
- ~errors are linked to fields with aria-describedby~
- ~contrast is checked~
- ~conscious decision about where to move focus after an action~

## Follow-up
- ~new tasks are created for pending or remaining tasks~
- ~no critical TODOs left to implement~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@ Changes since v2.4
 ### Fixes
 - Entitlement API
 - Search on the catalogue and admin pages did not support multiple search terms (#1541)
+- Printing application pages now works (except for drafts) (#1643)
 
 ## v2.5 "Maarintie" 2019-07-18
 
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -768,8 +768,16 @@
     [:body {:display :block}]
 
     ;; hide some unnecessary elements
-    [:.fixed-top {:display :none}])
+    ;; TODO: consider a hide-print class?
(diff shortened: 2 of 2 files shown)


opqdonut:
This is what print to pdf produces in firefox after the changes: [example-application-firefox.pdf](https://github.com/CSCfi/rems/files/3635353/example-application-firefox.pdf)

Ignore the re-frisk button in the lower right, it won't be present in real installations. (And I can't hide it easily because the elements don't have an id or class)

PR #1665, opened 2019-09-25
Fix user settings randomly not being loaded

### LINKED ISSUE #1664: User settings are randomly not loaded when the page is refreshed
Given the user is logged in and has selected FI as the language, when reload the page, then sometimes randomly the /api/user-settings request does not happen and the UI is shown in the default language. There are no errors in JS console.

Good page reload:
good reload

Bad page reload:
bad reload

luontola:
Fixes #1664

- Fixes a race condition between loading user settings and finding out whether the user is logged in
- If user is not logged in, return defaults as the user settings

## Reviewability
- [x] link to issue

## API


opqdonut:
I wonder, will this conflict with #1647?

PR #1674, opened 2019-10-01
move add-licenses button to licenses box

### LINKED ISSUE #1668: Move "add terms of use" button to "terms of use" box

opqdonut:
fixes #1668

## Reviewability
- [x] link to issue

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -362,6 +362,10 @@
                          :accepted (contains? accepted-licenses (:license/id license))
                          :readonly readonly?)
                   show-accepted-licenses?]))
+         (when (contains? permissions :application.command/add-licenses)
+           [:<>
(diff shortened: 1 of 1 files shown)


opqdonut:
I think this is minor enough not to warrant a changelog entry

src/cljs/rems/application.cljs
```diff
@@ -362,6 +362,10 @@
                          :accepted (contains? accepted-licenses (:license/id license))
                          :readonly readonly?)
                   show-accepted-licenses?]))
+         (when (contains? permissions :application.command/add-licenses)
```
Macroz: This assumes that the add-licences and accept-licenses functionality are never enabled for the same user, which is fine but it's a silent assumption.

opqdonut: Good point. This is what it looks like when both are enabled:
add-and-accept

PR #1678, opened 2019-10-01
Improve error messages

### LINKED ISSUE #1637: No error message if loading the application fails
For example visiting http://localhost:3000/#/application/999999 just shows an indefinitely spinning spinner.

luontola:
Closes #1637

- show the "Not found" page if the application doesn't exist (or any other ajax request returns 404)
- resolve flash message text at render time instead of the time when the message was created, to support changing the language or translations which are not yet loaded
- if loading translations fails, show the translation key as a fallback:
no translations
- wait for the translations, themes and config to be loaded before the first render, to avoid a quick flash of an incomplete page

## Reviewability
- [x] link to issue

### DIFF
--- a/src/cljs/rems/spa.cljs
+++ b/src/cljs/rems/spa.cljs
@@ -113,6 +113,12 @@
    {:dispatch [:set-active-page :forbidden]}))
 
 (rf/reg-event-fx
+ :not-found!
+ (fn [_ [_ current-url]]
+   (println "Received not-found from" current-url)
+   {:dispatch [:set-active-page :not-found]}))
+
+(rf/reg-event-fx
  :landing-page-redirect!
  (fn [{:keys [db]}]
    ;; do we have the roles set by set-identity already?
--- a/src/cljs/rems/spa.cljs
+++ b/src/cljs/rems/spa.cljs
@@ -441,12 +441,12 @@
 
 (defn fetch-translations! []
   (fetch "/api/translations"
-         {:handler #(rf/dispatch [:loaded-translations %])
+         {:handler #(rf/dispatch-sync [:loaded-translations %])
           :error-handler (flash-message/default-error-handler :top "Fetch translations")}))
 
 (defn fetch-theme! []
   (fetch "/api/theme"
-         {:handler #(rf/dispatch [:loaded-theme %])
+         {:handler #(rf/dispatch-sync [:loaded-theme %])
           :error-handler (flash-message/default-error-handler :top "Fetch theme")}))
 
 (defn mount-components []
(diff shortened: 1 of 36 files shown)


src/cljs/rems/util.cljs
```diff
@@ -75,9 +86,14 @@
...
-                  opts
-                  {:error-handler (wrap-default-error-handler (:error-handler opts))})))
+  ;; TODO: change also put! and post! to return a promise?
+  (p/create
+   (fn [resolve reject]
+     (GET url (-> (merge {:response-format :transit}
+                         opts
+                         {:error-handler (wrap-default-error-handler (:error-handler opts))})
+                  (update :handler append-handler resolve)
+                  (update :error-handler append-handler reject))))))
```
opqdonut: please document that fetch returns a promise (but that you can ignore the return value if you don't need it)

src/cljs/rems/spa.cljs
```diff
@@ -455,8 +456,9 @@
...
-  (fetch-theme!)
-  (config/fetch-config!)
-  (hook-browser-navigation!)
-  (mount-components))
+  (-> (p/all [(fetch-translations!)
+              (fetch-theme!)
+              (config/fetch-config!)])
+      (p/finally (fn []
+                   (hook-browser-navigation!)
+                   (mount-components)))))
```
Macroz: Doesn't this prevent any useful loading spinner from showing up?

How about not fetching any of these here after page load but instead already injecting them into the base page on the server side?

src/cljs/rems/util.cljs
```diff
@@ -67,17 +70,32 @@
...
+
 (defn fetch
   "Fetches data from the given url with optional map of options like #'ajax.core/GET.
 
   Has sensible defaults with error handler, JSON and keywords.
 
-  Additionally calls event hooks."
+  Additionally calls event hooks.
+
+  Returns a promise, but it's okay to ignore it if you prefer using the event hooks."
```
opqdonut: This is misleading, hooks refers to the window.rems.hooks below. You probably meant

> it's okay to ignore it if you prefer using `:handler` and/or `:error-handler`

luontola: Oh, that's what the previous comment about hooks meant. 🤣

src/cljs/rems/spa.cljs
luontola: The spinner will show up when it's loading the page content, which is what typically takes the longest time, for example in the catalogue page.

luontola: It's an option to switch to loading everything with ajax or injecting everything to the html. Currently it's a mixture of both. I created an idea ticket for it.

luontola: The slowest part of the initial page load is downloading and parsing the JavaScript, during which no JavaScript-generated spinner can anyways be shown. It would need to be in the static HTML.

PR #1681, opened 2019-10-02
rems.table parameter schemas and documentation

luontola:


src/cljs/rems/table.cljs
```diff
@@ -95,7 +169,10 @@
  (fn [[_ table] _]
    [(rf/subscribe (:rows table))])
  (fn [[rows] _]
-   (map apply-row-defaults rows)))
+   (->> rows
+        (s/validate Rows)
+        (map apply-row-defaults)
+        (s/validate Rows))))
```
opqdonut: nice double validation

luontola: The second validation could be done so that the optional keys are no more optional. Is there a simple way to do it?

PR #1683, opened 2019-10-03
Allow handler to close applications

luontola:
Closes #1669

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue

## Documentation
- [x] update changelog if necessary
- [x] documentation _at least_ for config options (i.e. docs folder)

PR #1684, opened 2019-10-03
Use GitHub Pages for /docs

luontola:
PR #1683 adds a .html file to the /docs folder, but GitHub only shows it with the text/plain content-type, so viewing it would require checking out the project. GitHub Pages is needed to support pure HTML in the docs. This PR updates the /docs folder to work with GitHub Pages/Jekyll.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,0 +1,5 @@
+defaults:
+  - scope:
+      path: ""
+    values:
+      layout: "default"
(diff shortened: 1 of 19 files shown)


luontola:
Tables with multiple rows in markdown:

| | | | |
|---|---|---|---|
| foo<br>bar<br>baz | foo  | **foo**<br>*bar*  |   |

PR #1650, opened 2019-09-20
Change event descriptions on application page

okahilak:
Related to #1634 

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue

## Documentation
- [x] update changelog if necessary

Issue #1634, opened 2019-09-17
Improvements to event list in application view


okahilak:
- [x] List events using applicant's real name instead of user id
- [x] Use active verb forms instead of passive forms
- [x] Be more consistent in the messages describing the events (e.g., use whole sentences, end with a full stop)

PR #1696, opened 2019-10-10
Don't leak internal user attributes in API

opqdonut:
Follow-up for #1634

This is a breaking change for the API, but it mostly breaks stuff that was just added by #1650. The other stuff that breaks (e.g. `/api/applications/:id` field `:application/applicant-attributes` was never documented.

## Reviewability
- [x] link to issue

## API

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -114,8 +114,8 @@
 
 (s/defschema UserWithAttributes
   {:userid UserId
-   :name s/Str
-   :email s/Str})
+   :name (s/maybe s/Str)
+   :email (s/maybe s/Str)})
 
 (s/defschema Workflow
   {:id s/Int
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -195,8 +195,7 @@
    (s/optional-key :application/copied-to) [{:application/id s/Int
                                              :application/external-id s/Str}]
    :application/last-activity DateTime
-   :application/applicant s/Str
-   :application/applicant-attributes UserWithAttributes
+   :application/applicant UserWithAttributes
    :application/members #{UserWithAttributes}
    :application/invited-members #{{:name s/Str
                                    :email s/Str}}
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -110,6 +110,13 @@
 
 (def UserId s/Str)
 
+(s/defschema User {:userid UserId})
+
+(s/defschema UserWithAttributes
+  {:userid UserId
+   :name s/Str
+   :email s/Str})
+
 (s/defschema Workflow
   {:id s/Int
    :organization s/Str
@@ -189,9 +196,8 @@
                                              :application/external-id s/Str}]
    :application/last-activity DateTime
    :application/applicant s/Str
-   :application/applicant-attributes {s/Keyword s/Str}
-   :application/members #{{:userid s/Str
-                           s/Keyword s/Str}}
+   :application/applicant-attributes UserWithAttributes
+   :application/members #{UserWithAttributes}
    :application/invited-members #{{:name s/Str
                                    :email s/Str}}
    :application/resources [V2Resource]
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -589,7 +589,8 @@
         (update :application/resources enrich-resources get-catalogue-item)
         (update :application/licenses enrich-licenses get-license)
         (update :application/events enrich-events get-user get-catalogue-item)
-        (assoc :application/applicant-attributes (get-user (:application/applicant application)))
+        ;; TODO application/applicant should be {:userid "x"} originally
+        (assoc :application/applicant (get-user (:application/applicant application)))
         (assoc :application/attachments (get-attachments-for-application (getx application :application/id)))
         (enrich-user-attributes get-user)
         (enrich-workflow-handlers get-workflow)
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -200,7 +200,7 @@
              :application/todo nil
              :application/created (:event/time event)
              :application/modified (:event/time event)
-             :application/applicant (:event/actor event)
+             :application/applicant {:userid (:event/actor event)}
              :application/members #{}
              :application/past-members #{}
              :application/invitation-tokens {}
@@ -589,8 +589,7 @@
         (update :application/resources enrich-resources get-catalogue-item)
         (update :application/licenses enrich-licenses get-license)
         (update :application/events enrich-events get-user get-catalogue-item)
-        ;; TODO application/applicant should be {:userid "x"} originally
-        (assoc :application/applicant (get-user (:application/applicant application)))
+        (assoc :application/applicant (get-user (get-in application [:application/applicant :userid])))
         (assoc :application/attachments (get-attachments-for-application (getx application :application/id)))
         (enrich-user-attributes get-user)
         (enrich-workflow-handlers get-workflow)
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -55,7 +55,7 @@
     :items [catalogue-item-id]}))
 
 (defn- valid-user? [userid]
-  (not (nil? (users/get-user-attributes userid))))
+  (not (nil? (users/get-user userid))))
 
 (def ^:private command-injections
   {:valid-user? valid-user?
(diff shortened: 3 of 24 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -189,9 +196,8 @@
                                              :application/external-id s/Str}]
    :application/last-activity DateTime
    :application/applicant s/Str
-   :application/applicant-attributes {s/Keyword s/Str}
-   :application/members #{{:userid s/Str
-                           s/Keyword s/Str}}
+   :application/applicant-attributes UserWithAttributes
```
luontola: We could replace these two with just `:application/applicant UserWithAttributes`

src/clj/rems/db/applications.clj
```diff
@@ -91,7 +91,7 @@
    :get-form-template form/get-form-template
    :get-catalogue-item catalogue/get-localized-catalogue-item
    :get-license licenses/get-license
-   :get-user users/get-user-attributes
+   :get-user users/get-user
```
luontola: Should `get-user-attributes` be removed or made private, if we don't want to use it?

src/cljc/rems/application_util.cljc
```diff
@@ -11,7 +11,7 @@
   (contains? (:application/permissions application)
              :application.command/save-draft))
 
-(def ^:private name-attribute-priority [:name :commonName :displayName :eppn])
+(def ^:private name-attribute-priority [:name :commonName :displayName :userid :eppn])
```
luontola: Could this be removed, if `:name` is anyways required?

src/clj/rems/db/applications.clj
opqdonut: Done!

src/cljc/rems/application_util.cljc
opqdonut: done

src/clj/rems/application/model.clj
```diff
@@ -589,7 +589,8 @@
         (update :application/resources enrich-resources get-catalogue-item)
         (update :application/licenses enrich-licenses get-license)
         (update :application/events enrich-events get-user get-catalogue-item)
-        (assoc :application/applicant-attributes (get-user (:application/applicant application)))
+        ;; TODO application/applicant should be {:userid "x"} originally
```
opqdonut: I'll fix this

src/clj/rems/poller/email.clj
```diff
@@ -39,11 +39,12 @@
    (when-not (empty? (:application/description application))
      (str ", \"" (:application/description application) "\""))))
 
+;; TODO user-for-email shouldn't need to call get-user since we add
+;; all this info to the application in rems.application.model/enrich-user-attributes
```
opqdonut: this is something for a follow-up PR

src/clj/rems/poller/email.clj
```diff
@@ -53,9 +54,10 @@
 ;; There's a slight inconsistency here: we look at current members, so
 ;; a member might get an email for an event that happens before he was
 ;; added.
+;; TODO this function occurs in so many places
```
opqdonut: so is this

src/clj/rems/db/applications.clj
```diff
@@ -55,7 +55,7 @@
     :items [catalogue-item-id]}))
 
 (defn- valid-user? [userid]
-  (not (nil? (users/get-user-attributes userid))))
+  (not (nil? (users/get-user userid))))
```
luontola: `not nil?` could be `some?`

src/clj/rems/application/model.clj
opqdonut: done

src/clj/rems/db/applications.clj
opqdonut: done

src/cljc/rems/application_util.cljc
```diff
@@ -11,13 +11,10 @@
...
 
-(def ^:private name-attribute-priority [:name :commonName :displayName :userid :eppn])
-
 (defn get-member-name [attributes]
   (when attributes
-    (->> (map attributes name-attribute-priority)
-         (remove nil?)
-         first)))
+    (or (:name attributes)
+        (:userid attributes))))
```
luontola: `when` is no more needed

opqdonut: done

src/cljs/rems/application.cljs
```diff
@@ -760,26 +759,26 @@
...
-                                        :application/applicant "developer"}
+                                        :application/applicant {:userid "developer"}}
                           :group? true}])
 
    (component-info applicants-info)
    (example "applicants-info"
             [applicants-info {:application/id 42
-                              :application/applicant "developer"
+                              :application/applicant {:userid "developer"}
                               :application/applicant-attributes {:userid "developer"
```
luontola: `:application/applicant-attributes` is still here

opqdonut: done

src/clj/rems/poller/email.clj
```diff
@@ -261,8 +263,8 @@
         email (assoc email-spec
                      :from (:mail-from env)
                      :to (or (:to email-spec)
-                             (util/get-user-mail
```
okahilak: nit: This could be cleaned up a bit further by deleting util/get-user-mail (as well as util/get-username) as neither is used after this patch. It also unifies things a bit since both functions use the "old-style" :eppn/:commonName access.

src/cljs/rems/application.cljs
```diff
@@ -387,7 +385,7 @@
      (application-list/format-application-id config application)]))
 
 (defn- format-event [event]
-  {:user (:commonName (:event/actor-attributes event))
+  {:user (:name (:event/actor-attributes event)) ;; TODO use get-member-name?
```
okahilak: Yeah, sorry, this is my bad, please change to get-member-name. I changed this recently to access :commonName there, but it should have been get-member-name all along.

src/clj/rems/db/users.clj
```diff
@@ -69,7 +70,11 @@
        (map :userid)
        (doall)))
 
-(defn get-user [userid]
-  (->> userid
-       get-user-attributes
-       format-user))
+(defn get-user
+  "Given a userid, returns a map with keys :userid, :mail and :name."
```
okahilak: nit: :mail -> :email

src/clj/rems/db/applications.clj
```diff
@@ -55,7 +55,7 @@
     :items [catalogue-item-id]}))
 
 (defn- valid-user? [userid]
-  (not (nil? (users/get-user-attributes userid))))
+  (some? (users/get-user userid)))
```
okahilak: Hmm, doesn't this always return true now that users/get-user augments the return value with the userid even if the user doesn't exist?

src/clj/rems/db/users.clj
```diff
@@ -69,7 +70,11 @@
...
+  [userid]
+  (-> userid
+      get-user-attributes
+      format-user
+      ;; hack for users without attributes
+      (assoc :userid userid)))
```
okahilak: I'm not sure offhand what this is needed for; can you clarify in the comment above?

opqdonut: get-user-attributes returns nil for unknown users, so we get `{:userid nil :name nil :email nil}` from format-user

src/clj/rems/db/applications.clj
opqdonut: good catch!

src/clj/rems/poller/email.clj
opqdonut: good catch!

src/clj/rems/db/users.clj
okahilak: Hmm, the behavior in the case of a non-existent user is slightly complicated, so it might be good to document it in the docstring. (I see that it is documented in the test case added for this function, but having it in the docstring would make it more explicit.)

src/clj/rems/poller/email.clj
opqdonut: done

src/clj/rems/db/users.clj
okahilak: Right, I mean overall: is this function ever legitimately called for non-existent user (e.g., in the case of a member invited from outside REMS)? If it is, could you add a clarification to the comment on L79 so that the purpose of having this hack would be obvious? Thanks!

src/cljs/rems/application.cljs
opqdonut: done

src/clj/rems/poller/email.clj
opqdonut: #1703

opqdonut:
related to #1726

PR #1708, opened 2019-10-10
Improve guide table

Macroz:
Improve both the guide as well as table component that is using it.

- links to the GitHub master source code from guide
- allows adding static text into the examples
- adds `namespace-info` component for guide
- example of empty table

Screenshot_2019-10-10_16-55-59

### DIFF
--- a/src/cljs/rems/guide_functions.cljs
+++ b/src/cljs/rems/guide_functions.cljs
@@ -15,21 +15,21 @@
                (str git/+master-url+ path))]
     [:a {:href href} link-text]))
 
+(defn- docstring [meta]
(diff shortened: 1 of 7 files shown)


src/cljs/rems/guide_functions.cljs
```diff
@@ -1,21 +1,37 @@
 (ns rems.guide-functions
   (:require [clojure.string :as str]))
 
+(def ^:private +rems-github-master+ "https://github.com/CSCfi/rems/tree/master")
```
opqdonut: link to current revision? it's available

PR #1710, opened 2019-10-11
attachment license in test data

### LINKED ISSUE #1686: Add attachment license to test data

opqdonut:
fixes #1686

## Reviewability
- [x] link to issue


src/clj/rems/db/test_data.clj
```diff
@@ -779,6 +804,7 @@
         _ (create-archived-form!)
         workflows (create-workflows! +fake-users+)]
     (create-disabled-license! (+fake-users+ :owner))
+    (create-attachment-license! {:actor (+fake-users+ :owner)})
```
luontola: Could the license be added to some resource, so that it would be in use? And while doing that, we could get rid of calling `db/create-resource-license!` in `rems.db.test-data/create-resource-license!` and instead rely on the license parameter to `rems.api.services.resource/create-resource!`.

Issue #1711, opened 2019-10-11
End-to-end test for email sending


opqdonut:
we need to prevent failures like #1709 somehow

PR #1714, opened 2019-10-11
Entitlement api fixes

opqdonut:
discovered as part of #1711 

could consider breaking backwards compatibility also, see TODO comment

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -60,11 +60,13 @@
 
 (s/defschema Entitlement
   {:resource s/Str
-   :userid UserId ;; TODO could have UserWithAttributes here instead of :userid and :mail
+   :user UserWithAttributes
    :application-id s/Int
    :start DateTime
    :end (s/maybe DateTime)
-   :mail s/Str})
+   :mail (rjs/field s/Str
+                    {:deprecate true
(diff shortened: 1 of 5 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -51,6 +60,7 @@
 
 (s/defschema Entitlement
   {:resource s/Str
+   :userid UserId ;; TODO could have UserWithAttributes here instead of :userid and :mail
```
luontola: 👍

luontola: For a transition period we could have both.

src/clj/rems/db/entitlements.clj
```diff
@@ -27,7 +28,7 @@
         (db/get-entitlements {:user (if (has-roles? :handler :owner :reporter)
                                       user-or-nil
                                       (getx-user-id))
-                              :resource resource-or-nil
+                              :resource-ext-id resource-or-nil
```
luontola: How did the tests pass before this change? 😟

src/clj/rems/api/schema.clj
opqdonut: hmm okay :userid is fresh so I can replace it with :user, and keep :mail with a deprecation comment

PR #1716, opened 2019-10-14
ADR 007: Browser support policy

### LINKED ISSUE #1364: Specify browser support policy
IE is not currently supported, but different versions of Edge have very different capabilities.

The policy should not mention versions explicitly, but have built in deprecation (such as supporting latest 3 versions, or versions from last 3 years etc.).

opqdonut:
closes #1364


Macroz:
Should we mention the platforms and A11N strategy as well?

opqdonut:
Probably yes, what would you say about them?

Issue #1705, opened 2019-10-10
Handler access to administration page


Macroz:
Handlers need to be able to add attachment licenses. Also the blacklist functionality should be available for them

- [ ] allow administration page access for handler role
- [ ] configure licenses page for handler role
- [ ] configure blacklist page for handler role (#1704)

opqdonut:
ended up showing all admin pages in read-only mode to handlers

PR #1729, opened 2019-10-23
refactor navigation

opqdonut:
Simplifies navigation code by checking the URL instead of re-frame `:page`. This means that the navigation hierarchy should be visible in the URLs, which is why some administration page URLs were changed.

In preparation for #1705.

## Reviewability
- [x] link to issue

## Documentation
- ~update changelog if necessary~

## Follow-up

### DIFF
--- a/src/cljs/rems/navbar.cljs
+++ b/src/cljs/rems/navbar.cljs
@@ -14,17 +14,22 @@
   [dest]
   (str (:root-path context) dest))
 
-(defn nav-link [path title & [active?]]
+(defn- nav-link-impl [path title & [active?]]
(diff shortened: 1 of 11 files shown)


src/cljs/rems/navbar.cljs
```diff
@@ -87,6 +98,6 @@
   [:div
    (component-info nav-link)
    (example "nav-link"
-            [nav-link "example/path" "link text"])
+            [nav-link-impl "example/path" "link text"])
```
Macroz: Add here static text comment about using `nav-link-impl` instead of `nav-link`

PR #1730, opened 2019-10-24
nicer kaocha configuration

opqdonut:
- no notifications
- progress bar instead of dots
- print all test names and descriptions on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ jobs:
           keys:
           - v2-dependencies-{{ checksum "project.clj" }}
       # verify that we can run unit tests without the database:
-      - run: DATABASE_URL=invalid lein kaocha unit
+      - run: DATABASE_URL=invalid lein kaocha --reporter kaocha.report/documentation unit
 
   test:
     <<: *defaults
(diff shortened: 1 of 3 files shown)


opqdonut:
This is what the test output now looks like on circle:

```
--- integration (clojure.test) ---------------------------
rems.api.test-blacklist
  test-blacklist
    initially no blacklist
    add three entries
    query parameters
    remove entry
    add entry again
    remove nonexistent entry

rems.api.test-forms
  form-edit-test
    form content before editing
      form content after editing
  forms-api-security-test
    without authentication
      list
      create
    without owner role
      list
      create
  form-enabled-archived-test
    disable
    archive
    fetch
    unarchive
    enable
    fetch again
  form-editable-test
    New form is editable
    Form is non-editable after in use by a catalogue item
  option-form-item-test
    create
      and fetch
  forms-api-filtering-test
  forms-api-all-field-types-test
    creating
      and fetching
  forms-api-test
    get all
    get one
    not found
    create
      invalid create
      valid create
        and fetch
          result matches input

opqdonut:
see here: https://circleci.com/gh/CSCfi/rems/17296

PR #1744, opened 2019-10-31
Warning about blacklisted members in application

### LINKED ISSUE #1742: Warning to handler when an applicant is blacklisted

opqdonut:
fixes #1742

## Reviewability
- [x] link to issue

## API

## Documentation
- ~update changelog if necessary~

## Testing

## Follow-up
- ~new tasks are created for pending or remaining tasks~

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -629,6 +629,7 @@
 
 (defn- hide-sensitive-information [application]
   (-> application
+      (dissoc :application/blacklisted-users)
       (update :application/events hide-sensitive-events)
       (update :application/workflow dissoc :workflow.dynamic/handlers)))
 
@@ -642,7 +643,6 @@
 (defn- hide-non-public-information [application]
   (-> application
       hide-invitation-tokens
-      (dissoc :application/blacklisted-users) ;; TODO expose this
       ;; these are not used by the UI, so no need to expose them (especially the user IDs)
       (dissoc ::latest-comment-request-by-user ::latest-decision-request-by-user)
       (dissoc :application/past-members)))
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -542,7 +542,7 @@
 
              {}))))
 
-(defn- get-blacklisted-users [application blacklisted?]
+(defn- get-blacklist [application blacklisted?]
   (let [all-members (application-util/applicant-and-members application)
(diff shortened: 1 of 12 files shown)


opqdonut:
Screenshot:
blacklist-warning

opqdonut:
I'll follow up with some refactorings I noticed along the way

src/clj/rems/application/model.clj
```diff
@@ -541,6 +542,18 @@
...
              {}))))
 
+(defn- get-blacklisted-users [application blacklisted?]
+  (let [all-members (application-util/applicant-and-members application)
+        all-resources (map :resource/ext-id (:application/resources application))]
+    (vec
+     (for [member all-members
+           resource all-resources
+           :when (blacklisted? (:userid member) resource)]
+       {:blacklist/user member :blacklist/resource {:resource/ext-id resource}}))))
```
Macroz: I would expect this function to group by user. Perhaps do that or rename the function to something like `get-blacklisted-user-resources`. The name does not imply that it returns the resources as well or that it returns one row per user-resource -pair.

src/cljs/rems/application.cljs
```diff
@@ -62,6 +62,16 @@
...
 
+(defn- blacklist-warning [application]
+  (let [blacklist (:application/blacklisted-users application)]
+    (when (not (empty? blacklist))
+      [:div.alert.alert-danger
+       (text :t.form/alert-blacklisted-users)
+       (into [:ul]
+             (for [entry blacklist]
+               [:li (get-member-name (:blacklist/user entry))
+                ": " (get-in entry [:blacklist/resource :resource/ext-id])]))])))
```
Macroz: I think from the UX perspective it doesn't make sense to list here the resource ids but we should instead use the catalogue items that are present in the application and their names.

src/clj/rems/application/model.clj
opqdonut: changed the function & field name to `blacklist`

src/cljs/rems/application.cljs
opqdonut: did this in the frontend

opqdonut:
Now the blacklist shows the catalogue item. I opted not to render the urn / more info links for clarity.

blacklist

opqdonut:
Hmm the same resource can point to multiple catalogue items, I'll need to fix that

Issue #837, opened 2019-01-15
vTHL4: Bulk change form of catalogue items


foxlynx:
Apply a new form to a set of catalogue items. REMS1 creates new catalogue items under the hood and disabled the old ones. In applications that are already in processing the old form is still used because they still use the old catalogue item.

List view should be done before this issue.

- [x] select multiple catalogue items (by clicking checkboxes next to the items)
- [x] select bulk operation (change form)
- [x] form with options for selectable form (autocomplete?)
- ~check if catalogue item name should be possibly updated here too (e.g. to add version or year number)~ decided not to do this yet
- [x] create new catalogue items for each catalogue item with the new form
- [x] disable & archive the old catalogue items
- ~link the new catalogue item to the old to track~ decided not to do this yet

jaakkocsc:
See #1763

PR #1756, opened 2019-11-04
feat: catalogue item form change API

Macroz:
Catalogue item form change API for #837 

## Reviewability
- [x] link to issue
- ~[ ] note if PR is on top of other PR~
- ~[ ] note if related change in rems-deploy repo~
- ~[ ] consider adding screenshots for ease of review~

## API
- ~[ ] Events are backwards compatible~

## Documentation
- ~[ ] components are added to guide page~
- ~[ ] documentation _at least_ for config options (i.e. docs folder)~
- ~[ ] ADR for major architectural decisions or experiments~

## Testing

## Follow-up
- ~[ ] new tasks are created for pending or remaining tasks~

### DIFF
--- a/src/clj/rems/api/services/catalogue.clj
+++ b/src/clj/rems/api/services/catalogue.clj
@@ -67,3 +67,32 @@
 
 (def get-localized-catalogue-items catalogue/get-localized-catalogue-items)
 (def get-localized-catalogue-item catalogue/get-localized-catalogue-item)
+
+(defn change-form!
+  "Changes the form of a catalogue item.
+
+  Since we don't want to modify the old item we must create
+  a new item that is the copy of the old item except for the changed form."
+  [item form-id]
+  ;; create a new item with the new form
+  (let [new-item (db/create-catalogue-item! {:enabled true
+                                             :archived false
+                                             :form form-id
+                                             :resid (:resource-id item)
+                                             :wfid (:wfid item)})]
+
+    ;; copy localizations
+    (doseq [[langcode localization] (:localizations item)]
+      (db/upsert-catalogue-item-localization! {:id (:id new-item)
+                                               :langcode (name langcode)
+                                               :title (:title localization)
+                                               :infourl (:infourl localization)}))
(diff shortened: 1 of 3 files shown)


src/clj/rems/api/services/catalogue.clj
```diff
@@ -67,3 +67,32 @@
...
+                                             :form form-id
+                                             :resid (:resource-id item)
+                                             :wfid (:wfid item)})]
+
+    ;; copy localizations
+    (doseq [[langcode localization] (:localizations item)]
+      (db/upsert-catalogue-item-localization! {:id (:id new-item)
+                                               :langcode (name langcode)
+                                               :title (:title localization)
+                                               :infourl (:infourl localization)}))
```
opqdonut: could you use create-catalogue-item! from this ns instead of these db/ functions?

src/clj/rems/api/services/catalogue.clj
```diff
@@ -67,3 +67,32 @@
...
+    ;; Reset cache so that next call to get localizations will get these ones.
+    (catalogue/reset-cache!)
+
+    ;; end the old catalogue item
+    (db/set-catalogue-item-enabled! {:id (:id item) :enabled false})
+    (db/set-catalogue-item-archived! {:id (:id item) :archived true})
+    (db/set-catalogue-item-endt! {:id (:id item) :end (:start new-item)})
```
opqdonut: could consider using the `set-` functions from this ns, but it doesn't really matter

src/clj/rems/api/services/catalogue.clj
Macroz: That one creates a whole API response so I'd rather not use it internally. Duplicated is ok here for now. Maybe can DRY this later.

src/clj/rems/api/services/catalogue.clj
Macroz: I think they suffer from the same as above, i.e. they almost give API level answer whereas the DB level functions only do the thing. At least there is no duplicated of code here.

PR #1760, opened 2019-11-05
Amendment to ADR 003: Namespaced keywords

opqdonut:

PR #1762, opened 2019-11-05
refactor: don't use ns aliases in keywords

opqdonut:
obeying ADR 003

see also #1760

### DIFF
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -378,7 +378,7 @@
            :application/comment (:comment cmd)})))
 
 (defn- actor-is-not-decider-error [application cmd]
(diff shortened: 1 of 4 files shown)


src/clj/rems/application/commands.clj
```diff
@@ -387,7 +387,7 @@
   (or (actor-is-not-decider-error application cmd)
       (when-not (contains? #{:approved :rejected} (:decision cmd))
         {:errors [{:type :invalid-decision :decision (:decision cmd)}]})
-      (let [last-request-for-actor (get-in application [::model/latest-decision-request-by-user (:actor cmd)])]
+      (let [last-request-for-actor (get-in application [:rems.applicaiton.model/latest-decision-request-by-user (:actor cmd)])]
```
Macroz: Typo

PR #1764, opened 2019-11-05
Changes to translations, from LBR

okahilak:


resources/translations/en.edn
```diff
@@ -170,15 +170,15 @@
          :apply "Apply"
          :apply-for-bundle "Apply for %1 resources as a bundle "
          :header "%1 resources in the cart"
-         :intro [:div.intro [:p "Add resources to the cart to apply for them. If several resources can be applied to with the same application, they will be automatically grouped into a bundle."]]
+         :intro [:div.intro [:p "Apply for access by adding resources to the cart. When applying for multiple rights at the same time, resources using the same application process are bundled together as one application."]]
          :remove "Remove"}
   :catalogue {:catalogue "Catalogue"
               :header "Resource"
-              :intro [:p "The catalogue contains all the resources that you can apply for in this system."]
+              :intro [:p "Access can be applied for the following resources."]
```
okahilak: Hm, I'd like to add a colon to the end of the sentence, but the problem is that the list of resources does not follow.

I think I actually prefer the old version, it is feels more general and more descriptive.

resources/translations/en.edn
```diff
@@ -170,15 +170,15 @@
          :apply "Apply"
          :apply-for-bundle "Apply for %1 resources as a bundle "
          :header "%1 resources in the cart"
-         :intro [:div.intro [:p "Add resources to the cart to apply for them. If several resources can be applied to with the same application, they will be automatically grouped into a bundle."]]
+         :intro [:div.intro [:p "Apply for access by adding resources to the cart. When applying for multiple rights at the same time, resources using the same application process are bundled together as one application."]]
```
opqdonut: "application process" is not a REMS term. I like our old formulation more

resources/translations/en.edn
opqdonut: this seems ungrammatical, maybe "Access to the following resources can be applied for"

resources/translations/fi.edn
```diff
@@ -137,7 +137,7 @@
                           :submitted "%1 lähetti hakemuksen käsiteltäväksi."
                           :unknown "Tuntematon"}
                  :id "Tunniste"
-                 :intro [:div.intro [:p "Täyttämällä ja lähettämällä tämän hakemuksen kaikki hakijat voivat saada oikeuden haettuihin resursseihin. Kunkin hakijan on lisäksi hyväksyttävä käyttöehdot viimeistään hakemuksen käsittelyn jälkeen."]]
+                 :intro [:div.intro [:p "Täytä hakemus ja lähetä se käsiteltäväksi. Hakemukseen voi kutsua jäseniksi useampia hakijoita, ja jokaisen on hyväksyttävä käyttöehdot viimeistään kun hakemus on hyväksytty."]]
```
opqdonut: this is missing the point from the english text that access is only granted after accepting the terms of use... but I guess that's fine

okahilak: I think there should be a comma before the last "kun" word.

resources/translations/fi.edn
```diff
@@ -168,15 +168,15 @@
          :apply "Hae"
          :apply-for-bundle "Hae %1 aineistoa yhdessä "
          :header "%1 aineistoa korissa"
-         :intro [:div.intro [:p "Lisää resursseja koriin hakeaksesi niitä. Samalla hakemuksella haettavat resurssit ryhmitellään automaattisesti yhteen."]]
+         :intro [:div.intro [:p "Hae käyttöoikeuksia lisäämällä resursseja koriin. Useampaa oikeutta kerralla haettaessa keskenään samaa hakemusmenettelyä noudattavat resurssit yhdistyvät yhdeksi hakemukseksi."]]
```
opqdonut: hakemusmenettely is not a REMS term, here too I like the previous version

resources/translations/en.edn
okahilak: This is the only place in the translations in which the word "right" is used to refer to the access rights. Using "access right" instead ("When applying for multiple access rights [...]") would add some context.

resources/translations/en.edn
okahilak: I think the problem is that "apply" needs the dependent preposition ("apply for"), so grammatically it should be "Access can be applied for for the following resources", but having to repeat "for" doesn't sound that good.

resources/translations/fi.edn
jaakkocsc: Yes, there should

okahilak: I think the second clause in the last sentence is dependent on the first clause, so it would be better to use subordinate conjunction there, e.g., "jolloin", instead of "ja". (I think the English translation has a similar slight issue.)

resources/translations/fi.edn
```diff
@@ -168,15 +168,15 @@
...
          :remove "Poista"}
   :catalogue {:catalogue "Aineistoluettelo"
               :header "Aineistoluettelo"
-              :intro [:p "Aineistoluettelossa on listattuna kaikki resurssit joita voit tästä järjestelmästä hakea."]
+              :intro [:p "Allaolevien resurssien käyttöoikeudet ovat haettavissa."]
               :more-info "Lisätietoja"
               :continue-existing-application "Jatka hakemusta"
-              :continue-existing-application-intro [:div.intro [:p "Sinulla on keskeneräisiä hakemuksia, joita voit jatkaa tästä:"]]
-              :apply-resources "Hae uusia resursseja"}
+              :continue-existing-application-intro [:div.intro [:p "Keskeneräiset hakemukset:"]]
```
okahilak: I really like this kind of compact expression, I think it's a definite improvement over the old one.

resources/translations/en.edn
jaakkocsc: "multiple rights" does not sound correct semantically. "multiple resources" would be more coherent phrase.

resources/translations/fi.edn
jaakkocsc: It could even be "Keskeneräiset hakemuksesi:"

resources/translations/fi.edn
jaakkocsc: "…useampia hakijoita, joista jokaisen…" and "…viimeistään silloin, kun…"

resources/translations/en.edn
jaakkocsc: "Apply for access rights by adding…" Term access by it self refers to physical access. "When applying for multiple resources…". An also, ""…using the same application process are bundled together as a single application" might be even more clear?

resources/translations/en.edn
UFOPOLI: The old version is partly redundant and doesn't sound like completely proper English.

resources/translations/fi.edn
jaakkocsc: Previous version is much more easier to read and understand.

resources/translations/fi.edn
UFOPOLI: I don't think there should. "Viimeistään kun" is an expression that shouldn't be broken in two. A comma is used to separate two clauses that could also exist separately. The kun clause meets this criterion but the one ending with "viimeistään" doesn't.

resources/translations/en.edn
UFOPOLI: Resources aren't "applied for", access rights are. How about: 

"Apply for access by adding resources to the cart. When applying for access to multiple resources at the same time, resources using the same application are bundled together."

resources/translations/en.edn
UFOPOLI: We had a long discussion about the "for" problem last week. "Access to the following resources can be applied for" would work.

resources/translations/fi.edn
UFOPOLI: I thought about that and came to the conclusion that accepting the terms is implied because an application can't be accepted otherwise.

resources/translations/fi.edn
UFOPOLI: Let's change it then: 

"Hae käyttöoikeuksia lisäämällä kielivaroja koriin. Useamman kielivaran käyttöoikeutta kerralla haettaessa keskenään samaa hakemusta käyttävät kielivarat yhdistyvät."

I have no idea where "resursseja" reappeared from here. It wasn't in the version I sent.

resources/translations/fi.edn
luontola: http://www.kielitoimistonohjepankki.fi/ohje/96

resources/translations/en.edn
UFOPOLI: "Apply for access by adding resources to the cart. When applying for access to multiple resources at the same time, resources using the same application are bundled together."

resources/translations/fi.edn
UFOPOLI: "joista jokaisen" is fine but "silloin" is redundant, and the comma was discussed above.

resources/translations/fi.edn
UFOPOLI: "Keskeneräiset hakemuksesi:" is fine. I generally try to avoid too much "you" in the texts but it can be there.

resources/translations/fi.edn
okahilak: I actually explained that (admittedly briefly and in parentheses) in the previous email I sent: these are the translations that are for the common use, so let's use "resource" here. After they're accepted, I'll make custom translations that are LBR-specific, in which "resurssi" has been changed to "kielivara" but are identical otherwise.

That is for technical reasons: LBR-specific translations are in our private repository, so having these discussions there would not work out so well.

resources/translations/en.edn
okahilak: Done

resources/translations/en.edn
okahilak: Done

resources/translations/fi.edn
okahilak: Opted to use Jaakko's proposal in this patch. Tero's proposal is a bit difficult to understand for me, most likely because of the participial phrase (lauseenvastike). Maybe it could be an actual clause instead?

resources/translations/en.edn
okahilak: Added "[Apply for access] rights" to the first sentence, as Jaakko proposed.

resources/translations/fi.edn
okahilak: Added "joista jokaisen", left out "silloin".

resources/translations/fi.edn
okahilak: Opted to not add the possessive suffix for consistency reasons: it's not used elsewhere in the translations either (e.g., "Jatka hakemusta").

resources/translations/en.edn
okahilak: I'm fine with how it ended up.

resources/translations/fi.edn
UFOPOLI: In the latter sentence, I guess (there's one in both)? We can change it to e.g. 

"Hae käyttöoikeuksia lisäämällä kielivaroja koriin. Samaa hakemusta käyttävät kielivarat yhdistyvät korissa yhdeksi hakemukseksi."

UFOPOLI: And in Engllish: 

"Apply for access by adding resources to the cart. Resources using the same application are bundled together in the cart."

okahilak: Yes, in the latter sentence. I like that proposal. Done.

PR #1786, opened 2019-11-11
feat: hide application search tips behind (?) icon

opqdonut:
closes #1767

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue
- [x] consider adding screenshots for ease of review

## Documentation
- [x] update changelog if necessary

## Accessibility
- [ ] all icons have the aria-label attribute
- [ ] all fields have a label
- [ ] errors are linked to fields with aria-describedby
- [ ] contrast is checked
- [ ] conscious decision about where to move focus after an action

## Follow-up
- [x] no critical TODOs left to implement

PR #1787, opened 2019-11-11
feat: adjust formatting of empty application list

opqdonut:
The success message "No applications to process." is misleading when
the user has entered a search term. Use a more neutral "No
applications." in all cases.

Noticed when implementing #1786 

## Reviewability

## Follow-up


opqdonut:
Before:
search-before
After:
search-after

PR #1788, opened 2019-11-11
Fix dropdown

Macroz:
Dropdown was misbehaving and selecting all items in the single case. This is due to mixing up JS and CLJS values. This is a fix that simplifies the situation.

The real problem is shown e.g. in create catalogue item page where you can select something from a dropdown. Then see that if you expand the same dropdown again you will see that all rows are rendered selected.

The examples also referred to `:userid` as id but obviously the example is meant to show `:id` as the id.

### DIFF
--- a/src/cljs/rems/dropdown.cljs
+++ b/src/cljs/rems/dropdown.cljs
@@ -48,6 +48,12 @@
                          :item-key :id
                          :item-label :userid
                          :on-change on-change}])
+     (example "dropdown menu, single-choice, selected item Bob"
+              [dropdown {:items items
+                         :item-key :id
+                         :item-label :userid
+                         :item-selected? #(= "Bob" (:userid %))
+                         :on-change on-change}])
      (example "dropdown menu, multi-choice, several values selected"
               [dropdown {:items items
                          :item-key :id
--- a/src/cljs/rems/dropdown.cljs
+++ b/src/cljs/rems/dropdown.cljs
@@ -29,28 +29,28 @@
 (defn guide
   []
   (let [on-change (fn [items] (println "items" items))
-        items [{:id 1 :userid "Alice"}
-               {:id 2 :userid "Bob"}
-               {:id 3 :userid "Carl"}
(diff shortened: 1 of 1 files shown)


src/cljs/rems/dropdown.cljs
```diff
@@ -9,29 +9,23 @@
...
+                 :inputId id
+                 :isMulti multi?
+                 :isClearable clearable?
+                 :isOptionDisabled #(item-disabled? %)
+                 :isOptionSelected #(item-selected? %)
+                 :maxMenuHeight 200
+                 :noOptionsMessage #(text :t.dropdown/no-results)
+                 :options (into-array items)
+                 :value (into-array (filter item-selected? items))
+                 :onChange #(do (prn %)
```
luontola: Debug print?

Macroz: Good catch

src/cljs/rems/dropdown.cljs
```diff
@@ -45,12 +39,18 @@
...
               [dropdown {:items items
-                         :item-key :userid
+                         :item-key :id
                          :item-label :userid
                          :on-change on-change}])
+     (example "dropdown menu, single-choice, selected item Bob"
+              [dropdown {:items items
+                         :item-key :id
+                         :item-label :userid
+                         :item-selected? #(= "Bob" (:userid %))
```
luontola: Renaming `:userid` to `:name` would make the example more understandable, when there is only one unique ID. Also, `:item-selected?` should check the `:id`.

Macroz: Yes I think I'll do that. Only for us `:userid` is a string obviously.

I decided to use `"Bob"` here so that you don't have to look for the id value mapping from source in the guide.

Macroz: Also it demonstrates that you get the full object. The last example uses the `:id` instead.

luontola: 👌

PR #1775, opened 2019-11-07
Email outbox

luontola:
Closes #1750

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue

## Documentation
- [x] update changelog if necessary
- [ ] add or update docstrings for namespaces and functions
- [x] documentation _at least_ for config options (i.e. docs folder)
- [ ] ADR for major architectural decisions or experiments

## Different installations
- [ ] ~new configuration options added to rems-deploy repository~

## Testing
- [x] complex logic is unit tested
- [x] valuable features are integration / browser / acceptance tested automatically

## Follow-up
- [x] new tasks are created for pending or remaining tasks
- [x] no critical TODOs left to implement

PR #1789, opened 2019-11-11
Email namespace reordering

luontola:
Follow up & on top of #1775


src/clj/rems/email/core.clj
```diff
@@ -0,0 +1,102 @@
...
+
+        :else
+        (try
+          (postal/send-message {:host host :port port} email)
+          nil
+          (catch Throwable e ; e.g. email address does not exist
+            (log/warn e "failed sending email:" (pr-str email))
+            (str "failed sending email: " e)))))))
+
+(defn run []
```
opqdonut: should this be either private or called something like `run-poller` or perhaps `process-outbox`?

PR #1797, opened 2019-11-13
maintenance commands: add-api-key, fix demo-data and test-data

opqdonut:
- Now you can add API keys from the command line:
  `lein run add-api-key 42 "an api key"`
- Generating test-data and demo-data got broken by #1775, fixed now


opqdonut:
related to #1530

PR #1798, opened 2019-11-13
docs: add AVOID comments to code snippets in ADRs

opqdonut:


opqdonut:
follow-up for #1760

luontola:
For some of the big code samples it would be nice if the problematic line would be highlighted e.g. with a comment saying what to avoid.

opqdonut:
Did you have something concrete in mind? I read through the examples and the text immediately prior to the AVOID snippets seemed to describe what to avoid pretty well.

luontola:
I added my ideas as a commit.

opqdonut:
good stuff :+1: (can't approve my own PR :)

Issue #1803, opened 2019-11-14
Follow when adding a new field


jaakkocsc:
When creating/editing a form and adding a new field, the view should scroll to show the field in full and focus should go to first text box.

PR #1810, opened 2019-11-19
Follow form item when moving

opqdonut:
Closes #1802

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue

## Documentation
- ~update changelog if necessary~ will update after #1803
- [x] add or update docstrings for namespaces and functions

## Accessibility
- [x] conscious decision about where to move focus after an action

## Follow-up
- [x] no critical TODOs left to implement

PR #1811, opened 2019-11-19
form editor: follow field when moving

### LINKED ISSUE #1802: Follow field when moving
When creating/editing a form and moving a field up/down, it would be nice if focus follows the item begin moved.

opqdonut:
Closes #1802

## Reviewability
- [x] link to issue

## Documentation
- ~update changelog if necessary~ (will update in #1803)

## Accessibility

## Follow-up


opqdonut:
let's go with #1810 instead

Issue #1784, opened 2019-11-11
Entitlement outbox instead of entitlement poller


opqdonut:
After #1750 we could also refactor the entitlements to work like this:

- entitlements are inserted into the db immediately (in the `command!` function)
- new entitlements are also inserted to a new entitlement_outbox table which tracks which entitlements have been POSTed (to `:entitlements-target`)
- the entitlement poller monitors the `entitlement_outbox` table and performs the POSTs, tracking success/failure in the entitlement_outbox table
- further ideas:
  - should we get rid of the `entitlement_post_log` table?

Pros of this approach:
- similar email and POST handling
- entitlements appear in db immediately

PR #1816, opened 2019-11-21
remove entitlement poller

opqdonut:
Instead, process entitlements synchronously in the command handler.
This has no change in behaviour since we never retried POSTing the
entitlements anyway.

To be followed with a proper entitlement outbox.

For #1784


src/clj/rems/api/services/command.clj
```diff
@@ -28,6 +34,11 @@
          (mapcat #(approver-bot/generate-commands %))
          doall)))
 
+(defn run-process-managers [new-events]
+  (concat
+   (run-entitlements new-events)
+   (run-approver-bot new-events)))
```
luontola: The call to `rems.email.core/generate-emails!` could be moved from `rems.api.services.command/command!` into this function the same way as `run-entitlements`.

opqdonut: heh just noticed the same

Issue #1706, opened 2019-10-10
UI for adding and removing users from the blacklist


Macroz:
Edit functionality for blacklist. The API exists already. This should be in the administration pages.

- [x] users can be added and removed to a per-resource blacklist by handlers

PR #1818, opened 2019-11-21
Removing users from blacklist

luontola:
First part of #1706

Should there be a commenting possibility when removing blacklist entries?

remove button

## Reviewability
- [x] link to issue

## Documentation

## Accessibility

### DIFF
--- a/src/cljs/rems/administration/blacklist.cljs
+++ b/src/cljs/rems/administration/blacklist.cljs
@@ -8,7 +8,7 @@
             [rems.spinner :as spinner]
             [rems.table :as table]
             [rems.text :refer [text localize-time]]
-            [rems.util :refer [fetch]]))
+            [rems.util :refer [fetch post!]]))
 
 (rf/reg-event-fx
  ::enter-page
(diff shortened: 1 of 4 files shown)


src/cljs/rems/administration/blacklist.cljs
```diff
@@ -26,6 +26,27 @@
...
    {:db (assoc db ::loading? true)}))
 
+(rf/reg-event-fx
+ ::remove-from-blacklist
+ (fn [{:keys [db]} [_ resource user]]
+   (let [description [text :t.administration/remove]]
+     (post! "/api/blacklist/remove"
+            {:params {:blacklist/resource (select-keys resource [:resource/ext-id])
+                      :blacklist/user (select-keys user [:userid])
+                      :comment ""}
```
opqdonut: since the blacklist is such a serious thing, perhaps we should support comments in removals too

luontola: Ok. I'll add it in the follow up PR.

PR #1820, opened 2019-11-22
refactor: use fully-qualified keywords in blacklist hugsql queries

opqdonut:
now that hugsql supports namespaced keywords, here's an example on how
to use them


luontola:
Needs manual merging.

PR #1823, opened 2019-11-25
Catalogue item change form

Macroz:
This implements a first (maybe rough) version of #837 and closes it.

The looping is done in the UI so that if there is a problem, it's stopped, but the user can opt to rerun the loop again. Also the new table select with filter options can be used creatively to select the forms to change e.g. by filtering to existing use of some form.

There can be some confusion for users which functions are "bulk edits" and which affect only one item. However re-rendering the buttons on the table is too slow and "jumpy". We could perhaps have an area for only bulk operations, but that we can decide based on feedback. At least it is possible to go to view an item and back, and not lose the selection.

Since this is a "power tool", we should spend extra time making sure this is properly tested and does not cause loss of work / ops work.

Screenshot_2019-11-25_13-01-49
Screenshot_2019-11-25_13-02-16
Screenshot_2019-11-25_13-02-31

## Reviewability
- [x] link to issue
- ~[ ] note if PR is on top of other PR~
- ~[ ] note if related change in rems-deploy repo~

## API
- ~[ ] Events are backwards compatible~

## Documentation
- ~[ ] documentation _at least_ for config options (i.e. docs folder)~
- ~[ ] ADR for major architectural decisions or experiments~

## Testing

## ~Accessibility~
Note, the accessibility is based on existing table, dropdown and table selection features so there is no change.

## Follow-up
- ~[ ] new tasks are created for pending or remaining tasks~

### DIFF
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -49,7 +49,6 @@
                    :back "Back"
                    :blacklist "Blacklist"
                    :change-form "Change form"
-                   :change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy with the chosen form will be activated."
                    :cancel "Cancel"
                    :catalogue-item "Catalogue item"
                    :catalogue-items "Catalogue items"
@@ -189,7 +188,8 @@
              :show "Show"
              :show-less "Show less"
              :show-more "Show more"}
-  :change-form {:form-selection "Form"}
+  :change-form {:change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy with the chosen form will be activated."
+                :form-selection "Form"}
   :create-catalogue-item {:form-selection "Form"
                           :infourl "More info URL (optional, defaults to resource URN)"
                           :resource-selection "Resource"
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -188,7 +188,7 @@
              :show "Show"
              :show-less "Show less"
              :show-more "Show more"}
-  :change-form {:change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy with the chosen form will be activated."
+  :change-form {:change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy will use the chosen form." 
                 :form-selection "Form"}
   :create-catalogue-item {:form-selection "Form"
                           :infourl "More info URL (optional, defaults to resource URN)"
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -359,6 +359,9 @@
            :search "Search"}
   :table {:ascending-order "Ascending order"
           :descending-order "Descending order"
+          :all-rows-selected "All rows selected"
+          :no-rows-selected "No rows selected"
+          :some-rows-selected "Some rows selected"
           :show-all-n-rows "Show all %1 rows"}
   :unauthorized-page {:unauthorized "Unauthorized"
                       :you-are-unauthorized "You're unauthorized to access this page."}
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -48,9 +48,13 @@
                    :archive "Archive"
                    :back "Back"
                    :blacklist "Blacklist"
+                   :change-form "Change form"
+                   :change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy with the chosen form will be activated."
                    :cancel "Cancel"
                    :catalogue-item "Catalogue item"
                    :catalogue-items "Catalogue items"
+                   :change "Change"
+                   :changed "Changed"
                    :comment "Comment"
                    :copy-as-new "Copy as new"
                    :create-catalogue-item "Create catalogue item"
@@ -186,6 +190,7 @@
              :show "Show"
              :show-less "Show less"
              :show-more "Show more"}
+  :change-form {:form-selection "Form"}
   :create-catalogue-item {:form-selection "Form"
                           :infourl "More info URL (optional, defaults to resource URN)"
                           :resource-selection "Resource"
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -54,7 +54,6 @@
                    :catalogue-item "Catalogue item"
                    :catalogue-items "Catalogue items"
                    :change "Change"
-                   :changed "Changed"
                    :comment "Comment"
                    :copy-as-new "Copy as new"
                    :create-catalogue-item "Create catalogue item"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -186,7 +186,7 @@
              :show "Näytä"
              :show-less "Näytä vähemmän"
              :show-more "Näytä lisää"}
-  :change-form {:change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto lopetetaan ja uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake."
+  :change-form {:change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto poistetaan käytöstä ja uusi kopio käyttää valitsemaasi lomaketta."
                 :form-selection "Lomake"}
   :create-catalogue-item {:form-selection "Lomake"
                           :infourl "Lisätietoja-URL (vapaaehtoinen, puuttuessa käytetään resurssin URN:ää)"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -48,9 +48,13 @@
                    :archive "Arkistoi"
                    :back "Takaisin"
                    :blacklist "Kieltolista"
+                   :change-form "Vaihda lomaketta"
+                   :change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto lopetetaan ja uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake."
                    :cancel "Peruuta"
                    :catalogue-item "Aineisto"
                    :catalogue-items "Aineistot"
+                   :change "Vaihda"
+                   :changed "Vaihdettu"
                    :comment "Kommentti"
                    :copy-as-new "Kopioi uudeksi"
                    :create-catalogue-item "Uusi aineisto"
@@ -184,6 +188,7 @@
              :show "Näytä"
              :show-less "Näytä vähemmän"
              :show-more "Näytä lisää"}
+  :change-form {:form-selection "Lomake"}
   :create-catalogue-item {:form-selection "Lomake"
                           :infourl "Lisätietoja-URL (vapaaehtoinen, puuttuessa käytetään resurssin URN:ää)"
                           :resource-selection "Resurssi"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -49,7 +49,6 @@
                    :back "Takaisin"
                    :blacklist "Kieltolista"
                    :change-form "Vaihda lomaketta"
-                   :change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto lopetetaan ja uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake."
                    :cancel "Peruuta"
                    :catalogue-item "Aineisto"
                    :catalogue-items "Aineistot"
@@ -187,7 +186,8 @@
              :show "Näytä"
              :show-less "Näytä vähemmän"
              :show-more "Näytä lisää"}
-  :change-form {:form-selection "Lomake"}
+  :change-form {:change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto lopetetaan ja uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake."
+                :form-selection "Lomake"}
   :create-catalogue-item {:form-selection "Lomake"
                           :infourl "Lisätietoja-URL (vapaaehtoinen, puuttuessa käytetään resurssin URN:ää)"
                           :resource-selection "Resurssi"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -54,7 +54,6 @@
                    :catalogue-item "Aineisto"
                    :catalogue-items "Aineistot"
                    :change "Vaihda"
-                   :changed "Vaihdettu"
                    :comment "Kommentti"
                    :copy-as-new "Kopioi uudeksi"
                    :create-catalogue-item "Uusi aineisto"
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -357,6 +357,9 @@
            :search "Etsi"} ; "Hae" would be more natural, but the catalogue page uses the same word for another meaning
   :table {:ascending-order "Nouseva järjestys"
           :descending-order "Laskeva järjestys"
+          :all-rows-selected "Kaikki rivit valittuna"
+          :no-rows-selected "Ei yhtään riviä valittuna"
+          :some-rows-selected "Joitain rivejä valittuna"
           :show-all-n-rows "Näytä kaikki %1 riviä"}
   :unauthorized-page {:unauthorized "Ei käyttöoikeutta"
                       :you-are-unauthorized "Sinulla ei ole tämän sivun käyttöoikeutta."}
--- a/src/clj/rems/api/catalogue_items.clj
+++ b/src/clj/rems/api/catalogue_items.clj
@@ -48,7 +48,7 @@
 
 (s/defschema ChangeFormResponse
   {:success s/Bool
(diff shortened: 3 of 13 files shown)


luontola:
I think the selected items should be reset when entering the catalogue page. 

Also, seem that you can select items from multiple non-overlapping searches, which might be ok as a design decision (e.g. if it's not easy to find all items to change with a single search), but it becomes confusing when they are still selected after visiting other pages. Resetting the selection on entering the catalogue page should fix that.

resources/translations/en.edn
```diff
@@ -185,6 +188,7 @@
              :show "Show"
              :show-less "Show less"
              :show-more "Show more"}
+  :change-form {:form-selection "Form"}
```
luontola: Should more translation keys be in this map? At least the intro text.

src/clj/rems/api/catalogue_items.clj
```diff
@@ -48,7 +48,7 @@
 
 (s/defschema ChangeFormResponse
   {:success s/Bool
-   :new-catalogue-item-id s/Int})
+   :catalogue-item-id s/Int})
```
luontola: This breaking API change should be mentioned in the changelog.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -60,12 +61,30 @@
...
    "/administration/catalogue-items/create"
    (text :t.administration/create-catalogue-item)])
 
-(defn- to-catalogue-item [catalogue-item-id]
+(defn- change-form-button [items]
+  [:button.btn.btn-primary
+   {:disabled (when (empty? items) :disabled)
+    :on-click (fn []
+                (rf/dispatch [:rems.administration.change-catalogue-item-form/enter-page items])
+                (navigate! "/administration/catalogue-items/change-form") )}
```
luontola: Why is enter-page dispatched here in the link instead of the router, like all other pages?

luontola: If it's because of the items parameter, that could also be transmitted as query parameters in the URL.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -104,8 +123,8 @@
                      {:td [:td.active
                            [readonly-checkbox {:value checked?}]]
                       :sort-value (if checked? 1 2)})
-           :commands {:td [:td.commands
-                           [to-catalogue-item (:id item)]
+           :commands {:td [:td.commands {:on-click #(.stopPropagation %)}
```
luontola: Why stopPropagation? A clarifying comment would be necessary.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -131,11 +150,16 @@
...
-                         :default-sort-column :name}]
+                         :default-sort-column :name
+                         :selectable? true
+                         :on-select #(rf/dispatch [::set-selected-items! %])}]
     [:div.mt-3
      [table/search catalogue-table]
      [table/table catalogue-table]]))
 
+(defn- items-by-id [items ids]
+  (filter (comp ids :id) items))
```
luontola: For me items-by-id would be the name for a map of format `{id item}`, but this function is something else. Maybe extract the predicate to a named function and use filter directly.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -144,7 +168,10 @@
         (if @(rf/subscribe [::loading?])
           [[spinner/big]]
           [[roles/when roles/show-admin-edit-buttons?
-            [to-create-catalogue-item]
-            [status-flags/display-archived-toggle #(rf/dispatch [::fetch-catalogue])]
+            [:div.commands.text-left.pl-0
+             [ create-catalogue-item-button]
```
luontola: Extra space.

src/clj/rems/api/catalogue_items.clj
Macroz: It's not a breaking API change when this is the PR where the API is introduced in.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: We **don't** want to reset everything on page enter i.e. when you change URL. That would disallow view of single item and back while having something selected.

Macroz: Query parameters would introduce a lot of duplication or coupling between the table where everything is selected and change form page.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: Otherwise clicking on a button link in commands will propagate to row selection.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: I think it's missing an s.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: > I think the selected items should be reset when entering the catalogue page.
> 
> Also, seem that you can select items from multiple non-overlapping searches, which might be ok as a design decision (e.g. if it's not easy to find all items to change with a single search), but it becomes confusing when they are still selected after visiting other pages. Resetting the selection on entering the catalogue page should fix that.

It would prevent the user from viewing specific rows while trying to select the right ones because the list resets on every navigation. Like you can read from commit messages I chose to explicitly allow multiple filters as a feature and disallow archive toggle here.

Macroz:
> I think the selected items should be reset when entering the catalogue page.
> 
> Also, seem that you can select items from multiple non-overlapping searches, which might be ok as a design decision (e.g. if it's not easy to find all items to change with a single search), but it becomes confusing when they are still selected after visiting other pages. Resetting the selection on entering the catalogue page should fix that.

It would prevent the user from viewing specific rows while trying to select the right ones because the list resets on every navigation. Like you can read from commit messages I chose to explicitly allow multiple filters as a feature and disallow archive toggle here.

resources/translations/en.edn
Macroz: I don't see the need for premature duplication of the translation texts. We don't intend to normalize / dedupe them anyway for other things.

luontola: I mean that `:t.administration/change-form-intro` is not a generic text used on many administration page, but specific to the change-form page, so a renaming it to `:t.change-form/intro` would be more cohesive.

luontola: Likewise `:t.administration/change-form` could be renamed to `:t.change-form/title`

src/clj/rems/api/catalogue_items.clj
luontola: Ok.

src/cljs/rems/administration/catalogue_items.cljs
luontola: Please add a comment which says that.

luontola: Clicking the resource/form/workflow links also selects the row, so each of those would also need the same treatment. Or then in the click handler for the row check what was the target element of the click event, and only select the row if it was a `td` element. Or then just rely on the checkboxes for row selection.

src/cljs/rems/administration/catalogue_items.cljs
luontola: Something like `filter-items-by-ids` or `select-items-with-ids` would make it more obvious what the function does.

src/cljs/rems/administration/change_catalogue_item_form.cljs
```diff
@@ -0,0 +1,140 @@
...
+
+(rf/reg-sub ::forms (fn [db _] (::forms db)))
+
+(defn- form-change-loop [items form]
+  (let [item (first items)]
+    (cond (empty? items)
+          (flash-message/show-default-success! :top [text :t.administration/change-form])
+
+          (not= (:formid item) (:form/id form))
+          (rf/dispatch [::change-catalogue-item-form! (:id item) form (partial form-change-loop (rest items) form)])
```
luontola: `(partial ` could be just `#(`, especially since all the arguments are already defined here, so it's not the typical case of partial application.

src/cljs/rems/administration/change_catalogue_item_form.cljs
```diff
@@ -0,0 +1,140 @@
...
+
+          :else (recur (rest items) form))))
+
+(defn- all-items-have-the-form-already? [items form]
+  (every? (comp #{(:form/id form)} :formid) items))
```
luontola: I'm having a hard time understanding this line of code, since it uses a set as a function together with comp, both of which are complicated enough to understand alone.

luontola: If I understand correctly, this could be refactored to `#(= (:form/id form) (:formid %))`

src/cljs/rems/administration/change_catalogue_item_form.cljs
```diff
@@ -0,0 +1,140 @@
...
+ (fn [[catalogue language] _]
+   (map (fn [item]
+          {:key (:id item)
+           :name {:value (get-localized-title item language)}
+           :form (let [value (:form-name item)]
+                   {:value value
+                    :td [:td.form
+                         [atoms/link nil
+                          (str "/administration/forms/" (:formid item))
+                          value]]})})
```
luontola: A better name for the `value` variable would be `form-name`.

src/cljs/rems/table.cljs
```diff
@@ -259,9 +267,63 @@
...
+ ::selected-rows
+ (fn [db [_ table]]
+   (get-in db [::selected-rows (:id table)])))
+
+(rf/reg-event-db
+ ::set-selected-rows
+ (fn [db [_ table rows]]
+   (let [new-db (assoc-in db [::selected-rows (:id table)] (set (map :key rows)))]
+     (when-let [on-select (:on-select table)]
+       (on-select (get-in new-db [::selected-rows (:id table)])))
```
luontola: Instead of looking up the selected rows from `new-db` right after putting them there, they could be saved to a local variable for reuse: `(let [selected-rows (set (map :key rows))]`

resources/translations/en.edn
Macroz: I will move the intro, that text is actually leftover from earlier time when it was still on the same page. I think we often use the more general texts in the specific sub-pages than vice versa so I think I will leave the latter to admin.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: I think I'll try improving the row click handler. That way the row selection click handling does not propagate to table users.

src/cljs/rems/administration/change_catalogue_item_form.cljs
Macroz: That is not the meaning of partial to me but I'll do the change.

src/cljs/rems/administration/change_catalogue_item_form.cljs
Macroz: Using a set for this is idiomatic usage in Clojure.  I think using `comp` is also idiomatic here. However the alternative is also pretty clear so I'll change it.

src/cljs/rems/table.cljs
Macroz: Yes, maybe not worth the change but I did it anyway.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: The click handler now checks that the target was a `td`, `th` or `tr` element which I think is good enough for this purpose as the default action.

resources/translations/en.edn
```diff
@@ -185,6 +187,8 @@
              :show "Show"
              :show-less "Show less"
              :show-more "Show more"}
+  :change-form {:change-form-intro "You are about to change the form of the following catalogue items. The changes will be made one at a time, so that each old item is terminated and a new copy with the chosen form will be activated."
```
okahilak: This is referred to as :t.administration/change-form-intro in the code, instead of :t.change-form/change-form-intro, as it's defined here.

resources/translations/fi.edn
```diff
@@ -183,6 +185,8 @@
              :show "Näytä"
              :show-less "Näytä vähemmän"
              :show-more "Näytä lisää"}
+  :change-form {:change-form-intro "Olet muuttamassa seuraavien aineistojen lomaketta. Muutokset tehdään yksi kerrallaan siten, että vanha aineisto lopetetaan ja uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake."
```
okahilak: Could it say "poistetaan käytöstä" instead of "lopetetaan"? I think it would be more aligned with the other terminology.

okahilak: Also, how about "uuteen aktivoitavaan kopioon laitetaan valitsemasi lomake" -> "uusi aktivoitava kopio käyttää valitsemaasi lomaketta"? I think using "laittaa" in the above meaning still has a somewhat casual style, so it's not very formal language.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -60,12 +61,30 @@
           :error-handler (flash-message/default-error-handler :top description)})
    {}))
 
-(defn- to-create-catalogue-item []
+(rf/reg-event-db
+ ::set-selected-items!
```
okahilak: nit: I checked the codebase for other similar cases and it looks like it's not customary in our codebase to use exclamation mark with these reg-event-db setters, so you might want to align this with that convention.

src/cljs/rems/administration/catalogue_items.cljs
okahilak: I have to say that I agree with Esko here. :) With its current name, I was expecting the return value to be a map with ids as keys.

okahilak:
Hmm, not sure if this is intended or a bug, but I couldn't figure out the logic behind this behavior:

1. First, no items are selected:

image

2. After I select one resource:

image

3. After I write something to the search field:

image

4. After I press the "minus", it looks like this:

image

5. After I press the "check" repeatedly, it stays looking like this:

image

okahilak:
Also, a simpler case that seems to work unintuitively:

1. An empty search:

image

2. After pressing the checkbox:

image

3. Repeatedly pressing the checkbox keeps it checked:

image

4. Emptying the search field removes the tick mark:

image

5. Writing something to the search field that empties the search results adds the tick mark back:

image

src/cljs/rems/administration/change_catalogue_item_form.cljs
```diff
@@ -0,0 +1,140 @@
...
+ (fn [db [_ old-catalogue-item-id new-catalogue-item-id new-form]]
+   (update db ::catalogue-items (fn [items]
+                                  (for [item items]
+                                    (if (= (:id item) old-catalogue-item-id)
+                                      (assoc item :id new-catalogue-item-id :formid (:form/id new-form) :form-name (:form/title new-form))
+                                      item))))))
+
+
+(rf/reg-event-fx
+ ::change-catalogue-item-form!
```
okahilak: Also here, it seems that the current convention in the code is to not have the exclamation mark in the event name, even when the effect calls a clojure function with side effects (such as post! here).

okahilak:
Two more comments here:

- I'd like to see "Back" button on "Change Form" page, for consistency with other pages, and to get more easily back to "Catalogue" page (it took me one or two seconds to figure out that I have to press "Catalogue" again on the tab listing to get back to Catalogue page => a bit too long).

- When having several archived Catalogue items after changing forms and I sort by resource name, the table looks like this:

image

You might want to consider having active catalogue item first on that list when there's a tie. It's kind of minor, not necessarily directly related to this patch, but I don't see anyone doing it later on either if it's not done now.

Before this patch, there were more rarely ties between entries, so subtleties of sorting mattered less.

src/cljs/rems/administration/catalogue_items.cljs
Macroz: Makes sense

Macroz:
> Two more comments here:
> 
> * I'd like to see "Back" button on "Change Form" page, for consistency with other pages, and to get more easily back to "Catalogue" page (it took me one or two seconds to figure out that I have to press "Catalogue" again on the tab listing to get back to Catalogue page => a bit too long).
> * When having several archived Catalogue items after changing forms and I sort by resource name, the table looks like this:
> 
> image

Added the button.

> 
> You might want to consider having active catalogue item first on that list when there's a tie. It's kind of minor, not necessarily directly related to this patch, but I don't see anyone doing it later on either if it's not done now.
> 
> Before this patch, there were more rarely ties between entries, so subtleties of sorting mattered less.

I added a tie breaker by reverse sorting by created so that latest is first. That is essentially the same as active but handles the history better.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -78,7 +98,9 @@
  (fn [[catalogue language] _]
    (map (fn [item]
           {:key (:id item)
-           :name {:value (get-localized-title item language)}
+           :name {:value (get-localized-title item language)
+                  :sort-value [(get-localized-title item language)
+                               (- (time-coerce/to-long (:start item)))]} ; secondary sort by created, reverse
```
luontola: This secondary sort doesn't work when sorting by resource, form or workflow.

luontola:
The colspan for the "show all" row is off by one. It's visible when hovering the row with the mouse.

Screenshot 2019-12-03 12 16 42

okahilak:
The patch is fine by me now, thanks!

src/cljs/rems/administration/catalogue_items.cljs
Macroz: Secondary support was not built to the table and this is only to improve the functionality of the default sort in this (arguably common) case. I'm not going to implement secondary sort to the table here.

PR #1824, opened 2019-11-25
refactor: generic outbox

opqdonut:
for #1784

### DIFF
--- a/src/clj/rems/db/outbox.clj
+++ b/src/clj/rems/db/outbox.clj
@@ -50,8 +50,15 @@
       coerce-outboxdata
       (assoc :outbox/id (:id row))))
 
-(defn- next-attempt-now? [entry]
-  (. (:outbox/next-attempt entry) isBefore (DateTime/now)))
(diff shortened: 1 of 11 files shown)


src/clj/rems/db/outbox.clj
```diff
@@ -0,0 +1,160 @@
...
+                                  (json/coercion-matcher schema)))))
+
+(defn- fix-row-from-db [row]
+  (-> (:outboxdata row)
+      json/parse-string
+      coerce-outboxdata
+      (assoc :outbox/id (:id row))))
+
+(defn- next-attempt-now? [entry]
+  (. (:outbox/next-attempt entry) isBefore (DateTime/now)))
```
luontola: Next attempt can be nil. It happens when deadline is reached. Possible NPE here?

Also here is an off by one race condition, because the comparison is < instead of <=, so it would not match a newly created entry where next attempt = now.  Adding one millisecond to now or using "not after" would fix it.

Issue #1830, opened 2019-11-26
new workflow for findata


opqdonut:
split out from #1806

Implement a new workflow that is like the dynamic workflow
  - except approve is done by the decider (instead of handler)
  - deciders should probably be listed in the workflow (just like handlers)

Questions to answer:
  - can we reuse command/event implementations from dynamic workflow?
  - how much duplicated code will there be?
  - what should the API look like? do we have one set of commands or a set per workflow?

Tentative task list:
- [x] backend implementation for new workflow
- [x] workflow wired up to API
- [ ] feature flag to hide new workflow from production
- [x] application can be submitted & approved from UI
- [x] creating workflow via administration UI

PR #1838, opened 2019-12-02
POC: Another workflow

luontola:
#1830

How does this look? Ideas on what direction to move forward?


luontola:
I discusssed this with @opqdonut. I'll make another POC with a different approach (one workflow + an option to disable commands), so that we can compare the approaches.

Issue #1829, opened 2019-11-26
server-side pdf generation


opqdonut:
split out from #1794

- [x] generate application pdfs on the server
  - probably best to use a headless browser and reuse the current print styles
- [x] serve pdfs from an API
- let's think about caching pdfs later, now they can be generated synchronously

opqdonut:
this was probably the wrong thing to do but it's done now, see #1794 :)

PR #1839, opened 2019-12-02
refactor authentication

opqdonut:
related to #1829


src/clj/rems/db/users.clj
```diff
@@ -47,7 +47,8 @@
 
   You should use get-user instead."
```
okahilak: This line could be updated now that there are direct users of get-user-attributes to better reflect who should use get-user.

PR #1843, opened 2019-12-04
feat: remove LDAP authentication

opqdonut:

PR #1844, opened 2019-12-04
update auth configuration docs, don't expose oidc secret in api

opqdonut:
~on top of #1843 for practical reasons~ rebased

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,11 +11,21 @@ The full list of available configuration options can be seen in [config-defaults
 Currently supported authentication methods are SAML2 and OpenId Connect (e.g. Auth0). Login method to be used can be defined with the key `:authentication` and the following values are recognized:
 
 * `:shibboleth` for SAML2
+* `:oidc` for OpenId Connect
 * `:fake-shibboleth` for development login
 
 ### SAML2 (`:shibboleth`)
 
-When using this option, login requests are directed to `/Shibboleth.sso/Login`.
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -11,11 +11,21 @@ The full list of available configuration options can be seen in [config-defaults
...
+When using this option, login requests are directed to `/Shibboleth.sso/Login`. A link to an alternative login endpoint can also be shown if the `:alternative-login-url` option is set.
+
+### OpenId Connect (`:oidc`)
+
+The `:oidc` authentication method has the following configuration options:
+
+* `:oidc-domain` – the openid connect configration is fetched from `https://{oidc-domain}/.well-known/openid-configuration`
+* `:oidc-client-id`
+* `:oidc-client-secret`
+* `:public-url` - the redirect uri sent to the openid endpoint is `{:public-url}/oidc-callback`
```
luontola: For consistency remove or add ":" in both `{oidc-domain}` and `{:public-url}`

src/clj/rems/api/public.clj
```diff
@@ -61,4 +61,6 @@
       (ok (assoc env
                  :authentication "HIDDEN"
                  :database-url "HIDDEN"
-                 :test-database-url "HIDDEN")))))
+                 :test-database-url "HIDDEN"
+                 :oidc-client-id "HIDDEN"
```
luontola: The client ID is public information, so it doesn't need to be hidden. It's present in the login redirect.

opqdonut: right! fixed

docs/configuration.md
opqdonut: fixed

Issue #1236, opened 2019-05-09
Other: Dockerize REMS


opqdonut:
after #1235 we can easily dockerize rems

preliminary task list, probably needs refining:

- [x] write Dockerfile for REMS
  - do we want a reverse proxy in the image? nginx instead of apache?
- [x] use java 11 in docker
  - we're using it for development already
- [x] build docker images in CI
- [x] publish docker images to a repository
  - latest build in master branch
  - versioned release builds
- [x] injecting config.edn, theme, etc.
  - build customized images with the theme bundled in?
- [x] run docker image on remsdev
  - do we terminate https inside the docker image or on a reverse proxy running on the host?
  - Rahti will handle https, the container may rely only on http
- [x] run postgres on remsdev, rems2demo etc. in docker also
  - db connections to external db cluster are available

PR #1853, opened 2019-12-10
fix redirects to relative urls

opqdonut:
we need to prefix relative redirects with :public-url instead of
letting jetty prefix them with the request url

related to #1236


src/clj/rems/auth/fake_shibboleth.clj
```diff
@@ -3,6 +3,7 @@
             [compojure.core :refer [GET defroutes]]
             [hiccup.page :refer [html5]]
             [hiccup.util :refer [url]]
+            [rems.config :refer [env]]
```
luontola: Unused require

Issue #1851, opened 2019-12-10
Updating items in admin menu takes effect with delay


okahilak:
One example of such update is adding a user to an existing workflow.

The delay is up to an hour, due to application cache being reset once in an hour.

Updating at least the following should trigger cache invalidation: catalogue item, workflow, blacklist. To lean on the safe side, all update operations could trigger cache invalidation.

PR #1854, opened 2019-12-10
Refactor blacklist by adding rems.api.services.blacklist

okahilak:
Part of #1851 

## Reviewability
- [x] link to issue

### DIFF
--- a/src/clj/rems/api/blacklist.clj
+++ b/src/clj/rems/api/blacklist.clj
@@ -2,7 +2,7 @@
   (:require [clj-time.core :as time]
             [compojure.api.sweet :refer :all]
             [rems.api.schema :as schema]
-            [rems.db.blacklist :as blacklist]
+            [rems.api.services.blacklist :as blacklist]
             [rems.db.users :as users]
             [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer [ok]]
@@ -10,8 +10,8 @@
   (:import [org.joda.time DateTime]))
 
 (s/defschema BlacklistCommand
-  {:blacklist/resource {:resource/ext-id blacklist/ResourceId}
-   :blacklist/user {:userid blacklist/UserId}
+  {:blacklist/resource {:resource/ext-id s/Str}
+   :blacklist/user {:userid schema/UserId}
    :comment s/Str})
 
 (s/defschema BlacklistEntryWithDetails
@@ -27,21 +27,6 @@
    :blacklist/comment (:event/comment entry)
    :blacklist/added-by (users/get-user (:event/actor entry))})
 
-(defn- command->event [command user-id]
-  {:event/actor user-id
-   :event/time (time/now)
-   :userid (get-in command [:blacklist/user :userid])
-   :resource/ext-id (get-in command [:blacklist/resource :resource/ext-id])
-   :event/comment (:comment command)})
-
-(defn- add! [user-id command]
-  (blacklist/add-event! (-> (command->event command user-id)
-                            (assoc :event/type :blacklist.event/add))))
-
-(defn- remove! [user-id command]
-  (blacklist/add-event! (-> (command->event command user-id)
-                            (assoc :event/type :blacklist.event/remove))))
-
 (def blacklist-api
   (context "/blacklist" []
     :tags ["blacklist"]
@@ -49,8 +34,8 @@
     (GET "/" []
       :summary "Get blacklist entries"
       :roles #{:handler :owner :reporter}
-      :query-params [{user :- blacklist/UserId nil}
-                     {resource :- blacklist/ResourceId nil}]
+      :query-params [{user :- schema/UserId nil}
+                     {resource :- s/Str nil}]
       :return [BlacklistEntryWithDetails]
       (->> (blacklist/get-blacklist {:userid user
                                      :resource/ext-id resource})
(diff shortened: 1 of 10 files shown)


src/clj/rems/application/events.clj
```diff
@@ -2,54 +2,59 @@
   (:require [clojure.test :refer :all]
             [schema-refined.core :as r]
             [schema.core :as s]
+            [rems.api.schema :refer [EventBase]]
```
opqdonut: I realise the need for code sharing but I'm not sure I like rems.application depending on rems.api. Maybe rems.schema would be a good place for "fundamental" schema definitions?

The reason I think there should be no dependency is that we want to keep the API stable, but we are free to change our internal representations as long as we produce the same API.

src/clj/rems/api/services/blacklist.clj
```diff
@@ -0,0 +1,51 @@
...
+
+(defn- check-foreign-keys [event]
+  ;; TODO: These checks could be moved to the database as (1) constraint checks or (2) fields with foreign keys.
+  (when-not (users/user-exists? (:userid event))
+    (throw (IllegalArgumentException. "user doesn't exist")))
+  (when-not (resource/ext-id-exists? (:resource/ext-id event))
+    (throw (IllegalArgumentException. "resource doesn't exist")))
+  event)
+
+(defn- command->event [command user-id]
```
opqdonut: the `user-id` parameter should probably be called `actor` here and elsewhere, otherwise it sounds like it is the user that is added to the blacklist

opqdonut:
discussed the separation of concerns between rems.db.blacklist and rems.api.services.blacklist on chat

src/clj/rems/api/services/blacklist.clj
okahilak: Done.

src/clj/rems/application/events.clj
okahilak: Done.

okahilak:
> discussed the separation of concerns between rems.db.blacklist and rems.api.services.blacklist on chat

Thanks, I think it's a lot better now.

src/clj/rems/db/blacklist.clj
```diff
@@ -39,6 +36,10 @@
   (s/validate BlacklistEvent event)
   (json/generate-string event))
 
+(defn- event-from-db [event]
+  (assoc (json->event (:eventdata event))
+         :event/id (:event/id event)))
+
 (defn- check-foreign-keys [event]
```
opqdonut: as I mentioned in my previous review, this function could be on the service level to reduce the dependencies of the db level. not insisting you move it though.

src/clj/rems/api/services/command.clj
```diff
@@ -27,11 +27,10 @@
     (when (= :application.event/revoked (:event/type event))
       (let [application (applications/get-unrestricted-application (:application/id event))]
         (doseq [resource (:application/resources application)]
-          (doseq [user (application-util/applicant-and-members application)]
-            (blacklist/add-to-blacklist! {:userid (:userid user)
-                                          :resource/ext-id (:resource/ext-id resource)
-                                          :actor (:event/actor event)
-                                          :comment (:application/comment event)}))))))
+          (blacklist/add-users-to-blacklist! {:users (application-util/applicant-and-members application)
```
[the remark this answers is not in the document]
okahilak: Right, so the loop is moved to service level in preparation for resetting the application cache after adding all the users to the blacklist. I'd like make sure that it is the service level that resets the cache, and it's probably best to reset the cache only after the loop is finished, thus the need to have the loop there, as well.

src/clj/rems/api/blacklist.clj
```diff
@@ -2,16 +2,16 @@
...
             [rems.api.schema :as schema]
-            [rems.db.blacklist :as blacklist]
+            [rems.api.services.blacklist :as blacklist]
             [rems.db.users :as users]
             [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer [ok]]
             [schema.core :as s])
   (:import [org.joda.time DateTime]))
 
 (s/defschema BlacklistCommand
```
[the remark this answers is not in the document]
okahilak: I think it's slightly more consistent with how defining the schemas is done in other rems.api.* namespaces if the schema for the command is defined here (for example, rems.api.workflows defines CreateWorkflowCommand, EditWorkflowCommand, and CreateWorkflowResponse schemas).

src/clj/rems/db/blacklist.clj
okahilak: Overlooked that, sorry. Yeah, that's a good idea. Done.

okahilak: Okay hm, it seems that part of the tests (namely, test-parameter-validation), currently located in rems.db.test-blacklist, would then have to be moved to rems.api.services.test-blacklist, which doesn't exist currently. I think it's cleaner to have all the tests in one place, and it's also good to have the check close to the database level, especially since we might want to eventually do that in the database (see the comment in check-foreign-keys function).

For these reasons, I ended up keeping the function where it is in this patch. Thanks for the suggestion, though.

okahilak:
Good suggestions, even though I didn't end up implementing any of them this time. I'll merge.

src/clj/rems/api/services/command.clj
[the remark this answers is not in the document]
opqdonut: fair enough!

PR #1856, opened 2019-12-12
Clear application cache after editing workflow, catalogue-item or blacklist

### LINKED ISSUE #1851: Updating items in admin menu takes effect with delay

okahilak:
Closes #1851 

## Reviewability
- [x] link to issue

## Documentation

### DIFF
--- a/test/clj/rems/api/test_catalogue_items.clj
+++ b/test/clj/rems/api/test_catalogue_items.clj
@@ -2,6 +2,7 @@
   (:require [clojure.string :as str]
             [clojure.test :refer :all]
             [rems.api.testing :refer :all]
+            [rems.db.applications :as applications]
             [rems.db.test-data :as test-data]
             [rems.handler :refer [handler]]
             [ring.mock.request :refer :all]))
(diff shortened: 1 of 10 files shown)


test/clj/rems/api/test_catalogue_items.clj
```diff
@@ -90,6 +97,13 @@
                              handler
                              read-ok-body)]
             (is (:success response) (pr-str response))
+            (testing "application is updated when catalogue item is edited"
+              (let [app-id (test-data/create-application! {:catalogue-item-ids [id]
+                                                           :actor "alice"})
+                    get-app #(applications/get-unrestricted-application app-id)]
+                (is (= {:fi "http://info.fi"}
```
okahilak: For some reason, this test doesn't fail even when the cache is not cleared in rems.api.services.catalogue, not sure why.

okahilak: Fixed, I had a bug in the test (it re-created the application after editing catalogue item).

PR #1867, opened 2019-12-16
refactor cljs actions a bit

opqdonut:


src/cljs/rems/actions/approve_reject.cljs
```diff
@@ -18,33 +16,21 @@
...
-            {:params {:application-id application-id
-                      :comment comment}
-             :handler (flash-message/default-success-handler
-                       :actions
-                       description
-                       (fn [_]
-                         (collapse-action-form action-form-id)
-                         (on-finished)))
-             :error-handler (flash-message/default-error-handler :actions description)}))
+   (command! "approve" {:application-id application-id
```
opqdonut: should the command name be a keyword? should we reuse the command names like `:application.command/approve`?

src/cljs/rems/actions/approve_reject.cljs
```diff
@@ -18,33 +16,21 @@
...
+                        :comment comment}
+             {:description [text :t.actions/approve]
+              :collapse action-form-id
```
opqdonut: should we just compose the collapse into on-finished here?

src/cljs/rems/actions/approve_reject.cljs
luontola: We could also have it under a `:type` key, the same way as in the backend.

luontola: Reusing the command names would make it easier to find every place where the command is used.

src/cljs/rems/actions/approve_reject.cljs
luontola: Maybe not. The `command!` is anyways very specific to actions.

PR #1868, opened 2019-12-17
Allow exporting applications as CSV in admin menu

### LINKED ISSUE #1857: Export all applications as csv on admin pages
Add a button to admin page, exporting all applications to CSV.

The CSV file could be structured as follows:

- Each application corresponds to one row,
- Columns contain application metadata (applicant, date, state, etc.) and the form fields.

In a later phase, we can add:

- Export to Excel in addition to (or instead of) CSV
- Allow filtering applications in backend -- in this first phase, the end user can filter the applications as desired using their tool of choice, after importing the CSV.

okahilak:
Closes #1857 

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Accessibility

### DIFF
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -45,6 +45,7 @@
                    :add "Lisää"
                    :added-at "Lisätty"
                    :added-by "Lisääjä"
+                   :applications "Hakemukset"
                    :archive "Arkistoi"
                    :back "Takaisin"
                    :blacklist "Kieltolista"
@@ -79,6 +80,7 @@
                             :resource-in-use "Resurssi on käytössä:"
                             :workflow-archived "Työvuo on arkistoitu:"
                             :workflow-in-use "Työvuo on käytössä:"}
+                   :export-applications "Vie hakemuksia"
                    :form "Lomake"
                    :forms "Lomakkeet"
                    :license "Lisenssi"
@@ -141,6 +143,12 @@
                           :revoked "%1 perui käyttöoikeuden."
                           :submitted "%1 lähetti hakemuksen käsiteltäväksi."
                           :unknown "Tuntematon"}
+                 :export {:applicant "Hakija"
+                          :external-id "Ulkoinen tunniste"
+                          :first-submitted "Lähetetty"
+                          :id "Tunniste"
+                          :resources "Resurssit"
+                          :state "Tila"}
                  :id "Tunniste"
                  :intro [:div.intro [:p "Täytä hakemus ja lähetä se käsiteltäväksi. Hakemukseen voi kutsua jäseniksi useampia hakijoita, joista jokaisen on hyväksyttävä käyttöehdot viimeistään kun hakemus on hyväksytty."]]
                  :last-activity "Muokattu"
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -192,10 +192,10 @@
       :return Commenters
       (ok (users/get-commenters)))
 
-    (GET "/export/:form-id" []
+    (GET "/export" []
       :summary "Export all submitted applications of a given form as CSV"
       :roles #{:owner}
-      :path-params [form-id :- (describe s/Int "form id")]
+      :query-params [form-id :- (describe s/Int "form id")]
       (-> (ok (applications/export-applications-for-form-as-csv (getx-user-id) form-id))
           (header "Content-Disposition" (str "filename=\"" (csv/applications-filename) "\""))
           (content-type "text/csv")))
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -12,6 +12,7 @@
             [rems.auth.util :refer [throw-forbidden]]
             [rems.config :as config]
             [rems.db.applications :as applications]
+            [rems.db.csv :as csv]
             [rems.db.users :as users]
             [rems.pdf :as pdf]
             [rems.util :refer [getx-user-id update-present]]
@@ -190,6 +191,14 @@
       :roles #{:handler}
       :return Commenters
       (ok (users/get-commenters)))
+
+    (GET "/export/:form-id" []
+      :summary "Export all submitted applications of a given form as CSV"
+      :roles #{:owner}
+      :path-params [form-id :- (describe s/Int "form id")]
+      (-> (ok (applications/export-applications-for-form-as-csv (getx-user-id) form-id))
+          (header "Content-Disposition" (str "filename=\"" (csv/applications-filename) "\""))
+          (content-type "text/csv")))
 
     (GET "/members" []
       :summary "Existing REMS users available for application membership"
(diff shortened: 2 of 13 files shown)


okahilak:
New admin navbar:

image

Selecting Applications tab:

image

After pressing "Export applications":

image

After selecting form:

image

After pressing "Export applications":

image

src/cljs/rems/administration/export_applications.cljs
```diff
@@ -0,0 +1,93 @@
...
+(rf/reg-sub ::loading? (fn [db _] (::loading? db)))
+
+(defn- cancel-button []
+  [atoms/link {:class "btn btn-secondary"}
+   "/administration/applications"
+   (text :t.administration/cancel)])
+
+(defn- export-button [form-id]
+  ;; Disabling <a> element is done using instructions from Bootstrap docs.
+  (let [disabled? (nil? form-id)]
```
okahilak: Hm, okay, apparently it is good practice to use question mark only with predicates and not boolean variables, will fix.

opqdonut: some people do use it for variables too, especially for parameter names

resources/translations/fi.edn
```diff
@@ -79,6 +80,7 @@
                             :resource-in-use "Resurssi on käytössä:"
                             :workflow-archived "Työvuo on arkistoitu:"
                             :workflow-in-use "Työvuo on käytössä:"}
+                   :export-applications "Vie hakemuksia"
```
opqdonut: Vie sounds weid but I guess it is the typical translation for export

src/clj/rems/db/csv.clj
```diff
@@ -0,0 +1,115 @@
+(ns rems.db.csv
+  "Utilities for exporting database contents as CSV"
+  (:require [clj-time.core :as time]
+            [clojure.string :as str]
+            [rems.config :refer [env]]
+            [rems.db.user-settings :as user-settings]
+            [rems.text :as text]))
+
+(defn print-to-csv [& {:keys [column-names rows
+                              enclose-strings-in-quotes separator]
```
opqdonut: could just call `enclose-strings-in-quotes` something shorter like `quote?` or `quote-strings?`

src/clj/rems/db/csv.clj
```diff
@@ -0,0 +1,115 @@
...
+                       :or {separator (:csv-separator env)}}]
+  (let [escape-quotes #(str/replace % "\"" "\\\"")
```
opqdonut: I'm not sure this gets all quotes right. What if the value already has an escaped quote `\"`? Did you try importing values quoted like this in excel? Not a huge thing, csv doesn't really have a spec so as long as we work on most data it's fine.

test/clj/rems/db/test_csv.clj
```diff
@@ -0,0 +1,112 @@
...
+  
+  (testing "escapes quotes inside strings when enclosed in quotes"
+    (is (= (csv/print-to-csv :column-names ["col1" "col2"]
+                             :rows [["\"so called\" value" "val2"]
+                                    ["val3" "val4"]]
+                             :separator ","
+                             :enclose-strings-in-quotes true)
+           "\"col1\",\"col2\"\n\"\\\"so called\\\" value\",\"val2\"\n\"val3\",\"val4\"\n")))
+
+  (testing "does not escape quotes inside strings when not enclosed in quotes"
```
opqdonut: [nit] one of these tests could change the separator

test/clj/rems/db/test_csv.clj
```diff
@@ -0,0 +1,112 @@
...
+    (is (= (csv/print-to-csv :column-names ["col1" "col2"]
+                             :rows [["\"so called\" value" "val2"]
+                                    ["val3" "val4"]]
+                             :separator ",")
+           "col1,col2\n\"so called\" value,val2\nval3,val4\n"))))
+
+(def ^:private applicant "applicant")
+(def ^:private test-time (DateTime. 1000))
+
+(deftest test-applications-to-csv
```
opqdonut: Btw, this could be a non-integration non-db test if you created the application from events. See something like `rems.application.test-commands/test-submit` for example.

opqdonut: However, a test that queried the API for applications in the db would be valuable.

src/clj/rems/api/applications.clj
```diff
@@ -191,6 +192,14 @@
       :return Commenters
       (ok (users/get-commenters)))
 
+    (GET "/export/:form-id" []
```
opqdonut: if we plan to support other forms of exports, form-id could be a query parameter, so something like `/api/applications/export?form-id=13` instead of `/api/applications/export/13`. Another option would be `/api/applications/export/form/13`.

resources/translations/fi.edn
okahilak: I double-checked and it indeed seems to be typical translation, see, e.g., https://docs.gimp.org/fi/gimp-file-export.html

I think the slight problem here is that there is less context for the verb "Vie" in the UI, compared to, e.g., a file menu containing a similar item "Vie..." where it is easier to interpret correctly.

Another option could be "Muunna hakemuksia", but the meaning would then be slightly different from "Export".

src/clj/rems/api/applications.clj
okahilak: Thanks, that's a good suggestion, especially since it's good to get the API right the first time around.

Decided to use a query parameter instead of defining the form in the URL because if we end up implementing filtering of applications, query parameters can be extended into accepting multiple parameters without "combining styles".

src/cljs/rems/administration/export_applications.cljs
okahilak: Right, it looks like the convention of using question mark also for booleans is used elsewhere in the code as well, so leaving it as it is.

src/clj/rems/db/csv.clj
okahilak: Done.

test/clj/rems/db/test_csv.clj
okahilak: Done.

src/clj/rems/db/csv.clj
okahilak: Thanks for the comment, I did not test importing the output to Excel (or rather, LibreOffice Calc) before, but I did now. Some observations:

- An escaped quote `\"` in the value seems to work fine when importing. It looks like Calc only treats the escape character specially if it directly precedes the text delimiter character (i.e., a quote).

- However, if the value ends with a quote `"`, Calc doesn't interpret it correctly, but instead messes up the subsequent columns. I think it is a problem in Calc because `"string\""` looks to me like it's correctly escaped and interpretable. The problem might be caused by there being an alternative way to escape `"` as `""`, which seems to be implemented at least in R's `write.csv` function. But I don't think this is actually our problem.

- At least Python and R seem to use a similar method of escaping only the text delimiter character as I did here, so this seems to be consistent with how it's done elsewhere.

test/clj/rems/db/test_csv.clj
okahilak: Added a test for API. I did not end up changing the existing test to non-integration, non-db right now, but added a TODO comment about it.

PR #1873, opened 2019-12-20
fix: don't render broken more info links in application

### LINKED ISSUE #1872: Application page has "More info" link on a resource even when the resource does not have "More info" field
E.g.,

image

opqdonut:
... by making urn-catalogue-item-link check that the resid is a urn.

fixes #1872

### DIFF
--- a/src/cljc/rems/catalogue_util.cljc
+++ b/src/cljc/rems/catalogue_util.cljc
@@ -1,9 +1,9 @@
 (ns rems.catalogue-util
   (:require [clojure.string :as str]))
 
-(defn urn-catalogue-item? [{:keys [resid]}]
+(defn urn? [resid]
   (and resid (str/starts-with? resid "urn:")))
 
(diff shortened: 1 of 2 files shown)


src/cljc/rems/catalogue_util.cljc
```diff
@@ -1,9 +1,9 @@
 (ns rems.catalogue-util
   (:require [clojure.string :as str]))
 
-(defn urn-catalogue-item? [{:keys [resid]}]
-  (and resid (str/starts-with? resid "urn:")))
+(defn urn? [resid]
```
okahilak: nit: This could be private now.

opqdonut: yeah, but it's a potentially useful util elsewhere too

opqdonut:
guide broke, made `urn?` tolerate nils

src/cljc/rems/catalogue_util.cljc
okahilak: Ok, fine by me to merge.

PR #1874, opened 2019-12-23
e2e test for decider workflow

luontola:
Follow up for #1830/#1855

### DIFF
--- a/test/clj/rems/api/test_applications.clj
+++ b/test/clj/rems/api/test_applications.clj
@@ -501,6 +501,31 @@
              (send-command user-id {:type :application.command/submit
                                     :application-id app-id}))))))
 
+(deftest test-decider-workflow
+  (let [applicant "alice"
+        handler "handler"
+        decider "carl"
+        wf-id (test-data/create-workflow! {:type :workflow/decider
+                                           :handlers [handler]})
+        cat-id (test-data/create-catalogue-item! {:workflow-id wf-id})
+        app-id (test-data/create-application! {:catalogue-item-ids [cat-id]
+                                               :actor applicant})]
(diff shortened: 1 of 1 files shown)


test/clj/rems/api/test_applications.clj
```diff
@@ -501,6 +501,31 @@
...
+      (is (= {:success true}
+             (send-command applicant {:type :application.command/submit
+                                      :application-id app-id}))))
+    (testing "request decision"
+      (is (= {:success true}
+             (send-command handler {:type :application.command/request-decision
+                                    :application-id app-id
+                                    :deciders [decider]
+                                    :comment ""}))))
+    (testing "approve"
```
okahilak: Hmm, I found these test cases from test-application-commands above:

```
      (testing "request-decision"
        (is (= {:success true} (send-command handler-id
                                             {:type :application.command/request-decision
                                              :application-id application-id
                                              :deciders [decider-id]
                                              :comment ""}))))
      (testing "decide"
        (is (= {:success true} (send-command decider-id
                                             {:type :application.command/decide
                                              :application-id application-id
                                              :decision :approved
                                              :comment ""}))))
````

In this test you added, how come the decider is actually allowed to approve the application? I thought that the role of the decider is to give a decision, which is shown to the handler who can then approve the application. Can you clarify a bit?

luontola: The difference is `(test-data/create-workflow! {:type :workflow/decider` - this is a different workflow where only the decider can approve/reject applications.

okahilak: Thanks, that clarifies, I was a bit hasty in reading the code.

Issue #1611, opened 2019-09-11
Weekly reminder email to handlers


opqdonut:
split out from #1366

PR #1877, opened 2019-12-23
Handler reminder emails

luontola:
#1611

Is still missing the scheduling mechanism. How about an API which is called from cron with API key?

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up


src/clj/rems/email/template.clj
```diff
@@ -193,3 +193,20 @@
                (text :t.email/footer))}])))
 
 ;; TODO member-joined?
+
+(defn handler-reminder-email [lang handler applications]
+  (with-language lang
```
okahilak: The function emails-to-recipients above does the language selection like this:

`(with-language (:language (user-settings/get-user-settings (:userid recipient)))`

Maybe this function could follow a similar template.

src/clj/rems/email/core.clj
```diff
@@ -20,12 +22,23 @@
...
-    (outbox/put! {:outbox/type :email
-                  :outbox/email email
-                  :outbox/deadline (-> (time/now) (.plus ^Period (:email-retry-period env)))})))
+    (enqueue-email! email)))
+
+;; TODO: scheduler or API+cron
+(defn generate-handler-reminder-emails! []
+  (doseq [email (->> (workflow/get-handlers)
+                     (map (fn [handler]
+                            (template/handler-reminder-email :en handler (todos/get-todos (:userid handler)))))
```
okahilak: Hard-coded language

test/clj/rems/email/test_template.clj
```diff
@@ -360,3 +360,38 @@
                 :subject "(2001/3, \"Application title\") Uusi hakemus"
                 :body "Hyvä Hannah Handler,\n\nAlice Applicant on lähettänyt käyttöoikeushakemuksen 2001/3, \"Application title\" resurss(e)ille fi title 11, fi title 21.\n\nVoit tarkastella hakemusta osoitteessa http://example.com/application/7\n\nTämä on automaattinen viesti. Älä vastaa."}
                (email-to "handler" mails)))))))
+
+(deftest test-handler-reminder-email
+  (with-redefs [rems.config/env (assoc rems.config/env :public-url "http://example.com/")
+                user-settings/get-user-settings (fn [userid]
```
okahilak: get-user-settings currently not used in the code paths run by this test.

resources/translations/fi.edn
```diff
@@ -281,6 +281,9 @@
           :decision-requested {:subject "(%3) Päätöspyyntö"
                                :message "Hyvä %1,\n\n%2 on pyytänyt sinulta päätöstä käyttäjän %4 hakemuksesta %3.\n\nVoit tarkastella hakemusta osoitteessa %6"}
           :footer "\n\nTämä on automaattinen viesti. Älä vastaa."
+          :handler-reminder {:application "%1, lähettänyt %2"
+                             :message "Hyvä %1,\n\nSeuraavat hakemukset ovat vaiheessa ja voivat vaatia toimiasi:\n\n%2\n\nVoit tarkastella hakemuksia osoitteessa %3"
```
okahilak: This message could have a slightly more formal tone, e.g., "Seuraavien hakemusten käsittely on kesken, ja ne saattavat vaatia toimenpiteitä:"

resources/translations/fi.edn
```diff
@@ -281,6 +281,9 @@
...
+                             :subject "Hakemuksia vaiheessa"}
```
okahilak: Similarly, I think "Hakemusten käsittely kesken" would be more formal here, although admittedly a bit long. But I'd prefer that in any case.

src/clj/rems/email/core.clj
```diff
@@ -20,12 +22,23 @@
...
+
+(defn generate-event-emails! [new-events]
   (doseq [event new-events
           email (event-to-emails event)]
-    (outbox/put! {:outbox/type :email
-                  :outbox/email email
-                  :outbox/deadline (-> (time/now) (.plus ^Period (:email-retry-period env)))})))
+    (enqueue-email! email)))
+
+;; TODO: scheduler or API+cron
```
okahilak: Hmm, I think there's potentially a fundamental problem here: I'd imagine that we wouldn't want to send out a reminder email for an application filed on the previous day, but having a weekly cron job or similar doesn't distinguish between actually old and recently filed applications.

Having a reminder email of all open applications at a particular moment also makes it difficult to see, which of the applications have been open long enough so that they warrant extra attention.

Also, if we had a filter here that would filter out recent enough applications and only leave applications older than, say, one week, depending on when the scheduler or cron job is run relatively to the time when the application was submitted, it could be take one week and six days before a reminder email for that application is sent.

I think the best solution would be to send out reminder emails on-application basis, one email per each application that is older than the specified time limit.

okahilak: Never mind, Leena enlightened me that this weekly reminder email for all open applications is exactly how it has been specified.

Resolving the conversation.

resources/translations/fi.edn
okahilak: Hmm, after thinking a bit, I'm not happy with the above proposal either -- how about "Keskeneräiset hakemukset"? I think it would have a close correspondence with the English translation ("Applications in progress").

okahilak: Okay, this came from Leena: "Avoimet hakemukset", probably the best of these proposals.

resources/translations/fi.edn
okahilak: Similarly here, following Leena's proposal, it could be "Seuraavat hakemukset ovat avoinna ja saattavat vaatia toimenpiteitä:"

PR #1880, opened 2019-12-30
Review reminder email

### LINKED ISSUE #1860: Reminder emails for outstanding review requests
- [ ] every 7 days
- probably need to make configurable
- related to #1611

luontola:
Closes #1860

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ Changes since v2.6
 - Reminder email about open applications to all handlers. The emails can
   be sent by calling `/api/email/send-handler-reminder` on a cron schedule
   or similar. The API requires an API key. (#1611)
+- Reminder email about applications with open review requests to reviewers.
+  The emails can be sent by calling `/api/email/send-reviewer-reminder` on a
+  cron schedule or similar. The API requires an API key. (#1860)
 
 ### Enhancements
 - Application search tips hidden behind question mark icon (#1767)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,11 @@ Changes since v2.6
   the same as before. See #837
 - Applications can be exported as CSV in admin menu (#1857)
 - Added a configuration option for setting a maximum number of days for handling a new application (#1861)
-- Reminder email about open applications to all handlers. The emails can
-  be sent by calling `/api/email/send-handler-reminder` on a cron schedule
-  or similar. The API requires an API key. (#1611)
-- Reminder email about applications with open review requests to reviewers.
-  The emails can be sent by calling `/api/email/send-reviewer-reminder` on a
-  cron schedule or similar. The API requires an API key. (#1860)
+- Added reminder emails. The emails can be sent by calling one of the following
+  APIs on a cron schedule or similar. The APIs require an API key. (#1611, #1860)
+  - `/api/email/send-handler-reminder` sends email about open applications to all handlers.
+  - `/api/email/send-reviewer-reminder` sends email about applications with open review requests to reviewers.
+  - `/api/email/send-reminders` sends all of the above emails.
 
 ### Enhancements
 - Application search tips hidden behind question mark icon (#1767)
--- a/src/clj/rems/email/core.clj
+++ b/src/clj/rems/email/core.clj
@@ -36,8 +36,20 @@
 (defn generate-handler-reminder-emails! []
   (doseq [email (->> (workflow/get-handlers)
                      (map (fn [handler]
-                            (let [lang (:language (user-settings/get-user-settings (:userid handler)))]
-                              (template/handler-reminder-email lang handler (todos/get-todos (:userid handler))))))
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ Changes since v2.6
 - Reminder email about open applications to all handlers. The emails can
   be sent by calling `/api/email/send-handler-reminder` on a cron schedule
   or similar. The API requires an API key. (#1611)
+- Reminder email about applications with open review requests to reviewers.
```
okahilak: I have a problem with the terminology here. The difference between 'reviewer' and 'handler' is probably not obvious to a user that is not completely familiar with REMS before. Also, offhand I would assume that when an application is submitted, it would open a review request for the handlers, but apparently the semantics of 'review request' is such that it is opened only if the original handler of the application explicitly makes a review request to someone who is not marked as a handler in the workflow.

I think those semantics are unnecessarily complicated, but I would need to think a bit to come up with better alternatives.

(For what it's worth, when looking in Github at the issue which this patch closes, I didn't quite understand what review request refers to, so I'm pretty sure it could be likewise confusing for an average user.)

src/clj/rems/email/core.clj
```diff
@@ -36,8 +36,20 @@
...
+                     (remove nil?))]
+    (enqueue-email! email)))
+
+(defn generate-reviewer-reminder-emails! []
+  (doseq [email (->> (applications/get-users-with-role :commenter) ; TODO: rename commenter to reviewer
+                     (map users/get-user)
+                     (map (fn [reviewer]
+                            (let [lang (:language (user-settings/get-user-settings (:userid reviewer)))
+                                  apps (->> (todos/get-todos (:userid reviewer))
+                                            (map #(= :waiting-for-your-review (:application/todo %))))]
```
okahilak: How about :waiting-for-your-decision, should it be included here, as well?

src/clj/rems/email/template.clj
```diff
@@ -210,3 +210,20 @@
...
+    (fn []
+      (when (not (empty? applications))
+        (let [list (->> applications
+                        (map (fn [application]
+                               (text-format :t.email.reviewer-reminder/application
+                                            (format-application-for-email application)
+                                            (application-util/get-member-name (:application/applicant application)))))
+                        (str/join "\n"))]
+          {:to-user (:userid reviewer)
+           :subject (text :t.email.reviewer-reminder/subject)
```
okahilak: minor: It seems that this new function only differs from the function handler-reminder-email such that :t.email.handler-reminder is changed to :t.email.reviewer-reminder. Maybe it could be given as a parameter to the function so less duplication would be needed?

test/clj/rems/api/test_email.clj
```diff
@@ -39,3 +40,42 @@
...
+(deftest test-send-reviewer-reminder
+  (let [api-key "42"
+        user-id "developer"
+        outbox-emails (atom [])
+
+        app-id (test-data/create-application! {:actor "alice"})]
+    (test-data/command! {:type :application.command/submit
+                         :application-id app-id
+                         :actor "alice"})
+    (test-data/command! {:type :application.command/request-comment
```
okahilak: Likewise here, even if request-comment is changed to request-review, it is still difficult to map the assumed semantics of the word to the conceptual difference between request-review and request-decision.

(I want to emphasize that it's by no means a problem introduced by this patch, but this patch kind of builds on the existing problem, so that's why I'm bringing it up here.)

At least, there could be a TODO comment here for changing the name of the command to request-review.

CHANGELOG.md
luontola: There are "Request review" and "Request decision" buttons in the UI, so they should be terms that the users are acquainted with.

src/clj/rems/email/core.clj
luontola: Maybe, but nobody has yet asked for an email reminder for deciders. Some sort of deduplication could then be needed.

src/clj/rems/email/template.clj
luontola: I'm waiting for the third duplicated copy before trying to deduplicate them.

test/clj/rems/api/test_email.clj
luontola: There's a comment about it in rems.application.events/CommentedEvent

luontola: Created #1887

okahilak:
If we're not sure about the terminology or the roles of reviewers/deciders/handlers in the future or for all instances, would it be more future-proof to use a single API call to send all reminder emails?

luontola:
I'll add an API that sends all reminders. I'll still keep the individual APIs to simplify testing and in case somebody wants more control in what messages are sent when.

Issue #1881, opened 2019-12-31
Add 'email address' type of form field


okahilak:
This is needed for Findata, which does not get applicants' emails from IdP.

- If IdP does not provide the email address for the applicant, use the email address specified in the form field for notifying the applicant.

- If the form does not have 'email address' field and IdP does not provide the email address, do not send email notifications to the applicant.

okahilak:
It turned out that there are quite many ways to specify the details of this feature:

- If IdP provides email address and it is asked for in the form, which one is used, or are notifications sent to both?
- If form has multiple email address fields, which one is used, or are notifications sent to all of them?
- Can email field be optional?
- Findata form asks for "Sähköpostiosoite" (required field) and "Yhteyshenkilön sähköpostiosoite, jos joku muu kuin hakija" (optional field). How are they going to use the latter field? Should the feature take that into account?
- Should the problem of IdP not providing email address be solved instead by being able to set email address as a part of the user settings? That would provide the benefit of, e.g., the user being able to disable email notifications.

okahilak:
Decided to implement user-settable email address in a separate user setting, see https://github.com/CSCfi/rems/issues/1884. Therefore, closing this issue.

PR #1885, opened 2020-01-02
Settings page from where users can change their email address

### LINKED ISSUE #1884: Allow users to change their email address
Findata does not get applicants' emails from IdP, so a workaround is to give users the option to set the email address themselves. It would likely be useful also for other users, who want to receive emails to another address.

This should make #1881 unnecessary.

- [x] UI for setting email in user settings
- [x] prefer the email address from user settings when sending notifications
- [x] force/warn users to input an email address, if IdP doesn't give one

luontola:
The email is saved in user settings (same as the user's language).

#1884, makes #1881 unnecessary

settings

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [x] all fields have a label
- [ ] contrast is checked

## Follow-up


src/clj/rems/db/user_settings.clj
```diff
@@ -30,4 +51,4 @@
      {:user user
       :settings (json/generate-string
                  (merge old-settings validated))})
-    {:success (some? validated)}))
+    {:success (not (empty? validated))}))
```
opqdonut: this seems to return `:success true` if e.g. language passes validation but email doesn't

src/clj/rems/email/core.clj
```diff
@@ -74,9 +74,8 @@
         email (assoc email-spec
                      :from (:mail-from env)
                      :to (or (:to email-spec)
-                             (:email
-                              (users/get-user
-                               (:to-user email-spec)))))
+                             (:email (user-settings/get-user-settings (:to-user email-spec)))
+                             (:email (users/get-user (:to-user email-spec)))))
```
opqdonut: hmm, so we use the email from user settings here, but not when rendering user details? I guess that's an ok design choice but it could be made more explicit

src/cljs/rems/settings.cljs
```diff
@@ -0,0 +1,72 @@
...
+     [document-title (text :t.navigation/settings)]
+     [flash-message/component :top]
+     (if (= ::loading form)
+       [spinner/big]
+       [:form
+        {:on-submit (fn [event]
+                      (.preventDefault event)
+                      (rf/dispatch [::save]))}
+
+        (let [id "email"]
```
opqdonut: should we render the language choice here? perhaps read-only?

okahilak:
If we're not completely sure about the terminology or the roles of reviewers/deciders/handlers in the future, would it be more future-proof to use a single API call to send all reminder emails?

src/clj/rems/db/user_settings.clj
luontola: You're right. As long as at least one parameter was valid, it returns true. Currently the API is used so that only one parameter is used at a time, so the problem doesn't come up. Should the function instead give an error and do nothing if even one parameter is invalid?

src/clj/rems/email/core.clj
luontola: Where are we rendering user details?

src/cljs/rems/settings.cljs
luontola: The language choice is already the language in which the page is shown and the selection is highlighted in the navigation menu. I don't think the user needs to know that language selection is persisted using the same method as the email.

src/clj/rems/email/core.clj
opqdonut: e.g. applicant details in applications, selecting users as reviewers, blacklist tables

src/clj/rems/db/user_settings.clj
opqdonut: yeah

src/cljs/rems/settings.cljs
```diff
@@ -0,0 +1,72 @@
...
+          [:div.form-group
+           [:label {:for id} (text :t.settings/email)]
+           [:input.form-control
+            {:type "email"
+             :id id
+             :value (:email form)
+             :placeholder (:email (:user identity))
```
luontola: The email from IDP should be shown explicitly instead of this placeholder.

Issue #1891, opened 2020-01-09
Hide "assign external id" command


opqdonut:
(leftover from #1858)

Not all installations need the assign external id command. Let's add a config option for hiding commands. (We can make hidden commands attributes of the workflow later if needed.)

PR #1900, opened 2020-01-10
feat: disable-commands configuration variable

opqdonut:
Needs rems-deploy changes.

For #1891

## Reviewability
- [x] link to issue

## API

## Documentation

## Different installations

## Testing

## Follow-up

### DIFF
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -95,9 +95,12 @@
  ;; on an application, starting from the first submission.
  :application-deadline-days nil
 
- ;; List of commands to disable. See the swagger schema of
- ;; /api/applications/{id} field application/permissions for possible
- ;; values.
+ ;; List of commands to disable. See the swagger model of
+ ;; /api/applications/{application-id} field application/permissions
+ ;; for all possible values.
+ ;;
+ ;; Example:
+ ;; :disable-commands [:application.command/assign-external-id :application.command/change-resources]
  :disable-commands []
 
  ;; Experimental features
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -95,5 +95,10 @@
  ;; on an application, starting from the first submission.
  :application-deadline-days nil
 
+ ;; List of commands to hide. See the swagger schema of
+ ;; /api/applications/{id} field application/permissions for possible
+ ;; values.
+ :hide-commands []
+
  ;; Experimental features
  :enable-pdf-api false}
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -95,10 +95,10 @@
  ;; on an application, starting from the first submission.
  :application-deadline-days nil
 
- ;; List of commands to hide. See the swagger schema of
+ ;; List of commands to disable. See the swagger schema of
  ;; /api/applications/{id} field application/permissions for possible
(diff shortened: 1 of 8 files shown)


resources/config-defaults.edn
```diff
@@ -95,5 +95,10 @@
  ;; on an application, starting from the first submission.
  :application-deadline-days nil
 
+ ;; List of commands to hide. See the swagger schema of
+ ;; /api/applications/{id} field application/permissions for possible
+ ;; values.
+ :hide-commands []
```
luontola: Swagger UI doesn't show the possible permissions. 😟

luontola: And neither does swagger.json:
```
"application/permissions": {
"type": "array",
"items": {
"type": "string"
},
"uniqueItems": true
}

Comment (luontola) [reply resources/config-defaults.edn]: Sorry, I was looking at the wrong branch.

File: src/clj/rems/application/model.clj [hunk 2]
```diff
@@ -572,10 +574,13 @@
...
   (not= #{:everyone-else} (permissions/user-roles application user-id)))
 
+(defn- hide-commands [commands]
+  (set/difference commands (set (:hide-commands env))))
+
 (defn apply-user-permissions [application user-id]
   (let [see-application? (see-application? application user-id)
         roles (permissions/user-roles application user-id)
-        permissions (permissions/user-permissions application user-id)
+        permissions (hide-commands (permissions/user-permissions application user-id))
```
luontola: The command could be truly disabled by using `rems.permissions/blacklist` in `rems.application.model/calculate-permissions`

opqdonut: done

resources/config-defaults.edn
```diff
@@ -95,5 +95,10 @@
  ;; on an application, starting from the first submission.
  :application-deadline-days nil
 
+ ;; List of commands to disable. See the swagger schema of
+ ;; /api/applications/{id} field application/permissions for possible
```
okahilak: Hmm, in my swagger, it shows only one possible permission in the example value for that endpoint:

image

Has that particular example application been approved, or what is the cause for that? In any case, I couldn't get a complete list of possible values of :disable-commands from there.

Also nit: the api endpoint is listed as `/api/applications/{application-id}` (instead of `/api/applications/{id}`) in swagger.

resources/config-defaults.edn
```diff
@@ -95,5 +95,10 @@
...
+ ;; values.
+ :disable-commands []
```
okahilak: Also, based on swagger, I'm not sure if I should write `:disable-commands ["application.command/copy-as-new"]`, `:disable-commands [:application.command/copy-as-new]`, or `:disable-commands [:copy-as-new]` here. Maybe an example would help? (Transforming the string to a keyword is probably already out of reach for most casual users of config.edn without further instructions.)

resources/config-defaults.edn
opqdonut: Check the "Model", not the "Example value". The example value for lists of enum values seems to be a list of length 1.

resources/config-defaults.edn
opqdonut: That's a good point. I'll think about it.

PR #1904, opened 2020-01-13
Allow API keys to have unavailable roles

### LINKED ISSUE #1662: API Key Limits
There is a need to limit what API users can do. I.e. now API key can represent any user and do any operation that is allowed for that user. There should be a possibility to limit the user, role or API function that can be used.

- [ ] API key class that can only create/close applications i.e. applicant role only
- [ ] API key for readonly entitlements

okahilak:
Closes #1662 

## Reviewability
- [x] link to issue

## Documentation

## Testing


src/clj/rems/db/api_key.clj
```diff
@@ -1,5 +1,14 @@
...
+  (:require [rems.db.core :as db]
+            [rems.json :as json]))
 
 (defn valid? [key]
   (not (nil? (db/get-api-key {:apikey key}))))
+
+(defn add-api-key! [key unavailable-roles comment]
+  (db/add-api-key! {:apikey key
+                    :comment comment
+                    :unavailableroles (json/generate-string unavailable-roles)}))
```
opqdonut: why a blacklist and not a whitelist?

src/clj/rems/middleware.clj
```diff
@@ -68,13 +68,19 @@
...
-                                context/*user* (set/union (roles/get-roles (getx-user-id))
-                                                          (applications/get-all-application-roles (getx-user-id)))
-                                (:uses-api-key? request) (conj :api-key))]
-      (with-mdc {:roles (str/join " " (sort context/*roles*))}
-        (handler request)))))
+    (let [unavailable-roles (-> request
+                                auth/get-api-key
+                                api-key/unavailable-roles)]
+      (binding [context/*root-path* (calculate-root-path request)
+                context/*roles* (cond-> #{}
```
opqdonut: this expression could probably be simplified, perhaps with some intermediate definitions in the let block above

src/clj/rems/db/api_key.clj
opqdonut: whitelists are typically better for security stuff like this. otherwise when we add a new role (e.g. domainowner) it will "leak" to all existing api keys when upgrading REMS

okahilak: Right, good point. I'll change it to a whitelist.

src/clj/rems/middleware.clj
okahilak: Yeah, it's a bit too complicated, I'll do something about it.

resources/migrations/20191023162400-permitted-roles-for-api-key.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE api_key ADD COLUMN permittedRoles jsonb;
```
okahilak: Okay hmm, this up-migration should probably now add all current roles as a default, right?

opqdonut: that's one option yes, the other is to treat a missing permittedRoles value as "everything"

test/clj/rems/api/test_api.clj
```diff
@@ -25,3 +25,26 @@
...
+      (let [resp (-> (request :get "/api/forms")
+                     (authenticate "42" "owner")
+                     handler)]
+        (is (= 200 (:status resp)))))
+    (testing "handler and owner roles unavailable"
+      (let [resp (-> (request :get "/api/forms")
+                     (authenticate "43" "owner")
+                     handler)]
+        (is (response-is-forbidden? resp)))))
+  (testing ":api-key role not available when using wrong API key"
```
okahilak: There was a bug previously, allowing the user to obtain :api-key role when using wrong API key but authenticating with cookie and csrf-token. Fixed in this patch, and added this test.

src/clj/rems/db/api_key.clj
```diff
@@ -1,5 +1,16 @@
 (ns rems.db.api-key
-  (:require [rems.db.core :as db]))
+  (:require [rems.db.core :as db]
+            [rems.json :as json]))
+
+(def +all-roles+ ["applicant" "decider" "handler" "logged-in" "owner" "past-reviewer" "reporter" "reviewer"])
```
opqdonut: these could be keywords, right? we also have rems.roles/+roles+ which might be deprecated now?

src/clj/rems/standalone.clj
```diff
@@ -66,7 +68,12 @@
      \"test-data\" -- insert test data into database
      \"demo-data\" -- insert data for demoing purposes into database
      \"validate\" -- validate data in db
-     \"add-api-key <api-key> [<description>]\" -- add api key to db"
+     \"add-api-key <api-key> [<description>] [<permitted-role 1>] ... [<permitted-role n>]\" -- add api key to db.
+        <description> is an optional text comment.
+        <permitted-role> is, e.g., owner or handler. If no permitted roles are
+          given, permit all roles.
```
opqdonut: excellent

test/clj/rems/api/test_api.clj
opqdonut: good catch

src/clj/rems/db/api_key.clj
okahilak: Thanks, changed them to keywords (as well as changing +all-roles+ to be a set instead of a vector), and changed +roles+ to +db-roles+ and added a comment there to distinguish it from +all-roles+ (the difference being that +db-roles+ only contains those roles that can be given to users in the database).

PR #1908, opened 2020-01-14
feat: highlight almost overdue applications on actions page

opqdonut:
fixes #1783

## Reviewability
- [x] link to issue

## API
  - changed here but new since last release

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/application_list.cljs
+++ b/src/cljs/rems/application_list.cljs
@@ -63,6 +63,13 @@
   (when-let [dl (:application/deadline application)]
     (time/after? (time/now) dl)))
 
+(defn- application-almost-overdue? [application]
+  (when-let [dl (:application/deadline application)]
+    (let [start (time/date-time (:application/first-submitted application))
+          seconds (time/seconds (* 0.75 (time/in-seconds (time/interval start dl))))
+          threshold (time/plus start seconds)]
+      (time/after? (time/now) threshold))))
+
 (rf/reg-sub
  ::table-rows
  (fn [[_ apps-sub] _]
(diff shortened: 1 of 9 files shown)


opqdonut:
application-almost-overdue

src/cljs/rems/application_list.cljs
```diff
@@ -57,6 +58,18 @@
                         :resubmitted-application}
                       (:application/todo app)))))
 
+;; could be in some util namespace, but only used here for now
+(defn- application-overdue? [application]
+  (when-let [dl (:application/deadline application)]
+    (time/after? (time/now) dl)))
```
okahilak: Doesn't moving this past deadline -check to front-end risk having the wrong time/timezone etc. in the user computer? That's actually the reason why I initially decided to have it in the back-end.

opqdonut: Timezones shouldn't affect it since we use timezoned datetimes, but yeah the client clock can be wrong. However since we use the client time only for rendering warnings, I think it's ok for now.

src/cljs/rems/application_list.cljs
```diff
@@ -90,19 +103,18 @@
            :todo (let [value (localize-todo (:application/todo app))]
                    {:value value
                     :td [:td.todo
-                         {:class (str (when (current-user-needs-to-do-something? app)
-                                        "text-highlight ")
-                                      (when (:application/past-deadline app)
-                                        "text-danger"))}
+                         {:class (when (current-user-needs-to-do-something? app)
```
okahilak: This is a nice change, I think it looks more clear now that only the :submitted column is highlighted.

src/cljs/rems/application_list.cljs
okahilak: Ok, I guess it's fine.

Issue #1893, opened 2020-01-09
Organizations


opqdonut:
Check NKR's specific needs before starting implementing. Some ideas:

- [x] add missing organization fields to db & api
  - [x] catalogue item (#1916)
  - [x] license (#1917)
  - ~license attachments~ no need for this yet
- [x] track organization of users (#1918)
- get user organization from aai-proxy / idp – split to #2035 and #2036
- [x] create `organization-owner` role that allows administration acccess only to users own organization
  - [x] Back-end: Filter by users's organization in listing APIs
    - ~resource, license, form, workflow~
      - [x] remove this feature (see ADR009 #2004) (#2034)
    - ~license attachments~
    - ~catalogue item~
  - [x] UI: Set organization when creating object
  - [x] Back-end: Check that user's organization matches object's organization when creating
  - Back-end: Check that user's organization matches object's organization when editing/archiving/disabling
    - [x] catalogue item (#2028)
    - [x] workflow (#2029)
    - [x] form (#2029)
    - [x] resource (#2029)
    - [x] license (#2031)
  - ~Back-end: Check that license's organization matches license attachment's organization when creating a license~ (not needed, see ADR009 #2004)
  - ~Back-end: Check that license, form, workflow, and resource organization match catalogue item's when creating a catalogue item (#2014)~
    - [x] remove this check (see ADR009 #2004) (#2028)
  - ~Back-end: Check that license organization matches resource's when creating a resource (#2022)~
    - [x] remove this check (see ADR009 #2004) (#2028)
  - ~UI: Only show licenses, forms, and workflows that match the catalogue item's organization when creating catalogue item~
    - [x] remove this feature (see ADR009 #2004) (#2038)
  - ~UI: Only show licenses that match the resource's organization when creating resource~
    - [x] remove this feature (see ADR009 #2004) (#2038)
  - [x] Add RDorganizationowner1/2 users to test data
- [x] list available organizations in config (#1973)
  - [x] configure available organizations for existing installations (do any even use organizations?)
- "show only my organization" checkbox in admin listing pages (split out to #2046)
- [x] document organizations feature (#1996, #2045)
- less clear tasks / ideas:
  - possibility to limit api key to organization?

opqdonut:
ADR009 (#2004) causes changes to organizations. Edited task description accordingly.

PR #1924, opened 2020-01-22
Add organization-owner role

okahilak:
In this patch, the role does not have privileges to create or edit items yet.

Part of #1893 

## Reviewability
- [x] link to issue

## API

## Documentation
   (Will do after the other changes related to the ticket are done.)

## Testing

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -44,6 +44,7 @@
   responds to requests from handlers
 - user-owner: a role that can create & update user details
 - owner: a role that can create and edit resources, workflows, catalogue items, etc.
+- organization-owner: like owner, but can only create and edit items belonging to the user's organization
 
 ## Events
 
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -28,6 +28,7 @@
    :applicant2 "malice"
    :approver1 "developer"
    :approver2 "handler"
+   :organization-owner "organization-owner"
    :owner "owner"
    :reporter "reporter"
    :reviewer "carl"
@@ -43,10 +44,12 @@
    "carl" {:eppn "carl" :mail "carl@example.com" :commonName "Carl Reviewer"}
    "elsa" {:eppn "elsa" :mail "elsa@example.com" :commonName "Elsa Roleless"}
    "frank" {:eppn "frank" :mail "frank@example.com" :commonName "Frank Roleless" :organization "frank"}
+   "organization-owner" {:eppn "organization-owner" :mail "organization-owner@example.com" :commonName "Organization Owner" :organization "organization"}
    "owner" {:eppn "owner" :mail "owner@example.com" :commonName "Owner"}
    "reporter" {:eppn "reporter" :mail "reporter@example.com" :commonName "Reporter"}
    approver-bot/bot-userid {:eppn approver-bot/bot-userid :commonName "Approver Bot"}})
 
+;; TODO: Add organization-owner to +demo-users+, as well as +oidc-users+.
 (def +demo-users+
   {:applicant1 "RDapplicant1@funet.fi"
    :applicant2 "RDapplicant2@funet.fi"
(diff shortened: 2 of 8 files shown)


src/clj/rems/api/public.clj
```diff
@@ -56,7 +56,7 @@
 
     (GET "/full" []
       :summary "Get (almost) full configuration"
-      :roles #{:owner}
+      :roles #{:owner :organization-owner}
```
opqdonut: this probably shouldn't be available to `:organization-owner` since it has info relating to the whole installation (e.g. oidc client secret)

src/clj/rems/api/applications.clj
```diff
@@ -171,7 +171,7 @@
 
     (GET "/export" []
       :summary "Export all submitted applications of a given form as CSV"
-      :roles #{:owner}
+      :roles #{:owner :organization-owner}
```
opqdonut: hmm I wonder if this should be available for reporter also

src/clj/rems/migrations/add_organization_owner_role.clj
```diff
@@ -0,0 +1,35 @@
+(ns rems.migrations.add-organization-owner-role
+  (:require [hugsql.core :as hugsql]
+            [rems.json :as json]))
+
+;; This migration goes through all api keys defined in api_key table and
+;; adds :organization-owner to all lists of permitted roles that already
+;; include :owner.
```
opqdonut: we might not need this migration, it might be enough to mention the new role in the changelog and installations can consider adding it. on the other hand, adding organization-owner can't hurt if somebody already has owner rights

docs/glossary.md
```diff
@@ -44,6 +44,7 @@
   responds to requests from handlers
 - user-owner: a role that can create & update user details
 - owner: a role that can create and edit resources, workflows, catalogue items, etc.
+- organization-owner: like owner, but can only create and edit items belonging to the user's organization
```
okahilak: This is a bit forward-leaning, technically not true yet. I can remove if desired.

src/clj/rems/api/resources.clj
```diff
@@ -38,7 +38,7 @@
 
     (GET "/" []
       :summary "Get resources"
-      :roles #{:owner :handler}
+      :roles #{:owner :organization-owner :handler}
```
opqdonut: if we add `:organization-owner` access before we implement the restrictions, the organization-owner feature is broken (= reveals more information than it should) in the interim. perhaps the roles should only get added to the endpoints at the same pace as the organization-owner access restrictions are implemented?

okahilak: Done.

src/clj/rems/api/public.clj
okahilak: Done.

src/clj/rems/api/applications.clj
okahilak: Removed organization-owner from here now, as the same considerations apply as with the other API endpoints. We can add both organization-owner and reporter here later.

src/clj/rems/migrations/add_organization_owner_role.clj
okahilak: True. I kept this migration here for now, as I believe (but am not sure) that we have at least one installation where we supposedly want to have organization-owner in the permitted roles, but it doesn't exist there currently. So this should save the step of editing the database manually in that case, and probably doesn't hurt elsewhere.

src/clj/rems/db/test_data.clj
```diff
@@ -63,6 +66,7 @@
    "RDapprover1@funet.fi" {:eppn "RDapprover1@funet.fi" :mail "RDapprover1.test@rems_example.org" :commonName "RDapprover1 REMSDEMO"}
    "RDapprover2@funet.fi" {:eppn "RDapprover2@funet.fi" :mail "RDapprover2.test@rems_example.org" :commonName "RDapprover2 REMSDEMO"}
    "RDreview@funet.fi" {:eppn "RDreview@funet.fi" :mail "RDreview.test@rems_example.org" :commonName "RDreview REMSDEMO"}
+   "RDorganizationowner@funet.fi" {:eppn "RDorganizationowner@funet.fi" :mail "RDorganizationowner.test@test_example.org" :commonName "RDorganizationowner REMSDEMO" :organization "organization"}
```
opqdonut: does this user exist in test haka?

okahilak: Apparently not, I amended the patch, not adding this user to this particular list.

PR #1937, opened 2020-01-24
Change Applications text on admin pages to Reports

### LINKED ISSUE #1933: Change "Applications" tab in admin menu to, e.g., "Reports"
"Applications" might be too non-specific. Also, the page may eventually end up including other reports than those for applications.

okahilak:
Closes #1933 

## Reviewability
- [x] link to issue


okahilak:
For some reason, I'm not completely happy with this after the change. Maybe it's because of the discrepancy in "level of abstraction" between "Reports" and "Export applications", see below:

image

Maybe the text "Export applications" would need to be changed then, as well? Ideally, there would be a subheading "Applications" or similar, but when there's only one button, it makes the page seem crowded.

Macroz:
I think this is fine like this for now. 

Some ideas
- There could be text about what kind of export this one is, e.g. it doesn't mention CSV which it could, or does it export closed ones too etc. Or that it will include the field values. 
- The page could have an intro text too. 
- As you said there could be a heading.

Pretty much all of these are useful but not necessary when this is more or less experimental still and has just one export.

okahilak:
Thanks, I'll add those ideas to an "idea" issue.

PR #1951, opened 2020-01-27
feat: allow handler to close applications in decider-workflow

### LINKED ISSUE #1938: Hide Close button for applicant  after submission
When catalog item is using decider workflow, the applicant should not be able to close an already submitted application. Only handler should still have the option available.

opqdonut:
for #1938

## Reviewability
- [x] link to issue

## Documentation

## Testing


opqdonut:
let's make sure this is actually what we want before merging

PR #1956, opened 2020-01-28
feat: allow applicant to close drafts in the decider workflow

### LINKED ISSUE #1938: Hide Close button for applicant  after submission

opqdonut:
fixes #1938

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -253,6 +253,7 @@
     {:permission :application.command/add-member}
     {:permission :application.command/assign-external-id}
     {:permission :application.command/change-resources}
+    {:permission :application.command/close}
     {:permission :application.command/copy-as-new}
     {:permission :application.command/create}
     {:permission :application.command/invite-member}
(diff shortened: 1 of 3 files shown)


src/clj/rems/application/model.clj
```diff
@@ -253,6 +253,7 @@
     {:permission :application.command/add-member}
     {:permission :application.command/assign-external-id}
     {:permission :application.command/change-resources}
+    {:permission :application.command/close}
```
okahilak: Hmm, does this actually give decider the permission to close applications, as well as the applicant? Is that what we want?

opqdonut: It let's the "usual" behaviour of close through from the master workflow. So: applicant can close drafts, handler can close submitted & approved applications

okahilak: Ok, thanks!

PR #1959, opened 2020-01-28
Allow organization-owner to fetch own organization's resources in API

okahilak:
Part of #1893 

## Reviewability
- [x] link to issue

## API

## Testing

### DIFF
--- a/src/clj/rems/api/resources.clj
+++ b/src/clj/rems/api/resources.clj
@@ -2,8 +2,7 @@
   (:require [compojure.api.sweet :refer :all]
             [rems.api.schema :refer :all]
             [rems.api.services.resource :as resource]
-            [rems.api.util] ; required for route :roles
-            [rems.auth.util :refer [throw-forbidden]]
+            [rems.api.util :refer [not-found-json-response]] ; required for route :roles
             [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer :all]
             [schema.core :as s])
@@ -44,16 +43,16 @@
                      {archived :- (describe s/Bool "whether to include archived resources") false}]
       :return Resources
       (ok (resource/get-resources (merge (when-not disabled {:enabled true})
-                                         (when-not archived {:archived false}))
-                                  (getx-user-id))))
+                                         (when-not archived {:archived false})))))
 
     (GET "/:resource-id" []
       :summary "Get resource by id"
       :roles #{:owner :organization-owner :handler}
       :path-params [resource-id :- (describe s/Int "resource id")]
       :return Resource
-      (ok (or (resource/get-resource resource-id (getx-user-id))
-              (throw-forbidden))))
+      (if-let [resource (resource/get-resource resource-id)]
+        (ok resource)
+        (not-found-json-response)))
 
     (POST "/create" []
       :summary "Create resource"
--- a/src/clj/rems/api/resources.clj
+++ b/src/clj/rems/api/resources.clj
@@ -3,6 +3,7 @@
             [rems.api.schema :refer :all]
             [rems.api.services.resource :as resource]
             [rems.api.util] ; required for route :roles
+            [rems.auth.util :refer [throw-forbidden]]
             [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer :all]
             [schema.core :as s])
@@ -38,19 +39,21 @@
 
     (GET "/" []
       :summary "Get resources"
-      :roles #{:owner :handler}
+      :roles #{:owner :organization-owner :handler}
       :query-params [{disabled :- (describe s/Bool "whether to include disabled resources") false}
                      {archived :- (describe s/Bool "whether to include archived resources") false}]
       :return Resources
       (ok (resource/get-resources (merge (when-not disabled {:enabled true})
-                                         (when-not archived {:archived false})))))
+                                         (when-not archived {:archived false}))
+                                  (getx-user-id))))
 
     (GET "/:resource-id" []
       :summary "Get resource by id"
-      :roles #{:owner :handler}
+      :roles #{:owner :organization-owner :handler}
       :path-params [resource-id :- (describe s/Int "resource id")]
       :return Resource
-      (ok (resource/get-resource resource-id)))
+      (ok (or (resource/get-resource resource-id (getx-user-id))
+              (throw-forbidden))))
 
     (POST "/create" []
       :summary "Create resource"
--- a/src/clj/rems/api/services/catalogue.clj
+++ b/src/clj/rems/api/services/catalogue.clj
@@ -9,16 +9,16 @@
             [rems.db.resource :as resource]
             [rems.db.workflow :as workflow]))
 
-(defn- dependencies-for-catalogue-item [id]
+(defn- dependencies-for-catalogue-item [id user-id]
   (let [{:keys [formid wfid resource-id]} (catalogue/get-localized-catalogue-item id)]
     {:form (form/get-form-template formid)
-     :resource (resource/get-resource resource-id)
+     :resource (resource/get-resource resource-id user-id)
      :workflow (workflow/get-workflow wfid)
      :licenses (licenses/get-licenses {:wfid wfid
                                        :items [id]})}))
 
 (defn create-catalogue-item! [{:keys [localizations organization] :as command} user-id]
-  (or (util/forbidden-organization? user-id organization)
+  (or (util/forbidden-organization-error? user-id organization)
       ;; TODO make :organization unoptional?
       (let [id (:id (db/create-catalogue-item! (merge {:organization ""}
                                                       (select-keys command [:form :resid :wfid :enabled :archived :organization]))))
@@ -49,8 +49,8 @@
   (db/set-catalogue-item-enabled! (select-keys command [:id :enabled]))
   {:success true})
 
-(defn set-catalogue-item-archived! [{:keys [id archived]}]
-  (let [{:keys [form resource workflow licenses]} (dependencies-for-catalogue-item id)
+(defn set-catalogue-item-archived! [{:keys [id archived]} user-id]
+  (let [{:keys [form resource workflow licenses]} (dependencies-for-catalogue-item id user-id)
         archived-licenses (filter :archived licenses)
         errors
         (remove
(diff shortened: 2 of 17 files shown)


src/clj/rems/api/services/catalogue.clj
```diff
@@ -9,16 +9,16 @@
...
     {:form (form/get-form-template formid)
-     :resource (resource/get-resource resource-id)
+     :resource (resource/get-resource resource-id user-id)
      :workflow (workflow/get-workflow wfid)
      :licenses (licenses/get-licenses {:wfid wfid
                                        :items [id]})}))
 
 (defn create-catalogue-item! [{:keys [localizations organization] :as command} user-id]
-  (or (util/forbidden-organization? user-id organization)
+  (or (util/forbidden-organization-error? user-id organization)
```
Macroz: If this returns an error it should not be with a `?`?

src/clj/rems/api/resources.clj
```diff
@@ -38,19 +39,21 @@
...
 
     (GET "/:resource-id" []
       :summary "Get resource by id"
-      :roles #{:owner :handler}
+      :roles #{:owner :organization-owner :handler}
       :path-params [resource-id :- (describe s/Int "resource id")]
       :return Resource
-      (ok (resource/get-resource resource-id)))
+      (ok (or (resource/get-resource resource-id (getx-user-id))
+              (throw-forbidden))))
```
Macroz: Why not throw in `get-resource` already?

test/clj/rems/api/services/test_resources.clj
```diff
@@ -36,28 +38,28 @@
                                         :enabled true})
       (is (= {:enabled true
               :archived false}
-             (status-flags res-id))))
+             (status-flags res-id owner))))
```
Macroz: I don't like having to pass the current user everywhere. It seems rather wasteful compared to a long time ago when we used the dynamic scope / context.

test/clj/rems/api/test_resources.clj
```diff
@@ -130,7 +130,32 @@
...
+                                 read-body)]
+                (is resource)
+                (is (= [licid] (map :id (:licenses resource))))))
+
+            (testing "fetch using incorrect organization owner"
+              (let [response (-> (request :get (str "/api/resources/" id))
+                                 (authenticate api-key "organization-owner2")
+                                 handler)]
+                (is (response-is-forbidden? response))
+                (is (= "forbidden" (read-body response)))))
```
Macroz: I'd rather have the API return empty so it doesn't leak information about resources etc. This is of course an API design choice.

src/clj/rems/api/services/catalogue.clj
```diff
@@ -49,8 +49,8 @@
   (db/set-catalogue-item-enabled! (select-keys command [:id :enabled]))
   {:success true})
 
-(defn set-catalogue-item-archived! [{:keys [id archived]}]
-  (let [{:keys [form resource workflow licenses]} (dependencies-for-catalogue-item id)
+(defn set-catalogue-item-archived! [{:keys [id archived]} user-id]
+  (let [{:keys [form resource workflow licenses]} (dependencies-for-catalogue-item id user-id)
```
opqdonut: hmmm an owner might've linked a form owned by `organizationA` to a catalogue item owned by `organizationB`. in that case we shouldn't let an organization owner of `organizationA` archive it (and so on for licenses, workflows, unarchiving, etc.). also, these cases should have tests (to document our decision about them)

src/clj/rems/api/services/resource.clj
```diff
@@ -59,5 +59,5 @@
                                     :archived archived})
         {:success true}))))
 
-(defn get-resource [id] (resource/get-resource id))
-(defn get-resources [filters] (resource/get-resources filters))
+(defn get-resource [id user-id] (resource/get-resource id user-id))
+(defn get-resources [filters user-id] (resource/get-resources filters user-id))
```
opqdonut: could user be in filters?

src/clj/rems/api/services/util.clj
```diff
@@ -4,18 +4,23 @@
             [rems.db.users :as users]))
 
 (defn forbidden-organization? [user-id organization]
+  ;; XXX: If called multiple times with the same user-id, queries database
+  ;;      again each time.
   (let [user (users/get-user user-id)
         not-owner? (not (contains? (roles/get-roles user-id) :owner))
```
opqdonut: hmm, even though you pass in user-id explicitly, you fetch the roles from the db

src/clj/rems/api/services/catalogue.clj
okahilak: Done.

src/clj/rems/api/resources.clj
okahilak: I thought it's more API level stuff than db level stuff to decide what to show exactly when the resource is not available.

src/clj/rems/api/services/catalogue.clj
okahilak: Okay, so this patch doesn't actually let organization-owner to archive items, user-id is passed here only because get-resource needs it in dependencies.-for-catalogue-item This patch only changes API endpoints /api/resources/ and /api/resources/[id]

opqdonut: there should probably be userless versions of get-resource and get-resources that can be used by the archiving logic

test/clj/rems/api/services/test_resources.clj
opqdonut: `status-flags` is only a util for this test. I guess one option would be to move `status-flags` inside the `let` so it can use `owner` directly :)

src/clj/rems/api/services/resource.clj
okahilak: Not sure, I looked at db/apply-filters and it seems to work by checking that certain key(s) exist in the given map, and filter out those elements for which they don't. I couldn't quite see how to implement the filtering by certain organization in that framework.

opqdonut: you're probably right, `filters` has a specific meaning in our db code

src/clj/rems/api/services/util.clj
okahilak: Hmm, not sure how to do it otherwise. Do you have a proposal?

okahilak: Ah, I forgot that the roles exist in the context, as well. Right, I'll have a look at that.

src/clj/rems/api/resources.clj
okahilak: I think it's also how it's done elsewhere in the API, e.g., /api/applications/[id] endpoint:

```
    (GET "/:application-id" []
      :summary "Get application by `application-id`"
      :roles #{:logged-in}
      :path-params [application-id :- (describe s/Int "application id")]
      :responses {200 {:schema Application}
                  404 {:schema s/Str :description "Not found"}}
      (if-let [app (applications/get-application (getx-user-id) application-id)]
        (ok app)
        (api-util/not-found-json-response)))

File: src/clj/rems/api/services/util.clj (hunk 7 above)
Comment (opqdonut) [reply src/clj/rems/api/services/util.clj]: Hmm, you also fetch the user details from the db...

I can't quite put my finger on what bothers me, but I kinda think that the API of this function should be either `(forbidden-organization? organization)` or `(forbidden-organization? user roles organization)` where `user` is a map of the user attributes, not just the user id... but now that I think about it, `(forbidden-organization? user-id organization)` is an ok API too. I guess I was just expecting something else.

File: src/clj/rems/api/resources.clj (hunk 2 above)
Comment (opqdonut) [reply src/clj/rems/api/resources.clj]: perhaps the weird part is `(ok (or (get-it) (throw-forbidden)))`, so how about:
```
(if-let [thing (get-it)]
  (ok thing)
  (forbidden-response)))
```

File: test/clj/rems/api/services/test_resources.clj (hunk 3 above)
Comment (okahilak) [reply test/clj/rems/api/services/test_resources.clj]: Done.

File: src/clj/rems/api/services/catalogue.clj (hunk 5 above)
Comment (Macroz) [reply src/clj/rems/api/services/catalogue.clj]: I actually think linking between organizations should not be allowed.

File: src/clj/rems/api/services/resource.clj (hunk 6 above)
Comment (Macroz) [reply src/clj/rems/api/services/resource.clj]: Key value filter would make sense as type though

File: src/clj/rems/api/services/util.clj (hunk 7 above)
Comment (okahilak) [reply src/clj/rems/api/services/util.clj]: Done, I'll change it to `(forbidden-organization? organization)`.

File: src/clj/rems/api/services/resource.clj (hunk 6 above)
Comment (okahilak) [reply src/clj/rems/api/services/resource.clj]: I think it matters less now that user-id will not be passed separately as a parameter.

File: test/clj/rems/api/test_resources.clj (hunk 4 above)
Comment (okahilak) [reply test/clj/rems/api/test_resources.clj]: I'll change it to not-found, as proposed by Joel, if that's ok?

File: src/clj/rems/api/resources.clj (hunk 2 above)
Comment (okahilak) [reply src/clj/rems/api/resources.clj]: Done.

File: test/clj/rems/api/test_resources.clj (hunk 4 above)
Comment (okahilak) [reply test/clj/rems/api/test_resources.clj]: Done.

File: src/clj/rems/api/services/catalogue.clj (hunk 5 above)
Comment (okahilak) [reply src/clj/rems/api/services/catalogue.clj]: Did not touch this part of the code at this point, so it can be decided later.

File: src/clj/rems/api/resources.clj [hunk 8]
```diff
@@ -47,10 +47,12 @@
...
     (GET "/:resource-id" []
       :summary "Get resource by id"
-      :roles #{:owner :handler}
+      :roles #{:owner :organization-owner :handler}
       :path-params [resource-id :- (describe s/Int "resource id")]
       :return Resource
-      (ok (resource/get-resource resource-id)))
+      (if-let [resource (resource/get-resource resource-id)]
+        (ok resource)
+        (not-found-json-response)))
```
Macroz: Maybe this could be named `not-found-response` since the API does not return JSON only?

Macroz: Forget this one. The actual code is a hack that should be improved elsewhere.

okahilak: Ok.

PR #1964, opened 2020-01-30
Allow only creating catalogue items compatible with organization in UI

okahilak:
Part of #1893 

## Reviewability
- [x] link to issue

## Documentation

## Testing

### DIFF
--- a/src/cljs/rems/administration/create_catalogue_item.cljs
+++ b/src/cljs/rems/administration/create_catalogue_item.cljs
@@ -88,7 +88,7 @@
   (let [request {:wfid (get-in form [:workflow :id])
                  :resid (get-in form [:resource :id])
                  :form (get-in form [:form :form/id])
-                 :organization (or (:organization form) "")
+                 :organization (:organization form)
                  :localizations (into {}
                                       (for [lang languages]
                                         [lang {:title (trim-when-string (get-in form [:title lang]))
--- a/src/cljs/rems/administration/create_catalogue_item.cljs
+++ b/src/cljs/rems/administration/create_catalogue_item.cljs
@@ -16,20 +16,16 @@
             [rems.util :refer [navigate! fetch post! put! trim-when-string]]))
 
 (defn- update-loading [db]
-  (let [progress (::loading-progress db)
-        catalogue-item-id (::catalogue-item-id db)]
-    (cond
-      (<= progress 1)
-      {:db (assoc db ::loading-progress (inc progress))}
+  (cond
+    (not (and (::resources db) (::workflows db) (::forms db)))
+    {:db db}
 
-      (nil? catalogue-item-id)
-      {:db (dissoc db ::loading-progress ::loading?)}
+    (::editing? db)
+    {:db (assoc db ::loading-catalogue-item? true)
+     ::fetch-catalogue-item (::catalogue-item-id db)}
 
-      :else
-      {:db (-> db
-               (dissoc ::loading-progress)
-               (assoc ::loading-catalogue-item? true))
-       ::fetch-catalogue-item catalogue-item-id})))
+    :else
+    {:db (dissoc db ::loading?)}))
 
 (rf/reg-event-fx
  ::enter-page
@@ -40,14 +36,14 @@
                  ::form (when (roles/disallow-setting-organization? roles)
                           {:organization organization})
                  ::catalogue-item-id catalogue-item-id
-                 ::loading? true
-                 ::loading-progress 0)
+                 ::editing? (not (nil? catalogue-item-id))
+                 ::loading? true)
       ::fetch-workflows nil
       ::fetch-resources nil
       ::fetch-forms nil})))
 
 (rf/reg-sub ::catalogue-item (fn [db _] (::catalogue-item db)))
-(rf/reg-sub ::editing? (fn [db _] (not (nil? (::catalogue-item-id db)))))
+(rf/reg-sub ::editing? (fn [db _] (::editing? db)))
 (rf/reg-sub ::loading? (fn [db _] (::loading? db)))
 
 (rf/reg-sub ::form (fn [db _] (::form db)))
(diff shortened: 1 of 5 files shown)


src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -15,10 +16,20 @@
...
-    (if (<= 2 progress)
-      (dissoc db ::loading-progress ::loading?)
-      (assoc db ::loading-progress (inc progress)))))
+  (let [progress (::loading-progress db)
+        catalogue-item-id (::catalogue-item-id db)]
+    (cond
+      (<= progress 1)
+      {:db (assoc db ::loading-progress (inc progress))}
+
+      (nil? catalogue-item-id)
```
opqdonut: what's this case?

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -44,42 +53,49 @@
...
-(defn- valid-request? [request languages]
-  (and (number? (:wfid request))
-       (number? (:resid request))
-       (number? (:form request))
-       (= (set languages)
-          (set (keys (:localizations request))))
-       (every? valid-localization? (vals (:localizations request)))))
+(defn- valid-request? [form request languages]
+  (let [selected-organization (:organization request)]
+    (and (= (get-in form [:workflow :organization]) selected-organization)
```
opqdonut: nit: could use multi-argument `=` here

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -15,10 +16,20 @@
...
 
 (defn- update-loading [db]
-  (let [progress (::loading-progress db)]
-    (if (<= 2 progress)
-      (dissoc db ::loading-progress ::loading?)
-      (assoc db ::loading-progress (inc progress)))))
+  (let [progress (::loading-progress db)
+        catalogue-item-id (::catalogue-item-id db)]
+    (cond
+      (<= progress 1)
```
opqdonut: instead of counting progress numbers, you could do something like

```
(cond
  (not (and (::resources db) (::workflows db) (::forms db)))
  {} ;; do nothing

  ...
```

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -44,42 +53,49 @@
...
+                 :form (get-in form [:form :form/id])
                  :organization (or (:organization form) "")
                  :localizations (into {}
                                       (for [lang languages]
                                         [lang {:title (trim-when-string (get-in form [:title lang]))
                                                :infourl (-> (get-in form [:infourl lang])
                                                             empty-string-to-nil
                                                             trim-when-string)}]))}]
-    (when (valid-request? request languages)
+    (when (valid-request? form request languages)
```
opqdonut: how about keeping `valid-request?` as-is and adding a separate check for organizations? not insisting.

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -174,37 +190,63 @@
...
+(defn- list-organizations []
+  (let [workflows @(rf/subscribe [::workflows])
+        resources @(rf/subscribe [::resources])
+        forms @(rf/subscribe [::forms])
+        items-to-organizations (fn [items key]
+                                 (set (distinct (mapv key items))))]
+    (sort
+     (vec (set/intersection (items-to-organizations workflows :organization)
+                            (items-to-organizations resources :organization)
+                            (items-to-organizations forms :form/organization))))))
```
opqdonut: nice way to do it

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -174,37 +190,63 @@
...
+
 (defn- catalogue-item-organization-field []
-  ;; TODO: Allow editing the organization of an existing catalogue item?
-  (let [readonly (or @(rf/subscribe [::editing?])
+  (let [organizations (list-organizations)
```
opqdonut: does the code now assume that you pick the organization first? what if you try to pick the resource first? maybe we can leave this for future development

src/cljs/rems/administration/create_catalogue_item.cljs
okahilak: Good idea, done.

src/cljs/rems/administration/create_catalogue_item.cljs
okahilak: If we end up on the page by creating a new catalogue item, instead of editing an existing one, it is indicated by missing catalogue-item-id. Maybe it should be more explicit, I'll think about it a bit.

src/cljs/rems/administration/create_catalogue_item.cljs
okahilak: Done.

src/cljs/rems/administration/create_catalogue_item.cljs
okahilak: I might prefer having the organization check as a part of `valid-request?` just by considering the name of the function, since I wouldn't count a request as "valid" if it has incompatible organizations.

src/cljs/rems/administration/create_catalogue_item.cljs
okahilak: As discussed, I'll change it so that the field is disabled if organization has not been selected.

okahilak:
Here are three different looks for the page after changing the fields to read-only when organization has not yet been set. Do you have any preferences between them?

image

image

image

opqdonut:
perhaps the middle one. should the wording be "Please select organization first"?

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -15,10 +16,16 @@
             [rems.util :refer [navigate! fetch post! put! trim-when-string]]))
 
 (defn- update-loading [db]
-  (let [progress (::loading-progress db)]
-    (if (<= 2 progress)
-      (dissoc db ::loading-progress ::loading?)
-      (assoc db ::loading-progress (inc progress)))))
+  (cond
+    (not (and (::resources db) (::workflows db) (::forms db)))
+    {:db db}
```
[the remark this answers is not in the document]
okahilak: Hmm, I tried it out and it seems to get stuck in the loading state, consistent with what one would except if the contents of the database were lost:

image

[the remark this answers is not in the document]
okahilak: Keeping it as it was, as discussed.

PR #1968, opened 2020-01-30
Field privacy

### LINKED ISSUE #1898: Control visibility of fields to reviewers
There is need to be able to limit which fields a reviewer sees because some data is not considered public. The simple version could work like this
- [ ] allow marking each form field as either in group public or private (in future there could be multiple groups but let's think about that more later)
- [ ] reviewer should see only public fields
- [ ] applicant, handler and decider should see all fields
- [ ] check the event history from API and make sure nothing private is seen by reviewers

Macroz:
Closes #1898.

approver_view
reviewer_view

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -148,10 +148,15 @@
    (s/optional-key :field/options) [{:key s/Str
                                      :label LocalizedString}]
    (s/optional-key :field/max-length) (s/maybe (s/constrained s/Int not-neg?))
-   (s/optional-key :field/privacy) (s/enum :public :private)
-   (s/optional-key :field/visibility) {:visibility/type (s/enum :always :only-if)
-                                       (s/optional-key :visibility/field) {:field/id FieldId}
-                                       (s/optional-key :visibility/values) [s/Str]}})
+   (s/optional-key :field/privacy) (rjs/field
+                                    (s/enum :public :private)
+                                    {:description "Public by default"})
+   (s/optional-key :field/visibility) (rjs/field
+                                       {:visibility/type (s/enum :always :only-if)
+                                        (s/optional-key :visibility/field) {:field/id FieldId}
+                                        (s/optional-key :visibility/values) [s/Str]}
+                                       {:description "Always visible by default"})
+   })
 
 (s/defschema NewFieldTemplate
   (-> FieldTemplate
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -148,6 +148,7 @@
    (s/optional-key :field/options) [{:key s/Str
                                      :label LocalizedString}]
    (s/optional-key :field/max-length) (s/maybe (s/constrained s/Int not-neg?))
+   (s/optional-key :field/privacy) (s/enum :public :private)
    (s/optional-key :field/visibility) {:visibility/type (s/enum :always :only-if)
                                        (s/optional-key :visibility/field) {:field/id FieldId}
                                        (s/optional-key :visibility/values) [s/Str]}})
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -162,6 +162,7 @@
   (assoc FieldTemplate
          :field/value s/Str
          :field/visible s/Bool
+         :field/private s/Bool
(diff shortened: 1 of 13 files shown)


src/clj/rems/application/model.clj
```diff
@@ -572,6 +572,20 @@
       (assoc :application/invited-members (set (vals (:application/invitation-tokens application))))
       (update :application/events (partial mapv #(dissoc % :invitation/token)))))
 
+(defn hide-answers-with-privacy [application roles]
+  (let [fields (get-in application [:application/form :form/fields])
+        fields-with-privacy (mapv (fn [field]
+                                    (if (or (some #{:applicant :member :decider :past-decider :handler :reporter}
+                                                  roles)
+                                            (= :public (:field/privacy field :public)))
+                                      (assoc field :field/private false)
```
Macroz: I wonder if it makes sense to call this enrichment or hiding ...

src/clj/rems/application/model.clj
```diff
@@ -572,6 +572,20 @@
       (assoc :application/invited-members (set (vals (:application/invitation-tokens application))))
       (update :application/events (partial mapv #(dissoc % :invitation/token)))))
 
+(defn hide-answers-with-privacy [application roles]
+  (let [fields (get-in application [:application/form :form/fields])
+        fields-with-privacy (mapv (fn [field]
```
opqdonut: map with large anonymous function => consider a `for` instead

src/clj/rems/application/model.clj
```diff
@@ -572,6 +572,20 @@
...
+                                    (if (or (some #{:applicant :member :decider :past-decider :handler :reporter}
```
opqdonut: should this be a function? something like `may-see-private-fields?`

src/cljc/rems/common/form.cljc
```diff
@@ -19,6 +19,9 @@
 (defn supports-options? [field]
   (contains? #{:option :multiselect} (:field/type field)))
 
+(defn supports-privacy? [field]
+  (not (contains? #{:label :header} (:field/type field))))
```
Macroz: I wonder if it makes sense to include everything here? I can see that it could be a little confusing if a label is shown but the field it is for is hidden, or an entire section and its header.

src/clj/rems/api/schema.clj
```diff
@@ -148,6 +148,7 @@
    (s/optional-key :field/options) [{:key s/Str
                                      :label LocalizedString}]
    (s/optional-key :field/max-length) (s/maybe (s/constrained s/Int not-neg?))
+   (s/optional-key :field/privacy) (s/enum :public :private)
```
opqdonut: schema should document that this defaults to public, right?

test/clj/rems/api/test_forms.clj
```diff
@@ -358,6 +358,68 @@
...
+                          handler
+                          read-ok-body
+                          :id)]
+          (is form-id)
+          (testing "and fetching"
+            (let [form (-> (request :get (str "/api/forms/" form-id))
+                           (authenticate api-key user-id)
+                           handler
+                           read-ok-body)]
+              (is (= (select-keys command [:form/organization :form/title])
```
Macroz: Maybe these are not worthwhile to test here. They are repeated elsewhere.

opqdonut: yeah I guess so

src/cljc/rems/common/form.cljc
opqdonut: let's leave it like this for now

src/clj/rems/application/model.clj
opqdonut: hide sounds nice, but it could also be something generic like apply-privacy...

src/clj/rems/application/model.clj
Macroz: Check the new refactoring.

src/clj/rems/api/schema.clj
Macroz: It's part of the opaque JSON blob so I guess not?

Macroz: Ok done now with description.

PR #1970, opened 2020-01-30
Allow only creating resources compatible with organization in UI

okahilak:
Part of #1893 

## Reviewability
- [x] link to issue

## Documentation

## Testing

### DIFF
--- a/src/cljs/rems/administration/create_resource.cljs
+++ b/src/cljs/rems/administration/create_resource.cljs
@@ -93,12 +93,8 @@
 (def ^:private organization-dropdown-id "organization-dropdown")
 (def ^:private licenses-dropdown-id "licenses-dropdown")
 
-(defn- list-organizations []
-  (let [licenses @(rf/subscribe [::licenses])]
-    (sort (distinct (mapv :organization licenses)))))
-
 (defn- resource-organization-field []
-  (let [organizations (list-organizations)
+  (let [organizations (:organizations @(rf/subscribe [:rems.config/config]))
         selected-organization @(rf/subscribe [::selected-organization])
         item-selected? #(= % selected-organization)
         readonly (roles/disallow-setting-organization? (:roles @(rf/subscribe [:identity])))]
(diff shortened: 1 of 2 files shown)


src/cljs/rems/administration/create_resource.cljs
```diff
@@ -85,34 +90,58 @@
...
+  (let [licenses @(rf/subscribe [::licenses])]
+    (sort (distinct (mapv :organization licenses)))))
+
 (defn- resource-organization-field []
-  (let [readonly (roles/disallow-setting-organization? (:roles @(rf/subscribe [:identity])))]
-    [text-field context {:keys [:organization]
-                         :label (text :t.administration/organization)
-                         :readonly readonly
-                         :placeholder (text :t.administration/organization-placeholder)}]))
+  (let [organizations (list-organizations)
```
opqdonut: hmm I think this is wrong. consider the case where an installation doesn't use resource licenses at all. not sure what the fix would be...

src/cljs/rems/administration/create_resource.cljs
```diff
@@ -85,34 +90,58 @@
...
      [:label {:for licenses-dropdown-id} (text :t.create-resource/licenses-selection)]
-     [dropdown/dropdown
-      {:id licenses-dropdown-id
-       :items available-licenses
-       :item-key :id
-       :item-label #(get-localized-title % language)
-       :item-selected? #(contains? (set selected-licenses) %)
-       :multi? true
-       :on-change #(rf/dispatch [::set-licenses %])}]]))
+     (cond
```
opqdonut: two-case cond could as well be an if

src/cljs/rems/administration/create_resource.cljs
okahilak: After discussion, getting the list of organizations from config now.

src/cljs/rems/administration/create_resource.cljs
okahilak: Done.

src/cljs/rems/administration/create_resource.cljs
```diff
@@ -16,35 +17,51 @@
...
+
+                        :else
+                        nil)]
      {:db (assoc db
-                 ::form (merge {:licenses #{}}
-                               (when (roles/disallow-setting-organization? roles)
+                 ::form (merge {:licenses []}
+                               (when organization
                                  {:organization organization}))
+                 ::organization-read-only? (not (nil? organization))
```
opqdonut: good idea to move the read-only logic here :+1:

PR #1971, opened 2020-01-30
feat: hide also previous answers

Macroz:
Leftover from #1898

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -1,6 +1,6 @@
 (ns rems.application.model
   (:require [clojure.test :refer [deftest is testing]]
-            [medley.core :refer [map-vals]]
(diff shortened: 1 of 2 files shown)


src/clj/rems/application/model.clj
```diff
@@ -580,9 +580,10 @@
   (if (or (= :public (:field/privacy field :public))
           (may-see-private-answers? roles))
     (assoc field :field/private false)
-    (assoc field
-           :field/value ""
-           :field/private true)))
+    (-> field
+        (assoc :field/private true)
+        (update-existing :field/value (fn [_] ""))
```
opqdonut: aka `(constantly "")`

Macroz: :+1:

PR #1973, opened 2020-01-31
:organizations config option

opqdonut:
for #1893

## Reviewability
- [x] link to issue

## API

## Documentation
  - will update once the organizations feature is complete
  - documentation is a task in #1893

## Different installations
  - need to do this for sites that use organizations

## Testing

## Follow-up

### DIFF
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -103,5 +103,8 @@
  ;; :disable-commands [:application.command/assign-external-id :application.command/change-resources]
  :disable-commands []
 
+ ;; List of allowed values for the organization field. Should contain at least one element.
+ :organizations [""]
+
  ;; Experimental features
  :enable-pdf-api false}
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -104,7 +104,7 @@
  :disable-commands []
 
  ;; List of allowed values for the organization field. Should contain at least one element.
- :organizations [""]
+ :organizations ["default"]
 
  ;; Experimental features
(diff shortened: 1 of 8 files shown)


test-config.edn
```diff
@@ -6,4 +6,6 @@
  :test-database-url "postgresql://localhost/rems_test?user=rems_test"
  :search-index-path "target/search-index-test"
  :authentication :fake-shibboleth
- :enable-pdf-api true}
+ :enable-pdf-api true
+ ;; list all organizations used in test-data
+ :organizations ["" "abc" "nbn" "organization1" "organization2"]}
```
okahilak: Hmm, is "perf" for performance data one of organizations that should be listed here?

resources/config-defaults.edn
```diff
@@ -103,5 +103,8 @@
  ;; :disable-commands [:application.command/assign-external-id :application.command/change-resources]
  :disable-commands []
 
+ ;; List of allowed values for the organization field. Should contain at least one element.
+ :organizations [""]
```
okahilak: After thinking a bit more, I'm not sure how "" will work as a default without thinking a few more steps ahead than I'm currently capable of. But I guess this is fine for now and we can change it later on as needed.

test-config.edn
opqdonut: good catch, I double-checked the list against the test db

resources/config-defaults.edn
opqdonut: after discussion, changed it to `["default"]`

PR #2003, opened 2020-02-13
feat: review and form model ADR

Macroz:


docs/architecture/009-review-and-form-model.md
```diff
@@ -0,0 +1,93 @@
...
+1. Require the handler to copy-paste the relevant data from private fields to each reviewer (status quo).
+   Application contains form fields of one form.
+2. Model each reviewer organization as a catalogue item with extra fields.
+   Application contains form fields of one form and all extra fields from all catalogue items.
+   Review request is sent with certain catalogue items selected and it controls which fields are shown to the reviewer.
+3. Model each reviewer organization as a catalogue item with a separate form.
+   Application contains form fields from all forms of all catalogue items.
+   Review request is sent with certain catalogue items selected and it controls which fields are shown to the reviewer.
+4. Model each reviewer organization as a catalogue item with same form.
+   Model private fields as repeating fields that repeat per catalogue item.
```
opqdonut: I don't understand option 4. Could you rephrase it?

docs/architecture/009-review-and-form-model.md
```diff
@@ -0,0 +1,93 @@
...
+   Application contains form fields of one form.
+   Review request is sent with certain catalogue items selected and it controls which fields are shown to the reviewer.
+5. Add a field type to the model where you can add multiple answers.
+   Model private fields as repeating fields that repeat per each answer to the multiple answer field.
```
opqdonut: I think I understand option 5 but I think the description could be clearer

docs/architecture/009-review-and-form-model.md
```diff
@@ -0,0 +1,93 @@
...
+
+All in all while multiple forms, and **option 3** are a sensible extension to our model, it certainly requires large changes to the logic all around the codebase.
+
+In the **option 4** we would make a new repeating functionality that would repeat per applied catalogue item. Modeling each reviewer organization as a catalogue item seems like a business decision, not something for us. It requires the organization to either be able to statically know each option, or have a process for adding missing ones. While the repeating logic is sound, we can also consider **option 5** as the same, but not having the need to couple this into catalogue items. So lets prefer **option 5** and not this one.
+
+A field type where you can "press add button" to add more rows sounds also like a natural extension to the types of fields we have. Also repeating fields based on the count of a previous answer sounds sensible. Some kind of grouping logic may be needed to repeat fields in a group (ABCABCABC and not AAABBBCCC). Therefore **option 5** seems like a sensible extension that we should consider.
```
opqdonut: I'm not sure what these repeated fields are. Perhaps they should be covered in Background?

docs/architecture/009-review-and-form-model.md
```diff
@@ -0,0 +1,93 @@
...
+
+Last but not least there is the **option 6**. So what if we want to model the registry keeping organizations as REMS organizations and catalogue items, and/or registries as catalogue items? We could have review requests targeted to organizations, where organization would define the email inboxes and field visibility. A reviewer would see all public form fields, and also private fields of the organization they represent. Common fields can be put as workflow form fields, if we add the capability. This is close to what we do with licenses at the moment, one can add them to resources as well as workflows.
+
+## Decision
+
+Let's try **option 6** because we anticipate that there is a need for multiple forms, as well as a wish to model the organizations better inside REMS.
+
+The tasks should be
+1. organization creation
+   - store id, name, description, owners and review-email
```
opqdonut: I'm not sure this concept for organizations is compatible with the current concept for organizations. First of all, currently we assume users are mapped to organizations by the idp, but in this case the mapping would be dynamic. Perhaps there wouldn't even be a need to map users to organizations since the review request constains the organization?

Could there be two different concepts for organization? This new one could be called a "visibility group" or something.

docs/architecture/009-review-and-form-model.md
Macroz: I have slightly adjusted the wording but we can try to fix it after f2f discussion.

docs/architecture/009-review-and-form-model.md
Macroz: Same as above.

docs/architecture/009-review-and-form-model.md
Macroz: They don't exist so they are not background. It is a concept that answer is repeated for some reason, not just one per field.

docs/architecture/009-review-and-form-model.md
Macroz: Here are some considerations:
- A review request can contain an organization but it may not contain a user. It is unknown by the handler who the reviewer will be. 
- Not all users logging into an organization should be able to see all reviews.
- Does IDP organization for a reviewer actually match expected organization? I think the mapping to a specific application must be done like the invite member.
- I don't see any problem in having one concept for an organization here and using it for two things. For example we can avoid having two separate but almost identical UI functions. 
- The organization that does reviews also likely wants to maintain their own forms and/or catalogue, which is exactly the organization model we have now?

docs/architecture/009-review-and-form-model.md
Macroz: Now with added wording and explanation before option list.

docs/architecture/009-review-and-form-model.md
Macroz: I think the leftover from this discussion was just that we slightly change organization concept and check with other users that there is no conflict there.

PR #2012, opened 2020-02-25
feat: log all user attributes when logging in via oidc

opqdonut:
related to #1893

## Reviewability
- [x] link to issue

### DIFF
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -22,6 +22,9 @@
  ;; Supported authentication methods: :shibboleth, :fake-shibboleth, :oidc
  :authentication :shibboleth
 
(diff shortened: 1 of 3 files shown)


resources/config-defaults.edn
```diff
@@ -22,6 +22,9 @@
  ;; Supported authentication methods: :shibboleth, :fake-shibboleth, :oidc
  :authentication :shibboleth
 
+ ;; Log additional details about authentication
+ :debug-authentication false
```
Macroz: I would rather have this named `:log-authentication-details` as some instance may want to do it anyway in production.

opqdonut: I'll change it

Issue #1890, opened 2020-01-08
Lock for search index is sometimes left on, causing search to not work


okahilak:
It has happened twice now, first at vm1446.kaj.pouta.csc.fi on Dec 18, and the second time at LBR on Jan 08.

The problem manifests itself by giving 500 error when searching, e.g., on actions page, as well as the following message in REMS log (e.g., catalina.out):

```
tore.NativeFSLockFactory@27162e7f): files: [write.lock]
org.apache.lucene.index.IndexNotFoundException: no segments* file found in LockValidatingDirectoryWrapper(NIOFSDirectory@/tmp/rems-search-index lockFactory=org.apache.lucene.store.NativeFSLockFactory@27162e7f): files: [write.lock]
```

When it happens, the directory `/tmp/rems-search-index` contains only the lock file (`write.lock`) but no actual index files (`segment*`).

It can be fixed momentarily by deleting the lock (`/tmp/rems-search-index/write.lock`), after which the search index is re-generated.

okahilak:
It happened again at Fairdata, noticed on Jan 15.

foxlynx:
Again at Fairdata, noticed on 17th of February, restarted 25th of February.

jaakkocsc:
Haven't seen this happening in a while. Reopen, if re-appears.

Macroz:
Happened again with `nkr-rems-test`

Macroz:
No news, is good news?

PR #2021, opened 2020-02-28
fix: empty search-index on startup by removing files

opqdonut:
instead of via lucene api

this helps with the situation where the index was created with a
previous version of lucene, and might also help with #1890

## Reviewability
- [x] link to issue

## Testing

### DIFF
--- a/src/clj/rems/application/search.clj
+++ b/src/clj/rems/application/search.clj
@@ -8,7 +8,8 @@
             [rems.db.applications :as applications]
             [rems.db.events :as events]
             [rems.text :as text])
-  (:import [org.apache.lucene.analysis Analyzer]
+  (:import [com.google.common.io MoreFiles RecursiveDeleteOption]
+           [org.apache.lucene.analysis Analyzer]
            [org.apache.lucene.analysis.standard StandardAnalyzer]
            [org.apache.lucene.document Document StringField Field$Store TextField]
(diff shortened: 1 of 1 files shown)


src/clj/rems/application/search.clj
```diff
@@ -26,14 +27,16 @@
...
-             (with-open [writer (IndexWriter. directory (-> (IndexWriterConfig. analyzer)
-                                                            (.setOpenMode IndexWriterConfig$OpenMode/CREATE)))]
-               (.deleteAll writer)))
-           (atom {::directory directory
-                  ::searcher-manager (SearcherManager. directory (SearcherFactory.))
-                  ::last-processed-event-id 0}))
+             ;; delete old index
+             (MoreFiles/deleteDirectoryContents index-dir (into-array [RecursiveDeleteOption/ALLOW_INSECURE]))
+             (let [directory (NIOFSDirectory. index-dir)]
+               ;; create new empty index by creating and closing an indexwriter, otehrwise SearcherManager will fail
```
Macroz: Small typo here

src/clj/rems/application/search.clj
```diff
@@ -26,14 +27,16 @@
...
+               (.close (IndexWriter. directory (IndexWriterConfig. analyzer)))
```
Macroz: I suppose any exception here will cause the start to fail and leave the files etc. for forensic purposes.

opqdonut: Yeah

PR #2009, opened 2020-02-24
WIP: Multiple form support

Macroz:
For #2003 

Ready for discussion. Missing still, e.g. showing multiple forms in UI, migrations. Should have discussions about backwards compatibility etc.

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue
- [ ] consider adding screenshots for ease of review

## API
- [x] API is documented and shows up in Swagger UI
- [ ] API is backwards compatible or completely new
- [ ] Events are backwards compatible

## Documentation
- [ ] update changelog if necessary
- [ ] add or update docstrings for namespaces and functions
- [ ] components are added to guide page
- [ ] documentation _at least_ for config options (i.e. docs folder)
- [ ] ADR for major architectural decisions or experiments

## Different installations
- [ ] new configuration options added to rems-deploy repository
- [ ] instance specific translations (i.e. LBR kielivara)

## Testing
- [ ] complex logic is unit tested
- [ ] valuable features are integration / browser / acceptance tested automatically

## Accessibility
- [ ] all icons have the aria-label attribute
- [ ] all fields have a label
- [ ] errors are linked to fields with aria-describedby
- [ ] contrast is checked
- [ ] conscious decision about where to move focus after an action

## Follow-up
- [ ] new tasks are created for pending or remaining tasks
- [ ] no critical TODOs left to implement

PR #2033, opened 2020-03-05
Preparation

Macroz:
Cherry-picked the independent parts of #2009.


src/cljc/rems/common/util.cljc
```diff
@@ -26,6 +26,32 @@
...
+              \"y\" :b}}
+
+  In case of non-unique keys, `build-index` picks the first value, e.g.
+
+    (build-index [:a] identity [{:a 1 :b \"x\"} {:a 1 :b \"y\"}])
+      ==> {1 {:a 1 :b \"x\"}}"
+  [ks f coll]
+  (if (empty? ks)
+    (f (first coll))
+    (->> coll
```
opqdonut: could use destructuring instead of first and rest below

src/cljc/rems/common/util.cljc
```diff
@@ -26,6 +26,32 @@
...
+         (group-by (first ks))
+         (map (fn [[k v]] [k (build-index (rest ks) f v)]))
```
opqdonut: could use map-vals here

Issue #1928, opened 2020-01-23
Attachments to comments


opqdonut:
- [x] user can add an attachment every time he sees a comment field
- [x] arbitrary number of attachments supported
- [x] attachment visibility is same as comment visibility
- [x] show attachment name when attaching file?

PR #2047, opened 2020-03-09
API for adding attachments to comments

opqdonut:
for #1928

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up


src/clj/rems/application/events.clj
```diff
@@ -106,40 +104,34 @@
...
-  (assoc EventBase
+  (assoc EventWithComment
          :event/type (s/enum :application.event/remarked)
-         :application/comment s/Str
          :application/public s/Bool))
 (s/defschema ResourcesChangedEvent
   (assoc EventBase
          :event/type (s/enum :application.event/resources-changed)
-         (s/optional-key :application/comment) s/Str
+         (s/optional-key :application/comment) s/Str ;; TODO
```
opqdonut: should we just make this mandatory?

src/clj/rems/api/services/command.clj
```diff
@@ -41,14 +42,21 @@
    (rejecter-bot/run-rejecter-bot new-events)
    (approver-bot/run-approver-bot new-events)))
 
+;; TODO move somewhere else?
+(defn attachment-for? [application-id attachment-id]
+  (let [attachment (attachments/get-attachment attachment-id)]
```
opqdonut: this always fetches the full attachment contents, which is inefficient

src/clj/rems/application/commands.clj
```diff
@@ -270,6 +262,19 @@
 (defn- ok [& events]
   (ok-with-data nil events))
 
+(defn- invalid-attachments-error [application injections cmd]
+  (let [invalid-ids (for [id (:attachments cmd)
+                          :when (not ((:attachment-for? injections) (:application/id application) id))]
+                      id)]
```
opqdonut: should also disallow using attachments that aren't your own

src/clj/rems/api/services/command.clj
opqdonut: fixed by adding `get-attachment-metadata`

src/clj/rems/application/commands.clj
opqdonut: done now

src/clj/rems/application/model.clj
```diff
@@ -532,6 +532,7 @@
       (update :application/licenses enrich-licenses get-license)
       (update :application/events (partial mapv #(enrich-event % get-user get-catalogue-item)))
       (assoc :application/applicant (get-user (get-in application [:application/applicant :userid])))
+      ;; TODO this contains even attachments that user might not have access to:
```
opqdonut: should probably do something about this

opqdonut: fixed now

src/clj/rems/application/events.clj
opqdonut: instead made all comments optional :)

test/clj/rems/api/test_applications.clj
```diff
@@ -724,15 +747,169 @@
...
+    (testing "handler uploads an attachment"
+      (let [attachment-id (add-attachment handler-id (file "handler-remark.txt"))]
+        (is (number? attachment-id))
+        (testing "and attaches it to a public remark"
+          (is (= {:success true} (send-command handler-id
+                                               {:type :application.command/remark
+                                                :application-id application-id
+                                                :comment "see attachment"
+                                                :public true
+                                                :attachments [attachment-id]}))))))
```
opqdonut: needs a test for not being able to remove the attachment after it is in use

opqdonut: no, there's no api for removing attachments (I was thinking about license attachments, which can be removed if they're not in use)

src/clj/rems/application/events.clj
```diff
@@ -16,28 +16,29 @@
    :event/actor UserId
    :application/id s/Int})
 
-(s/defschema ApprovedEvent
+(s/defschema EventWithComment
   (assoc EventBase
+         (s/optional-key :application/comment) s/Str
+         (s/optional-key :event/attachments) [s/Int]))
```
opqdonut: should this be `[{:attachment/id s/Int}]` instead?
we might want to return something like `{:attachment/id 1 :attachment/filename "boing.txt"}` from the API, but that can be handled in `enrich-event` perhaps

src/clj/rems/application/commands.clj
```diff
@@ -1,6 +1,6 @@
 (ns rems.application.commands
   (:require [clojure.test :refer [deftest is testing]]
-            [medley.core]
+            [medley.core :as medley]
```
Macroz: I think I prefer using medley through refers since they are so "core"

src/clj/rems/application/commands.clj
```diff
@@ -263,6 +259,23 @@
 (defn- ok [& events]
   (ok-with-data nil events))
 
+(defn- invalid-attachments-error [injections cmd]
+  (let [invalid-ids (for [id (:attachments cmd)
+                          :let [attachment ((:get-attachment-metadata injections) id)]
```
Macroz: I think we could start using the `getx` pattern here (or introduce `callx`).

src/clj/rems/application/events.clj
Macroz: I put it like that now for `:application/forms` for the events. Maybe should do that.

test/clj/rems/api/test_applications.clj
```diff
@@ -741,15 +764,169 @@
...
+      (testing "applicant"
+        (is (= ["handler-remark.txt"
+                "handler-approve.txt"
+                "handler-close1.txt"
+                "handler-close2.txt"]
+               (mapv :attachment/filename (:application/attachments (get-application application-id applicant-id))))))
+      (testing "handler"
+        (is (= ["handler-remark.txt"
+                "reviewer-review.txt"
+                "handler-private.txt"
```
Macroz: Maybe name the `handler-remark.txt` and `handler-private.txt` as `handler-public-remark.txt` and `handler-private-remark.txt` so it's obvious that both are from remarks.

test/clj/rems/application/test_model.clj
```diff
@@ -1499,3 +1499,22 @@
...
+                                                         :field/value "1" :field/previous-value "2"}]}
+                                         {:form/fields [{:field/type :text
+                                                         :field/value "3" :field/previous-value "4"}]}]
+                     :application/events [{:event/type :application.event/remarked
+                                           :comment "4"
+                                           :attachments [5 6]}]}]
+    (is (= [{:attachment/id 1 :attachment/filename "1.txt"}
+            {:attachment/id 2 :attachment/filename "2.txt"}]
+           (:application/attachments
+            (#'model/hide-attachments application))))))
```
Macroz: Why so?

src/clj/rems/application/commands.clj
opqdonut: good point, I'll add a getx

src/clj/rems/application/commands.clj
opqdonut: fair enough

src/clj/rems/api/services/attachment.clj
```diff
@@ -11,15 +13,29 @@
...
+                            attachment-id)
+      attachment
+
+      :else
+      (throw-forbidden))))
 
 (defn add-application-attachment [user-id application-id file]
   (let [application (applications/get-application user-id application-id)]
-    (when-not (application-util/form-fields-editable? application)
+    (when-not (some #{:application.command/save-draft :application.command/remark}
```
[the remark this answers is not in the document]
opqdonut: you're looking at an old diff, it's better now

test/clj/rems/api/test_applications.clj
opqdonut: done

test/clj/rems/application/test_model.clj
opqdonut: What was I thinking when I wrote this? Fixed.

src/clj/rems/application/events.clj
opqdonut: yeah, I changed it

src/clj/rems/api/services/attachment.clj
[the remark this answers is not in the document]
Macroz: I actually wrote this comment last week and for some reason it got published now ... :D I think we discussed this in person already then.

PR #2050, opened 2020-03-10
Add automatic rahti demo rems image upgrade

vvaltia:
Signed-off-by: Valtteri Valtia <valtteri.valtia@csc.fi>

## Reviewability
- [ ] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ workflows:
               only: /.*/
             branches:
               only:
-                - rahti-demo
+                - master
       - rahti-dev:
           requires:
             - build
@@ -174,7 +174,7 @@ workflows:
           filters:
             branches:
               only:
-                - rahti-demo
+                - master
       - rahti-demo:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,7 +108,7 @@ jobs:
           paths:
             - target/uberjar/rems.jar
 
-  rahti:
+  rahti-dev:
     <<: *defaults
     steps:
       - checkout
@@ -129,6 +129,28 @@ jobs:
             -H "Authorization: Bearer $rahti_dev_pod_delete_token" \
             -H 'Accept: application/json' \
             https://rahti.csc.fi:8443/api/v1/namespaces/rems-dev/pods/${DEV_CONTAINER_NAME}
+
+  rahti-demo:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run: docker build --pull --tag docker-registry.rahti.csc.fi/rems/rems:release --tag docker-registry.rahti.csc.fi/rems/rems:$(git describe --abbrev=0 --tags) .
+      - run: docker login -p $rahtitoken -u unused docker-registry.rahti.csc.fi
+      - run: docker push docker-registry.rahti.csc.fi/rems/rems:release
+      - run: docker push docker-registry.rahti.csc.fi/rems/rems:$(git describe --abbrev=0 --tags)
+      - run: |
+          DEMO_CONTAINER_NAME=$(curl -k -s \
+            -H "Authorization: Bearer $rahti_demo_pod_delete_token" \
+            -H 'Accept: application/json' \
+            https://rahti.csc.fi:8443/api/v1/namespaces/rems-demo/pods | \
+            jq '.items | .[].metadata.name' | grep rems-demo | tr -d \")
+          curl -k -s \
+            -X DELETE \
+            -H "Authorization: Bearer $rahti_demo_pod_delete_token" \
+            -H 'Accept: application/json' \
+            https://rahti.csc.fi:8443/api/v1/namespaces/rems-demo/pods/${DEMO_CONTAINER_NAME}
 
   # pseudo job to post a single ok status to github after all the tests
   ok:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -183,14 +205,24 @@ workflows:
...
+      - rahti-demo:
+          requires:
+            - build
+            - war
+          filters:
+            tags:
+              only: /v[0-9]\.[0-9].*/
+            branches:
+              only:
+                - master
```
opqdonut: this means it will also deploy every version of master. I think you want `only: []` to disable deploying any branches to rahti-demo.

.circleci/config.yml
```diff
@@ -130,6 +130,28 @@ jobs:
             -H 'Accept: application/json' \
             https://rahti.csc.fi:8443/api/v1/namespaces/rems-dev/pods/${DEV_CONTAINER_NAME}
 
+  rahti-demo:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - run: docker build --pull --tag docker-registry.rahti.csc.fi/rems/rems:release --tag docker-registry.rahti.csc.fi/rems/rems:$(git describe --abbrev=0 --tags) .
```
opqdonut: do we really need to rebuild the image here? can't we just retag the existing one? perhaps we could have three jobs:
- rahti: builds image, pushes it to rahti, tags with various tags (e.g. release if we are on a `v9.9` tag)
- rahti-dev: updates remsdev to run latest `circle` tag
- rahti-demo: updates rems-demo to run lates `release` tag

vvaltia: I get what you mean and my logic here was off. While we could separate the build to it's own job it would complicate the process somewhat. How do we know if circle was triggered my a build or a tag? I tried to modify the logic so that only one of either rahti-dev or rahti-demo would be run. This does achieve the same end result, but arguably is not as "neat".

opqdonut: you can check if a tag points to master using e.g. `git describe --exact-match`

opqdonut: relatedly, the tagging & pushing could be in its own script under `.circle` for easier development

vvaltia: Good idea. I moved the commands to an other file.

.circleci/config.yml
```diff
@@ -172,25 +166,31 @@ workflows:
...
           requires:
             - build
             - war
           filters:
-            tags:
-              only: /.*/
             branches:
               only:
-                - master
+                - rahti-demo
```
opqdonut: should probably be master?

.circleci/rahti-ci.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
opqdonut: please consider

`setopt -xeuo pipefail`

.circleci/rahti-ci.sh
```diff
@@ -0,0 +1,46 @@
...
+
+instance=${1}
+
+if [ "${instance}" == "dev" ] ; then
+    tag1=circle
+    tag2=$(git rev-parse HEAD)
+    token=${rahti_dev_pod_delete_token}
+elif [ "${instance}" == "dev" ] ; then
```
opqdonut: should this be "demo"?

.circleci/rahti-ci.sh
```diff
@@ -0,0 +1,46 @@
...
+    tag1=release
+    tag2=$(git describe --abbrev=0 --tags)
```
opqdonut: Are you sure this is right? If there is no tag for HEAD it will give you the name of the previous reachable tag. This will override the previous image built with that tag, which might not be what we want. Would it be better to use `git describe --tags --exact-match` and fail if there is no tag?

.circleci/config.yml
vvaltia: Yes it should

.circleci/rahti-ci.sh
vvaltia: Yes. Good catch.

.circleci/rahti-ci.sh
vvaltia: Your command seems safer. But if we expect to have cases where the new tag does not point to the head of a branch, we could do something like:

tag=$(git describe --abbrev=0 --tags)
commit=$(git rev-list -n 1 $tag)
git checkout $commit
lein uberwar

Then war step might be better to split into two -> war-tag and war-master. What do you think? Or would this complicate things with no added value?

opqdonut: checking out something else than circle is building for us seems wrong. when you push a new tag `foobar`, even if it doesn't point at the head of a branch, circle will build it

vvaltia: So circle has checked out the tag for us already? Then I see no case where the tag would not point to the head of the branch. Anyways I see no downside using "git describe --tags --exact-match" command, so I'll swap to it

.circleci/rahti-ci.sh
vvaltia: Yes good idea. But should I drop the "x" so we would not print out tokens?

opqdonut: hmm, good point. I'm pretty sure circle's _secrets masking_ will take care of that...
https://circleci.com/docs/2.0/env-vars/

vvaltia: Yes seems like so, but not completly: "The value of the environment variable is still accessible to users debugging builds with SSH." But this might not be an issue for us, or is it?

opqdonut: I think that's fine. To be able to rerun a job with SSH access you need to have write access to the repository on github.

PR #2051, opened 2020-03-12
Multiple forms support

Macroz:
Support for bundling of catalogue items with same workflow but different forms. See #2043  and #2044.
# Definition of Done / Review checklist

## Reviewability
- [x] link to issue
- [ ] note if PR is on top of other PR
- [ ] note if related change in rems-deploy repo
- [x] consider adding screenshots for ease of review

## API
- [x] API is documented and shows up in Swagger UI
- [ ] API is backwards compatible or completely new
- [ ] Events are backwards compatible

## Documentation
- [x] update changelog if necessary
- [ ] add or update docstrings for namespaces and functions
- [ ] components are added to guide page
- [ ] documentation _at least_ for config options (i.e. docs folder)
- [x] ADR for major architectural decisions or experiments

## Different installations
- [ ] new configuration options added to rems-deploy repository
- [ ] instance specific translations (i.e. LBR kielivara)

## Testing
- [x] complex logic is unit tested
- [x] valuable features are integration / browser / acceptance tested automatically

## Accessibility
- [ ] all icons have the aria-label attribute
- [ ] all fields have a label
- [ ] errors are linked to fields with aria-describedby
- [ ] contrast is checked
- [ ] conscious decision about where to move focus after an action

## Follow-up
- [x] new tasks are created for pending or remaining tasks
- [ ] no critical TODOs left to implement

PR #2054, opened 2020-03-16
fix: code examples for rems.fields/guide

opqdonut:
They were not rendering as code because of the wrapping form tag. I
checked that removing the form tag doesn't affect appearance.


opqdonut:
will be part of #2051

PR #2072, opened 2020-03-23
doc: clarify & fix documentation of :extra-pages config option

### LINKED ISSUE #2070: config-default.edn contains syntaxt error
Hi,

Instead of `file` it should be `filename` I believe. `file` results in application startup errors.
https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn#L82

I'd also recommend not using highly edn specific syntax in this example file (or document it better) as most admins will likely not know what `#_ ` stands for and are confused when the settings are not applied.

opqdonut:
fixes #2070

### DIFF
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -76,16 +76,20 @@
 
  ;; Optional extra pages shown in the navigation bar.
  ;;
- ;; Example: defines an external page and a markdown page:
- ;;   [{:id "hello"
- ;;     :url "http://example.org/hello.html"
- ;;     :translations {:fi {:title "Hei"}
- ;;                    :en {:title "Hello"}}}
- ;;    {:id "about"
- ;;     :translations {:fi {:title "Info"
- ;;                         :file "about-fi.md"}
- ;;                    :en {:title "About"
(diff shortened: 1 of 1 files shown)


resources/config-defaults.edn
```diff
@@ -75,18 +75,18 @@
...
-                                 :en {:title "Hello"}}}
-
-               ;; example of a local markdown file
-               #_{:id "about"
-                  :translations {:fi {:title "Info"
-                                      :file "about-fi.md"}
-                                 :en {:title "About"
-                                      :file "about-en.md"}}}]
+ ;;
+ ;; Example: defines an external page and a markdown page:
```
Macroz: How about something like this:
`;; Example: define an external static HTML page that will be linked to`
And the other with
`;; Example: a page generated from local Markdown files`

opqdonut: I wanted to demonstrate how you can have multiple definitions in an array. I'll try to make it clearer....

Macroz: Yeah just thought that it would be clearer if you split them and described what they actually are

opqdonut: how about now?

Issue #1949, opened 2020-01-27
Improvements to Conditional fields


Macroz:
- [x] document field/id vs. field/index and data-field-index
- [x] move create form validations to API for visibility

PR #2076, opened 2020-03-24
doc: field/id vs field/index

Macroz:
Closes https://github.com/CSCfi/rems/issues/1949


src/cljs/rems/administration/create_form.cljs
```diff
@@ -1,4 +1,15 @@
 (ns rems.administration.create-form
+  "Form editor for administrators used for editing and creating new forms.
+
+  NB: Each field has a generated distinct `:field/id` so that we can track moving fields,
+  their answers and especially references between fields. The id can be given in the API,
+  or it can be generated by the backend too. We also maintain a separate `:field/index`
```
opqdonut: not quite sure what you mean by "the id can be given in the api, or it can be generated by the backend too"

perhaps:

> The field/id is optional in the API. The backend will generate an id if it is missing. However, or frontend always sends a field/id.

Macroz: How about now?

PR #2077, opened 2020-03-24
fix attachment type checking, allow excel and csv

### LINKED ISSUE #2023: Allow excel files as attachments

opqdonut:
fixes #2023

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Accessibility

## Follow-up

### DIFF
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -345,6 +345,7 @@
          :submit "Send application"
          :success "Success"
          :upload "Upload an attachment"
+         :upload-extensions "Allowed file extensions"
          :validation {:invalid-email "Invalid email address."
                       :invalid-user "Invalid user."
                       :invalid-value "Invalid value."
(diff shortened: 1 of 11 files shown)


resources/translations/en.edn
```diff
@@ -331,6 +331,7 @@
          :events "Events"
          :failed "Failed"
          :has-accepted-licenses "Terms of use accepted."
+         :invalid-attachment "Attachment not acceptable"
```
Macroz: Would be nice to list here the allowed types. 

Also isn't it more like "Attachment type not acceptable"? Also there is now attachment being acceptable and allowed file extensions. How about unifying and using "Acceptable file types" and "Attachment type not acceptable".

src/cljs/rems/fields.cljs
```diff
@@ -271,7 +273,20 @@
       {:id id
        :type :button
        :on-click (fn [e] (.click (.getElementById js/document upload-id)))}
-      (text :t.form/upload)]]))
+      (text :t.form/upload)]
+     [:a.application-search-tips.btn.btn-link.collapsed
```
Macroz: Doesn't seem like the content is focused after opening. Comparing to the search info.

PR #2084, opened 2020-03-25
copy attachments when copying application

### LINKED ISSUE #2056: "copy as new" loses attachments

opqdonut:
fixes #2056

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -498,6 +498,23 @@
                                    {:event/type :application.event/member-uninvited
                                     :application/member (:member cmd)})))
 
+(defn- copied-field-values [application new-application-id injections]
+  (vec
+   (for [form (:application/forms application)
+         field (:form/fields form)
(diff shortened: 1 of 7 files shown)


src/clj/rems/application/commands.clj
```diff
@@ -509,17 +526,14 @@
       created-event-or-errors
       (let [created-event (:event created-event-or-errors)
             old-app-id (:application/id application)
-            new-app-id (:application/id created-event)]
+            new-app-id (:application/id created-event)
+            values (copied-field-values application new-app-id injections)]
```
Macroz: Hmm should this rather be named `copy-field-values!`

Macroz: Maybe it would also make sense to get the `copy-attachment!` injection out so that it's obvious which fn the `new-app-id` is for. It would also make it much clearer what injection is needed here and why.

Issue #1319, opened 2019-06-05
Implement ELIXIR Permissions API


jaakkocsc:
Implement the common permissions endpoint of the ELIXIR Permissions API.

It should be a new API endpoint, `/api/permissions`, which has
- [x] a GET method similar to existing entitlements endpoint outputting GA4GH Visas
- [x] a POST method for pushing new entitlements as GA4GH Visas to a configurable receiver

GA4GH Passport specification defines the [format of Visas](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#passport-visa). In the context of the [specification overview](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#overview), REMS is considered to combine functionalities of Passport Visa Assertion Repository and Passport Visa Issuer.

hukka:
https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/RI_Claims_V1.md is a bit different from the swagger; authoriser is not used, and there's a possibility to list accepted licences.

This would need some discussion with a domain expert. I suppose that the source could be https://grid.ac/institutes/grid.20709.3c for LBR (that's CSC's grid URL) and https://grid.ac/institutes/grid.14758.3f for THL. by would be "dac", I suppose. Biggest hurdles are:
- linking users with the Elixir ID — I suppose we need #1234 for that
- figuring out URLs for resources, as currently we have just resIds
- deciding what we will use as expiry dates, as now we only use them when revoking entitlements
- do we need to list conditions that bind the access not just to the user's ID, but their status in their organization (e.g. if they stop being faculty, they won't have access)

jaakkocsc:
Any differences from the RI spec in the swagger is an add-on feature requested by European Genome-Phenome Archive, and should be ignored — such as authoriser field. REMS should only implement specification for maximum compatibility.

opqdonut:
see also #794

opqdonut:
Waiting for the spec to stabilize. DL for new spec was 2019-08-20.

jaakkocsc:
Current:
`
curl -X GET http://rems/api/entitlements -H "x-rems-api-key: demo" -H "x-rems-user-id: 08ea84fde95aca412f572d7b84823391beed05e8@elixir-europe.org"
[
{
"resource": "urn:hg:example-controlled",
"application-id": 11,
[...]

PR #2092, opened 2020-03-30
Disable permissions api

foxlynx:
Related to #1319 

## Reviewability
- [ ] link to issue

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up

### DIFF
--- a/src/clj/rems/api/permissions.clj
+++ b/src/clj/rems/api/permissions.clj
@@ -22,6 +22,6 @@
       :return GetPermissionsResponse
       (if (not (:enable-permissions-api config/env))
         (not-implemented "permissions api not implemented")
-        (do (when-not (empty? user)
-              (api-util/not-found-json-response))
-            (ok (entitlements/get-entitlements-for-permissions-api user nil expired)))))))
+        (if (empty? user)
+          (api-util/not-found-json-response)
+          (ok (entitlements/get-entitlements-for-permissions-api user nil expired)))))))
--- a/src/clj/rems/api/permissions.clj
+++ b/src/clj/rems/api/permissions.clj
@@ -15,7 +15,7 @@
     :tags ["permissions"]
 
     (GET "/:user" []
-      :summary "Returns user's permissions in ga4gh visa format. Currently signed with fake key. See https://github.com/ga4gh-duri/ga4gh-duri.github.io/"
+      :summary "Experimental. Returns user's permissions in ga4gh visa format. Currently signed with fake key. See https://github.com/ga4gh-duri/ga4gh-duri.github.io/"
       :roles #{:handler :owner}
       :path-params [user :- (describe s/Str "return permissions for this user, required")]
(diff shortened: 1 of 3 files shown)


src/clj/rems/api/permissions.clj
```diff
@@ -14,11 +15,13 @@
     :tags ["permissions"]
 
     (GET "/:user" []
-      :summary "Returns user's permissions in ga4gh visa format. See https://github.com/ga4gh-duri/ga4gh-duri.github.io/"
+      :summary "Returns user's permissions in ga4gh visa format. Currently signed with fake key. See https://github.com/ga4gh-duri/ga4gh-duri.github.io/"
```
opqdonut: should the config key be mentioned here?

test/clj/rems/api/test_permissions.clj
```diff
@@ -59,6 +62,16 @@
             body (read-body response)]
         (is (= "forbidden" body))))))
 
+(deftest permissions-test-api-disabled
+  (with-redefs [rems.config/env (assoc rems.config/env :enable-permissions-api false)]
```
opqdonut: the key should be added to `config-defaults.edn` for documentation

src/clj/rems/api/permissions.clj
foxlynx: I would think no, it is not mentioned in the other experimental endpoint either that is enabled/disabled through config file.

opqdonut: well at least add a "(EXPERIMENTAL)" or something

src/clj/rems/api/permissions.clj
```diff
@@ -14,11 +15,13 @@
...
       :path-params [user :- (describe s/Str "return permissions for this user, required")]
       :query-params [{expired :- (describe s/Bool "whether to include expired permissions") false}]
       :return GetPermissionsResponse
-      (when-not (empty? user)
-        (api-util/not-found-json-response))
-      (ok (entitlements/get-entitlements-for-permissions-api user nil expired)))))
+      (if (not (:enable-permissions-api config/env))
+        (not-implemented "permissions api not implemented")
+        (do (when-not (empty? user)
+              (api-util/not-found-json-response))
```
opqdonut, line 26: Heh, I just noticed that this `when` does nothing. `(api-util/not-found-json-response)` is a value which just gets thrown away. You probably meant:

```
(if-not (empty? user)
  (api-util/not-found-json-response)
  (ok (entitlements/get-entitlements-for-permissions-api user nil expired)))
```

but even that has the condition flipped around, so probably

```
(if (empty? user)
  (api-util/not-found-json-response)
  (ok (entitlements/get-entitlements-for-permissions-api user nil expired)))
```

On the other hand, since this `when-not` is not doing anything and the tests pass, you can probably just remove it :sunglasses:

PR #2119, opened 2020-04-14
fix: truncate long attachment filenames in UI

### LINKED ISSUE #2118: Long filename prevents removing attachment from an application
If an attachment has a very long filename the "Remove attachment" button is rendered under the filename and the attachment cannot be removed from the application

opqdonut:
fixes #2118

## Reviewability
- [x] link to issue

## Documentation

## Follow-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ have notable changes.
 
 Changes since v2.11
 
+### Fixes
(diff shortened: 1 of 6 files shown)


opqdonut:
Here's what the a long action attachment filename looks like at two different zoom levels:

long1
long2

PR #2141, opened 2020-04-24
make footer work with longer content, clean up styles

opqdonut:

## Reviewability

### DIFF
--- a/src/cljs/rems/spa.cljs
+++ b/src/cljs/rems/spa.cljs
@@ -271,18 +271,18 @@
 
   Useful while developing and changing e.g. translations."
   []
-  [:button.btn.btn-secondary.btn-sm
-   {:on-click #(do (fetch-translations!)
-                   (fetch-theme!)
-                   (config/fetch-config!))}
-   [:i.fas.fa-redo]])
+  [:div.dev-reload-button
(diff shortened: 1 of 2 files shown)


opqdonut:
## Normal footer
normal-footer

## Footer with lots of content
big-footer

src/cljs/rems/spa.cljs
```diff
@@ -280,11 +280,10 @@
...
   [:footer.footer
    [:div.container
-    [:div.navbar
-     [:div.navbar-text (text :t/footer)]
-     (when (config/dev-environment?)
-       [:div.dev-only
-        [dev-reload-button]])]]])
+    (when (config/dev-environment?)
+      [:div.dev-reload-button.dev-only
+       [dev-reload-button]])
```
Macroz, line 285: I think it doesn't make sense to wrap `dev-reload-button` in a use site to `.dev-reload-button`, it should be instead rolled into the button component.

opqdonut: Do you think I should keep the `dev-only` class? It doesn't seem to have a use right now.

opqdonut: I removed it too.

PR #2143, opened 2020-04-27
feat: validate :default-language config option

opqdonut:

## Reviewability
- [ ] link to issue

## Testing

### DIFF
--- a/src/clj/rems/config.clj
+++ b/src/clj/rems/config.clj
@@ -52,7 +52,7 @@
   (when-let [url (:public-url config)]
     (assert (.endsWith url "/")
             (str ":public-url should end with /:" (pr-str url))))
-  (assert (some (partial = (:default-language config)) (:languages config))
+  (assert (contains? (set (:languages config)) (:default-language config))
           (str ":default-language should be one of :languages: "
                (pr-str (select-keys config [:default-language :languages]))))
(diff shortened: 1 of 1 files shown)


opqdonut:
example:
```
Execution error (AssertionError) at rems.config/validate-config (config.clj:55).
Assert failed: :default-language should be one of :languages: {:default-language :sv, :languages [:en :fi]}
(some (partial = (:default-language config)) (:languages config))
```

src/clj/rems/config.clj
```diff
@@ -52,6 +52,9 @@
   (when-let [url (:public-url config)]
     (assert (.endsWith url "/")
             (str ":public-url should end with /:" (pr-str url))))
+  (assert (some (partial = (:default-language config)) (:languages config))
```
Macroz, line 55: I like the idiom `#{(:default-language config)}` for these.

opqdonut: I actually had that at first, but since the expression gets printed in the exception, I switched to the "clearer" `partial =`. Now I'm not so sure...

opqdonut:
This is perhaps more readable:
```
Execution error (AssertionError) at rems.config/validate-config (config.clj:55).
Assert failed: :default-language should be one of :languages: {:default-language :sv, :languages [:en :fi]}
(contains? (set (:languages config)) (:default-language config))
```

PR #2144, opened 2020-04-27
audit log

### LINKED ISSUE #2057: Viewing an application should be logged as an event
Currently, only actions are logged as events on an application. However, it would be necessary to log also who has opened the application just to view it (handlers, specifically) in order to maintain a good audit log.

opqdonut:
for #2057

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/resources/migrations/20200427092452-audit-log.down.sql
+++ b/resources/migrations/20200427092452-audit-log.down.sql
@@ -1,0 +1,1 @@
+DROP TABLE audit_log CASCADE;
--- a/src/clj/rems/api.clj
+++ b/src/clj/rems/api.clj
@@ -97,7 +97,7 @@
 
 ;; This should be run outside transaction-middleware since we want to
 ;; write even on GET queries. We're only running one insert statement
-;; so we don't need a separte transaction for logging.
+;; so we don't need a separate transaction for logging.
 (defn audit-log-middleware [handler]
   (fn [request]
     (let [response (handler request)]
--- a/src/clj/rems/api.clj
+++ b/src/clj/rems/api.clj
@@ -1,5 +1,6 @@
 (ns rems.api
-  (:require [clojure.stacktrace :refer [print-cause-trace]]
+  (:require [clj-time.core :as time]
+            [clojure.stacktrace :refer [print-cause-trace]]
             [clojure.tools.logging :as log]
             [compojure.api.exception :as ex]
             [compojure.api.sweet :refer :all]
@@ -20,7 +21,10 @@
             [rems.api.user-settings :refer [user-settings-api]]
             [rems.api.users :refer [users-api]]
             [rems.api.workflows :refer [workflows-api]]
+            [rems.auth.auth :as auth]
+            [rems.db.core :as db]
             [rems.json :refer [muuntaja]]
+            [rems.util :refer [get-user-id]]
             [ring.middleware.cors :refer [wrap-cors]]
             [ring.util.http-response :refer :all]
             [ring.util.response :as response]
@@ -90,6 +94,23 @@
 (defn- read-only? [request]
   (not (contains? #{:put :post} (:request-method request))))
 
+;; This should be run outside transaction-middleware since we want to
+;; write even on GET queries. We're only running one insert statement
+;; so we don't need a separte transaction for logging.
+(defn audit-log-middleware [handler]
+  (fn [request]
+    (let [response (handler request)]
+      (try
+        (db/add-to-audit-log! {:time (time/now)
+                               :path (:uri request)
+                               :method (name (:request-method request))
+                               :apikey (auth/get-api-key request)
+                               :userid (get-user-id)
+                               :status (str (:status response))})
+        (catch Throwable t
+          (log/error "Adding to audit log failed:" t)))
+      response)))
+
 (defn transaction-middleware [handler]
   (fn [request]
     (conman/with-transaction [rems.db.core/*db* {:isolation :serializable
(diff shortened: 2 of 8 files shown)


resources/migrations/20200427092452-audit-log.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE audit_log CASCADE;
```
Macroz, line 1: I wonder if we should just stop writing these (potentially data lossing) down migrations especially for high value data.

src/clj/rems/api.clj
```diff
@@ -90,6 +95,23 @@
 (defn- read-only? [request]
   (not (contains? #{:put :post} (:request-method request))))
 
+;; This should be run outside transaction-middleware since we want to
+;; write even on GET queries. We're only running one insert statement
+;; so we don't need a separte transaction for logging.
```
Macroz, line 100: Typo

src/clj/rems/api.clj
```diff
@@ -101,7 +123,10 @@
   request)
 
 (def api-routes
-  (api
+  ;; we wrap audit-log-middleware outside compojure-api since we want
+  ;; to see the status codes produced by our exception handlers
+  (audit-log-middleware
```
Macroz, line 128: I wonder if it should be a config option to audit log or not?

Macroz: Possibly need to implement some configurable deleting / vacuuming of old data.

test/clj/rems/api/test_api.clj
```diff
@@ -1,6 +1,7 @@
 (ns ^:integration rems.api.test-api
   (:require [clojure.test :refer :all]
             [rems.api.testing :refer :all]
+            [rems.db.core :as db]
```
Macroz, line 4: Why is this added?

test/clj/rems/api/test_audit_log.clj
```diff
@@ -0,0 +1,110 @@
...
+                                              (header "Cookie" cookie)
+                                              (header "x-csrf-token" csrf)
+                                              (json-body {:id 9999999 :archived true})
+                                              handler))))))))
+    (testing "user can't get log"
+      (is (response-is-forbidden? (-> (request :get "/api/audit-log")
+                                      (authenticate "42" "alice")
+                                      handler))))
+    (testing "reporter can get log"
+      (is (= [{:userid nil :apikey nil :method "get" :path "/api/unknown" :status "404"}
```
Macroz, line 70: One of the acceptance test cases is that can the reporter see who has viewed a specific application? I don't think this is covered here yet. Moreover, isn't it a requirement that someone can figure out who has seen someone's data? How would one figure out which applications a person is in, and then filter from this log all the people who have seen such an application? That could be its own API I guess.

resources/migrations/20200427092452-audit-log.down.sql
opqdonut: That's a good question. However we use migratus reset to empty the db in some tests, so perhaps a middle ground would be to rename the table here?

src/clj/rems/api.clj
opqdonut: Cleaning data is luckily easy to do on the DB level. Let's wait for what the users need. I'm a bit hesitant to add auto-cleaning to REMS, as e.g. a system clock glitch might wipe the whole audit log.

test/clj/rems/api/test_api.clj
opqdonut: Leftover from earlier commit, removed.

test/clj/rems/api/test_audit_log.clj
opqdonut: Yeah some further work on querying might be needed. Let's discuss and make a new issue? I think this feature makes sense to merge already in this state.

resources/migrations/20200427092452-audit-log.down.sql
Macroz: I guess for tests it may make sense to have some other reset function that is faster i.e. truly start from an empty DB or db dump. I think this can be left here for now.

src/clj/rems/api.clj
Macroz: It could be provided as an ops script sometime later.

test/clj/rems/api/test_audit_log.clj
Macroz: Ok as long as the concern is in some task on the board.

opqdonut: I created #2147, and added filtering by application to this PR

test/clj/rems/api/test_audit_log.clj
```diff
@@ -29,8 +31,15 @@
             (testing "> authorized"
               (is (= 200 (:status (-> (request :get "/api/users/active")
                                       (authenticate "42" "owner")
+                                      handler)))))
+            (reset! time-a (time/now))
+            (testing "> application"
+              (is (= 200 (:status (-> (request :get (str "/api/applications/" app-id))
```
Macroz, line 37: Should we prefer `api-call` and `api-response` for these in the future? We also have the `(true? :success)` and `(= {:success true}` variants. Could add some idea task to unify these a bit.

opqdonut: I think here it's more natural to use the explicit forms since some calls lack `authenticate`. I'll clean up a bit otherwise.

PR #2167, opened 2020-05-06
Browser tests

Macroz:
Improved browser tests with easier to use for development.

Going to update changelog, document its use in docs in this PR but please have a look already so I can correct any leftover code issues.

I will implement the last tests in the organization PR as they touch most of the admin UI a little bit.

## Documentation

### DIFF
--- a/test/clj/rems/browser_test_util.clj
+++ b/test/clj/rems/browser_test_util.clj
@@ -42,6 +42,7 @@
          " "
          (mod-nth ["leopard" "gorilla" "turtle" "orangutan" "elephant" "saola" "vaquita" "tiger" "rhino" "pangolin"] (mod t 123)))))
 
+;; TODO these could use more of our wrapped fns if we reordered
 (defn init-driver!
   "Starts and initializes a driver. Also stops an existing driver.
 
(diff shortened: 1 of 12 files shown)


test/clj/rems/test_browser.clj
```diff
@@ -315,83 +424,170 @@
...
+      (is (= {"Organization" "nbn"
+              "Title (EN)" (str (:license-name @test-context) " EN")
+              "Title (FI)" (str (:license-name @test-context) " FI")
+              "Title (SV)" (str (:license-name @test-context) " SV")
+              "Type" "link"
+              "External link (EN)" "https://www.csc.fi/home"
+              "External link (FI)" "https://www.csc.fi/etusivu"
+              "External link (SV)" "https://www.csc.fi/home"
+              "Active" ""}
+             (slurp-fields driver :license))))))
```
opqdonut, line 555: nice!

test/clj/rems/test_browser.clj
```diff
@@ -315,83 +424,170 @@
...
+              "Active" ""}
+             (slurp-fields driver :resource)))
+      (is (= (str "License \"" (:license-name @test-context) " EN\"")
+             (et/get-element-text driver [:licenses {:class :license-title}]))))))
+
+(deftest test-create-catalogue-item
+  (let [driver (get-driver)]
+    (et/with-postmortem driver {:dir reporting-dir}
+      (login-as driver "owner")
+      (swap! test-context assoc
```
opqdonut, line 584: since you're actually calling `test-create-license`, why not just pass the license name in?

test/clj/rems/test_browser.clj
```diff
@@ -1,45 +1,100 @@
...
+    (io/delete-file file true)))
+
+(def reporting-dir
+  (doto (io/file "browsertest-errors")
+    (.mkdirs)
+    (delete-files)))
+
+(defn- random-seed []
+  (let [t (System/currentTimeMillis)
+        counts ["one" "two" "three" "four" "five" "six" "seven" "eight" "nine" "ten"]
```
opqdonut, line 51: why not something like
```
(defn random-element [rng v]
  (get v (.nextInt (count v))))
```

test/clj/rems/test_browser.clj
```diff
@@ -1,45 +1,100 @@
...
+        count-i (mod (Math/floor (/ t 60000)) (count counts))
+        adjectives ["amusing" "comic" "funny" "laughable" "hilarious" "witty" "jolly" "silly" "ludicrous" "wacky"]
+        adjectives-i (mod (Math/floor (/ t 1000)) (count adjectives))
+        animals ["leopard" "gorilla" "turtle" "orangutan" "elephant" "saola" "vaquita" "tiger" "rhino" "pangolin"]
+        animals-i (mod (/ t 123) (count animals))]
+    (str/join [(nth counts count-i)
```
opqdonut, line 57: [golf] couldn't this be just `(str x " " y " " z (if ...))`?

test/clj/rems/test_browser.clj
```diff
@@ -153,157 +253,166 @@
...
-                      ["label" ""]
-                      ["text" ""]
-                      ["texta" ""]]
-                     (for [field (select [:application/forms ALL :form/fields ALL] application)]
-                       ;; TODO could test other fields here too, e.g. title
-                       [(:field/type field)
-                        (:field/value field)]))))
-            (testing "after navigating to the application view again"
-              (scroll-and-click *driver* [{:css "table.my-applications"}
+  (let [driver (get-driver)]
```
opqdonut, line 307: I can see the need for consistently passing in the driver as the first argument for everything
... but I'm not sure if `(let [driver (get-driver)] (doto driver ...))` is any better than `(doto *driver* ...)`

Did you consider wrapping all etaoin functions so that they get a `*driver*` parameter? This is what we do with `*db*` and it works nicely enough.

Not insisting on any changes, just thinking out loud.

test/clj/rems/test_browser.clj
```diff
@@ -315,83 +424,170 @@
...
+(defn test-create-license []
+  (let [driver (get-driver)]
+    (et/with-postmortem driver {:dir reporting-dir}
+      (doto driver
+        (go-to-admin-licenses)
+        (scroll-and-click  :create-license)
+        (et/wait-visible  {:tag :h1 :fn/text "Create license"})
+        (select-option  "Organization" "nbn")
+        (scroll-and-click :licensetype-link)
+        (fill-form-field "License name" (str (:license-name @test-context) " EN") {:index 1})
```
opqdonut, line 534: do you just want to demo the `:index` feature here or is there a reason you can't `(fill-form-field "License name (EN)" ...)`?

test/clj/rems/test_browser.clj
Macroz: I was trying to simplify calling these functions manually one by one. Let's test and see if this is workable or not.

test/clj/rems/test_browser.clj
Macroz: Like discussed in the meeting, I will dump the utilities into their own namespace and then refactor it so that we can use a dynamic driver again and don't need the `doto` or `let` `get-driver` anymore. It will be much better.

test/clj/rems/test_browser.clj
Macroz: It's always called "License name" since the language is in the header before this spot. This is because the type selection affects the content of each place. A possible refactoring of the screen is to not use the header but embed the language in each label instead. Then there is no need for the index. I'll not do this yet.

test/clj/rems/browser_test_util.clj
```diff
@@ -0,0 +1,150 @@
...
+  (:require [clj-http.client :as http]
+            [clojure.java.io :as io]
+            [clojure.test :refer :all]
+            [clojure.tools.logging :as log]
+            [etaoin.api :as et]
+            [medley.core :refer [assoc-some]]
+            [rems.api.testing :refer [standalone-fixture]]
+            [rems.config]
+            [rems.standalone]
+            [rems.browser-test-util :as btu])
```
opqdonut, line 14: would `browser/` or perhaps `b/` be a better prefix?

Issue #2052, opened 2020-03-12
Workflow forms


Macroz:
- [x] allow selection of form for a workflow
- ~support many workflow forms in data model, but support only one in the admin UI~
  - turns out UI support for many forms was easy
- [x] workflow form should be the first form for an application
- [x] use description field from first form
- [x] only allow setting workflow form when creating workflow, no need to edit
- [x] don't allow editing a form if its used by a workflow

PR #2178, opened 2020-05-13
refactor administration dependency tracking

opqdonut:
on top of #2172

part of #2052 

## Reviewability
- [x] link to issue

## API
- [x] API is backwards compatible

## Documentation

## Testing

## Follow-up


src/clj/rems/api/services/dependencies.clj
```diff
@@ -0,0 +1,131 @@
...
+(defn dependencies []
+  (when-not @dependencies-cache
+    (reset! dependencies-cache (compute-dependencies)))
+  @dependencies-cache)
+
+(defn reset-cache! []
+  (reset! dependencies-cache nil))
+
+;; TODO change format of errors so we can get rid of this conversion
+(defn- format-deps [deps]
```
Macroz, line 72: Consider naming this like `format-deps-for-errors`

src/clj/rems/api/services/dependencies.clj
```diff
@@ -0,0 +1,131 @@
...
+  (merge dep
+         (select-keys (enrich-dependency dep) [:archived :enabled])))
+
+(defn- only-id [item]
+  (select-keys item [:resource/id :license/id :catalogue-item/id :form/id :workflow/id]))
+
+(defn- list-to-maps [lst]
+  {:dependencies
+   (map-vals (comp set (partial map :to)) (group-by :from lst))
+   :reverse-dependencies
```
Macroz, line 46: Could better names for these be `depends-on` and `used-by` or something along those lines?

src/clj/rems/api/services/dependencies.clj
```diff
@@ -0,0 +1,131 @@
...
+       (when workflows
+         [{:type :t.administration.errors/workflow-archived
+           :workflows workflows}])
+       (when forms
+         [{:type :t.administration.errors/form-archived
+           :forms forms}])
+       ;; case not possible:
+       (when catalogue-items
+         [{:type :catalogue-items-archived
+           :catalogue-items catalogue-items}])))))
```
Macroz, line 131: This feels like code that can perhaps be removed? At least having separate cases for all the different types could perhaps be removed by restructuring the errors. How about we had a common `:type` for all archived errors? The items themselves could tell which kinds were wrong. I think the UI will just display a list of all items and does not care about grouping by types.

src/clj/rems/api/services/form.clj
```diff
@@ -133,9 +121,11 @@
...
 (defn set-form-archived! [{:keys [id archived]}]
   (util/check-allowed-organization! (:form/organization (get-form-template id)))
-  (let [catalogue-items (catalogue-items-for-form id false)]
-    (or (and archived (form-in-use-error id false))
-        (do
-          (db/set-form-template-archived! {:id id
-                                           :archived archived})
-          {:success true}))))
+  (if-let [errors (when archived
+                    (dependencies/archive-errors :t.administration.errors/form-in-use {:form/id id}))]
```
Macroz, line 125: I don't understand why the error key needs to be sent here.

src/clj/rems/api/services/licenses.clj
```diff
@@ -23,6 +24,8 @@
                                         :title (:title localization)
                                         :textcontent (:textcontent localization)
                                         :attachmentId (:attachment-id localization)}))
+    ;; reset-cache! not strictly necessary since forms don't depend on anything, but here for consistency
```
Macroz, line 27: `forms -> licenses`

src/clj/rems/api/services/resource.clj
```diff
@@ -18,6 +19,8 @@
     (doseq [licid licenses]
       (db/create-resource-license! {:resid id
                                     :licid licid}))
+    ;; reset-cache! not strictly necessary since forms don't depend on anything, but here for consistency
```
Macroz, line 22: `forms -> resources`

src/clj/rems/api/services/resource.clj
```diff
@@ -28,25 +31,12 @@
...
-        {:success true}))))
+  (if-let [errors (if archived
+                    (dependencies/archive-errors :t.administration.errors/resource-in-use {:resource/id id})
+                    (dependencies/unarchive-errors {:resource/id id}))]
+    {:success false
+     :errors errors}
+    (do
+      (db/set-resource-archived! {:id id
+                                  :archived archived})
+      {:success true})))
```
Macroz, line 42: These lines of code are not quite consistent between the different items but I wonder if that is necessary. Seems like there is some boilerplate here that could be removed.

src/clj/rems/api/services/workflow.clj
```diff
@@ -43,47 +45,28 @@
   (applications/reload-cache!)
   {:success true})
 
+(defn get-workflow [id]
+  (workflow/get-workflow id))
+
+(defn get-workflows [filters]
+  (workflow/get-workflows filters))
+
```
Macroz, line 53: These functions exist in master now?

test/clj/rems/api/services/test_catalogue.clj
```diff
@@ -118,8 +118,9 @@
       (archive-resource! true)
       (archive-license! true)
       (let [errors (:errors (archive-catalogue-item! false))]
+        ;; TODO indirect catalogue item -> resource -> license dep not tracked right now
```
Macroz, line 121: I guess it is quite easy to just expand the transitive dependencies? Instead of using the `(dependencies)` as a direct map you could calculate them on the fly in a helper function or just expand them directly to the cached map?

src/clj/rems/api/services/dependencies.clj
```diff
@@ -0,0 +1,131 @@
...
+         (select-keys (enrich-dependency dep) [:archived :enabled])))
+
+(defn- only-id [item]
+  (select-keys item [:resource/id :license/id :catalogue-item/id :form/id :workflow/id]))
+
+(defn- list-to-maps [lst]
+  {:dependencies
+   (map-vals (comp set (partial map :to)) (group-by :from lst))
+   :reverse-dependencies
+   (map-vals (comp set (partial map :from)) (group-by :to lst))})
```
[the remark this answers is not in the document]
opqdonut: Yeah build-index collects only the first hit, I want all. I'll consider it.

src/clj/rems/api/services/dependencies.clj
opqdonut: I had a TODO about that but ended up not doing it for some reason. It's a good idea.

src/clj/rems/api/services/form.clj
opqdonut: This refactoring kept the errors as-is, and the keys were different in different service files...

test/clj/rems/api/services/test_catalogue.clj
opqdonut: Yes, but I didn't want to change the behaviour (yet)

src/clj/rems/api/services/dependencies.clj
opqdonut: as discussed, I'll do this in a follow-up PR

src/clj/rems/api/services/dependencies.clj
opqdonut: I considered that, but dependencies/reverse dependencies are standard terms and also match the name of the ns nicely.

Macroz: One thing to consider would be also to avoid using the map directly so you can arrange it differently after the fact. Of course if all the uses are in this ns and not public it'll be easy anyway.

Macroz: I was mainly thinking about how to read the uses of the map.

test/clj/rems/api/services/test_catalogue.clj
Macroz: We can definitely do that in a follow-up PR.

src/clj/rems/api/services/dependencies.clj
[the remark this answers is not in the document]
opqdonut: done!

src/clj/rems/api/services/resource.clj
opqdonut: Yeah now the code depends on whether unarchive-errors are possible. I'll refactor this in the next PR when I simplify the error structures.

src/clj/rems/api/services/form.clj
Macroz: Maybe the other part is that the merge with the `:type key` could happen here and not in `archive-errors`?

PR #2199, opened 2020-05-25
validation for /api/workflows/create

### LINKED ISSUE #2182: validation in workflow create API
- [ ] check that handlers exist
- [ ] check that forms exist

opqdonut:
fixes #2182

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/services/workflow.clj
+++ b/src/clj/rems/api/services/workflow.clj
@@ -15,20 +15,28 @@
 (def ^:private validate-workflow-body
   (s/validator workflow/WorkflowBody))
 
+(defn invalid-forms-error [forms]
+  (let [invalid (seq (filter (comp nil? form/get-form-template :form/id) forms))]
+    (when invalid
+      {:success false
+       :errors [{:type :invalid-form
+                 :forms invalid}]})))
+
 (defn create-workflow! [{:keys [user-id organization type title handlers forms]}]
   (util/check-allowed-organization! organization)
-  (let [body {:type type
-              :handlers handlers
-              :forms forms} ;; TODO missing validation for handlers and forms, see #2182
-        id (:id (db/create-workflow! {:organization (:organization/id organization)
-                                      :owneruserid user-id
-                                      :modifieruserid user-id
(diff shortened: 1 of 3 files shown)


src/clj/rems/api/services/workflow.clj
```diff
@@ -15,20 +15,28 @@
 (def ^:private validate-workflow-body
   (s/validator workflow/WorkflowBody))
 
+(defn invalid-forms-error [forms]
+  (let [invalid (seq (filter (comp nil? form/get-form-template :form/id) forms))]
```
Macroz: This is quite a mouthful even though it's basically right. Perhaps `invalid` should be named `invalid-forms`?

src/clj/rems/api/services/workflow.clj
```diff
@@ -15,20 +15,28 @@
...
+    (when invalid
+      {:success false
+       :errors [{:type :invalid-form
+                 :forms invalid}]})))
+
 (defn create-workflow! [{:keys [user-id organization type title handlers forms]}]
   (util/check-allowed-organization! organization)
```
Macroz, line 33: I wonder if we should put this inside the `or` though it's more of the assert type. Perhaps instead of `check-allowed-organization! it should be called `assert-allowed-organization!` or something.

src/clj/rems/api/services/workflow.clj
Macroz: Perhaps this could be written with `remove` and not `nil?` to simplify? (see the user one)

Macroz: Oh you did it in the later commits ... :+1:

src/clj/rems/api/services/workflow.clj
opqdonut: It's outside the `or` in lots of places like `rems.api.services.form`. I think the `check-` name was chosen to be consistent with `check-user` and `check-roles`. I think `assert-` would sound too severe.

Issue #2189, opened 2020-05-19
Browser tests for admin screens


Macroz:
- [x] create license, view licenses, view license
- [x] create resource, view resources, view resource
- [x] create catalogue item, view catalogue items, view catalogue item
- [x] create workflow, view workflows, view workflow
- [x] edit workflow
- [x] create form, view forms, view form
- [x] create form with all field types
- [x] edit form
- [x] blacklist
- [x] report smoke test
- [ ] archiving/disabling logic?
- [x] view categories, create category, edit category

Macroz:
See also #2218 that is for finishing the report browser test

Macroz:
We decided not to cover each enable/archive function in browser tests. We have coverage of the catalogue items, because they need to be enabled before they can be used. The others do not seem to add enough value as browser tests.

PR #2203, opened 2020-05-26
more admin browser tests

opqdonut:
for #2189

- creating catalogue items & dependencies
- form editor

## Reviewability
- [x] link to issue

## Testing

## Follow-up

### DIFF
--- a/test/clj/rems/test_browser.clj
+++ b/test/clj/rems/test_browser.clj
@@ -547,7 +547,7 @@
       (go-to-admin "Catalogue items")
       (btu/scroll-and-click :create-catalogue-item)
       (select-option "Organization" "nbn")
-      (fill-form-field "Name" (str (btu/context-get :catalogue-item-name) " EN") {:index 1})
+      (fill-form-field "Name" (btu/context-get :catalogue-item-name) {:index 1})
       (fill-form-field "Name" (str (btu/context-get :catalogue-item-name) " FI") {:index 2})
       (fill-form-field "Name" (str (btu/context-get :catalogue-item-name) " SV") {:index 3})
       (select-option "Workflow" (btu/context-get :workflow-name))
@@ -560,7 +560,7 @@
       (btu/screenshot (io/file btu/reporting-dir "created-catalogue-item.png"))
       (is (str/includes? (btu/get-element-text {:css ".alert-success"}) "Success"))
       (is (= {"Organization" "NBN"
-              "Title (EN)" (str (btu/context-get :catalogue-item-name) " EN")
+              "Title (EN)" (btu/context-get :catalogue-item-name)
               "Title (FI)" (str (btu/context-get :catalogue-item-name) " FI")
               "Title (SV)" (str (btu/context-get :catalogue-item-name) " SV")
               "More info (EN)" ""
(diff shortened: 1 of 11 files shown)


test/clj/rems/test_browser.clj
```diff
@@ -72,6 +72,24 @@
...
+  (btu/wait-page-loaded)
+  (btu/screenshot (io/file btu/reporting-dir "administration-forms-page.png")))
+
+(defn go-to-admin-workflows []
+  (click-administration-menu "Workflows")
+  (btu/wait-visible {:tag :h1 :fn/text "Workflows"})
+  (btu/wait-page-loaded)
+  (btu/screenshot (io/file btu/reporting-dir "administration-workflows-page.png")))
+
+(defn go-to-admin-catalogue []
```
Macroz, line 87: Perhaps we could use a slightly more generic helper that would be like `(go-to-admin "Workflows")` that would screenshot `administration-workflows-page.png`?

test/clj/rems/test_browser.clj
```diff
@@ -489,15 +508,120 @@
...
+              "Title (EN)" (str (btu/context-get :catalogue-item-name) " EN")
+              "Title (FI)" (str (btu/context-get :catalogue-item-name) " FI")
+              "Title (SV)" (str (btu/context-get :catalogue-item-name) " SV")
+              "More info (EN)" ""
+              "More info (FI)" ""
+              "More info (SV)" ""
+              "Workflow" (btu/context-get :workflow-name)
+              "Resource" (btu/context-get :resid)
+              "Form" (btu/context-get :form-name)
+              "Active" ""
```
Macroz, line 594: Would be nice to add some helper to slurp to load something based on the icons here.

test/clj/rems/test_browser.clj
```diff
@@ -489,15 +508,120 @@
...
+              "End" ""}
+             (dissoc (slurp-fields :catalogue-item)
+                     "Start")))
+      (btu/scroll-and-click {:fn/text "Enable"})
```
Macroz, line 598: This enabling could also be put outside this function.

test/clj/rems/test_browser.clj
```diff
@@ -489,15 +508,120 @@
...
+                          :resid (str "browser.testing.resource/" (btu/get-seed))
+                          :form-name (str "Browser Test Form " (btu/get-seed))
+                          :workflow-name (str "Browser Test Workflow " (btu/get-seed))
+                          :catalogue-item-name (str "Browser Test Catalogue Item " (btu/get-seed)))
+      (create-license)
+      (create-resource)
+      (create-form)
+      (create-workflow)
+      (create-catalogue-item))
+    (testing "check that catalogue item is visible for applicants"
```
Macroz, line 623: Enabling could happen here. We certainly don't use it elsewhere, but we could check that the catalogue item is visible only after enabling. That sounds valuable enough to test because looks like it's easy!

test/clj/rems/test_browser.clj
```diff
@@ -625,3 +625,108 @@
...
+
+    (btu/scroll-and-click :save)
+    (btu/wait-visible {:tag :h1 :fn/text "Form"})
+    (btu/wait-page-loaded)
+
+    (is (= {"Organization" "NBN"
+            "Title" "Form editor test"
+            "Active" ""}
+           (slurp-fields :form)))
+    ;; the text is split into multiple DOM nodes so we need to get fancy
```
Macroz, line 680: I think this is something that should be fixed. We could for example have a structured localized text for title so that maxlength and required are just parameters. Or since this part mostly does text manipulation wrap it into a `str` but that would prevent using elements in the parts.

Macroz: Perhaps you could put here an explanation that this is checking the preview?

test/clj/rems/test_browser.clj
opqdonut: done

test/clj/rems/test_browser.clj
opqdonut: maybe once we actually care about the activity status in some test

test/clj/rems/test_browser.clj
opqdonut: ok

test/clj/rems/test_browser.clj
opqdonut: ok

test/clj/rems/test_browser.clj
opqdonut: I'll add an explanation.

I would rather implement a generic helper that loops through elements and uses get-element-text: otherwise we need to jump through hoops in the implementation every time

Macroz: Maybe extract this part to its own localized text?
```
      title " "
      (when max-length
        (text-format :t.form/maxlength (str max-length)))
      " "
      (if optional
        (text :t.form/optional)
        (text :t.form/required))]
```

Macroz: Ideally `:fn/has-text` would work the same as `has-text?` in etaoin.

Macroz: Looks like this would be possible in xpath and `has-text?` uses internally `xpath/node-by-text` whereas `fn/has-text` uses a simpler xpath that does not "do the right thing". Could try patching this.

Macroz: I've fixed some sites in the `application.cljs` previously to make the DOM simpler.

opqdonut: I'll see if `has-text?` works

test/clj/rems/test_browser.clj
Macroz: I think the catalogue item could be one.

PR #2209, opened 2020-05-28
multiple attachments

### LINKED ISSUE #2122: Support for multiple attachments for application
Allow applicant to add multiple attachments to an application instead of just one.

opqdonut:
fixes #2122

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/form_validation.clj
+++ b/src/clj/rems/form_validation.clj
@@ -35,6 +35,10 @@
     (invalid-option-value? field) {:field-id (:field/id field)
                                    :type     :t.form.validation/invalid-value}))
 
+;; TODO: validation for attachment fields
+;; a) validate that answer is sequence of numbers
+;; b) check that numbers are valid? (needs injection / side effect!)
+
 (defn- validate-field-submit [field]
   (if (required? field)
     {:field-id (:field/id field)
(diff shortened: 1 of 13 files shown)


src/clj/rems/form_validation.clj
```diff
@@ -26,14 +27,21 @@
   (and (= (:field/type field) :option)
        (not (option-value-valid? field))))
 
+;; TODO: validate that attachments are actually valid?
+(defn- invalid-attachment-value? [field]
```
foxlynx, line 31: How about renaming this method so that it doesn't get forgotten to the comment that there is no "real" validation to attachments? Sth like validate-attachment-id?

src/cljc/rems/common/form.cljc
```diff
@@ -409,3 +409,18 @@
...
                                          :visibility/values ["yes"]}))))))))
+
+(defn parse-attachment-ids [string]
+  (mapv parse-int (when-not (empty? string)
+                    (str/split string #","))))
+
+(defn unparse-attachment-ids [ids]
+  (str/join "," ids))
+
+(deftest test-parse-unparse-attachment-ids
```
foxlynx, line 420: Might be just my brain wiring, but I would prefer if all the parse cases were first and then unparse. :)

Issue #2183, opened 2020-05-14
Sorting order of application ids


opqdonut:
- [ ] rems ids like 2020/3 aren't sorted properly, e.g. 2020/12 comes before 2020/3
- might need to support diaarinumeros like "THL/53/14.01.00/2002-rems/27" also

PR #2223, opened 2020-06-10
feat: pin table sorting function

marharyta:

## Reviewability
- [x] link to issue: https://github.com/CSCfi/rems/issues/2183

## Documentation

## Testing

## Follow-up


marharyta:
Unit tests not ready yet, but coming

marharyta:
After the initial comments from @Macroz 
[ ] the diff is not clean, there seems to be unrelated changes (more later)
[ ] the commits look a bit strange (e.g. three merges in a row), consider rebasing on top of the current main, after fixing the things
[x]a debug console log is still there - removed
[x] almost duplicate functions in the parse-month-to-number and parse-year-to-number - removed and renamed
[ ] why do the numbers need to have these special parsing functions in the first place? the solution should likely be more generic than just supporting a year and a number (i.e. the real diaarinumero have multiple parts)
[ ] there are changes in table.cljs that I don't understand and that should probably not be there
[ ] unrelated changes in browser tests?
[x] unrelated changes in test-util?
[ ] if/when using the javascript parseInt, always specify a radix of 10 for backwards compatibility
[x] application-util I think is the correct place for the parser 👍

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,15 @@
 
 (defn is-handler? [application user]
   (contains? (workflow-handlers application) user))
+
+(defn- str-to-int [s]
+  #?(:clj (Integer/parseInt s)
+     :cljs (js/parseInt s 10)))
+
+(defn parse-sequence-number-within-year-to-number [external-id]
+  (str-to-int (get (str/split external-id #"/") 1)))
```
Macroz, line 38: This function assumes that the format will always be `year/number` and the external-id could be something like `THL/2492/4.00.00/2020` so this should be written a bit more generic, and the name would then change. I would name it  something like `parse-sortable-external-id`

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,15 @@
...
+
+;; put test here
+(deftest test-validate-external-id
```
Macroz, line 41: The unit test should always be named `test-the-function-name` so here `test-parse-sequence-number-within-year-to-number`.

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,15 @@
...
+  (is (= 10 (parse-sequence-number-within-year-to-number "2020/10")))
+  (is (= 6 (parse-sequence-number-within-year-to-number "2020/6"))))
```
Macroz, line 43: I don't see a case for `nil`, for different years, and different formats like `THL/2492/4.00.00/2020`

src/cljs/rems/table.cljs
```diff
@@ -181,17 +181,21 @@
...
     (rf/subscribe [::sorting table])])
  (fn [[rows sorting] _]
    (->> rows
-        (sort-by #(get-in % [(:sort-column sorting) :sort-value])
-                 (case (:sort-order sorting)
-                   :desc #(compare %2 %1)
-                   #(compare %1 %2))))))
+        (sort-by (fn [row]
+                   (get-in row [(:sort-column sorting) :sort-value]))
+                 #(normal-sort sorting %1 %2)))))
```
Macroz, line 198: Why this change?

src/cljs/rems/table.cljs
```diff
@@ -181,17 +181,21 @@
         ;;       (or then we could just not validate these internal row representations)
         (s/validate Rows))))
 
+(defn- normal-sort [sorting option1 option2]
+  (case (:sort-order sorting)
+    :desc (compare option2 option1)
+    :asc (compare option1 option2)))
+
```
Macroz, line 188: This is not identical to the old code. Why?

test/clj/rems/test_browser.clj
```diff
@@ -906,19 +906,19 @@
...
+                             (q (get-in application [:application/applicant :name]))
+                             (q (text/localize-time (get-in application [:application/first-submitted])))
+                             (q "Applied")
+                             (q "")
+                             (q "Tämä on monimutkainen arvo skandein varusteltuna!")])]
+                 (->> #"applications_.*\.csv"
+                      btu/downloaded-files
+                      first
+                      slurp
+                      str/split-lines)))))))
```
Macroz, line 924: It's ok to fix some formatting problem but it's likely that there will be conflict with this soon so I'd just remove it because you are not touching the file otherwise.

src/cljs/rems/table.cljs
marharyta: I believed it would look cleaner this way, I might be wrong

src/cljs/rems/table.cljs
marharyta: I believed it would look cleaner this way, I might be wrong

src/cljc/rems/common/application_util.cljc
marharyta: is there any place I can look at other possible formats?

src/cljc/rems/common/application_util.cljc
marharyta: ok!

marharyta:
Commit squash in progress.

marharyta:
Update:
After the initial comments from @Macroz
[x] the diff is not clean, there seems to be unrelated changes (more later)
[ ] the commits look a bit strange (e.g. three merges in a row), consider rebasing on top of the current main, after fixing the things
[x]a debug console log is still there - removed
[x] almost duplicate functions in the parse-month-to-number and parse-year-to-number - removed and renamed
[ x] why do the numbers need to have these special parsing functions in the first place? the solution should likely be more generic than just supporting a year and a number (i.e. the real diaarinumero have multiple parts)
[x] there are changes in table.cljs that I don't understand and that should probably not be there - to me adding a separate function to do sorting seemed like a cleaner solution
[x] unrelated changes in browser tests?
[x] unrelated changes in test-util?
[x ] if/when using the javascript parseInt, always specify a radix of 10 for backwards compatibility
[x] application-util I think is the correct place for the parser 👍

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,25 @@
...
+
+(defn- printToConsole [s]
+  #?(:clj (println s)
+     :cljs (js/console.log s)))
+
+(defn parse-sortable-external-id [external-id]
+  ;; the idea is to parse all numbers from the string, then add them up to form an external id - 
+  ;; comes from a theory that it is somehow tied to the date or the numbers are created in sequence
+  (cond
+    (if (seq (re-seq #"\d+" external-id)) "true" "false") (reduce + (map (fn [str] (str-to-int str)) (re-seq #"\d+" external-id)))
```
opqdonut, line 45: Why sum the numbers? This leads to funny sort orders, for example:

- ABC/2000.1
- ABC/2003.1
- ABC/2001.5

It's of course hard to sort arbitrary identifiers in general, but perhaps sorting the number sequences lexicographically would be a better solution. Luckily clojure vectors work nicely for this:

```
(map #(vec (re-seq #"\d+" %)) ["ABC/2000.1" "ABC/2003.1" "ABC/2001.5" "ABC/2001.11"])
  ==> (["2000" "1"] ["2003" "1"] ["2001" "5"] ["2001" "11"])
(sort [[2000 1] [2003 1] [2001 5] [2001 11]])
  ==> ([2000 1] [2001 5] [2001 11] [2003 1])
```

In any case, a test that sorts various lists of identifiers would be useful both to document the behaviour and as a place to add regression cases.

marharyta: ok, Ill redo it :)

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Changes since v2.12
 - OIDC access tokens are now revoked on logout if the OIDC server provides a `revocation_endpoint`. (#2176)
 - Application attachment fields now accept multiple attachments. (#2122)
 - It's now possible to add a text to the login page after the login button using extra translations (:t.login/intro2) (#2214)
+- It is possible to sort applications based on the application external id accumulated number (#2183)
```
Macroz, line 36: It was possible before? This should say what it actually does (i.e. tries to sort them correctly).

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,25 @@
 
 (defn is-handler? [application user]
   (contains? (workflow-handlers application) user))
+
+(defn- str-to-int [s]
+  #?(:clj (Integer/parseInt s)
+     :cljs (js/parseInt s 10)))
+
+(defn- printToConsole [s]
```
Macroz, line 37: naming-convention-is-kebab-case also this function is not required?

src/cljc/rems/common/application_util.cljc
Macroz: Doesn't this also lose all the strings for comparison purposes? Also test what happens with 0-prefixed numbers i.e. 001 and stuff like 4.00.00. See `THL/2492/4.00.00/2020`

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,25 @@
...
+  ;; the idea is to parse all numbers from the string, then add them up to form an external id - 
+  ;; comes from a theory that it is somehow tied to the date or the numbers are created in sequence
+  (cond
+    (if (seq (re-seq #"\d+" external-id)) "true" "false") (reduce + (map (fn [str] (str-to-int str)) (re-seq #"\d+" external-id)))
+    :else external-id))
+
+;; put test here
+(deftest test-parse-sortable-external-id
+  (is (= 2030 (parse-sortable-external-id "2020/10")))
+  (is (= 2092 (parse-sortable-external-id "THL/54/14.01.00/2002-rems/21")))
```
Macroz, line 51: This should probably be `["THL/" 54 "/14.01.00/" 2002 "-rems/" 21]`

And THL/2492/4.00.00/2020 would be `["THL/" 2492 "/4.00.00/" 2020]`

src/cljs/rems/table.cljs
Macroz: But you don't handle the default case here like the previous code did.

src/cljs/rems/application_list.cljs
```diff
@@ -81,7 +81,7 @@
    (map (fn [app]
           {:key (:application/id app)
            :id {:value (:application/id app)}
-           :external-id {:value (:application/external-id app)}
+           :external-id {:value (:application/external-id app) :sort-value (application-util/parse-sortable-external-id (:application/external-id app))}
```
Macroz, line 84: I think you should split the line at `:sort-value` like in the next line.

src/cljc/rems/common/application_util.cljc
Macroz: I searched some with Google and several are in this format.

src/cljc/rems/common/application_util.cljc
opqdonut: 0-prefixed numbers should work fine when parsed into ints.

src/cljc/rems/common/application_util.cljc
opqdonut: Why not
```
["THL/" 54 "/" 14 "." 1 "." 0 "/" 2002 "-rems/" 21]
["THL/" 2492 "/" 4 "." 0 "." 0 "/" 2020]
```
more consistent handling of numbers and results in the same ordering (assuming 0-padding is consistent)

src/cljc/rems/common/application_util.cljc
Macroz: Yes but we need tests to cover them.

src/cljc/rems/common/application_util.cljc
Macroz: I'm fine with any if there are clear cases that show what the behavior is. It's a bit difficult to say if how some of these number-like parts should be sorted, and if they only ever have numbers.

Macroz: It's quite likely that we have to amend the cases later.

Macroz: One of the possible complications is that we have to detect which part is a year and sort by that first.

src/cljc/rems/common/application_util.cljc
marharyta: "Doesn't this also lose all the strings for comparison purposes?" - sorry @Macroz do you mean my solution?

CHANGELOG.md
marharyta: right, should have formulated more specifically.

src/cljc/rems/common/application_util.cljc
marharyta: yes, I will delete it.

src/cljc/rems/common/application_util.cljc
Macroz: Any solution with `re-seq` of just digits.

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,26 @@
...
+(defn- str-to-int [s]
+  #?(:clj (Integer/parseInt s)
+     :cljs (js/parseInt s 10)))
+
+;; (defn- printToConsole [s]
+;;   #?(:clj (println s)
+;;      :cljs (js/console.log s)))
+
+(defn parse-sortable-external-id
+  "The idea is to parse all numbers from the string to vector and then compare the values in the vector"
```
Macroz, line 42: This is the wrong idea in that it doesn't handle now the possible differing string parts?

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,26 @@
...
+  [external-id]
+  (when external-id
+    (when-let [number-sequence  (seq (re-seq #"\d+" external-id))]
+      (vec (map str-to-int
```
Macroz, line 46: `mapv`

src/cljc/rems/common/application_util.cljc
```diff
@@ -27,3 +29,26 @@
...
+                number-sequence)))))
+
+(deftest test-parse-sortable-external-id
+  (is (= [2020 10] (parse-sortable-external-id "2020/10")))
+  (is (= [54 14 01 00 2002 21] (parse-sortable-external-id "THL/54/14.01.00/2002-rems/21")))
```
Macroz, line 51: This should not lose "THL" or "-rems" since those can be used in the comparison ...

src/cljc/rems/common/application_util.cljc
marharyta: fair enough :)

PR #2238, opened 2020-06-16
feat: don't set a transaction timeout for dev

Macroz:
Don't set a transaction timeout for development use so that you can single-step while debugging.

### DIFF
--- a/test/clj/rems/db/test_transactions.clj
+++ b/test/clj/rems/db/test_transactions.clj
@@ -69,7 +69,10 @@
         wrapped-handle-command (fn [cmd application injections]
                                  (mark-observed-app-version (old-handle-command cmd application injections)
                                                             application))]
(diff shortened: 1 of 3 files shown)


test/clj/rems/db/test_transactions.clj
```diff
@@ -69,7 +69,10 @@
         wrapped-handle-command (fn [cmd application injections]
                                  (mark-observed-app-version (old-handle-command cmd application injections)
                                                             application))]
-    (with-redefs [commands/handle-command wrapped-handle-command]
+    (with-redefs [commands/handle-command wrapped-handle-command
+                  rems.config/env (assoc rems.config/env
+                                         :database-lock-timeout "4s"
+                                          :database-idle-in-transaction-session-timeout "8s")]
```
opqdonut, line 75: [nit] indentation :)

Issue #2039, opened 2020-03-09
Organization creation


Macroz:
- [x] store the organization data in db
- [x] store the review-email address
- [x] minimal CRUD UI 
  - [x] list
  - [x] view
  - [x] create
  - [x] edit (emails+)
- store members as they log-in through invite links?
- name and description for organization used in the organization dropdowns?
- [x] localize organization name
- [x] localize review email name
- [x] make organizations unoptional in API user return values
- [x] short name i.e. abbreviation for organization names?
- [x] browser test the organization UI
- [ ] check documentation and installation instructions

PR #2240, opened 2020-06-16
feat: organization enable and archive

Macroz:
Add organization enable and archive function for #2039 

- all items depend on organizations, even unused
- organizations don't depend on anything

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -44,7 +44,9 @@
           (s/optional-key :organization/last-modified) DateTime
           (s/optional-key :organization/owners) [User]
           (s/optional-key :organization/review-emails) [{:name LocalizedString
-                                                         :email s/Str}]}))
+                                                         :email s/Str}]
+          (s/optional-key :organization/enabled) s/Bool
+          (s/optional-key :organization/archived) s/Bool}))
 
 (s/defschema CatalogueItemLocalizations
   {s/Keyword {;; TODO :id (it's the catalogue item id) and :langcode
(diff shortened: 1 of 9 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -44,7 +44,9 @@
           (s/optional-key :organization/last-modified) DateTime
           (s/optional-key :organization/owners) [User]
           (s/optional-key :organization/review-emails) [{:name LocalizedString
-                                                         :email s/Str}]}))
+                                                         :email s/Str}]
+          (s/optional-key :organization/enabled) s/Bool
+          (s/optional-key :organization/archived) s/Bool}))
```
opqdonut, line 49: am I missing something or do existing organizations default to disabled and unarchived? wouldn't this need a migration?

Macroz: I don't think anyone is running new organization code yet so it should be ok? Also the disabling and archiving is not functional anywhere yet?

opqdonut: Ok fair enough. But there should probably still be defaults for enabled & archived when creating an organization, right?

Macroz: Do you mean this part?

https://github.com/CSCfi/rems/blob/master/src/clj/rems/db/organizations.clj#L16

opqdonut: Yeah!

PR #2249, opened 2020-08-11
Indicate items in cart

### LINKED ISSUE #2228: Indicate catalog items already in the cart
When a catalog item is added to the shopping cart it should be indicated that the item is now in the cart and should not be added again. Possible solutions:

- Change "Add to cart" button to "Remove from cart"
- Remove "Add to cart" button when item in cart
- Change colouring of the catalog item

Last time the problem was poor performance for 1000 item table updates when removing the entire row so the feature was removed. With changing the button to a "toggle" or hiding it could certainly work faster.

Macroz:
Closes #2228 

indicate-items

## Reviewability
- [x] link to issue

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Changes since v2.12
 - OIDC access tokens are now revoked on logout if the OIDC server provides a `revocation_endpoint`. (#2176)
 - Application attachment fields now accept multiple attachments. (#2122)
 - It's now possible to add a text to the login page after the login button using extra translations (:t.login/intro2) (#2214)
+- Indicate which items are in shopping cart by changing add button to remove (#2228)
 
 ## v2.12 "Merituulentie" 2020-05-04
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ Changes since v2.12
 - OIDC access tokens are now revoked on logout if the OIDC server provides a `revocation_endpoint`. (#2176)
 - Application attachment fields now accept multiple attachments. (#2122)
 - It's now possible to add a text to the login page after the login button using extra translations (:t.login/intro2) (#2214)
+- Indicate which items are in shopping cart by changing add button to remove (#2228)
```
opqdonut, line 36: the button now says just "poista", but something like "poista korista" might be more clear in the context

Macroz: It's using the same button so I wonder if it would be good to simplify the add-text or complicate the remove-text?

Macroz: Changed them to match.

PR #2253, opened 2020-08-13
feat: use enabled and archived without organization ns

Macroz:
Technical change to rename `:organization/enabled` to `:enabled` and `:organization/archived` to `:archived`. Afterwards they are in line with other items and the state-flags -concept and don't complicate it.

### DIFF
--- a/src/clj/rems/api/services/dependencies.clj
+++ b/src/clj/rems/api/services/dependencies.clj
@@ -51,7 +51,7 @@
 
 (defn- add-status-bits [dep]
   (merge dep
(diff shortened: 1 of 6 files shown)


src/clj/rems/api/services/dependencies.clj
```diff
@@ -51,7 +51,7 @@
 
 (defn- add-status-bits [dep]
   (merge dep
-         (select-keys (enrich-dependency dep) [:archived :enabled :organization/archived :organization/enabled])))
+         (select-keys (enrich-dependency dep) [:archived :enabled])))
```
opqdonut, line 54: :tada:

Issue #2015, opened 2020-02-27
Blacklisting extended


jaakkocsc:
If a user is blacklisted in the resource view then all related, open applications should be rejected automatically.

PR #2261, opened 2020-08-18
rejecting existing applications when blacklisting

opqdonut:
for #2015

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/blacklist.clj
+++ b/src/clj/rems/api/blacklist.clj
@@ -2,7 +2,9 @@
   (:require [clj-time.core :as time]
             [compojure.api.sweet :refer :all]
             [rems.api.schema :as schema]
+            [rems.api.services.command :as command]
             [rems.api.services.blacklist :as blacklist]
+            [rems.application.rejecter-bot :as rejecter-bot]
             [rems.db.users :as users]
             [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer [ok]]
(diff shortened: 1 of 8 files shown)


src/clj/rems/api/blacklist.clj
```diff
@@ -56,6 +58,8 @@
       :body [command BlacklistCommand]
       :return schema/SuccessResponse
       (blacklist/add-user-to-blacklist! (getx-user-id) command)
+      (doseq [cmd (rejecter-bot/reject-all-applications-by (get-in command [:blacklist/user :userid]))]
+        (command/command! cmd))
```
Macroz, line 62: Should there be one function in the service layer that does all of this?

Macroz (approved):
Looks pretty good.

src/clj/rems/api/blacklist.clj
opqdonut: The problem is that services.command already depends on services.blacklist, so we can't have services.blacklist depend on services.command...

But now that I look at this piece of code I notice that I should check the return value of `command!`

Macroz: Yes it's not atomic. Also it feels a bit strange to have blacklist stuff in command.

opqdonut: added the error handling

src/clj/rems/api/services/command.clj
```diff
@@ -25,16 +25,17 @@
     (entitlements/update-entitlements-for-event event))
   [])
 
+;; TODO should this be in the revoke command handler instead?
```
opqdonut, line 28: see my comment here

PR #2269, opened 2020-08-21
feat: support multiple resources in /apply-for redirect

opqdonut:

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/home.clj
+++ b/src/clj/rems/home.clj
@@ -1,5 +1,6 @@
 (ns rems.home
-  (:require [clojure.tools.logging :as log]
+  (:require [clojure.string :as str]
+            [clojure.tools.logging :as log]
             [mount.core :as mount]
             [compojure.core :refer [GET defroutes routes]]
             [rems.api.services.attachment :as attachment]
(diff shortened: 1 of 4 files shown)


src/clj/rems/home.clj
```diff
@@ -14,15 +15,22 @@
...
+      (= 0 (count items)) :not-found
+      (< 1 (count items)) :not-unique
+      :else (:id (first items)))))
+
+(defn- apply-for-resources [resources]
+  (let [items (map resource-to-item resources)]
+    (cond
+      (some #{:not-found} items) (-> (not-found "Resource not found")
+                                     (content-type "text/plain"))
+      (some #{:not-unique} items) (-> (not-found "Resource ID is not unique")
```
Macroz, line 31: Possibly we could use the 400 bad request here?

src/clj/rems/home.clj
```diff
@@ -40,11 +48,15 @@
   (GET "/" []
     (layout/home-page))
 
+  ;; TODO should these redirects have swagger documentation?
```
Macroz, line 51: I suppose it's not very easy unless they are moved under `/api`? I think https://github.com/CSCfi/rems/blob/master/docs/linking.md is enough.

src/clj/rems/home.clj
```diff
@@ -40,11 +48,15 @@
...
+
   (GET "/accept-invitation" [token]
     (redirect (str "/application/accept-invitation/" token)))
 
-  (GET "/apply-for" [resource]
-    (apply-for-resource resource))
+  (GET "/apply-for" [resource] ;; can specify multiple resources
+    (if (vector? resource)
```
Macroz, line 57: Could this break if compojure decides to return just a seq?

src/clj/rems/home.clj
opqdonut: The ring wiki says it's a vector. https://github.com/ring-clojure/ring/wiki/Parameters

src/clj/rems/home.clj
opqdonut: The ring wiki says it's a vector. https://github.com/ring-clojure/ring/wiki/Parameters

src/clj/rems/home.clj
opqdonut: why not!

src/clj/rems/home.clj
opqdonut: Yeah it's probably enough, but it would  be nice to document these (and e.g. the exact parameter formats and HTTP return codes) as swagger. We might be able to merge two swagger docs or something? But maybe let's not worry about it now.

Issue #2130, opened 2020-04-16
Configurable extra user attributes


opqdonut:
- [x] configurable list of additional attributes to read from oidc claim
- [x] localized text for the "name"
- [x] store attributes in db
- [x] return attributes from API
  - [x] only show additional user attributes in applications to owner/reporter/handler (can probably use the application `:see-everything` permission)
- [x] docs

opqdonut:
wait for #2035

opqdonut:
this is ready for development now

PR #2270, opened 2020-08-21
refactor: separate add-user! and add-user-raw! functions

opqdonut:
previously, we only had add-user! which took in raw attributes

related to #2130

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/db/users.clj
+++ b/src/clj/rems/db/users.clj
@@ -38,12 +38,16 @@
     (or (str/blank? (:userid user))
         (str/blank? (:name user)))))
 
-(defn add-user-raw! [user userattrs]
-  (assert user)
+(defn add-user-raw!
+  "Create or update a user given a userid and a map of raw user attributes."
+  [userid userattrs]
+  (assert userid)
   (assert userattrs)
-  (db/add-user! {:user user :userattrs (json/generate-string userattrs)}))
+  (db/add-user! {:user userid :userattrs (json/generate-string userattrs)}))
 
-(defn add-user! [user]
+(defn add-user!
+  "Create or update a user given formatted user attributes."
+  [user]
   (add-user-raw! (:userid user) (unformat-user user)))
 
 (defn get-raw-user-attributes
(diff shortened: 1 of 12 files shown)


src/clj/rems/db/users.clj
```diff
@@ -38,14 +38,17 @@
...
         (str/blank? (:name user)))))
 
-(defn add-user! [user userattrs]
-  (assert user)
+(defn add-user-raw!
+  "Create or update a user given a userid and a map of raw user attributes."
+  [userid userattrs]
+  (assert userid)
   (assert userattrs)
-  (db/add-user! {:user user :userattrs (json/generate-string userattrs)}))
```
foxlynx, line 44: What happens here if userattrs are not valid? Or is it possible?

test/clj/rems/db/test_user_settings.clj
```diff
@@ -89,7 +89,7 @@
            (user-settings/validate-new-settings {:notification-email nil})))))
 
 (deftest test-notification-email-visible
-  (users/add-user! "pekka" {:eppn "pekka" :commonName "Pekka" :mail "pekka@example.com"})
+  (users/add-user! {:userid "pekka" :name "Pekka" :email "pekka@example.com"})
```
foxlynx, line 92: Ia there on purpose :mail --> :email ?

foxlynx (approved):
See comments

test/clj/rems/db/test_user_settings.clj
opqdonut: Yes, previously `add-user!` wanted raw attributes which use `:eppn` and `:mail`, now it takes formatted attributes which use `:userid` and `:email`. I guess we support `:mail` since that's what shibboleth gives us...

src/clj/rems/db/users.clj
opqdonut: We just store the attributes. The only thing user attributes really need is the userid, which is asserted here. We do some filtering when reading the DB with `invalid-user?` above.

PR #2279, opened 2020-08-27
small improvements to docker-entrypoint.sh

opqdonut:


docker-entrypoint.sh
```diff
@@ -1,10 +1,8 @@
 #!/bin/bash
 
-cd rems
```
opqdonut, line 3: rahti might need this. @vvaltia knows better

vvaltia: I think this is fine. It was an unnecessary command failing every time

vvaltia (approved):

PR #2280, opened 2020-08-27
doc: document who implements /assets/

opqdonut:


marharyta (commented):
maybe ```https://www.webjars.org/``` link could be beneficial?

marharyta (approved):

PR #2288, opened 2020-08-28
feat: configurable oidc userid attribute

### LINKED ISSUE #2281: Make OIDC userid attribute configurable
Currently it defaults to `:sub`.

Probably best to add a separate configuration key for it, `:oidc-userid-attribute` instead of using `:oidc-extra-attributes`.

opqdonut:
fixes #2281

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/auth/oidc.clj
+++ b/src/clj/rems/auth/oidc.clj
@@ -45,7 +45,7 @@
         ;; locale – could be used to set preferred lang on first login
         ;; email – non-unique (!) email
         id-data (jwt/validate id-token issuer audience now)
-        identity-base {:eppn (:sub id-data)
+        identity-base {:eppn (get id-data (keyword (getx env :oidc-userid-attribute)))
                        ;; need to maintain a fallback list of name attributes since identity
                        ;; providers differ in what they give us
                        :commonName (some id-data [:name :unique_name :family_name])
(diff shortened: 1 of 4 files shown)


opqdonut:
marking as WIP, it seems we need more features

opqdonut:
rebased now that #2277 is in

src/clj/rems/auth/oidc.clj
```diff
@@ -45,7 +45,7 @@
         ;; locale – could be used to set preferred lang on first login
         ;; email – non-unique (!) email
         id-data (jwt/validate id-token issuer audience now)
-        identity-base {:eppn (:sub id-data)
+        identity-base {:eppn (get id-data (keyword (getx env :oidc-userid-attribute)))
```
Macroz, line 48: Couldn't the attribute be a keyword instead of a string in the config?

Macroz (approved):

src/clj/rems/auth/oidc.clj
opqdonut: Yeah, it could, but I picked a string so that it would map more directly to the JSON oidc sends us

opqdonut: One further advantage is that a string works for a key like `":boing foo"`  which can be impossible to express as a keyword literal

PR #2305, opened 2020-09-08
refactor: better name for event-type-specific-application-view

opqdonut:


opqdonut:
another option:
```
event-type-specific-application-view => application-content-view
calculate-permissions => application-permission-view
```

marharyta (approved):

Macroz (changes requested):
I think `misc` sounds too much like unimportant. `state` too vague and perhaps conflicts with `application/state`. I think I would like `application-base-view` more.

opqdonut:
base sounds good :+1:

Macroz (approved):

PR #2306, opened 2020-09-08
Organization browser test

Macroz:
Missing browser test for #2039.


opqdonut (approved):
looks good to me but it seems the test fails

Macroz:
> looks good to me but it seems the test fails

I think it required some more waits for reliable testing.

PR #2308, opened 2020-09-10
feat: UI for picking entitlement end date

### LINKED ISSUE #2123: support for entitlement end dates
- [ ] configurable default length for entitlements
- form can have a field for proposed end date
- [ ] handler/decider can (optionally) select end date as part of approve action
  - [ ] the end date is set in entitlement immediately when it is created
  - [ ] event log should also show the end date: "Handler has approved the application. The entitlement ends 2021-03-04"
- no need for automation that closes the application: applications can stay in the approved state

opqdonut:
closes #2123

## Reviewability
- [x] link to issue

## Documentation

## Different installations

## Testing

## Accessibility
- [x] all fields have a label
- [ ] contrast is checked

## Follow-up


opqdonut:
approve-end-action
approve-end-event

src/cljc/rems/text.cljc
```diff
@@ -125,40 +136,36 @@
 
 (defn localize-event [event]
   (let [event-type (:event/type event)]
-    (text-format
-     (get event-types event-type :t.applications.events/unknown)
-     (application-util/get-member-name (:event/actor-attributes event))
+    (str
```
Macroz, line 139: I think instead of `str` all the bits together a better design would use clear slots either with `text-format` (because languages could vary) or more hiccup here.

src/cljs/rems/actions/approve_reject.cljs
```diff
@@ -1,25 +1,37 @@
...
  (fn [{:keys [db]} _]
-   {:db (assoc db ::comment "")
+   {:db (assoc db ::comment "" ::end (default-end (get-in db [:config :entitlement-default-length-days])))
     :dispatch [:rems.actions.action/set-attachments action-form-id []]}))
 
 (rf/reg-sub ::comment (fn [db _] (::comment db)))
 (rf/reg-event-db ::set-comment (fn [db [_ value]] (assoc db ::comment value)))
 
+(rf/reg-sub ::end (fn [db _] (::end db)))
+(rf/reg-event-db ::set-end (fn [db [_ value]] (assoc db ::end value)))
```
Macroz, line 26: I would prefer this file to use `entitlement-end`, `::entitlement-end` etc. here, it's slightly more greppable for the entirety of the feature.

Macroz (approved):
Looks good to me, a couple ideas to think about

src/cljs/rems/actions/approve_reject.cljs
opqdonut: that's a good idea

src/cljc/rems/text.cljc
opqdonut: Unfortunately localize-event needs to return a string because it's also used in the pdf code. I was thinking of having something like `"%1 approved the application. %2" as the format string for the approved event, but that didn't seem useful.

PR #2322, opened 2020-09-11
fixes & browser test for catalogue item editor

### LINKED ISSUE #2321: catalogue item editor is broken

opqdonut:
closes #2321

## Reviewability
- [x] link to issue

## Documentation

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Changes since v2.12
 - Fixed enabling a catalogue item after changing its form. (#2283)
 - Added missing decision text to pdf event list.
 - More compatible CSV reports. Line returns are removed from field values and CSV lines are separated with CRLF. (#2311)
+- Fixed editing a catalogue item. (#2322)
 
 ### Additions
 - The form administration pages now flag forms that have missing localizations. REMS also logs a warning on startup for these forms. (#2098)
--- a/resources/translations/fi.edn
+++ b/resources/translations/fi.edn
@@ -92,6 +92,7 @@
                    :license-name "Lisenssin nimi"
                    :licenses "Lisenssit"
                    :modifier "Muokkaaja"
+                   :more-info "Lisätietoja-URL (vapaaehtoinen, puuttuessa käytetään resurssin URN:ää)"
                    :name "Nimi"
                    :no-licenses "Ei lisenssejä"
                    :organization "Organisaatio"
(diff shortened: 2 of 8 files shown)


src/cljs/rems/administration/organization.cljs
```diff
@@ -71,13 +71,12 @@
...
-                    (map display-localized-review-email)
-                    (interpose [:br]))
-               {:box? false
-                :solid? true}]
+              [:div
+               [:label (text :t.administration/review-emails)]
+               [:div.solid-group
+                (->> (:organization/review-emails organization)
+                     (map display-localized-review-email)
+                     (interpose [:br]))]]
```
Macroz, line 79: The original intent here was to move into a direction where our detail views i.e. info-fields could be nested. Also sequences should be handled by rows or maps as tables. I think this moves a bit more to the hand-written direction but clearly there is no complete solution for the nesting yet.

resources/translations/fi.edn
```diff
@@ -92,6 +92,7 @@
                    :license-name "Lisenssin nimi"
                    :licenses "Lisenssit"
                    :modifier "Muokkaaja"
+                   :more-info "Lisätietoja-URL (vapaaehtoinen, puuttuessa käytetään resurssin URN:ää)"
```
Macroz, line 95: Isn't it so that URN is used only if there is no link and resource id looks like a URN? Otherwise no link.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Changes since v2.12
 - Fixed enabling a catalogue item after changing its form. (#2283)
 - Added missing decision text to pdf event list.
 - More compatible CSV reports. Line returns are removed from field values and CSV lines are separated with CRLF. (#2311)
+- Fixed editing a catalogue item. (#2322)
```
Macroz, line 37: Maybe "Fixed not all fields showing up when editing a catalogue item" ?

Macroz (approved):

src/cljs/rems/administration/organization.cljs
opqdonut: yeah... I thought about changing slurp-fields so that it only looks at the immediate children, but that was difficult and would've been fragile in another way

resources/translations/fi.edn
opqdonut: yeah

CHANGELOG.md
opqdonut: It was completely broken, no fields showed up for me and I couldn't press save

Macroz: Ok in that case it's better like it is

resources/translations/fi.edn
Macroz: Maybe the text is slightly misleading now. "vapaaehtoinen, puuttuessa kokeillaan URN"?

src/cljs/rems/administration/organization.cljs
Macroz: Something to keep in mind for later as the model gets more complex.

resources/translations/fi.edn
opqdonut: I'd rather not try to translate a complex sentence into swedish, but if you insist... :grimacing:

src/cljs/rems/administration/organization.cljs
opqdonut: I guess one solution would be to look for all the text boxes and then get their labels, instead of looking for form-groups

opqdonut: ...in slurp-fields that is

resources/translations/fi.edn
Macroz: This could really use some info text help or link to manual.

Macroz: Let's keep the description for now.

Issue #2313, opened 2020-09-10
More rights for reporter


opqdonut:
A user with only the reporter role doesn't see the administration UI and can't navigate to the application exports.

- [x] reporter should be able to navigate to export applications page
- [x] reporter should be able to export applications via UI
- [x] reporter should have read-only access to other administration pages (just like handler does)

PR #2325, opened 2020-09-15
introduce +admin-read-roles+ and +admin-write-roles+

opqdonut:
follow-up to #2313

## Reviewability
- [x] link to issue

## API

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/administration/blacklist.cljs
+++ b/src/cljs/rems/administration/blacklist.cljs
@@ -6,7 +6,7 @@
             [rems.atoms :as atoms]
             [rems.dropdown :as dropdown]
             [rems.flash-message :as flash-message]
-            [rems.roles :as roles]
+            [rems.common.roles :as roles]
             [rems.spinner :as spinner]
             [rems.table :as table]
             [rems.text :refer [text text-format localize-time]]
--- a/src/cljs/rems/administration/blacklist.cljs
+++ b/src/cljs/rems/administration/blacklist.cljs
@@ -6,10 +6,14 @@
             [rems.atoms :as atoms]
             [rems.dropdown :as dropdown]
             [rems.flash-message :as flash-message]
+            [rems.roles :as roles]
             [rems.spinner :as spinner]
             [rems.table :as table]
             [rems.text :refer [text text-format localize-time]]
             [rems.util :refer [fetch post!]]))
+
+(defn show-blacklist-add? [roles]
+  (some #{:owner :handler} roles)) ;; same roles as in rems.api.blacklist
 
 (rf/reg-event-fx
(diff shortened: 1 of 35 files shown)


src/cljs/rems/administration/blacklist.cljs
```diff
@@ -64,10 +68,11 @@
 
 (rf/reg-event-fx
  ::fetch-users
- (fn []
-   (fetch "/api/blacklist/users"
-          {:handler #(rf/dispatch [::fetch-users-result %])
-           :error-handler (flash-message/default-error-handler :top "Fetch users")})
+ (fn [{:keys [db]} _]
+   (when (show-blacklist-add? (get-in db [:identity :roles]))
```
Macroz, line 72: This puts the ownership of the roles in this ns where it belongs, but there is an extra dispatch from the user side to here. The other places seem to so far avoid the dispatch entirely at the use site (i.e. never call `::fetch-users`). I'm not sure if one is better but maybe good to follow the same pattern everywhere? Ideally when you use a certain stateful component it could fetch its own data and check the rights too.

opqdonut: rems.administration.blacklist is an experiment in modularizing components a bit more, so I just followed the existing architecture

src/cljs/rems/administration/form.cljs
```diff
@@ -74,11 +74,12 @@
    (let [id (:form/id form)]
      [:div.col.commands
       [administration/back-button "/administration/forms"]
-      [roles/when roles/show-admin-edit-buttons?
-       [edit-button id]
-       [copy-as-new-button id]
-       [status-flags/enabled-toggle form #(rf/dispatch [:rems.administration.forms/set-form-enabled %1 %2 [::enter-page id]])]
-       [status-flags/archived-toggle form #(rf/dispatch [:rems.administration.forms/set-form-archived %1 %2 [::enter-page id]])]]])
+      [roles/when roles/+admin-write-roles+
+       [:<>
```
Macroz, line 78: I think this is not an improvement. Most hiccup assumes you can just have multiple things in the body. If you want to keep `roles/when` close to Clojure `when` then I would suggest we have `roles/<>`, `roles/when-<>`, `roles/show-when` or something that keeps the `[:<>]` like before and uses the `roles/when`.

Macroz (changes requested):
Should we have a small test for this new `roles/when`?

src/cljs/rems/administration/form.cljs
opqdonut: I like show-when

Macroz (approved):
Much better!

PR #2338, opened 2020-09-17
new, simpler, docker-based circle configuration

opqdonut:
- use docker to run tests
- upgrade to circle config format 2.1
- better names for jobs: unit-test, integration-test, browser-test
- better parallelism: integration & browser tests are in separate jobs


Macroz (approved):
Wow, thanks a lot for doing this! Configuring CI is always quite hairy but the new configuration is simpler and cleaner.

PR #2339, opened 2020-09-18
improve entitlement end ui behaviour

### LINKED ISSUE #2331: Entitlement end date enhancements
In the application events the entitlement end date shows the exact time although the choice is about dates. Let's show only the date of the end date.

It is also likely that the end date is interpreted to end on the end of that date. So let's add one date to the end date chosen.

- [x] only show the date, not time
- [x] end the entitlement on the end of the day (i.e. select 2020-09-17, entitlement will be valid on 17th and end on midnight 23:59:59... when the day turns to 18th.

opqdonut:
- entitlement now ends at the end of the specified day (in UTC)
- only date is rendered for end

fixes #2331

## Reviewability
- [x] link to issue

## Documentation
  - fix for entry already in changelog, no need

## Testing

## Follow-up

### DIFF
--- a/src/cljc/rems/text.cljc
+++ b/src/cljc/rems/text.cljc
@@ -2,9 +2,9 @@
   (:require [clojure.test :refer [deftest is]]
             #?(:clj [clj-time.core :as time]
                :cljs [cljs-time.core :as time])
-            #?(:clj [clj-time.format :as format]
-               :cljs [cljs-time.format :as format])
-            #?(:cljs [cljs-time.coerce :as coerce])
+            #?(:clj [clj-time.format :as time-format]
+               :cljs [cljs-time.format :as time-format])
+            #?(:cljs [cljs-time.coerce :as time-coerce])
             [clojure.string :as str]
             #?(:cljs [re-frame.core :as rf])
             [rems.common.application-util :as application-util]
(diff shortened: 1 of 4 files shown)


src/cljc/rems/text.cljc
```diff
@@ -1,8 +1,10 @@
 (ns rems.text
-  (:require #?(:clj [clj-time.core :as time]
+  (:require [clojure.test :refer [deftest is]]
+            #?(:clj [clj-time.core :as time]
                :cljs [cljs-time.core :as time])
             #?(:clj [clj-time.format :as format]
                :cljs [cljs-time.format :as format])
+            #?(:cljs [cljs-time.coerce :as coerce])
```
Macroz, line 7: I like to import these as `time-coerce` etc. these days just because they obviously go together.

src/cljc/rems/text.cljc
```diff
@@ -98,6 +100,23 @@
              (when time
                (format/unparse-local (time-format) (time/to-default-time-zone time))))))
 
+(defn localize-utc-date
```
Macroz, line 103: Would we actually want to have the server time-zone defined in the config for this? So that in Finland for Finnish officials you get more or less what you expect?

Macroz (approved):
Something to consider but looks good.

src/cljc/rems/text.cljc
opqdonut: Yeah something like that would be good to think about in the future. So far REMS has only referred to instants of time when something happened (event, entitlement), so storing those in UTC and rendering them in the browser timezone was the right thing to do.

Macroz: We decided to go with this and wait for the first installation that has a clear requirement about this.

PR #2340, opened 2020-09-18
Organization fixes

### LINKED ISSUE #2332: Organization fixes
- [x] don't show the review emails box if it's empty
- [x] show the username of the modifier, show like on other screens
- [x] you are not supposed to be able to edit and empty the short name and title (allows save currently)
- [x] show which fields are optional

Macroz:
Closes #2332

Screenshot_2020-09-18_10-08-56

Screenshot_2020-09-18_10-11-35

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/organizations.clj
+++ b/src/clj/rems/api/organizations.clj
@@ -1,6 +1,6 @@
 (ns rems.api.organizations
   (:require [compojure.api.sweet :refer :all]
-            [rems.api.schema :refer [OrganizationArchivedCommand OrganizationEnabledCommand OrganizationFull SuccessResponse UserId UserWithAttributes]]
+            [rems.api.schema :refer [OrganizationArchivedCommand OrganizationEnabledCommand OrganizationFull SuccessResponse User UserWithAttributes]]
             [rems.api.util] ; required for route :roles
             [rems.api.services.organizations :as organizations]
             [rems.util :refer [getx-user-id]]
@@ -8,7 +8,10 @@
             [schema.core :as s]))
 
 (s/defschema CreateOrganizationCommand
-  OrganizationFull)
+  (-> OrganizationFull
+      (dissoc :organization/modifier
+              :organization/last-modifier)
+      (assoc (s/optional-key :organization/owners) [User])))
 
 (s/defschema CreateOrganizationResponse
   {:success s/Bool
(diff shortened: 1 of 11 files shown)


src/clj/rems/api/organizations.clj
```diff
@@ -1,22 +1,28 @@
...
   {:success s/Bool
    (s/optional-key :organization/id) s/Str
    (s/optional-key :errors) [s/Any]})
 
 (s/defschema EditOrganizationCommand
-  OrganizationFull)
+  (-> OrganizationFull
+      (dissoc :organization/modifier
+              :organization/last-modifier)
+      (assoc (s/optional-key :organization/owners) [User])))
```
opqdonut, line 25: so this is the same as CreateOrganizationCommand?

src/clj/rems/db/organizations.clj
```diff
@@ -35,11 +36,16 @@
 (defn get-organizations-raw []
   (->> (db/get-organizations)
        (map parse-organization)
+       (map #(update % :organization/owners (partial mapv (comp users/get-user :userid))))
+       (map #(update % :organization/modifier (comp users/get-user :userid)))
```
opqdonut, line 40: shouldn't this be done in the service layer according to our architecture? also, is this `-raw` any more?

opqdonut (approved):
I'm fine with resolving my comments in a follow-up PR

src/clj/rems/api/organizations.clj
Macroz: Yeah since it is easy to copy that schema to the name. I'll do that. They ended up the same.

src/clj/rems/db/organizations.clj
Macroz: I had a look at the others and they seemed to use the `get-user` from `db.users` in the db-layer whereas for e.g. organization we have `join-organization` in the service layer. Here the `UserWithAttributes` schema coercing requires us to load them already at this level.

You are right about the `-raw` that perhaps it's not correct anymore. We can have a think about how we want to continue with this.

opqdonut (approved):

src/clj/rems/db/organizations.clj
Macroz: Let's do this one in another PR once we have the idea. https://github.com/CSCfi/rems/issues/2342

PR #2347, opened 2020-09-22
misc. browser test improvements

opqdonut:
Now that we have kaocha which captures stdout, it's fine to have more
verbose test logs: you only see them on failure. Verbose logs make
diagnosing CI failures easier.


opqdonut:
Browser tests passed 5 times on circle. I'll rebase to remove the debug commit, and then this can go in IMO.

test/clj/rems/browser_test_util.clj
```diff
@@ -171,6 +170,21 @@
...
+(def +character-delay+ 0.1)
+
+(defn fill-human [q text]
+  (doseq [c text]
+    (et/wait +character-delay+)
+    (et/fill (get-driver) q c))
+  (et/wait +character-delay+)
+  (let [value (get-element-attr q "value")]
+    (when-not (= text value)
+      (log/warn "Failed to fill field to" (pr-str text) "got" (pr-str value) "instead."))))
```
Macroz, line 186: I would like to see the typos and random delays here too. Is there a reason why you did not just copy them too?

Macroz (approved):
Looks good. Great catch!

test/clj/rems/browser_test_util.clj
opqdonut: etaoin is under EPL, which is a copyleft license, so I didn't want to copy any code from it

opqdonut: I can reimplement typos if you want :)

Macroz: I think it would be nice.

Macroz (approved):

PR #2361, opened 2020-09-25
fix: refresh UI organizations after creating organization

### LINKED ISSUE #2359: Newly created org not visible
A newly created organization is not visible in organization dropdown menus when creating a new resource, licence, etc. Browser refresh or logout/login is needed for it to appear.

opqdonut:
fixes #2359

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/administration/organizations.cljs
+++ b/src/cljs/rems/administration/organizations.cljs
@@ -6,6 +6,7 @@
             [rems.atoms :as atoms :refer [readonly-checkbox document-title]]
             [rems.flash-message :as flash-message]
             [rems.common.roles :as roles]
+            [rems.config :as config]
             [rems.spinner :as spinner]
             [rems.table :as table]
             [rems.text :refer [text]]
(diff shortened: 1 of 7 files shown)


src/cljs/rems/config.cljs
```diff
@@ -24,3 +27,39 @@
 (defn dev-environment? []
   (let [config @(rf/subscribe [::config])]
     (boolean (:dev config))))
+
+;; organizations
```
opqdonut, line 31: should we also move fetching themes and translations to this file?

Macroz: Could be but maybe it also should be named after the APIs i.e. public? Or is that wrong too? Organizations are not exactly config so they could be in their own ns. They are also not public.

src/cljs/rems/administration/organizations.cljs
```diff
@@ -21,6 +22,11 @@
  ::fetch-organizations
  (fn [db]
    (let [description [text :t.administration/organizations]]
+     ;; Refresh :organizations subscription. We do this here
+     ;; since it would be surprising if an organization was visible on
+     ;; this page but not selectable e.g. when creating a resource.
+     (config/fetch-organizations!)
```
Macroz, line 28: I think this is a bit odd thing to do. It's not strictly necessary after the previous change?

Macroz (approved):

src/cljs/rems/administration/organizations.cljs
opqdonut: I think it's useful if e.g. you've created an org using the API and then want to use it via the UI, or perhaps somebody else just created an org and tells you to check it out.

Issue #2330, opened 2020-09-17
Add margin between "My applications" "Search" and "No applications" text.


Macroz:
Add a margin between "My applications" "Search" and "No applications" text. Now they are stuck together while the table of data would have a margin. This is reproducible with the reporter user as in the screenshot.

![reporter](https://user-images.githubusercontent.com/823661/94020713-e6bbcb00-fdbb-11ea-996a-e7fa5305e094.png)

- [ ] add margin
- [ ] bonus: check the other lists that they have the proper margin as well

PR #2362, opened 2020-09-28
style: add margin to search-field

marharyta:

## Reviewability
- [x] link to issue - https://github.com/CSCfi/rems/issues/2330

After:
Screenshot 2020-09-28 at 10 05 00

## API - na

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -791,7 +791,9 @@
      :.application-applicant {:max-width "10em"}]]
    [:.search-field {:display :flex
                     :flex-wrap :nowrap
-                    :align-items :center}
+                    :align-items :center
+                    :margin-bottom "1rem"
+                    :margin-top "1rem"}
     [:label {:margin-bottom 0}] ; override the default from Bootstrap
     [:div.input-group {:width "17em"}]]
    [:.search-tips {:font-size "0.9rem"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Changes since v2.13
 ### Changes
 
 ### Fixes
+- Add vertical margins around search field for better readability (#2330)
 
 ### Additions
 - All fields can have an info text, shown if the small icon is clicked. (#1863)


Macroz (commented):
This seems to cause a double-margin for the lower table so perhaps you could do it in some other way. You can have a look at the screenshot to see what I mean.

marharyta:
> This seems to cause a double-margin for the lower table so perhaps you could do it in some other way. You can have a look at the screenshot to see what I mean.

I did that on purpose

Macroz:
> > This seems to cause a double-margin for the lower table so perhaps you could do it in some other way. You can have a look at the screenshot to see what I mean.
> 
> I did that on purpose

Ok, but the margin in the lower table will be different than the tables in Catalogue or Administration pages then? Shouldn't it all be the same or what did you have in mind?

Catalogue margin is small
catalogue

Administration is small
admin

But here the lower table has twice the margin.
applications

marharyta:
> > > This seems to cause a double-margin for the lower table so perhaps you could do it in some other way. You can have a look at the screenshot to see what I mean.
> > 
> > 
> > I did that on purpose
> 
> Ok, but the margin in the lower table will be different than the tables in Catalogue or Administration pages then? Shouldn't it all be the same or what did you have in mind?
> 
> Catalogue margin is small
> catalogue
> 
> Administration is small
> admin
> 
> But here the lower table has twice the margin.
> applications

That is exactly what I had in mind, the margin should be bigger everywhere since search is semantically separate element  from the lists

Macroz:
> That is exactly what I had in mind, the margin should be bigger everywhere since search is semantically separate element from the lists

The search box belongs to the list, it affects the contents of the list and makes no sense by itself. That is one point. But the margin can be changed to be bigger if it looks better.

However, here the different lists from the different pages end up with a different margin between the search and the list. I do not understand this part. Consistency is good? The margin is not bigger everwhere?

marharyta:
@Macroz "Here the different lists from the different pages end up with a different margin between the search and the list. I do not understand this part"  - that would be a good reason to do some CSS refactoring then, I think

Macroz:
> @Macroz "Here the different lists from the different pages end up with a different margin between the search and the list. I do not understand this part" - that would be a good reason to do some CSS refactoring then, I think

I suggested initially of looking another way so it doesn't break, but if that change was intentional then fixing the other lists to match too is what needs to be done.

marharyta:
ok, @Macroz I can take a look

marharyta:
It seems to not be causing any layout changes anywhere except for the http://localhost:3000/applications page
I ll investigate

marharyta:
Looks like it needed a ```.mt-3``` class

marharyta:
@Macroz do you think you can take a look again now?

src/cljs/rems/applications.cljs
```diff
@@ -43,7 +43,7 @@
                                      :default-sort-column :created
                                      :default-sort-order :desc}]
         (when (roles/show-all-applications? (:roles identity))
-          [:<>
+          [:div.mt-3
```
Macroz, line 46: This will not fix any of the other tables, like in the screenshots the problem exists at least in the Catalogue page. Possibly every page that has search-component followed by the list-component. We should look for a solution that sets the margin to them and not to any wrapper div outside which is different for each page. And I think we have to check manually every table to see that the margin is correct and possible margin collapse is working correctly.

Macroz (changes requested):
I think we have to look for a solution still. And we really need to check every table.

src/cljs/rems/applications.cljs
marharyta: Ok, what do you sugggest/.

Macroz: I think there is no avoiding of going through all the tables by hand and checking which change would fix them all. Certainly adding `.mt-3` sounds like wrong, it should perhaps work if you added that to the list component but if it doesn't then have to figure out why.

src/clj/rems/css/styles.clj
```diff
@@ -473,6 +473,7 @@
    [:.alert-dark {:color (get-theme-attribute :alert-dark-color)
                   :background-color (get-theme-attribute :alert-dark-bgcolor)
                   :border-color (get-theme-attribute :alert-dark-bordercolor :alert-dark-color)}]
+   [:.applications.alert-secondary {:margin-top "1rem"}]
```
Macroz, line 476: How about instead of this, we just add `mt-3` to this line https://github.com/CSCfi/rems/blob/master/src/cljs/rems/application_list.cljs#L181

The we would know with certainty that it will ever only affect that one place that needed fixing.

src/cljs/rems/applications.cljs
```diff
@@ -43,7 +43,7 @@
                                      :default-sort-column :created
                                      :default-sort-order :desc}]
         (when (roles/show-all-applications? (:roles identity))
-          [:<>
+          [:div
```
Macroz, line 46: I think this change is leftover, we could revert it and have the `:<>` back?

Macroz (approved):
Here are some small things to look at. Otherwise it's about ready.

src/clj/rems/css/styles.clj
Macroz: I think this line can just be removed now?

marharyta: This line is what fixes the problem, why remove it?

Macroz (approved):

Issue #2354, opened 2020-09-24
Field description fixes


Macroz:
- [x] saving a field without all localizations for the field description loses the texts. It should rather be a validation error.
- [x] remove support for field description for the label type
- [x] check that all other field types have a working field description

Macroz:
@marharyta have you done the third bullet too? I mean did you (manually) check that each field type is sensible now and displays the field description correctly? Label obviously is done but I'm wondering if we have some type that doesn't work yet.

marharyta:
@Macroz yes, I did check them manually
What do you suggest? Should we write more tests or?

Macroz:
I was asking only to confirm that this task is completely done. It can be then checked and closed. I think we don't need tests for all of them, the importance is quite low.

PR #2365, opened 2020-09-28
test: add validation on info-text field in case if not label

marharyta:

## Reviewability
- [x] link to issue - https://github.com/CSCfi/rems/issues/2354

## API

## Documentation

## Different installations

## Testing

## Accessibility
- [ ] all fields have a label
- [ ] contrast is checked

## Follow-up


src/cljs/rems/administration/create_form.cljs
```diff
@@ -134,15 +134,19 @@
   (into {} (for [language languages]
              [language (trim-when-string (get lstr language ""))])))
 
+(defn- string-is-not-empty [s]
+  (not (str/blank? s)))
+
```
opqdonut, line 139: unused function here

test/cljs/rems/administration/test_create_form.cljs
```diff
@@ -212,7 +223,17 @@
...
+      (is (= [{:field/id "fld1"
+               :field/title {:en "en title"
+                             :fi "fi title"}
+               :field/optional true
+               :field/type :text
+               :field/max-length 12
+               :field/placeholder {:en "en placeholder"
+                                   :fi "fi placeholder"}}]
+
+             (mapv #(build-request-field % languages) fields-one-empty-info-text))))))
```
opqdonut, line 236: Hmm, I would prefer for `:field/info-text` to be present in this case, which would trigger a validation error for it from the backend, instead of just silently dropping it.

This isn't such a big deal because we also run the validations in the frontend so you should never get this far.

opqdonut (approved):
I'm fine with merging this after removing the unused function.

If you feel like it, improving the behaviour of build-request would be nice, but it's not a blocker since it doesn't affect actual use.

test/cljs/rems/administration/test_create_form.cljs
Macroz: Yeah I think not sending it is the right thing for the UI to do, but it does leave the backend untested. Backend could have its own test case for this but maybe not worth it.

opqdonut: We have a test for the common form-validation code anyway.

src/cljs/rems/administration/create_form.cljs
marharyta: I see, I pushed it by mistake

test/cljs/rems/administration/test_create_form.cljs
marharyta: Do you think I could do that in a different PR?
I do not think that we currently test any field at this stage?

opqdonut: it's fine like this, as I said it's not a big deal, we're just thinking out loud with @Macroz

opqdonut (approved):

test/cljs/rems/administration/test_create_form.cljs
marharyta: what do you think @Macroz ?

Macroz (approved):
Works for me :+1:

PR #2387, opened 2020-10-05
chore: bump etaoin to latest

Macroz:
I didn't get any browser test error so let's try in a PR.

Original comment for 0.3.6 was:
```
  ;; TODO 0.3.10 breaks browser tests for some reason
```


opqdonut (approved):
if tests pass, let's just merge

PR #2389, opened 2020-10-05
fix: try a longer timeout for booting driver

Macroz:
We have the the CI env sometimes timing out here. Let's try with a
longer timeout to see if it helps. NB: default wait has decreased from
20s to 7s in the 0.3.9.

See also https://github.com/igrishaev/etaoin/issues/338


Macroz:
I will try rerunning the test a few times to see if I can still make CI break.

opqdonut (approved):

Issue #2142, opened 2020-04-24
Reading bona fide status from IDP


jaakkocsc:
Upon user login, the ELIXIR AAI provides REMS with a GA4GH Passport (an OIDC claim containing the Passport). The Passport contains information about user's bona fide status, which may be used to automate certain application processing.

Depends on #2130

jaakkocsc:
OP, OpenID Provider = ELIXIR AAI
RP, Relying Party = REMS

1. Upon login RP receives an ID Token and an AccessToken
2. RP queries OP's UserInfo endpoint with the AccessToken
3. RP receives GA4GH Passport as a response
4. RP looks for a Visa with type "ResearcherStatus" from the Passport (there may be multiple)
5. Value for the Visa must be "so" or "system" for the logged in person having a university granted research status (i.e. is able to accept bona fide requests in REMS)

PR #2395, opened 2020-10-06
feat: configurable oidc scopes, fetch oidc userinfo

opqdonut:
some preparatory work for #2142

on top of #2393

## Reviewability
- [x] link to issue

## Documentation

## Different installations

## Follow-up

### DIFF
--- a/src/clj/rems/auth/oidc.clj
+++ b/src/clj/rems/auth/oidc.clj
@@ -51,9 +51,15 @@
                        ;; providers differ in what they give us
                        :commonName (some id-data [:name :unique_name :family_name])
                        :mail (:email id-data)}
-        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))]
+        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))
+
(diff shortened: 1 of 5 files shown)


src/clj/rems/auth/oidc.clj
```diff
@@ -51,9 +51,15 @@
                        ;; providers differ in what they give us
                        :commonName (some id-data [:name :unique_name :family_name])
                        :mail (:email id-data)}
-        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))]
+        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))
+
+        user-info (when-let [url (:userinfo_endpoint oidc-configuration)]
```
Macroz, line 56: Why the underscore in the key?

Macroz (approved):

src/clj/rems/auth/oidc.clj
opqdonut: The `oidc-configuration` is read from `https://{oidc-domain}/.well-known/openid-configuration`, and we don't do `snake_case` to `kebab-case` conversion for the keys. I think it's good to use the standard keys here instead of changing their format.

PR #2398, opened 2020-10-08
initial implementation of reading bona fide status from idp

opqdonut:
for #2142

This implementation has been tested on remsdev against the
elizir-czech test authentication service.

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/ga4gh.clj
+++ b/src/clj/rems/ga4gh.clj
@@ -36,7 +36,7 @@
    :sub s/Str
    :exp s/Int
    :iat s/Int
-   :scope (s/eq "openid") ;; could also list multiple space-separated scopes, as long as one is "openid"
+   (s/optional-key :scope) s/Str ;; Embedded Document Tokens "MUST NOT contain 'openid'"
    (s/optional-key :jti) s/Str
    :ga4gh_visa_v1 VisaObject})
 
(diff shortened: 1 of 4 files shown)


opqdonut:
Some questions for reviewers:
- Is `:bonafide` a good key in user attributes? Should we use something else?
- Do you think this needs more tests? I can always write tests that use mock JWTs but that won't guarantee compatibility with the actual tokens.
- There is still some refactoring & implementing to do, but I think this is a mergeable piece.

opqdonut:
commit https://github.com/CSCfi/rems/pull/2398/commits/3f66dfb8b6d145369d034c0a1977bd598821e262 tested against elixir-czech login

src/clj/rems/ga4gh.clj
```diff
@@ -34,10 +37,12 @@
    :sub s/Str
    :exp s/Int
    :iat s/Int
-   :scope (s/eq "openid") ;; could also list multiple space-separated scopes, as long as one is "openid"
+   (s/optional-key :scope) s/Str ;; Embedded Document Tokens "MUST NOT contain 'openid'"
```
Macroz, line 40: This would be the right place for `;`

Macroz (approved):

src/clj/rems/ga4gh.clj
opqdonut: I don't like how emacs indents `;`....

opqdonut: hmm no wait that's changed, it no longer indents the `;` to a fixed column

Macroz: it's only if you put it on a newline?

PR #2403, opened 2020-10-09
render researcher status in frontend

opqdonut:
for #2142

## Reviewability
- [x] link to issue

## API

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -602,7 +602,7 @@
   [{:keys [element-id attributes application group? can-remove? accepted-licenses?]}]
   (let [application-id (:application/id application)
         user-id (:userid attributes)
-        other-attributes (dissoc attributes :name :userid :email :organizations :notification-email)
+        other-attributes (dissoc attributes :name :userid :email :organizations :notification-email :researcher-status-by)
         title (cond (= (:userid (:application/applicant application)) user-id) (text :t.applicant-info/applicant)
                     (:userid attributes) (text :t.applicant-info/member)
                     :else (text :t.applicant-info/invited-member))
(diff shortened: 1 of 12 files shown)


src/cljs/rems/application.cljs
```diff
@@ -630,7 +630,9 @@
                        (when-let [mail (:email attributes)]
                          [info-field (text :t.applicant-info/email) mail {:inline? true}])
                        (when-let [organizations (seq (:organizations attributes))]
-                         [info-field (text :t.applicant-info/organization) (str/join ", " (map organization-name-if-known organizations)) {:inline? true}])]
+                         [info-field (text :t.applicant-info/organization) (str/join ", " (map organization-name-if-known organizations)) {:inline? true}])
+                       (when (#{:so :system} (:researcher-status-by attributes))
+                         [info-field (text :t.applicant-info/researcher-status) [readonly-checkbox {:value true}] {:inline? true}])]
```
Macroz, line 635: I wonder if this should be shown as false when it does not exist? This status is kind of important for the REMS instance running the bona fide stuff so it may be better to explicitly show false?

Macroz (approved):
Looks good.

One additional thing that came to my mind is that perhaps it would be useful for the verified researcher to be able to confirm their own status somewhere. Could we put the user's attributes to the user's own info page or something?

src/cljs/rems/application.cljs
opqdonut: Yeah I thought about that, but then it would show an uninteresting field for other installations. I guess I could make it configurable though...

opqdonut:
seeing your own attributes is a good idea, I'll make a follow-up PR with it

src/cljs/rems/application.cljs
Macroz: We can always add it later. We can have a user experience check when the feature is complete.

PR #2408, opened 2020-10-19
feat: show user attributes on settings page

opqdonut:
follow-up to #2403
for #2142

## Reviewability
- [x] link to issue

## Documentation

### DIFF
--- a/src/cljs/rems/user.cljs
+++ b/src/cljs/rems/user.cljs
@@ -1,0 +1,31 @@
+(ns rems.user
+  (:require [clojure.string :as str]
+            [re-frame.core :as rf]
+            [rems.atoms :refer [info-field readonly-checkbox]]
+            [rems.common.util :refer [index-by]]
+            [rems.text :refer [text localized]]))
+
+(defn attributes [attributes]
+  (let [language @(rf/subscribe [:language])
+        organization-by-id @(rf/subscribe [:organization-by-id])
+        organization-name-if-known (fn [organization]
+                                     (if-let [known-organization (organization-by-id (:organization/id organization))] ; comes from idp, maybe unknown
+                                       (get-in known-organization [:organization/short-name language])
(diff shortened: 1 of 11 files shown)


opqdonut:
details

opqdonut:
This is kind of the MVP for users seeing their own attributes. I think the "Settings" page could probably be renamed to something else, perhaps "Your account"?

src/cljs/rems/user.cljs
```diff
@@ -0,0 +1,70 @@
...
+            [rems.common.application-util :refer [get-member-name]]
+            [rems.common.util :refer [index-by]]
+            [rems.text :refer [text localized]])
+  (:require-macros [rems.guide-macros :refer [component-info example]]))
+
+(defn username [attributes]
+  (when-let [name (get-member-name attributes)]
+    [info-field (text :t.applicant-info/name) name {:inline? true}]))
+
+(defn attributes [attributes]
```
Macroz, line 14: This could have a docstring. I'm kind of missing the detail about where this is supposed to be used.

Macroz (approved):
I think after adding this details section we should brush up the notification settings a bit too. I'm not sure it is the most comprehensible one at the moment. At minimum it could be e.g.under heading "Notifications".

Also I don't quite like that this page has its own look, e.g. have a look at this screenshot. Perhaps it would work with two regular collapses with the "Notifications" and "Your user details" headings? Small change but rather more unified look.

three

opqdonut:
So how about calling the page something like "User" or "Profile" and then having "Notification settings" and "User Attributes" as subtitles?

Macroz:
> So how about calling the page something like "User" or "Profile" and then having "Notification settings" and "User Attributes" as subtitles?

I think that's a good idea too.

opqdonut:
profile

Macroz (approved):
I like it more!

PR #2413, opened 2020-10-20
test: remove redundant case from test-invite-member

opqdonut:


test/clj/rems/application/test_commands.clj
```diff
@@ -1086,24 +1086,6 @@
...
-                :application/member {:name "Member Applicant 1"
-                                     :email "member1@applicants.com"}
-                :invitation/token "very-secure"}
-               (ok-command application
-                           {:type :application.command/invite-member
-                            :actor handler-user-id
-                            :member {:name "Member Applicant 1"
-                                     :email "member1@applicants.com"}}
-                           injections)))))
     (testing "other users cannot invite members"
```
Macroz, line 1089: I think the original intent was probably that a handler could invite a member even before submitting. However I think that as a feature also doesn't make sense as draft should not be seen by anyone but the members. And the test used submitted state here for some weird reason.

Macroz (approved):

test/clj/rems/application/test_commands.clj
opqdonut: Yeah. Looking at the history the test has been this way for quite some time, at least since bad0a693dec3d56496438131612ce4a84e3184fd

PR #2421, opened 2020-10-30
fix: use UserId instead of plain s/Str

Macroz:
I noticed that we have two spots in the event schemas where we use `s/Str` though we could use the `UserId` alias. Eventually perhaps we would use the `{:userid xxx}` structure but this is better than nothing.


opqdonut (approved):
this doesn't affect the swagger schema, right?

Macroz:
> this doesn't affect the swagger schema, right?

It should not as `UserId` here is just a regular def.

Issue #2367, opened 2020-09-29
Fix focus in create resource


Macroz:
When choosing `Administration`, then `Resources` and finally `Create`, the focus should be in the first field but it seems to disappear.

- [ ] fix focus to go to the first field
- [ ] bonus: check the other create navigations too

BadAlgorithm:
Hi @Macroz I'm happy to tackle this one. 

I've had a quick look at the codebase and can see where the focus is going. Did you have a specific solution in mind or should I just try open a PR and get your thoughts?

Macroz:
Great! No specific solution but I urge you to have a look at what we have done elsewhere by looking for `focus` in the code. Please, ask us if you have any questions!

BadAlgorithm:
Hi @Macroz,

I've got a few questions about this issue, it seems to be a little trickier than I originally anticipated.

I've identified a few problems which makes this a little challenging.

Firstly, I noticed that there is a root `on-update` function that calls focus on the `#main-content` element. I did remove this since it would often steal focus from the dropdowns. However, I want to ask why was this being focused by default? I'm hoping this wouldn't break existing functionality. My current solution is to extend the dropdown component to take in an `auto-focus?` flag that will forward to the Select component. This works, but I would've loved to be able to use the centralised on-update function.

My initial solution was to add custom focus classes to each of the first children in the forms and query for that class (via `document.getElementsBySelector`). This solution was nice since it allowed me to use the centralised `on-update` function and implement it in a backwards-compatible way. However it seems like the lifecycle methods don't allow me to do this. It would call the update function before the children had a chance to mount.

What are your thoughts?

Macroz:
Good questions and thinking.

I think the idea was to move the focus initially to a relevant element. However, it is not working correctly. Typically forms would/should(?) focus the first input element, and obviously here it didn't work so. I think for pages where there are no inputs, the idea was to focus something else instead. However I know that there should be the "skip to main content link" that you can navigate to backwards, so I wonder why the autojump is there.

These are all related to accessibility requirements. I'll dig around for the answer a bit. I couldn't find anything quickly that would specify how the initial focus should be set. Only that e.g. navigation and focus must be logical. The logic can really only be tested with NVDA on Windows or such setup and I can do that out once we have some solution to try.

I'll get back to you!

Macroz:
[...]

PR #2422, opened 2020-10-30
Focus on h1 element when triggered by user-navigation

BadAlgorithm:
Partially fixes
https://github.com/CSCfi/rems/issues/2367

This is the basic solution to getting headers to focus when navigating between pages. Falls back to `
#main-container` if it cannot find any headers. Due to the semantic nature of `h1`, there should be only 1 present on the page at any given moment so it should be okay. 

Although I've mostly been out of action for a couple of weeks I'm happy to continue with the next step which is getting focus on first load (due to the lifecycle management of the application, this may need bigger modifications).

### DIFF
--- a/src/cljs/rems/spa.cljs
+++ b/src/cljs/rems/spa.cljs
@@ -101,6 +101,7 @@
     :languages [:en]
     :default-language :en
     :translations {}
+    ::grab-focus? true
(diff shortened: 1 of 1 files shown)


Macroz (approved):
Looks good to me :+1: 

I think we have to have a look at the focus indicators at some point :smile:

Issue #2040, opened 2020-03-09
Reviewer/decider invitation


Macroz:
When doing a request a reviewer or decider can be invited by email even if they are not an existing REMS user
- [x] send invitation flow
- [x] send email with token
- [x] accept invitation flow
  - [x] connect invited person to correct application (and/or organization) and role upon logging in and accepting the token
  - [x] localized messages (accepting and problems)
- details like in member invitation
- [x] share what is possible with member invitation too

PR #2426, opened 2020-11-04
refactor: factor out user-selection widget for actions

opqdonut:
related to #2040

## Reviewability
- [x] link to issue

### DIFF
--- a/resources/translations/en.edn
+++ b/resources/translations/en.edn
@@ -40,7 +40,6 @@
             :remove-member "Remove member"
             :request-decision "Request decision"
             :request-review "Request review"
-            :request-selection "Send request to"
             :request-selections "Send requests to"
             :resources-selection "Resources included in the application:"
             :return "Return to applicant"
--- a/src/cljs/rems/actions/request_decision.cljs
+++ b/src/cljs/rems/actions/request_decision.cljs
@@ -1,42 +1,17 @@
 (ns rems.actions.request-decision
   (:require [re-frame.core :as rf]
-            [rems.actions.components :refer [action-attachment action-button comment-field action-form-view button-wrapper command!]]
-            [rems.atoms :refer [enrich-user]]
-            [rems.dropdown :as dropdown]
-            [rems.flash-message :as flash-message]
-            [rems.text :refer [text]]
-            [rems.util :refer [fetch]]))
+            [rems.actions.components :refer [action-attachment action-button comment-field action-form-view button-wrapper command! user-selection]]
+            [rems.text :refer [text]]))
 
 (def ^:private action-form-id "request-decision")
-(def ^:private dropdown-id "request-decision-dropdown")
-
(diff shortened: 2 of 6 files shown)


resources/translations/en.edn
```diff
@@ -40,7 +40,6 @@
             :remove-member "Remove member"
             :request-decision "Request decision"
             :request-review "Request review"
-            :request-selection "Send request to"
```
Macroz, line 43: But this singular is used for request decision as that is always to a single person?

src/cljs/rems/actions/request_decision.cljs
```diff
@@ -57,39 +32,28 @@
...
                         :application-id application-id}]
-    [:div.form-group
-     [:label {:for dropdown-id} (text :t.actions/request-selection)]
-     [dropdown/dropdown
-      {:id dropdown-id
-       :items potential-deciders
-       :item-key :userid
-       :item-label :display
-       :item-selected? #(contains? (set selected-deciders) %)
-       :multi? true
```
opqdonut, line 81: See here, decider selection has had `multi? true`. The event & command also support multiple users.

Macroz: Well it could potentially be wrong. We don't have it as configurable and I guess that's ok for now, but the assumption we had in an earlier discussion couple weeks ago (when doing the wording change) was that there would only ever be one decider.

opqdonut: Which wording change? Looking at the history, the label has been `:t.actions/request-selection` and `multi?` has been `true` at least since 2019-08-01, commit a6c15aae1c5afcbeb66706434b0b30f36a6a646b.

Macroz: https://github.com/CSCfi/rems/pull/2397

opqdonut: Right! What do you propose? I find that the UI is clear enough even with multiselection, and it doesn't really do any harm to support multiple deciders.

Macroz: Well I would just like to confirm from @jaakkocsc  that it's so far ok to have multiple deciders. The changes look fine otherwise.

Macroz (approved):
Let's just check that the change wrt. decider requests / wording is ok but otherwise let's merge.

src/cljs/rems/actions/request_decision.cljs
jaakkocsc: I'd say the decision on having only one decider or multiple deciders is organisational, by the "organisation-owner", not a technical limitation set by REMS.

Macroz: It simplifies REMS to have all the invites/requests to be available for multiple people.

PR #2425, opened 2020-11-03
refactor: nicer rems.actions.components

opqdonut:
related to #2040

# Definition of Done / Review checklist

## Reviewability
- [x] link to issue

## Documentation
- [ ] add or update docstrings for namespaces and functions
- [ ] components are added to guide page

## Follow-up
- [x] no critical TODOs left to implement

PR #2432, opened 2020-11-05
fix: actions with empty comment

opqdonut:
Before #2425 the action comment defaulted to "", which was sent to the
backend if no comment was entered. After, it defaulted to nil, which
violated the API schema and resulted in a HTTP 400. Return to the old
behaviour, but add TODOs for cases where it might make sense to
require a nonempty comment.

## Reviewability
- [ ] link to issue

## Testing
  - these could have browser tests

## Follow-up

### DIFF
--- a/src/cljs/rems/actions/add_licenses.cljs
+++ b/src/cljs/rems/actions/add_licenses.cljs
@@ -23,7 +23,7 @@
    {:db (assoc db
                ::potential-licenses #{}
                ::selected-licenses #{})
-    :dispatch [:rems.actions.components/set-comment action-form-id nil]
+    :dispatch [:rems.actions.components/set-comment action-form-id ""]
     ::fetch-licenses #(rf/dispatch [::set-potential-licenses %])}))
 
 (defn- assoc-all-titles
--- a/src/cljs/rems/actions/review.cljs
+++ b/src/cljs/rems/actions/review.cljs
@@ -8,7 +8,7 @@
 (rf/reg-event-fx
  ::open-form
  (fn [_ _]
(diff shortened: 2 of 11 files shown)


src/cljs/rems/actions/review.cljs
```diff
@@ -35,7 +35,7 @@
    [[button-wrapper {:id "review-button"
                      :text (text :t.actions/review)
                      :class "btn-primary"
-                     :on-click on-send}]]
+                     :on-click on-send}]] ;; TODO disable submit if comment field is empty?
```
opqdonut, line 38: #2433

src/cljs/rems/actions/add_licenses.cljs
```diff
@@ -23,7 +23,7 @@
    {:db (assoc db
                ::potential-licenses #{}
                ::selected-licenses #{})
-    :dispatch [:rems.actions.components/set-comment action-form-id nil]
+    :dispatch [:rems.actions.components/set-comment action-form-id ""]
```
Macroz, line 26: Do you have plans to also remove the empty string requirement at some point (and change Schema?)

Macroz (approved):

src/cljs/rems/actions/add_licenses.cljs
opqdonut: I guess the right choice would be to not set the `:comment` field in the command if there's no command.

... but I guess we could also support nil values for the field in the backend.

opqdonut: But no, no concrete plans right now

PR #2435, opened 2020-11-09
feat: initial invite decider/reviewer UI

opqdonut:
this has all the functionality, but we might want to simplify the
actions pane by combining request-review with invite-reviewer etc.

for #2040

## Reviewability
- [x] link to issue

## Documentation

## Testing
  - will add browser tests once UI stabilises

## Accessibility
- [x] all fields have a label

## Follow-up


opqdonut:
invite-reviewer
invite-events

marharyta (approved):

src/cljs/rems/actions/components.cljs
```diff
@@ -76,6 +76,32 @@
...
+                         :on-change #(on-change (.. % -target -value))}]])
+
+(defn name-field [{:keys [field-key]}]
+  [text-field-view {:id (str "name-" field-key)
+                    :label (text :t.actions/member-name)
+                    :value @(rf/subscribe [::name field-key])
+                    :on-change #(rf/dispatch [::set-name field-key %])}])
+
+(defn email-field [{:keys [field-key]}]
+  [text-field-view {:id (str "email-" field-key)
```
Macroz, line 100: Could this be of type `email ?

resources/translations/en.edn
```diff
@@ -27,7 +27,9 @@
                      :submission-failed "Submission failed. Check the following errors:"}
             :export-entitlements "Export entitlements as CSV"
             :handled-applications "Processed applications"
+            :invite-decider "Invite decider by email"
             :invite-member "Invite member"
+            :invite-reviewer "Invite reviewer by email"
```
Macroz, line 32: I wonder if these should rather be called `Request review by email`? From the user's point of view it does the review request.

Macroz: Originally I thought that we'd have just one Request review action and there you could choose how you want to invite by e.g. existing user autocomplete, email address list or organization reviewer autocomplete. I don't really want to see three different review actions. I think it could be made to be one action form with option on how/who to send it to?

Macroz (approved):
Code looks good. I'd like to confirm the UX side of having a separate action for this. Also perhaps you can see if the email type would work.

src/cljs/rems/actions/components.cljs
opqdonut: good catch, done

resources/translations/en.edn
opqdonut: yeah "Request review by email" is ok, but I'll try to combine the UIs also

opqdonut:
This is what it looks like now:

request-dropdown

request-via-email

src/cljs/rems/actions/invite_decider_reviewer.cljs
```diff
@@ -76,12 +76,12 @@
      :decider decider-form-id
      :reviewer reviewer-form-id)
    (case role
-     :decider (text :t.actions/invite-decider)
-     :reviewer (text :t.actions/invite-reviewer))
+     :decider (str (text :t.actions/request-decision-menu) " " (text :t.actions/invite-decider-suffix))
+     :reviewer (str (text :t.actions/request-review-menu) " " (text :t.actions/invite-decider-suffix)))
```
Macroz, line 80: Also here it would be better to not concatrnate strings but use templates as they are from tempura or with format parameters.

Macroz (commented):
I'll check the UX still

src/cljs/rems/actions/invite_decider_reviewer.cljs
opqdonut: Would you prefer (text :t.actions/request-decision-menu) had a parameter at the very end? That feels like overkill when we can just catenate. The localizations this UI needs are

- The action buttons: "Request review", "Request decision"
- The dropdown elements: "... from user", "... via email"
- The form titles "Request review from user", "Request review by email" etc.
- The submit buttons: "Request review", "Request decision"

I feel the more verbose form titles are important so the user can see that his selection was honoured. But if the form title were just "Request review", we wouldn't need string operations for translations.

src/cljs/rems/actions/invite_decider_reviewer.cljs
```diff
@@ -59,15 +59,15 @@
...
+                :text (str "... " (text :t.actions/invite-decider-suffix))
+                :on-click #(rf/dispatch [::open-form :decider])}])
 
-(defn invite-reviewer-action-button []
-  [action-button {:id reviewer-form-id
-                  :text (text :t.actions/invite-reviewer)
-                  :on-click #(rf/dispatch [::open-form :reviewer])}])
+(defn invite-reviewer-action-link []
+  [action-link {:id reviewer-form-id
+                :text (str "... " (text :t.actions/invite-reviewer-suffix))
```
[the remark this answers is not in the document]
Macroz: Adding via CSS is wrong too in the same way. None of these use the translation features and would not work if the language specifies the use of ellipsis character, different word order, right-to-left or weirder non-English behavior.

src/cljs/rems/actions/invite_decider_reviewer.cljs
Macroz: The point is just leaving these details open, and to be within the translation system. I think there is not much advantage in trying to DRY the translations. And clever use of `text-format`, especially multiple rounds of translations, does not feel like right too. I think the simplest, and within the translation system still, would be to duplicate the translations themselves.

Maybe one option is to make the `...` operation a helper in the text namespace so that we could mostly avoid the CSS hack and the string concatenation and leave it to our translation system from where we can fix it easily if we need to.

opqdonut: fair enough

opqdonut: fixed

src/cljs/rems/actions/invite_decider_reviewer.cljs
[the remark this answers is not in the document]
opqdonut: fixed this, didn't fix the existing btn-opens-more thing

src/cljs/rems/actions/invite_decider_reviewer.cljs
Macroz: I think it looks pretty good like this :+1:

PR #2447, opened 2020-11-18
Improve browser tests

Macroz:
Trying to improve the speed of running browser tests by reusing the browser instance.

Also try waiting before inputting so that we would not get the lost initial character error.


opqdonut (approved):
code looks great, let's see if it's actually faster

PR #2450, opened 2020-11-19
fix: don't allow handlers to edit all organizations

### LINKED ISSUE #2441: check-allowed-organization! is broken
... it's looking at the user's organizations, instead of the organization's owners

opqdonut:
This leftover case in forbidden-organization? caused a bug where a
handler that was the owner of one organization could edit items in any
organization.

The special case for handlers was needed in the old design
(see PR #1959) that only allowed organization owners to see items in
their own organization. A new design (issue #1893, PR #2034, ADR 009)
allows everyone (handlers, owners, organization owners) to see
items in all organizations.

Fixes #2441.

## Reviewability
- [x] link to issue

## Documentation
  - entry already exists

## Testing
  - couldn't figure out a nice way to regression test this, unless we
    want to add a very specific test

## Follow-up

### DIFF
--- a/src/clj/rems/api/services/util.clj
+++ b/src/clj/rems/api/services/util.clj
@@ -7,11 +7,9 @@
 
 (defn- forbidden-organization? [organization]
   (let [not-owner? (not (contains? context/*roles* :owner))
-        not-handler? (not (contains? context/*roles* :handler)) ;; handlers have read-only access to all orgs
         organization-owners (set (map :userid (:organization/owners organization)))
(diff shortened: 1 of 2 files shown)


Macroz (approved):
We could perhaps rename this `forbidden-organization?` to better reflect where it is used. For read or writes for example.

opqdonut:
Replaced the private `forbidden-organization?` with a `may-edit-organization?`. Considered renaming `check-allowed-organization?` with `check-allowed-to-edit-organization?` but that felt too verbose.

Macroz (approved):

Issue #1523, opened 2019-08-22
Bona fide referer workflow; implement bonafidebot


opqdonut:
We need to support applying for a bona fide status. The application should work roughly like this:

- User applies for a bona fide catalogue item in REMS and names a referer (who should already have an _idp-provided bona fide status_)
  - Form should have: referer email address, free text field, and a license
- Referer should get an email, and be able to approve/reject the user once logged in to REMS
- If the referer approves, the application is approved and an entitlement to the bona fide resource is granted
- Elixir AAI reads the bona fide entitlement from REMS (via permission API #1319)
  - This should be a ResearcherStatus visa with `"by": "peer"`
- Sometime later a super-user (or rather, a human handler for the workflow in question) can log in and close or revoke the application thus ending the entitlement and bona fide status (e.g. does not trust person anymore)

Plan:
- [x] bona fide status is modeled as a catalogue item with a _default workflow_ that has bonafidebot as handler
 - [x] bonafidebot sends a decision request by email to the given address (like review request #2040)
- [x] the referer clicks on the link in his email, logs in, and his bona fide status (or lack of it) is sent by Elixir AAI to REMS (requires #2130 #2142).
  - The referer is given the decider role for the application.
- [x] bonafidebot waits for the decision from the referer, verifies the situation and responds with a clear message
  - possibility for bots to return (error) messages to the user directly (split out to #2511)
  - [x] bonafidebot approves when decider approves and has bona fide status
  - [x] bonafidebot rejects when decider rejects
  - [x] bonafidebot rejects when decider does not have bona fide status
    - or perhaps the application should be left waiting and the referer can try to get a bona fide status and try again?
- customize email text for decision request (currently it's a generic "Bonafide Bot has requested a decsion from you ...") (split out to #2511)
  - one option is to just use extra-translations to override the email text for the bonafide rems instance
- generating ResearcherStatus visas (in addition to the current ControlledAccessGrants visas) (split out to #2513)

Notes and questions:
- Later, bonafidebot could be an external integration that uses event notifications (instead of a REMS core feature) (#2095)
- No bot is needed if the decision request command handler checks the bona fide status
however, the bot solution allows us to isolate the bona fide -related code in one place.
- Display a message that the user does not have bona fide status when trying to approve? This would require specific code and break the isolation. A bot message would not.
- A review request or decision request? A decider is more natural to have the other close/reject commands than just a reviewer.
- There are separate *idp-provided bona fide status*, and *REMS-provided bona fide status*
[...]

PR #2456, opened 2020-11-24
 initial implementation of bona-fide-bot

opqdonut:
for #1523

## Reviewability
- [x] link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/docs/bots.md
+++ b/docs/bots.md
@@ -54,3 +54,26 @@ Example of creating the bot user with the API.
 ```sh
 curl -X POST -H "content-type: application/json" -H "x-rems-api-key: 42" -H "x-rems-user-id: owner" http://localhost:3000/api/users/create --data '{"userid": "rejecter-bot", "name": "Rejecter Bot", "email": null}'
 ```
+
+## Bona Fide bot
+
+User id: `bonafide-bot`
+
+The Bona Fide bot is used for granting peer-verified GA4GH
+ResearcherStatus visas. See [ga4gh-visas.md](ga4gh-visas.md) for more
+background.
+
+The Bona Fide bot is designed to be used with
+- a _default workflow_ that has the bot as a handler (and optionally some human handlers)
+- a catalogue item that has a form that has an email field
+
+When an application gets submitted for the catalogue item, the bot
+sends a _decision request_ to the email address it extracts from the
+application. Then the bot waits until the recipient of the request
+logs in to rems and performs the _decide_ action. At this point:
+
+- If the decider has a ResearcherStatus visa (with `"by": "so"` or
+  `"by": "system"`, see from their IDP, [ga4gh-visas.md](ga4gh-visas.md)):
+  - and if the decider posted an approve decision: the bot approves the application
+  - and if the decider posted a reject decision: the bot rejects the application
+- If the decider doesn't have a ResearcherStatus visa, nothing happens.
--- a/docs/bots.md
+++ b/docs/bots.md
@@ -57,7 +57,7 @@ curl -X POST -H "content-type: application/json" -H "x-rems-api-key: 42" -H "x-r
 
 ## Bona Fide bot
 
-User id: `bonafide-bot`
+User id: `bona-fide-bot`
 
 The Bona Fide bot is used for granting peer-verified GA4GH
 ResearcherStatus visas. See [ga4gh-visas.md](ga4gh-visas.md) for more
--- a/docs/bots.md
+++ b/docs/bots.md
@@ -65,7 +65,7 @@ background.
 
 The Bona Fide bot is designed to be used with
 - a _default workflow_ that has the bot as a handler (and optionally some human handlers)
-- a catalogue item that has a form that has an email field
+- a catalogue item that has a form that has an email field (in case of multiple email fields, the _first one_ is used)
 
 When an application gets submitted for the catalogue item, the bot
 sends a _decision request_ to the email address it extracts from the
--- a/docs/bots.md
+++ b/docs/bots.md
@@ -76,4 +76,4 @@ logs in to rems and performs the _decide_ action. At this point:
   `"by": "system"`, see from their IDP, [ga4gh-visas.md](ga4gh-visas.md)):
   - and if the decider posted an approve decision: the bot approves the application
   - and if the decider posted a reject decision: the bot rejects the application
-- If the decider doesn't have a ResearcherStatus visa, nothing happens.
+- If the decider doesn't have a ResearcherStatus visa, the bot rejects the application
--- a/src/clj/rems/application/bonafide_bot.clj
+++ b/src/clj/rems/application/bonafide_bot.clj
@@ -1,6 +1,8 @@
 (ns rems.application.bonafide-bot
   "A bot that enables workflows where a user can ask another user to vouch
-   for their bona fide researcher status."
+   for their bona fide researcher status.
+
+   See also: docs/bonafide.md, docs/ga4gh-visas.md"
   (:require [clojure.test :refer [deftest is testing]]
             [clj-time.core :as time]
             [rems.common.application-util :as application-util]
--- a/src/clj/rems/application/bonafide_bot.clj
+++ b/src/clj/rems/application/bonafide_bot.clj
@@ -19,6 +19,62 @@
         (mapcat :form/fields
                 (:application/forms application))))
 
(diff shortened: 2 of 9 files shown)


src/clj/rems/application/bonafide_bot.clj
```diff
@@ -0,0 +1,46 @@
...
+  "A bot that enables workflows where a user can ask another user to vouch
+   for their bona fide researcher status."
+  (:require [clj-time.core :as time]
+            [rems.common.application-util :as application-util]
+            [rems.db.applications :as applications]
+            [rems.db.users :as users]))
+
+(def bot-userid "bonafide-bot")
+
+(defn- find-email-address [application]
```
Macroz, line 15: Should we add some protection about having just one email address? Or just use `find-first` to be more explicit about which to return?

test/clj/rems/api/test_end_to_end.clj
```diff
@@ -607,3 +609,97 @@
...
+                            :researcher-status-by "so"}
+        bot-attributes {:userid bonafide-bot/bot-userid
+                        :email nil
+                        :name "bonafide bot"}]
+    (testing "create users"
+      (api-call :post "/api/users/create" applicant-attributes api-key owner-id)
+      (api-call :post "/api/users/create" referer-attributes api-key owner-id)
+      (api-call :post "/api/users/create" bot-attributes api-key owner-id))
+    (let [resource-id
+          (extract-id
```
Macroz, line 633: I would prefer to have these on the same line. It does not change indentation much but at least it reserves the first column for names and second for values.

In `cond` we at least add blanks after each "section" and here that is both followed and not.

src/clj/rems/application/bonafide_bot.clj
```diff
@@ -18,29 +20,96 @@
...
+                                  :email "refer2000@example.com"
+                                  :name "Ref Errer"}]
+          (testing "bot not handler"
+            (is (empty? (generate-commands event
+                                           referer-attributes
+                                           {:application/workflow {:workflow.dynamic/handlers [{:userid "handler"}]}}))))
+          (testing "bot is handler,"
+            (let [application {:application/workflow {:workflow.dynamic/handlers [{:userid bot-userid}]}}]
+              (testing "referer does not have researcher status"
+                (is (empty? (generate-commands event referer-attributes application))))
```
Macroz, line 95: Should this rather reject the application completely if the referer is invalid?

docs/bots.md
```diff
@@ -54,3 +54,26 @@ Example of creating the bot user with the API.
 ```sh
 curl -X POST -H "content-type: application/json" -H "x-rems-api-key: 42" -H "x-rems-user-id: owner" http://localhost:3000/api/users/create --data '{"userid": "rejecter-bot", "name": "Rejecter Bot", "email": null}'
 ```
+
+## Bona Fide bot
+
+User id: `bonafide-bot`
```
Macroz, line 60: We use "bona fide" and "bonafide". I wonder if `bona-fide-bot` would make more sense?

docs/bots.md
```diff
@@ -54,3 +54,26 @@ Example of creating the bot user with the API.
...
+
+The Bona Fide bot is used for granting peer-verified GA4GH
+ResearcherStatus visas. See [ga4gh-visas.md](ga4gh-visas.md) for more
+background.
+
+The Bona Fide bot is designed to be used with
+- a _default workflow_ that has the bot as a handler (and optionally some human handlers)
+- a catalogue item that has a form that has an email field
```
Macroz, line 68: Maybe we should document here that there must be only one email field (or if we change so that the _first_ will always be used)?

docs/bots.md
```diff
@@ -54,3 +54,26 @@ Example of creating the bot user with the API.
...
+When an application gets submitted for the catalogue item, the bot
+sends a _decision request_ to the email address it extracts from the
+application. Then the bot waits until the recipient of the request
+logs in to rems and performs the _decide_ action. At this point:
+
+- If the decider has a ResearcherStatus visa (with `"by": "so"` or
+  `"by": "system"`, see from their IDP, [ga4gh-visas.md](ga4gh-visas.md)):
+  - and if the decider posted an approve decision: the bot approves the application
+  - and if the decider posted a reject decision: the bot rejects the application
+- If the decider doesn't have a ResearcherStatus visa, nothing happens.
```
Macroz, line 79: I'm wondering if this is the right thing to do or rather should we reject here too?

Macroz (approved):
Some ideas for discussion. Looks good!

src/clj/rems/application/bonafide_bot.clj
opqdonut: could be an assertion too I guess, but it's a bit impolite if it fails only when the bot tries to do something, not when creating the workflow

test/clj/rems/api/test_end_to_end.clj
opqdonut: I guess I forgot to add the blank lines, this style would definitely need them. This same two-line-let style is used in the other end-to-end tests and it makes it easier to see the repetition between the statements, especially in `test-end-to-end` which creates a lot of stuff. In this case everything is a bit simpler so I can switch to the standard let style.

docs/bots.md
opqdonut: Yeah, it seems "bona fide" is more correct, even though "bonafide" is also seen. However I wonder if this should be called something like "referral-bot" or "researcher-status-bot"? I'm not sure we want to spread the bona fide term in the code since it isn't even used by the GA4GH standards (which talk about "researcher status").

docs/bots.md
opqdonut: I'll document the current behaviour for now.

src/clj/rems/application/bonafide_bot.clj
opqdonut: added a test

docs/bots.md
opqdonut: @jaakkocsc says let's go with "bona fide"

Macroz: Yes so with `bona-fide-bot` too then right?

src/clj/rems/application/bonafide_bot.clj
Macroz: Yeah ideally there would be some collaboration with the form editor or you would pick the email field in the workflow/bot settings. Or this could be a report of its own, data validation sort of things. We don't have to do it now as long as the docs match the exact behavior.

opqdonut: We could use field ids for this btw, fetching the unique field with id "referrer-email" or something?

Of course if there are multiple forms there might be multiple fields with the same field id, but we could use the first form.

docs/bots.md
opqdonut: I'll rename everything :+1:

src/clj/rems/application/bonafide_bot.clj
opqdonut: The ticket actually wonders about this too.

> - bonafidebot rejects when decider does not have bona fide status
>   - or perhaps the application should be left waiting and the referer can try to get a bona fide status and try again?

I'll change it to reject for now.

docs/bots.md
opqdonut: It rejects now.

test/clj/rems/api/test_end_to_end.clj
```diff
@@ -607,3 +609,92 @@
...
+                            :researcher-status-by "so"}
+        bot-attributes {:userid bona-fide-bot/bot-userid
+                        :email nil
+                        :name "bona fide bot"}]
+    (testing "create users"
+      (api-call :post "/api/users/create" applicant-attributes api-key owner-id)
+      (api-call :post "/api/users/create" referer-attributes api-key owner-id)
+      (api-call :post "/api/users/create" bot-attributes api-key owner-id))
+    (let [resource-id
+          (extract-id (api-call :post "/api/resources/create" {:organization {:organization/id "default"}
```
Macroz, line 633: Here I would expect the `resource-id` to start the line and continue with `(extract-id ...)`

Macroz (approved):

test/clj/rems/api/test_end_to_end.clj
opqdonut: ah missed a spot, fixed

PR #2461, opened 2020-11-30
ADR 014: User identities

opqdonut:
for #2464

### DIFF
--- a/docs/architecture/014-userids.md
+++ b/docs/architecture/014-userids.md
@@ -137,6 +137,9 @@ the userid.
 
 This is backwards compatible with current behaviour.
 
+[Issue #2366](https://github.com/CSCfi/rems/issues/2366),
+[PR #2471](https://github.com/CSCfi/rems/pull/2471)
+
 ## 2. Workaround for THL instance
 
 The `sub` attributes are going to change when the identity provider
--- a/docs/architecture/014-userids.md
+++ b/docs/architecture/014-userids.md
@@ -4,17 +4,23 @@ Authors: @opqdonut @Macroz @jaakkocsc
 
 # Background
 
-TODO
+Currently REMS uses the userid received from the OIDC server (from the
+`sub` claim or another claim based on configuration), and uses this
+userid as the internal database key for that user. The userid is
+embedded in application events, workflows, etc.
 
 ## Problems with the current approach
 
 - The `sub` attribute might be opaque and not usable between systems, e.g. when sending entitlements to other systems
-- Different login methods may have different formats for externally referrable fields
+- Different login methods may have different formats for external user identifiers
 - Different login methods might use conflicting values for claims
   - E.g. `foo@csc.fi` via HAKA or CSC SSO (TODO: do we know of cases like this?)
 - A single user might have multiple external identifiers (TODO: do we know of cases like this?)
 
 ## Use cases for user identities
+
+A user identifier can be used for multiple things. We might want to
+use different identifiers for different purposes.
 
 - Referring to users inside REMS data (e.g. workflow handler)
 - Referring to users in API responses (e.g. workflow handler)
@@ -28,7 +34,9 @@ We'll add internal random user ids to REMS. This internal user id will
 be the key that users are referred to within REMS. The user's external
 id will be stored in the attributes JSON blob.
 
-TODO: a good index is needed for performance
+This will allow us more flexibility in the future when identity
+requirements and use cases change, and might also make all sorts of
+migrations easier (since internal user ids don't need to be touched).
 
 ## Login
 
@@ -142,7 +150,32 @@ ids to be used when pushing entitlements to other Elixir services.
 
 This can be accomplished by setting `:oidc-userid-attributes` to `["elixirId" "old_sub"]`.
 
-## 3. Add internal ids
+## 3. Use internal ids everywhere
 
-Implement the full internal ids solution described in this ADR in
-2021Q1.
+Implement the "Login" section of this ADR, and also write the
+necessary database migrations (see also Migration under Open questions
+below).
+
+## 4. Support for external ids in APIs
+
+Implement the "User identity resolution" section of this ADR, one API
+at a time. This should probably be in the same release as step 3.
+
+# Open questions
+
+## API format
+
+Should we switch all APIs that take userids to a structured
+`{"userid": "abc123"}` form? Then we could specify users more
+explicitly like `{"eppn": "user@example.com"}`.
+
+## Migration
+
+Should we migrate existing userids to new random values or keep them
+as-is?
+
+## Database indexes
+
+A good index is needed for performance when looking up users based on
+attributes. Is storing the data in JSONB fine or should we also add
+new columns?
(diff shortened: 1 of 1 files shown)


docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+## User identity resolution
+
+When somebody uses the userid `xyz` in an API call, REMS searches for the user like this:
+
+1. If there is a user with a REMS internal user id `xyz`, use that user
+2. Iterate through every attribute `attr` in `:oidc-user-attributes`, in order
+   1. If there is a user with `attr: xyz`, use that user
+   2. If not, keep iterating
+3. If no user was found, return HTTP 400 Bad Request
+
```
Macroz, line 61: Should we also support search with `"attr: xyz"` so that you can use any attribute for the user?

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+## User identity resolution
+
+When somebody uses the userid `xyz` in an API call, REMS searches for the user like this:
+
+1. If there is a user with a REMS internal user id `xyz`, use that user
+2. Iterate through every attribute `attr` in `:oidc-user-attributes`, in order
+   1. If there is a user with `attr: xyz`, use that user
+   2. If not, keep iterating
+3. If no user was found, return HTTP 400 Bad Request
```
Macroz, line 60: Should we also return 400 if many users are found but one was expected?

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+Implement the "User identity resolution" section of this ADR, one API
+at a time. This should probably be in the same release as step 3.
+
+# Open questions
+
+## API format
+
+Should we switch all APIs that take userids to a structured
+`{"userid": "abc123"}` form? Then we could specify users more
+explicitly like `{"eppn": "user@example.com"}`.
```
Macroz, line 173: Or have a complementary API call with the structured values? Or put the JSON in a string?

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+
+## `x-rems-user-id`
+
+Should we allow external user ids in `x-rems-user-id`?
```
Macroz, line 177: I think it could be nice, as long as we don't leak info in it.

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+
+## User search API
+
+Should we have an API for searching / resolving users?
```
Macroz, line 181: :+1:

Macroz: Maybe we don't need the UI for it at first.

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Migration
+
+Should we migrate existing userids to new random values or keep them
+as-is?
```
Macroz, line 186: I first thought that we'd keep them as is, treat them as randomly generated to their current value. Possibly the migration is not that hard but certainly some extra work.

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Database indexes
+
+A good index is needed for performance when looking up users based on
+attributes. Is storing the data in JSONB fine or should we also add
+new columns?
```
Macroz, line 192: We also have Lucene that we could use for the search.

Macroz (approved):
Added some comments. Looks good to go on.

docs/architecture/014-userids.md
opqdonut: Yes!

docs/architecture/014-userids.md
opqdonut: I'll write that up as a future option.

opqdonut: Ah that's actually already mentioned in Open questions > API format

docs/architecture/014-userids.md
opqdonut: Hmm yeah I guess we could do something like

```
(defschema User (s/or s/String {s/Keyword s/String})
```

I'll add a note about that.

docs/architecture/014-userids.md
opqdonut: Good point, I'll add a note about that.

docs/architecture/014-userids.md
Macroz: Yeah, was just thinking that it should be as a string here but it's commented also later.

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,198 @@
...
+
+```
+GET /api/applications/1234
+{"application/id": 1234,
+ "application/events":
+ [...
+  {"event/type": "application.event/review-requested",
+   "application/id": 1234,
+   "event/actor": "5e8edd851d2fdfbd7415232c67367cc3",
+   "event/actor-attributes": {"userid": "5e8edd851d2fdfbd7415232c67367cc3",
```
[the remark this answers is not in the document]
opqdonut: Ah I see what you mean. They way REMS is written at the moment is that any `userid` attribute from the id token would be overridden by what REMS deduces the userid to be (based on `:oidc-userid-attributes`). In that sense the `userid` attribute name is reserved by REMS, and a user organization can't configure REMS to pass through an attribute called `userid` from the IDP.

If this were a problem we could switch to a less easily conflicting id like `x-rems-user-id`, but that would require changes all over the code base.

docs/architecture/014-userids.md
```diff
@@ -0,0 +1,198 @@
...
+Problem: A user has logged in using some identity provider and created
+some applications. However now he has started logging in via another
+identity provider and the entitlements should be associated with his
+new identity.
+
+Solution: Copy the users attributes from their new `users` row to
+their old `users` row. Remove the new `users` row. Now when the user
+logs in they get their old identity, which the entitlements are
+associated with. The external id related to those entitlements has
+also changed.
```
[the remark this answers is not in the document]
opqdonut: The entitlements and the event log would refer to the internal user id. However in the API we would output the external user id. So this would effectively change the external user id associated with their entitlements and events. This might break some consistency requirements in systems that integrate with REMS, but would be ok from the standpoint of REMS's internal consistency.

Anyway, we probably shouldn't do this to userids that have gotten entitlements. Instead the old applications should be closed and the user should just reapply for whatever is needed using their right identity.

docs/architecture/014-userids.md
jaakkocsc: For impersonating I would think the internal id would be a bit safer option as it is much harder to guess.

docs/architecture/014-userids.md
jaakkocsc: We could generate new random internal ids, but then we would need to guess the `{"attr": "value"}` pairs for those ids. Probably with an excellent prdictability, but still a guess is a guess.

docs/architecture/014-userids.md
Macroz: Internal ids require for the user of the API to figure out the internal id, store it somewhere and then call with it in the future. The external id is much more convenient because likely it is known without the extra steps.

I'm not sure there is any safety advantage here, the API-Key is the one that defines who can actually call the API.

jaakkocsc (approved):

Issue #2363, opened 2020-09-28
Improve application UI and UX by placing logo in the upper left corner as it is a common prcatice


marharyta:
Currently the logo is in the center.
The research suggests it is bad UX practice: [https://www.nngroup.com/articles/centered-logos/](https://www.nngroup.com/articles/centered-logos/)

I suggest to follow a better UX convention and place the REMS logo in the upper left corner of navigation header as a link to homepage. 

Should support hero image for some customers and left-upper logo for navigation otherwise

- check themes from deploy repo

Macroz:
Have to remember that each instance of REMS has their own logo, colors etc. that may or may not work the same. This like other major (cosmetic) changes to the UI, should be tested against all the instances(themes, and provided that it changes a lot, approved by each of the instances.

PR #2483, opened 2020-12-09
feat: add logo to navigation bar

marharyta:
# Checklist for author

## Reviewability
- [x] Link to issue - [https://github.com/CSCfi/rems/issues/2363](https://github.com/CSCfi/rems/issues/2363) 
- [x]  Based on  [https://github.com/CSCfi/rems-deploy/pull/212](https://github.com/CSCfi/rems-deploy/pull/212)

master
Screenshot 2021-01-13 at 14 52 20
feat-2363 with navigation logo disabled
Screenshot 2021-01-13 at 14 54 54
feat-2363 with navigation menu enabled
Screenshot 2021-01-13 at 14 56 53

fake login is the same in master and feat-2363
Screenshot 2021-01-13 at 14 57 37

Screenshot 2021-01-13 at 14 59 34

Screenshot 2021-01-13 at 15 13 35

Screenshot 2021-01-13 at 15 23 58

## Documentation

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
-- The navigation menu brand logo is now configurable. You can see `:theme :navbar-logo  ` configuration variable that if set to true adds logo to the top navigation menu. [configuration.md](docs/configuration.md). (#2363)
+- The navigation menu brand logo is now configurable. You can see ` :logo-name-navigation ` configuration variable that if provided with alogo image adds logo to the top navigation menu. [configuration.md](docs/configuration.md). (#2363)
 
 ## v2.15 "Tapiolantie" 2021-01-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
-- The navigation menu brand logo is now configurable. You can see `:theme :navbar-logo` configuration variable that if set to true adds logo top navigation menu. [configuration.md](docs/configuration.md). (#2363)
+- The navigation menu brand logo is now configurable. You can see `:theme :navbar-logo  ` configuration variable that if set to true adds logo to the top navigation menu. [configuration.md](docs/configuration.md). (#2363)
 
 ## v2.15 "Tapiolantie" 2021-01-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
+- The navigation menu brand logo is now configurable. You can see `:theme :navbar-logo` configuration variable that if set to true adds logo top navigation menu. [configuration.md](docs/configuration.md). (#2363)
+
 ## v2.15 "Tapiolantie" 2021-01-08
 
 ### Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
-- The navigation menu brand logo is now configurable. You can see ` :logo-name-navigation ` configuration variable that if provided with alogo image adds logo to the top navigation menu. [configuration.md](docs/configuration.md). (#2363)
+- The navbar can be configured to show a logo image. When the `:navbar-logo-name` config is provided, the logo is shown in the navbar (top navigation menu). This logo also can be customized per language like the regular logo. (#2363)
 
 ## v2.15 "Tapiolantie" 2021-01-08
 
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -123,8 +123,7 @@
  :theme {:color1 "#eee"
          :color2 "#555"
          :color3 "#000"
-         :color4 "#000"
-         :navbar-logo false}
+         :color4 "#000"}
 
  ;; Optional extra static resources directory.
  :extra-static-resources nil
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -125,7 +125,7 @@
          :color3 "#000"
          :color4 "#000"
          ;; navigation type options
-         :upper-left-logo false}
+         :navbar-logo false}
 
  ;; Optional extra static resources directory.
  :extra-static-resources nil
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -124,7 +124,6 @@
          :color2 "#555"
          :color3 "#000"
          :color4 "#000"
-         ;; navigation type options
          :navbar-logo false}
 
  ;; Optional extra static resources directory.
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -123,7 +123,9 @@
  :theme {:color1 "#eee"
          :color2 "#555"
          :color3 "#000"
-         :color4 "#000"}
+         :color4 "#000"
+         ;; navigation type options
+         :upper-left-logo false}
 
  ;; Optional extra static resources directory.
  :extra-static-resources nil
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -52,8 +52,8 @@
 (defn get-logo-name-sm [lang]
   (resolve-image (get-theme-attribute (keyword (str "logo-name-" (name lang) "-sm")) :logo-name-sm)))
 
-(defn get-logo-navigation [lang]
-  (resolve-image (get-theme-attribute (keyword (str "logo-name-" (name lang) "-navigation")) :logo-name-navigation)))
+(defn get-navbar-logo [lang]
+  (resolve-image (get-theme-attribute (keyword (str "navbar-logo-name-" (name lang))) :navbar-logo-name)))
 
 (defn- generate-at-font-faces []
   (list
@@ -589,7 +589,7 @@
                                  :background-origin (get-theme-attribute :logo-content-origin)}]
    [(s/descendant :.logo-menu :.img) {:height "100%"
                                       :background-color (get-theme-attribute :logo-bgcolor)
-                                      :background-image (get-logo-navigation context/*lang*)
+                                      :background-image (get-navbar-logo context/*lang*)
                                       :-webkit-background-size :contain
                                       :-moz-o-background-size :contain
                                       :-o-background-size :contain
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -35,6 +35,10 @@
       (or attr-value (recur (rest attr-names))))))
 
 (def content-width (u/px 1200))
+(def logo-height-menu (u/px 40))
+(def logo-height (u/px 150))
+(def menu-height 56)
+(def recalculated-menu-height (u/px (int (+ menu-height 12))))
 
 (defn resolve-image [path]
   (when path
@@ -96,14 +100,16 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo
-                          {:height (u/px 150)}]))
+                         [:.logo {:height logo-height
+                                  :min-height logo-height}]))
    (stylesheet/at-media {:max-width (u/px 870)}
                         [:.user-widget [:.icon-description {:display "none"}]])
    (stylesheet/at-media {:min-width (u/px 480)}
                         [:.commands {:white-space "nowrap"}])
    (stylesheet/at-media {:prefers-reduced-motion :reduce}
-                        [:body {:scroll-behavior :auto}])))
+                        [:body {:scroll-behavior :auto}])
+   (stylesheet/at-media {:max-width (u/px 480)}
+                        [:.navbar-brand {:display "none"}])))
 
 (defn- generate-phase-styles []
   [:.phases {:width "100%"
@@ -350,12 +356,13 @@
     [:label.form-check-label {:cursor :pointer}]]
    [:html {:position :relative
            :min-width (u/px 320)
+           :min-height "100vh"
            :height (u/percent 100)}]
    [:body {:font-family (get-theme-attribute :font-family "'Lato', sans-serif")
            :min-height (u/percent 100)
            :display :flex
            :flex-direction :column
-           :padding-top (u/px 56)
+           :padding-top recalculated-menu-height
            :scroll-behavior :smooth}]
    [:h1 :h2 {:font-weight 400}]
    [:h1 {:margin-bottom (u/rem 2)}]
@@ -369,7 +376,7 @@
    [:.fixed-top {:background-color "#fff"
                  :border-bottom (get-theme-attribute :header-border "3px solid #ccc")
                  :box-shadow (get-theme-attribute :header-shadow :table-shadow)
-                 :min-height (u/px 56)}]
+                 :min-height menu-height}]
    [:.skip-navigation {:position :absolute
                        :left (u/em -1000)}
     [:&:active
@@ -381,11 +388,11 @@
                     :min-height (u/px 300)
                     :max-width content-width
                     :flex-grow 1
-                    ;; Height of navigation + logo, to avoid page content going under
+                    ;; Height of navigation, to avoid page content going under
                     ;; the navigation bar when the main content is focused.
                     ;; See https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors
-                    :padding-top (u/px 212)
-                    :margin-top (u/px -212)}]
+                    :padding-top recalculated-menu-height
+                    :margin-top (u/px (- (int (+ menu-height 6))))}]
    [(s/> :.spaced-sections "*:not(:first-child)") {:margin-top (u/rem 1)}]
    [:.btn {:white-space :nowrap
            :font-weight (button-navbar-font-weight)}]
@@ -555,6 +562,25 @@
                    :flex-direction "row"
                    :justify-content "space-between"
                    :min-width "100%"}]
+
+   ;; Logo, login, etc.
+   [:.logo-menu {:height logo-height-menu
+                 :background-color (get-theme-attribute :logo-bgcolor)
+                 :width "100px"}]
+   [:.logo {:height logo-height
+            :background-color (get-theme-attribute :logo-bgcolor)
+            :width "100%"
+            :margin "0 auto"}]
+   [(s/descendant :.logo :.img) (s/descendant :.logo-menu :.img) {:height "100%"
+                                                                  :background-color (get-theme-attribute :logo-bgcolor)
+                                                                  :background-image (get-logo-image context/*lang*)
+                                                                  :-webkit-background-size :contain
+                                                                  :-moz-o-background-size :contain
+                                                                  :-o-background-size :contain
+                                                                  :background-size :contain
+                                                                  :background-repeat :no-repeat
+                                                                  :background-position [[:center :center]]
+                                                                  :background-origin (get-theme-attribute :logo-content-origin)}]
    ;; Footer
    (let [footer-text-color (get-theme-attribute :footer-color :table-heading-color "#fff")]
      [:footer {:width "100%"
@@ -571,24 +597,6 @@
                             :bottom 0
                             :right (u/px 140)}]])
 
-   ;; Logo, login, etc.
-   [:.logo {:height (u/px 140)
-            :background-color (get-theme-attribute :logo-bgcolor)
-            :padding "0 20px"
-            :margin-bottom (u/em 1)}]
-   [(s/descendant :.logo :.img) {:height "100%"
-                                 :background-color (get-theme-attribute :logo-bgcolor)
-                                 :background-image (get-logo-image context/*lang*)
-                                 :-webkit-background-size :contain
-                                 :-moz-o-background-size :contain
-                                 :-o-background-size :contain
-                                 :background-size :contain
-                                 :background-repeat :no-repeat
-                                 :background-position [[:center :center]]
-                                 :background-origin (get-theme-attribute :logo-content-origin)
-                                 :padding-left (u/px 20)
-                                 :padding-right (u/px 20)
-                                 :padding-top (u/px 8)}]
    [:.jumbotron
     {:background-color "#fff"
      :text-align "center"
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -52,6 +52,9 @@
 (defn get-logo-name-sm [lang]
   (resolve-image (get-theme-attribute (keyword (str "logo-name-" (name lang) "-sm")) :logo-name-sm)))
 
+(defn get-logo-navigation [lang]
+  (resolve-image (get-theme-attribute (keyword (str "logo-name-" (name lang) "-navigation")) :logo-name-navigation)))
+
 (defn- generate-at-font-faces []
   (list
    (stylesheet/at-font-face {:font-family "'Lato'"
@@ -526,9 +529,13 @@
      :letter-spacing (u/rem 0.015)
      :padding-left 0
      :padding-right 0
-     :color (get-theme-attribute :navbar-color "#111")}
+     :color (get-theme-attribute :navbar-color "#111")
+     :justify-content "space-between"}
     [:.nav-link :.btn-link
      {:background-color :inherit}]]
+   [:#administration-menu
+    [:.nav-link
+     {:padding ".5rem 0"}]]
    [:.navbar-top-bar {:width (u/percent 100)
                       :height (u/px 4)
                       :display :flex
@@ -582,7 +589,7 @@
                                  :background-origin (get-theme-attribute :logo-content-origin)}]
    [(s/descendant :.logo-menu :.img) {:height "100%"
                                       :background-color (get-theme-attribute :logo-bgcolor)
-                                      :background-image (get-logo-name-sm context/*lang*)
+                                      :background-image (get-logo-navigation context/*lang*)
                                       :-webkit-background-size :contain
                                       :-moz-o-background-size :contain
                                       :-o-background-size :contain
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -103,7 +103,7 @@
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
                          [:.logo {:height logo-height}]
-                         [:.navbar-brand {:display "none"}]))
+                         [:.logo-menu {:display "none"}]))
    (stylesheet/at-media {:max-width (u/px 870)}
                         [:.user-widget [:.icon-description {:display "none"}]])
    (stylesheet/at-media {:min-width (u/px 480)}
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -38,7 +38,7 @@
 (def logo-height-menu (u/px 40))
 (def logo-height (u/px 150))
 (def menu-height 56)
-(def recalculated-menu-height (u/px (int (+ menu-height 12))))
+(def recalculated-menu-height (u/px (+ menu-height 12)))
 
 (defn resolve-image [path]
   (when path
@@ -100,16 +100,14 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo {:height logo-height
-                                  :min-height logo-height}]))
+                         [:.logo {:height logo-height}]))
    (stylesheet/at-media {:max-width (u/px 870)}
                         [:.user-widget [:.icon-description {:display "none"}]])
    (stylesheet/at-media {:min-width (u/px 480)}
-                        [:.commands {:white-space "nowrap"}])
+                        [:.commands {:white-space "nowrap"}]
+                        [:.navbar-brand {:display "none"}])
    (stylesheet/at-media {:prefers-reduced-motion :reduce}
-                        [:body {:scroll-behavior :auto}])
-   (stylesheet/at-media {:max-width (u/px 480)}
-                        [:.navbar-brand {:display "none"}])))
+                        [:body {:scroll-behavior :auto}])))
 
 (defn- generate-phase-styles []
   [:.phases {:width "100%"
@@ -356,7 +354,6 @@
     [:label.form-check-label {:cursor :pointer}]]
    [:html {:position :relative
            :min-width (u/px 320)
-           :min-height "100vh"
            :height (u/percent 100)}]
    [:body {:font-family (get-theme-attribute :font-family "'Lato', sans-serif")
            :min-height (u/percent 100)
@@ -365,7 +362,8 @@
            :padding-top recalculated-menu-height
            :scroll-behavior :smooth}]
    [:h1 :h2 {:font-weight 400}]
-   [:h1 {:margin-bottom (u/rem 2)}]
+   [:h1 {:margin-bottom (u/rem 2)
+         :margin-top (u/rem 2)}]
    [:#app {:min-height (u/percent 100)
            :flex 1
            :display :flex}]
@@ -388,11 +386,11 @@
                     :min-height (u/px 300)
                     :max-width content-width
                     :flex-grow 1
-                    ;; Height of navigation, to avoid page content going under
+                    ;; Height of navigation + logo, to avoid page content going under
                     ;; the navigation bar when the main content is focused.
                     ;; See https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors
-                    :padding-top recalculated-menu-height
-                    :margin-top (u/px (- (int (+ menu-height 6))))}]
+                    :padding-top (u/px 212)
+                    :margin-top (u/px -212)}]
    [(s/> :.spaced-sections "*:not(:first-child)") {:margin-top (u/rem 1)}]
    [:.btn {:white-space :nowrap
            :font-weight (button-navbar-font-weight)}]
@@ -570,7 +568,9 @@
    [:.logo {:height logo-height
             :background-color (get-theme-attribute :logo-bgcolor)
             :width "100%"
-            :margin "0 auto"}]
+            :margin "0 auto"
+            :padding "0 20px"
+            :margin-bottom (u/em 1)}]
    [(s/descendant :.logo :.img) (s/descendant :.logo-menu :.img) {:height "100%"
                                                                   :background-color (get-theme-attribute :logo-bgcolor)
                                                                   :background-image (get-logo-image context/*lang*)
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -38,7 +38,6 @@
 (def logo-height-menu (u/px 40))
 (def logo-height (u/px 150))
 (def menu-height 56)
-(def recalculated-menu-height (u/px (+ menu-height 12)))
 
 (defn resolve-image [path]
   (when path
@@ -100,12 +99,12 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo {:height logo-height}]))
+                         [:.logo {:height logo-height}]
+                         [:.navbar-brand {:display "none"}]))
    (stylesheet/at-media {:max-width (u/px 870)}
                         [:.user-widget [:.icon-description {:display "none"}]])
    (stylesheet/at-media {:min-width (u/px 480)}
-                        [:.commands {:white-space "nowrap"}]
-                        [:.navbar-brand {:display "none"}])
+                        [:.commands {:white-space "nowrap"}])
    (stylesheet/at-media {:prefers-reduced-motion :reduce}
                         [:body {:scroll-behavior :auto}])))
 
@@ -359,7 +358,7 @@
            :min-height (u/percent 100)
            :display :flex
            :flex-direction :column
-           :padding-top recalculated-menu-height
+           :padding-top (u/px (+ menu-height 12))
            :scroll-behavior :smooth}]
    [:h1 :h2 {:font-weight 400}]
    [:h1 {:margin-bottom (u/rem 2)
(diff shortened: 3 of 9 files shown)


resources/config-defaults.edn
```diff
@@ -123,7 +123,9 @@
  :theme {:color1 "#eee"
          :color2 "#555"
          :color3 "#000"
-         :color4 "#000"}
+         :color4 "#000"
+         ;; navigation type options
+         :upper-left-logo false}
```
Macroz, line 128: How about we name this `:navbar-logo` because it may not always be in the upper or left?

src/clj/rems/css/styles.clj
```diff
@@ -35,6 +35,10 @@
       (or attr-value (recur (rest attr-names))))))
 
 (def content-width (u/px 1200))
+(def logo-height-menu (u/px 40))
+(def logo-height (u/px 150))
+(def menu-height 56)
+(def recalculated-menu-height (u/px (int (+ menu-height 12))))
```
Macroz, line 41: Why is the recalculation done? I don't get the purpose from the name so we should consider a better name.

Also there should be no need to cast an int to an int.

src/clj/rems/css/styles.clj
```diff
@@ -96,14 +100,16 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo
-                          {:height (u/px 150)}]))
+                         [:.logo {:height logo-height
+                                  :min-height logo-height}]))
```
Macroz, line 104: Why is this `:min-height` needed now?

src/clj/rems/css/styles.clj
```diff
@@ -96,14 +100,16 @@
...
    (stylesheet/at-media {:max-width (u/px 870)}
                         [:.user-widget [:.icon-description {:display "none"}]])
    (stylesheet/at-media {:min-width (u/px 480)}
                         [:.commands {:white-space "nowrap"}])
    (stylesheet/at-media {:prefers-reduced-motion :reduce}
-                        [:body {:scroll-behavior :auto}])))
+                        [:body {:scroll-behavior :auto}])
+   (stylesheet/at-media {:max-width (u/px 480)}
+                        [:.navbar-brand {:display "none"}])))
```
Macroz, line 112: This rule could be placed along with the `.commands` just above. They share the breakpoint (and should share the same).

src/clj/rems/css/styles.clj
```diff
@@ -350,15 +356,17 @@
     [:label.form-check-label {:cursor :pointer}]]
    [:html {:position :relative
            :min-width (u/px 320)
+           :min-height "100vh"
```
Macroz, line 359: What does this do?

src/clj/rems/css/styles.clj
```diff
@@ -381,11 +389,11 @@
...
                     :max-width content-width
                     :flex-grow 1
-                    ;; Height of navigation + logo, to avoid page content going under
+                    ;; Height of navigation, to avoid page content going under
                     ;; the navigation bar when the main content is focused.
                     ;; See https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors
-                    :padding-top (u/px 212)
-                    :margin-top (u/px -212)}]
+                    :padding-top recalculated-menu-height
+                    :margin-top (u/px (- (int (+ menu-height 6))))}]
```
Macroz, line 396: This does not anymore take into account the hero logo being there so can this work?

src/cljs/rems/administration/administration.cljs
```diff
@@ -19,7 +19,7 @@
      (text :t.administration/back)]))
 
 (defn navigator []
-  [:div#administration-menu.navbar.mb-4.mr-auto.ml-auto
+  [:div#administration-menu.navbar.mb-3.mr-auto.ml-auto
```
Macroz, line 22: This doesn't actually seem to change anything. Perhaps because the changed `h1` margins.

src/cljs/rems/navbar.cljs
```diff
@@ -60,6 +60,10 @@
               text (get-in page [:translations language :title] (text :t/missing))]
           [nav-link url text])))))
 
+(defn logo [menu-navigation?]
+  [:div {:class (str "navbar-brand" (if menu-navigation? " logo-menu" " logo"))}
```
Macroz, line 64: I think this is probably wrong. For example now you are putting the class `navbar-brand` also on the hero logo.

src/cljs/rems/navbar.cljs
```diff
@@ -80,13 +84,17 @@
...
-     {:type :button :data-toggle "collapse" :data-target "#small-navbar"}
-     "\u2630"]
-    [navbar-items :div#big-navbar.collapse.navbar-collapse.mr-3 identity]]
-   [:div.navbar [user-widget (:user identity)]]])
+  (let [theme @(rf/subscribe [:theme])]
+    [:nav.navbar-flex
+     [:div.navbar.navbar-expand-sm.flex-fill
+      (if (:upper-left-logo theme)
+        [logo (:upper-left-logo theme)]
+        nil)
```
Macroz, line 92: This can be written with `when` as you don't have the else clause.

src/cljs/rems/navbar.cljs
```diff
@@ -100,7 +108,8 @@
   (let [identity @(rf/subscribe [:identity])]
     [:div.fixed-top
      [skip-navigation]
-     [:div.navbar-top-bar [:div.navbar-top-left] [:div.navbar-top-right]]
+     [:div.navbar-top-bar
+      [:div.navbar-top-left] [:div.navbar-top-right]]
```
Macroz, line 112: I think we should either keep the old format or have `top-left` and `top-right be on their own lines. This is kind of in the middle of those two.

src/cljs/rems/spa.cljs
```diff
@@ -336,7 +334,6 @@
     (lazy-load-data!)
     [:div
      [nav/navigation-widget]
-     [logo]
```
Macroz, line 339: This breaks backwards compatibility so we should keep it until it is configurable.

src/cljs/rems/spa.cljs
```diff
@@ -196,6 +196,7 @@
 ;;;; Pages
 (defn login-intro []
   [:div
+   [nav/logo]
```
Macroz, line 199: This is broken with LBR theme because the element is not full width.

Macroz (changes requested):
Found some things to change and some questions. Ping me if you need screenshots of them. I think the biggest problem is that the change is not backwards compatible so the PR can't be merged before that is changed.

resources/config-defaults.edn
marharyta: sounds good!

src/clj/rems/css/styles.clj
marharyta: "cast an int to an int": well, I think this is the way it was done elsewhere in the styles file and also this is the way it doesn`t throw the error for me 

"Why is the recalculation done? I don't get the purpose from the name so we should consider a better name.' - I agree, the name could be better

src/clj/rems/css/styles.clj
marharyta: I believe I was tinkering with nav styles and arrived at a conclusion it was needed, but after a few iterations lost the sight of it. It neither fixes nor breaks anything.

src/clj/rems/css/styles.clj
opqdonut: It's probably this garden bug that needs a workaround:

https://github.com/CSCfi/rems/blob/master/src/clj/rems/css/styles.clj#L794

However I think it just got fixed so we could just upgrade garden.

src/clj/rems/css/styles.clj
marharyta: it makes sure that footer is not lifter above the end of screen

src/clj/rems/css/styles.clj
marharyta: Yes, this is true, let me re-add it

src/clj/rems/css/styles.clj
opqdonut: Yep, https://github.com/noprompt/garden/issues/120 is fixed, but no release yet. So let's go with the workaround for now.

src/cljs/rems/administration/administration.cljs
marharyta: that is the point. In cases where we have administration navbar and h1 below there is no too much space between them and in turn when h1 we have decent space on top.

long story short, it does not change anything in the places where admin navbar meets the h1

src/cljs/rems/navbar.cljs
marharyta: ok, I do not get how this is wrong

src/cljs/rems/navbar.cljs
marharyta: true. Good point, will fix

src/cljs/rems/navbar.cljs
marharyta: hmmm, ok, I am pretty sure cljfmt fixed it, but I can try

src/cljs/rems/spa.cljs
marharyta: may i ask how? this is not obvious to me since i am just putting the logo component into another

src/clj/rems/css/styles.clj
Macroz: Yeah but these are all ints. There is no decimal place.

src/cljs/rems/spa.cljs
marharyta: Let me check quickly

src/clj/rems/css/styles.clj
Macroz: In that case the less there is of diff the better so we should remove it.

src/clj/rems/css/styles.clj
Macroz: Why is it needed in this PR specifically? What has changed that has broken the page? To me it seems like everything works fine even without so in what case does it not work?

src/cljs/rems/administration/administration.cljs
Macroz: All the pages should have the navbar and the `h1` at the top so we could just remove this change?

src/cljs/rems/navbar.cljs
Macroz: The logo in the main content area is not in the navbar, so it should not have `navbar-brand` class which is for the logo in the navbar?

src/cljs/rems/spa.cljs
Macroz: This removes the hero logo from the main content. Then it is not the same as before for those installations that do not want this logo in the navbar. The default should be that if someone doesn't do the config change for the new feature then nothing changes. I.e. the hero logo stays where it has been.

src/clj/rems/css/styles.clj
marharyta: Right, it seems to work without int... weird, but ok

src/clj/rems/css/styles.clj
marharyta: I sometimes would see the footer slightly lifted so I thought It would not hurt to add it
But it can be a separate change if this is important

src/cljs/rems/navbar.cljs
marharyta: hah, good point

src/cljs/rems/spa.cljs
marharyta: ok, I am confused, @opqdonut can you maybe help me?

opqdonut: Looks to me like @marharyta has changed the behaviour so that the hero image shows up only in the login page. I think that's a reasonable change. However @Macroz would prefer the look of the site to _stay the same_ if the theme does not take `:upper-left-logo` into use. I think that's also reasonable and will make deploying the change easier (won't need to get an OK for the changed visuals from user organizations).

Would this work for you both?

- if `:upper-left-logo` is set
  - display hero image on login page
  - don't display hero on other pages
- if `:upper-left-logo` is not set
  - display hero image on all pages

src/cljs/rems/administration/administration.cljs
marharyta: not every page has administration navbar

src/clj/rems/css/styles.clj
marharyta: thank you @opqdonut

src/cljs/rems/administration/administration.cljs
marharyta: without administration navbar
Screenshot 2021-01-14 at 13 07 33

with administration navbar
Screenshot 2021-01-14 at 13 07 51

src/clj/rems/css/styles.clj
marharyta: so, would you like me to remove it?

src/clj/rems/css/styles.clj
marharyta: actually, now that i remember, min-height was actually meant for this part

src/cljs/rems/spa.cljs
marharyta: i am sorry, but i still do not get it, the look is the same

opqdonut: this is the difference I see.

master:
master-catalogue

this PR:
branch-catalogue

marharyta: that is weird, it should show everywhere (home, fake login, every screen) or then home, fake login and navbar

marharyta: let me check

src/clj/rems/css/styles.clj
Macroz: Can you reproduce it without this? If you can then please let me know how. If it is random are you sure this fixes it?

src/clj/rems/css/styles.clj
Macroz: What `min-height`?

src/cljs/rems/administration/administration.cljs
Macroz: But you are adding bottom margin to the admin navbar so surely you only need to consider the pages that have administration navbar (and all pages have `h1`)?

src/cljs/rems/spa.cljs
Macroz: I would rather have these two logos configurable separately. There is no need to couple them together.

marharyta: i see, I agree

src/clj/rems/css/styles.clj
marharyta: Ok, ill try!

src/cljs/rems/administration/administration.cljs
marharyta: ok, I have completely lost trail of your point

src/cljs/rems/spa.cljs
marharyta: I fixed the part that where the logo should show the logo on home

Macroz: The fix has a problem in that it duplicates the logo. These two `when` statements are unrelated so the component can appear twice.
doubled

Macroz (changes requested):
Also noticed that the logo is now different size compared to the previous one?

different_size

src/cljs/rems/spa.cljs
```diff
@@ -330,10 +329,15 @@
...
-        grab-focus? @(rf/subscribe [::grab-focus?])]
+        grab-focus? @(rf/subscribe [::grab-focus?])
+        theme @(rf/subscribe [:theme])]
     (lazy-load-data!)
     [:div
      [nav/navigation-widget]
+     (when-not (:navbar-logo theme)
+       [nav/logo])
+     (when (= page-id :home)
+          [nav/logo])
```
opqdonut, line 340: btw the logo can get duplicated because both of these conditions can be active at once

marharyta: that is true, I did notice it today 
Screenshot 2021-01-21 at 16 37 14
hahah, two is better than one :) ahah

src/cljs/rems/spa.cljs
marharyta: yes, that is true, I am working on fixing it :)

src/clj/rems/css/styles.clj
```diff
@@ -381,11 +387,11 @@
                     :min-height (u/px 300)
                     :max-width content-width
                     :flex-grow 1
-                    ;; Height of navigation + logo, to avoid page content going under
+                    ;; Height of navigation, to avoid page content going under
```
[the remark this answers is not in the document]
marharyta: On it now

resources/config-defaults.edn
```diff
@@ -123,7 +123,9 @@
  :theme {:color1 "#eee"
          :color2 "#555"
          :color3 "#000"
-         :color4 "#000"}
+         :color4 "#000"
+         ;; navigation type options
```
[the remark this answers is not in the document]
marharyta: agreed.

src/clj/rems/css/styles.clj
marharyta: yes, I did

src/cljs/rems/spa.cljs
marharyta: I did fix that, pushing changes now

src/clj/rems/css/styles.clj
marharyta: should be back at this point

src/clj/rems/css/styles.clj
```diff
@@ -96,12 +100,12 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo
-                          {:height (u/px 150)}]))
+                         [:.logo {:height logo-height}]))
```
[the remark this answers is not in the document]
marharyta: I think I just replaced ```(u/px 150)``` with ```logo-height``` here

marharyta:
I think I can summarize:
- add config options for the image
- add config option to dev config
- make sure LBR looks good
- reassure that the logo is aligned with the title on the homepage

@opqdonut @Macroz anything else?

opqdonut:
I'd also add

- clean up style (see Macroz's comments)
- make sure site appearance stays the same (as current master / latest release) when `:navbar-logo` is not set

marharyta:
Screenshot 2021-01-23 at 0 28 29
Screenshot 2021-01-23 at 0 28 48

@Macroz I believe my style fixes were implemented correctly

marharyta:
Screenshot 2021-01-23 at 0 49 58
Some text adjustment for the menu
@Macroz I added the m-3 here because now that the margin from the hero component is gone, we can clearly see from my screenshots

src/cljs/rems/administration/administration.cljs
marharyta: the problem only occurs when navbar meets h1, there is no point of changing it on every page.

src/clj/rems/css/styles.clj
```diff
@@ -555,6 +566,37 @@
...
+   [(s/descendant :.logo-menu :.img) {:height "100%"
+                                      :background-color (get-theme-attribute :logo-bgcolor)
+                                      :background-image (get-logo-navigation context/*lang*)
+                                      :-webkit-background-size :contain
+                                      :-moz-o-background-size :contain
+                                      :-o-background-size :contain
+                                      :background-size :contain
+                                      :background-repeat :no-repeat
+                                      :background-position [[:center :center]]
+                                      :background-origin (get-theme-attribute :logo-content-origin)}]
```
Macroz, line 599: We could remove some duplication from these two nearly identical blocks of styling.

src/clj/rems/css/styles.clj
```diff
@@ -571,24 +613,6 @@
...
-                                 :-webkit-background-size :contain
-                                 :-moz-o-background-size :contain
-                                 :-o-background-size :contain
-                                 :background-size :contain
-                                 :background-repeat :no-repeat
-                                 :background-position [[:center :center]]
-                                 :background-origin (get-theme-attribute :logo-content-origin)
-                                 :padding-left (u/px 20)
-                                 :padding-right (u/px 20)
-                                 :padding-top (u/px 8)}]
```
Macroz, line 591: Where has this padding gone? Is it not needed?

src/cljs/rems/administration/administration.cljs
Macroz: I'm not sure what problem you are referring to here. What I'm thinking is that the margins of the `navbar` and `h1` collapse so there is not too much space.

src/cljs/rems/navbar.cljs
```diff
@@ -60,6 +60,10 @@
               text (get-in page [:translations language :title] (text :t/missing))]
           [nav-link url text])))))
 
+(defn logo [menu-navigation?]
+  [:div {:class (str (if menu-navigation? "navbar-brand logo-menu" "logo"))}
+   [:div.img]])
+
```
Macroz, line 66: We could make this much simpler by having two different components instead of this parameterized component. But what I do not like is that this `menu-navigation?` parameter does a very unsymmetrical choice between adding two or one completely unrelated class. Why are the classes not `logo-navbar` and `logo`? Why do we have these `navbar-brand` and `logo-menu` here? Is the terminology clear? Why is there also `:logo-name-navigation`? Can we not have one term for the component in the navbar and use the same for the config etc?

src/cljs/rems/navbar.cljs
```diff
@@ -80,13 +84,16 @@
...
-   [:div.navbar.navbar-expand-sm.flex-fill
-    [:button.navbar-toggler
-     {:type :button :data-toggle "collapse" :data-target "#small-navbar"}
-     "\u2630"]
-    [navbar-items :div#big-navbar.collapse.navbar-collapse.mr-3 identity]]
-   [:div.navbar [user-widget (:user identity)]]])
+  (let [theme @(rf/subscribe [:theme])]
+    [:nav.navbar-flex
+     [:div.navbar.navbar-expand-sm.flex-fill
+      (when-not (str/blank? (:logo-name-navigation theme))
```
Macroz, line 90: We can make this simpler with
```clj
(when (:logo-name-navigation theme)
  ...)
```

But also notice that this would be wrong because one can specify `:logo-name-fi-navigation` and not use `:logo-name-navigation`? See what the `styles.clj` does with the `get-theme-attribute`.

src/clj/rems/css/styles.clj
```diff
@@ -96,8 +102,8 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo
-                          {:height (u/px 150)}]))
+                         [:.logo {:height logo-height}]
+                         [:.navbar-brand {:display "none"}]))
```
Macroz, line 106: Do we really need to set this to `display: none` because aren't we using the `navbar-small` component for the smallest breakpoint?

src/cljs/rems/spa.cljs
```diff
@@ -332,11 +330,13 @@
...
   (let [page-id @(rf/subscribe [:page])
-        grab-focus? @(rf/subscribe [::grab-focus?])]
+        grab-focus? @(rf/subscribe [::grab-focus?])
+        theme @(rf/subscribe [:theme])]
     (lazy-load-data!)
     [:div
      [nav/navigation-widget]
-     [logo]
+     (when (or (= page-id :home) (str/blank? (:logo-name-navigation theme)))
+       [nav/logo])
```
Macroz, line 339: This test does not work. The hero version of the logo should appear if it is configured on all pages and not just home. It should also not depend on the navbar logo at all. Also this test is wrong (see the previous comment).

Also I'm not sure why the logo component makes sense to be in the navbar namespace where it has been moved in this PR. I think it should probably be in the atoms.

Macroz (changes requested):

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
+- The navigation menu brand logo is now configurable. You can see ` :logo-name-navigation ` configuration variable that if provided with alogo image adds logo to the top navigation menu. [configuration.md](docs/configuration.md). (#2363)
+
```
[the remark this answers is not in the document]
marharyta: "This logo also can be customized per language like the regular logo."? @Macroz

src/clj/rems/css/styles.clj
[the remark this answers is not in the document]
marharyta: as per bootstrap docs: [https://getbootstrap.com/docs/4.0/components/navbar/](https://getbootstrap.com/docs/4.0/components/navbar/)

src/clj/rems/css/styles.clj
marharyta: I was planning to submit the refactoring PR later, would that work for you?

src/cljs/rems/navbar.cljs
marharyta: right, I did not take that one into account. this gets pretty complicated then.

CHANGELOG.md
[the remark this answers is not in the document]
marharyta: ok, reading another comment, I see your point

src/cljs/rems/administration/administration.cljs
marharyta: can we just agree that it looks ok and works with current UI and move on?

src/cljs/rems/navbar.cljs
marharyta: Bootstrap docs again

src/clj/rems/css/styles.clj
Macroz: For this specific refactoring yes.

src/cljs/rems/administration/administration.cljs
Macroz: I think I was saying that the old one also looked ok and worked and therefore the diff was unnecessary. We can leave this like it is now, the change is minor.

src/cljs/rems/navbar.cljs
Macroz: We can solve it with a simple utility function that both the `styles.clj` and frontend files use.

src/cljs/rems/navbar.cljs
marharyta: https://getbootstrap.com/docs/4.0/components/navbar/

Macroz: I can accept using `navbar-brand` here but I think we shouldn't need the other reference to it (from the `styles.clj`).

I also suggest using `:logo-name-navbar` so that we don't need to remember `navbar` vs. `navigation` for this. We should be able to do away with `logo-menu` entirely. Then we would always refer to `navbar` and `logo` for these things (except the one reference to Bootstrap styling).

Macroz: Also
- `logo-name` is the name of the logo
- `logo-name-sm` is the name of the logo in small breakpoint

I suggest `navbar-logo-name` because it is not a variant of the logo but a separate thing for the navbar.

marharyta: A lot of Bootstrap native classes are currently being referenced in our ```styles.clj```  file.

src/clj/rems/css/styles.clj
marharyta: probably, if I removed it? Would you like me to go and double check? Sure, I will.

marharyta: I did check, yes, all good

src/cljs/rems/navbar.cljs
marharyta: As of now, I did delete the reference to ```navbar-brand``` from ```styles.clj```

marharyta:
@Macroz I believe I have addressed all the comments at this point

src/cljs/rems/navbar.cljs
```diff
@@ -60,6 +60,14 @@
               text (get-in page [:translations language :title] (text :t/missing))]
           [nav-link url text])))))
 
+(defn logo []
+  [:div {:class "logo"}
+   [:div.img]])
```
Macroz, line 65: This is not the best place for the logo component (navbar.cljs), atoms.cljs would perhaps be better.

src/cljs/rems/spa.cljs
```diff
@@ -93,6 +94,11 @@
  (fn [db _]
    (:theme db)))
 
+(rf/reg-sub
+ :user-language
+ (fn [db _]
+   (:language (:user-settings db))))
+
```
Macroz, line 101: We should rather use the existing sub `:language` (from `user_settings.cljs`). This will not work.

src/cljs/rems/spa.cljs
```diff
@@ -332,11 +335,16 @@
...
+        theme @(rf/subscribe [:theme])
+        lang @(rf/subscribe [:user-language])]
     (lazy-load-data!)
     [:div
      [nav/navigation-widget]
-     [logo]
+     (when (or (= page-id :home)
+               (and (not ((keyword (str "navbar-logo-name-" (name lang))) theme))
+                    (not (:navbar-logo-name theme))))
+       [nav/logo])
```
Macroz, line 347: This logic is still wrong. The regular logo should be there always when configured so. It should not depend on the page id or navbar logo.

Macroz (changes requested):
The two problems still need to be fixed before this can be merged.

src/cljs/rems/navbar.cljs
marharyta: agreee

src/cljs/rems/spa.cljs
marharyta: What do you mean will not work? it works in the UI

src/cljs/rems/spa.cljs
marharyta: what do you refer to as the regular logo? the hero logo element?

marharyta: Screenshot 2021-01-26 at 16 36 49

src/cljs/rems/spa.cljs
Macroz: If you look at the definition of `:language` subscription it is different to this definition and it is for a reason (it needs to e.g. check cookies and such).

src/cljs/rems/spa.cljs
Macroz: Well my thinking is that ideally you can choose only navbar logo, only hero logo, both or none. That is orthogonal, flexible and the features are not coupled. I can accept the navbar logo overriding the hero logo too, but in that case I'd like to see the helper fn for this navbar logo feature detection (that is duplicated above in the navbar).

opqdonut: I thought we agreed on the spec already many revisions of this PR ago:

- Home page always shows the hero image
- Other pages don't show the hero image if the navbar image is defined

If the options were just navbar, hero, both or none, you couldn't have the hero just on the front pages, which _I think_ we imagine most users will want.

So let's finish this PR with that spec, and discuss changes to the spec afterwards, right?

Macroz: Unfortunately I think we didn't have an agreement after all but we can live with this interpretation as long as the bug is fixed.

Macroz: Bug as in the language sub.

marharyta: Isn`t this what I did @opqdonut ?
I mean I had the same thought, that we reaches the understanding like you described

marharyta: https://github.com/CSCfi/rems/pull/2483/files#r564566475 
I just went through the entitre messaging history and I swear I cannot find a single place where you would suggest this specific pattern

opqdonut: @marharyta  yeah, I was replying to @Macroz

src/cljs/rems/spa.cljs
marharyta: I can only find subscribtion for  ``` :language ```

Macroz: https://github.com/CSCfi/rems/blob/master/src/cljs/rems/user_settings.cljs#L28

src/cljs/rems/spa.cljs
Macroz: We obviously talked past each other in this. The idea for this feature was pretty clear to me in the beginning already but all the work around the various issues is enough to confuse and miss the details. This specific solution fulfills the main point (that it does not break backwards compatibility) so like I said we can go on with this as long as the bug is fixed.

hero_or_navbar

Macroz: Pretty much yes, it is implied. If two features are completely separate they are said to be orthogonal. They can be both turned on or off independently i.e. they can be only one, only other, both or none. An ideal feature is orthogonal from others, otherwise it is said to be coupled (and cannot be understood separately). For a white-label product such as REMS, I think it is desirable to have a set of features that work independently and flexibly, because the needs of different installations vary and we cannot predict everything. This is a tiny case in that sense but I think the principle is valuable overall in software development.

marharyta: ok, so how would that functionality align with the implementation?
most of out users have links to hero image in the theme, so if the rules work as true for the hero, it will always be false for the nav logo. Are you implying there should also be an extra setting for turn nav logo on? Turn hero off? a combo?

Macroz: The idea was about both of the features having their own configs to turn them on separately. For the specific case of navbar logo but only front page hero logo there needs to be a new config or solution though but we don't need to go there now. Like I said this logic is fine now because it is backwards compatible and the improvement would be tiny.

src/cljs/rems/spa.cljs
marharyta: I guess we are done to one problem at this point. May I just ask about the language thing a bit more? @opqdonut @Macroz  isnt  ```(:user-settings db)``` supposed to be synchronized with the settings?

Macroz: Yes but that is not the same as what the chosen language is.

Macroz: You can check what the existing subscription does and see what it needs to check.

marharyta: now that this part has been tackled, I am kindly asking @Macroz to take a look again

Macroz: Sure I'll have a look

src/clj/rems/css/styles.clj
```diff
@@ -96,8 +102,8 @@
                            :background-size :contain
                            :background-repeat :no-repeat
                            :background-position [[:center :center]]}]
-                         [:.logo
-                          {:height (u/px 150)}]))
+                         [:.logo {:height logo-height}]
+                         [:.logo-menu {:display "none"}]))
```
Macroz, line 106: I think we need to adjust these breakpoints more. It doesn't quite work with this. I would rebase on top of the master (with the removed Profile-link) and then fix the breakpoint but we can also merge this change and do another PR with these fixes. Definitely should adjust the breakpoint with the latest change to the navbar.

breakpoint_failure

Also the menu here should be on the left so we should hide the logo earlier.
slightly_bigger_logged_in

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ Changes since v2.15
 
 - The OIDC configuration is now validated, and REMS refuses to start without a valid OIDC configuration. See the `:oidc-metadata-url` configuration variable and [configuration.md](docs/configuration.md). (#2519)
 
+- The navbar can be configured to show a logo image. When the `:navbar-logo-name` config is provided, the logo is shown in the navbar (top navigation menu). This logo also can be customized per language like the regular logo. (#2363)
```
Macroz, line 25: I think something that was lost in the chat was the idea that we should showcase this feature in the `example-theme/theme.edn`. So draw a new REMS small logo for the navbar (with localizations).

Macroz (approved):
Hey, looks pretty good. I noticed some things while clicking the UI, but I think they don't block the merging of this PR. Need to create them as a follow-up task though.

src/clj/rems/css/styles.clj
marharyta: I personally do not rely on Chrome simulator much, been known to display layout with 'quirks'
I would go and test it properly with Xcode and Android studio for the different PR.
@joel and @Macroz do we have any usage stats? Like desktop vs. mobile usage I can fall back on to make a decision how to test?

CHANGELOG.md
marharyta: Draw? you mean create a logo option for REMS default theme?

src/clj/rems/css/styles.clj
opqdonut: No stats :/

CHANGELOG.md
opqdonut: I think we can also get by with just using `resources/public/img/rems_logo_*.png` as the navbar logo for now.

marharyta: right, so you mean add the logo to the default theme? Is that right?

src/clj/rems/css/styles.clj
marharyta: ok, so no analytics I take it?

opqdonut: Yeah, REMS doesn't include any analytics, and if any users have included their own, we haven't seen the data.

CHANGELOG.md
opqdonut: Yep, a navbar logo for the example theme.

marharyta: Just to be clear and make sure we are in agreement? I reeeeaaallly want to be done with the ticket.

marharyta: done

src/clj/rems/css/styles.clj
Macroz: Simulator or not a person can easily have a window of inopportune width, you can try the same with resizing your browser window. The responsiveness of the page should not be broken.

CHANGELOG.md
Macroz: Why do you only add it for Swedish though? There should be one for each language? 

Also I would just resize the logo image for the navbar (or make new ones, these are just for demo purposes anyway) and use separate files so that it is clear what is meant to happen and why.

opqdonut: I was surprised by having the logo only for one languge too, but having it only for some languages makes it easy to preview both looks in dev, so it might work!

marharyta: yes, @opqdonut got my idea!

Macroz: I think it's ok for a demo for ourselves but not for code/config that we leave for others to read.

marharyta: What would you suggest could be the best approach then (also taking into account we need to demo both options)?
Have explicitly stated it in the theme file? create an alternative theme file?

Macroz: We have plenty of servers with the old config / other configs that we can ourselves use. As far as the future goes we want to advertise this new better option so it should be the complete default with sane values.

marharyta: "sane values"?

Macroz: Sane values for example config means that someone will get a good idea what are valid example values. I.e. the likes of which can be used in production.

marharyta: ok, done

src/clj/rems/css/styles.clj
marharyta: I think I can create an issue from this discussion

PR #2494, opened 2020-12-14
ADR 015: Accessibility

opqdonut:


marharyta:
@opqdonut yeah, looks like it

docs/architecture/015-accessibility.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Screen readers
+
+The most important things for a nice screen reader experience are:
+
+- we offer a skip to main content link
+- we try to order the DOM in an order that makes sense
+  (e.g. the Actions menu comes after the other application information)
+
+### Mac OS specific settings
```
opqdonut, line 61: now it's under `## Screen readers`, not under `## Focus`

docs/architecture/015-accessibility.md
```diff
@@ -0,0 +1,65 @@
...
+
+It has bee discovered that in order to properly test/use a11y features, such as  `:focus` outline, on Mac OS Catalina, you should have Accessibility settings explicitly enabled in your OS settings. Please, refer to these resources for instructions:
```
Macroz, line 63: Typo `bee` also I would prefer spelling out accessibility here.

docs/architecture/015-accessibility.md
marharyta: hm, where should I place it then?

opqdonut: I'll try to clarify. Now the sequence of headers is

```
## Use cases
## Focus
## Screen readers
### Mac OS specific settings
```

I'd prefer it to be

```
## Use cases
## Focus
### Mac OS specific settings
## Screen readers
```

since "Mac OS specific settings" is related to Focus, not to Screen readers.

Macroz (approved):
I think this is good enough to merge. Can add details later.

docs/architecture/015-accessibility.md
marharyta: makes sense

opqdonut: I did the change already BTW

Issue #2484, opened 2020-12-10
HTTP caching issues


opqdonut:
1. app.js is served without any cache-control headers, which makes browsers cache it very aggressively. Solutions include:
  - set a Cache-Control max-age header
  - use content-addressing or some sort of cache-busting, for example having `app-<hash>.js` or `app-<version>.js` or fetching `app.js?<hash>`
2. The `/` resource has `Cache-control: private` for some reason. Check where does this come from and if it is really what we want.

PR #2499, opened 2020-12-17
HTTP caching improvements + refactoring

opqdonut:
for #2484

- Advise browsers to never cache the REMS home page (the / resource)
  with a Cache-Control: no-store
- Cache-bust app.js and screen.css by referring to them as
  app.js?<git-hash-or-timestamp> from the / resource
- If REMS has specified no Cache-Control header, fall back to
  a max-age 23h setting. This will let users see e.g. updates to the
  theme files latest the next day.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing


src/clj/rems/handler.clj
```diff
@@ -22,7 +22,9 @@
 (defn public-routes []
   (routes
    (home/home-routes)
-   (auth/auth-routes)))
+   ;; never cache authentication results
+   ;; TODO this is a slightly hacky place to do this
+   (middleware/wrap-no-cache (auth/auth-routes))))
```
opqdonut, line 27: I wonder if we should actually default to `Cache-Control: no-store` in our middleware and explicitly set `Cache-Control: max-age=23h` for static resources

Macroz: Possibly that would be the safest.

src/clj/rems/middleware.clj
```diff
@@ -236,7 +236,8 @@
       (assoc-in [:session :cookie-attrs] {:http-only true, :same-site :lax})))
 
 (defn wrap-base [handler]
-  (-> ((:middleware +defaults+) handler)
+  (-> handler
+      ((if (:dev env) wrap-dev identity))
```
Macroz, line 231: So much better :+1:

src/clj/rems/handler.clj
```diff
@@ -67,6 +67,13 @@
       (apply-for-resources resource)
       (apply-for-resources [resource])))
 
+  (GET "/landing_page" [] ; DEPRECATED: legacy url redirect
```
Macroz, line 53: Perhaps we could put that to the release notes as deprecated too and then remove it, since we dropped the legacy auth support too?

src/clj/rems/handler.clj
```diff
@@ -136,8 +115,17 @@
...
 (defn app-routes []
   (routes
+   (wrap-login-redirect (routes (GET "/" []
+                                     (layout/home-page))
+                                attachment-routes
+                                redirects))
+   css-routes
+   ;; never cache authentication results
+   ;; TODO this is a slightly hacky place to do this
+   (middleware/wrap-no-cache (auth/auth-routes))
```
Macroz, line 115: Perhaps we should use without the ns-prefix here like the other wraps?

src/clj/rems/handler.clj
```diff
@@ -136,8 +115,17 @@
 
 (defn app-routes []
```
Macroz, line 103: The end result is better than before but this is perhaps slightly confusing route definition still. I don't have clear ideas on how to improve yet but the things are not exactly at the same level of abstraction (inlined stuff, wrappers, vars and not vars)

Macroz (approved):
I'm happy to merge this and start finding out about the corner cases still hidden. 

Perhaps it would be a good idea still to change the default to no cache and explicitly cache only static resources. Or what do you think?

src/clj/rems/handler.clj
opqdonut: yeah I guess we could just drop it, the any links to /landing_page should be long gone now

src/clj/rems/handler.clj
opqdonut: fine by me

src/clj/rems/handler.clj
opqdonut: Yeah but now the complexity is visible. I plan to simplify it further in the future (get rid of entitlements/routes for example)

opqdonut:
How about merging this and following up with a PR that changes the default?

src/clj/rems/handler.clj
opqdonut: (will change in the next PR)

PR #2500, opened 2020-12-18
more HTTP caching improvements

opqdonut:
for #2484

- change default behaviour to uncacheable, explicitly tag cacheable
  handlers
- workaround for app.js being cached when using figwheel
- refactorings

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
  - entry already there

## Testing


src/clj/rems/handler.clj
```diff
@@ -100,6 +101,11 @@
     (route/files "/" {:root path})
     never-match-route))
 
+(def ^:private webjar-handler
+  ;; Serves our webjar (https://www.webjars.org/) dependencies as /assets/<webjar>/<file>
+  ;; Weirdly ring-webjars only exposes a middleware and not a route.
```
Macroz, line 107: Could have been a docstring too?

src/clj/rems/handler.clj
```diff
@@ -113,25 +113,24 @@
...
-   (extra-script-routes (:extra-scripts env))
-   (static-resources (:extra-static-resources env))
-   (static-resources (:theme-static-resources env))
-   webjar-handler
+   (wrap-cacheable
+    (routes
+     resource-handler
+     (extra-script-routes (:extra-scripts env))
+     (static-resources (:extra-static-resources env))
+     (static-resources (:theme-static-resources env))
```
Macroz, line 140: We could perhaps say something about caching in the docs and say how to best handle these resources updated by admins (i.e. that perhaps adding the new file under a new name is best)

src/clj/rems/middleware.clj
```diff
@@ -202,19 +202,20 @@
      (users/format-user identity))))
 
 (defn wrap-cache-control
-  "In case a Cache-Control header is missing, add a default of 23h"
+  "In case a Cache-Control header is missing, add a default of no-store"
```
Macroz, line 205: The name of this could perhaps indicate what it does i.e. `wrap-cache-no-store-if-no-cache-specified`, `wrap-cache-no-store-when-not-specified` or `wrap-cache-default-no-store`.

src/clj/rems/handler.clj
```diff
@@ -110,6 +110,15 @@
 (def ^:private resource-handler
   (route/resources "/" {:root "public"}))
 
+(defn- dev-js-handler
```
Macroz, line 113: Do we need the same for CSS or was it that Figwheel does the busting there?

Macroz (approved):
Looks good to me. Some ideas for future PRs you can read.

src/clj/rems/handler.clj
opqdonut: done in followup PR

src/clj/rems/handler.clj
opqdonut: done

src/clj/rems/handler.clj
opqdonut: Hmm, it looks like CSS live-reloading isn't working for me currently. If I reload my browser it does get the fresh screen.css, but editing rems.css.styles and restarting the mount component doesn't do anything.

In any case, we need the `dev-js-handler` to avoid hitting the `(wrap-cacheable ... resource-handler)` which would serve app.js otherwise. CSS is always served by `styles/css-routes` which only serves uncacheable results.

... That should probably be fixed. I'll look into it.

opqdonut: Fixed in follow-up PR

PR #2502, opened 2020-12-20
Scrollable actions

### LINKED ISSUE #2501: Better actions area for small screens
The actions area for some actions does not work well with the sticky position – it becomes too tall to fit the screen. The user must then scroll to the end of the entire application page before it starts scrolling downwards. We can try to improve this with a scrollable area.

Macroz:
Fixes #2501 

This is a workaround because the browsers will not handle overflow in
a sticky element properly. We don't want to add yet another JS
component for just this right now. Perhaps once we upgrade to
Bootstrap 5 we can work out better layout or solution.

https://user-images.githubusercontent.com/823661/102723315-a236c180-430f-11eb-89a3-91e226391880.mp4

Pretty difficult to test so maybe not worthwhile now.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -147,7 +147,7 @@
 (defn- generate-actions-float-menu
   "The #actions floating menu can be too long for some screens. There is no clean solution for this in pure CSS
   and to avoid yet another random JS-library we make the element scrollable with a dynamic max-height. This can
-  break if the 150px space is not enough anymore but works for now.
+  break if the 105px space is not enough anymore but works for now.
 
   We also do not want these styles to affect the mobile layout (i.e. more narrow than 992) where the actions
   is at the bottom of everything and has any height available."
@@ -163,7 +163,7 @@
    (stylesheet/at-media {:min-width (u/px 992)}
                         [:#actions {:overflow-x :hidden
                                     :overflow-y :auto
-                                    :max-height "calc(100vh - 150px)"}])))
+                                    :max-height "calc(100vh - 105px)"}])))
 
 (defn- button-navbar-font-weight []
   ;; Default font-weight to 700 so the text is considered
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -144,6 +144,27 @@
                    :border-color (get-theme-attribute :phase-bgcolor-completed "#ccc")
                    :color (get-theme-attribute :phase-color-completed :phase-color)}]]])
 
+(defn- generate-actions-float-sizes
+  "The #actions floating menu can be too long for some screens. There is no clean solution for this in pure CSS
+  and to avoid yet another random JS-library we make the element scrollable on certain sizes.
+
+  We also do not want these styles to affect the mobile layout (i.e. more narrow than 992) where the actions
+  is at the bottom of everything and has any height available."
+  []
+  (list
+   (stylesheet/at-media {:min-width (u/px 992)
+                         :max-height (u/px 1080)}
+                        [:#actions ; make buttons here smaller on small screens
+                         [:.btn {:font-size "0.875rem" ; duplicates bootstrap style btn-sm
+                                 :padding "0.25rem 0.5rem"
+                                 :line-height 1.5
+                                 :border-radius "0.2rem"}]])
+   (for [height (range 1080 500 -100)]
+     (stylesheet/at-media {:min-width (u/px 992)
+                           :max-height (u/px height)}
+                          [:#actions {:overflow-y :auto
+                                      :max-height (u/px (- height 200))}]))))
+
 (defn- button-navbar-font-weight []
   ;; Default font-weight to 700 so the text is considered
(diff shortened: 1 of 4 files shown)


src/clj/rems/css/styles.clj
```diff
@@ -144,6 +144,27 @@
...
+                        [:#actions ; make buttons here smaller on small screens
+                         [:.btn {:font-size "0.875rem" ; duplicates bootstrap style btn-sm
+                                 :padding "0.25rem 0.5rem"
+                                 :line-height 1.5
+                                 :border-radius "0.2rem"}]])
+   (for [height (range 1080 500 -100)]
+     (stylesheet/at-media {:min-width (u/px 992)
+                           :max-height (u/px height)}
+                          [:#actions {:overflow-y :auto
+                                      :max-height (u/px (- height 200))}]))))
```
opqdonut, line 166: did you try something like `100vh - 200px` before settling on this max-height trick?

Macroz: No I actually didn't. That one may work. Setting `bottom` etc. just don't.

Macroz: You should probably try this one with some manual testing to see if you can break it.

src/clj/rems/css/styles.clj
```diff
@@ -144,6 +144,26 @@
                    :border-color (get-theme-attribute :phase-bgcolor-completed "#ccc")
                    :color (get-theme-attribute :phase-color-completed :phase-color)}]]])
 
+(defn- generate-actions-float-menu
+  "The #actions floating menu can be too long for some screens. There is no clean solution for this in pure CSS
+  and to avoid yet another random JS-library we make the element scrollable on certain sizes.
+
+  We also do not want these styles to affect the mobile layout (i.e. more narrow than 992) where the actions
+  is at the bottom of everything and has any height available."
```
opqdonut, line 152: this docstring probably could be reworded a bit now that the hack is gone

src/clj/rems/css/styles.clj
```diff
@@ -144,6 +144,26 @@
...
+                         :max-height (u/px 1080)}
+                        [:#actions ; make buttons here smaller on small screens
+                         [:.btn {:font-size "0.875rem" ; duplicates bootstrap style btn-sm
+                                 :padding "0.25rem 0.5rem"
+                                 :line-height 1.5
+                                 :border-radius "0.2rem"}]])
+   (stylesheet/at-media {:min-width (u/px 992)
+                         :max-height (u/px 1080)}
+                        [:#actions {:overflow-y :auto
+                                    :max-height "calc(100vh - 150px)"}])))
```
opqdonut, line 165: This works fine on chrome, but on firefox the `position: sticky` is not active in any screen size: the actions pane doesn't follow the viewport. The overflow works fine though.

opqdonut: nevermind, now that I'm rechecking it works fine in FF too

opqdonut: Could we drop the max-height from this rule now?

Macroz: You mean the breakpoint? Yes I suppose so, it's then less easier to break with suddenly high content.

opqdonut:
I can sometimes get an unnecessary y overflow to happen in chrome, but not reliably. This only happens when I switch to reactive mode (& back). It's probably not an issue we need to spend time fixing now.
actions-y-overflow

opqdonut (approved):
consider rewriting the docstring of generate-actions-float-menu now that the hack is gone

Macroz:
> consider rewriting the docstring of generate-actions-float-menu now that the hack is gone

Well I think all the text in it still applies. There is no mention of the breakpoints and `calc(100vh - 150px)` is still a hack that can break when we change the layout. A good solution would not require setting any sizes here.

src/clj/rems/css/styles.clj
opqdonut: to clarify, I think this no longer applies:

> There is no clean solution for this in pure CSS and to avoid yet another random JS-library we make the element scrollable on certain sizes

Macroz:
> I can sometimes get an unnecessary y overflow to happen in chrome, but not reliably. This only happens when I switch to reactive mode (& back). It's probably not an issue we need to spend time fixing now.
> actions-y-overflow

I added one mitigation for this, i.e. forcing that no horizontal scrollbar can appear. The vertical scrollbar does not happen on my chrome or firefox but also is not problematic as it's the normal behavior for this. Perhaps the vertical scrollbar only happened due to the horizontal one taking some space.

src/clj/rems/css/styles.clj
Macroz: :+1: I reworded it a bit to clarify

opqdonut (approved):

marharyta:
Looks good to me!

marharyta (approved):

Issue #2506, opened 2020-12-30
Incorrect config documentation for oidc?


talavis:
In the documentation at [1] and the default values at [2] you refer to `:oidc-configuration-url` replacing `:oidc-domain`. As far as I can tell, this is supposed to the option `:oidc-metadata-url` instead.

[1] https://github.com/CSCfi/rems/blob/master/docs/configuration.md
[2] https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn

PR #2507, opened 2020-12-30
Fix the reference to :oidc-metadata-url

talavis:
Fix the inaccurate configuration option references I mentioned in #2506.
Also fix for the bullet point for the `oidc-metadata-url` option.


Macroz (approved):
Thanks. Good eyes!

PR #2514, opened 2021-01-08
Add docs for fixing OIDC hostname issues when using docker namespaces

katpavlova:
Adding some documentation pertaining to using REMS with Keycloak and Docker. This issue is no fault of REMS, but it took a lot of very uninteresting work to resolve it, so I wanted to share it in case anyone ran into this problem in the future.
Feel free to move these notes elsewhere if the REMS `docs` are not the appropriate place for them.


docs/configuration.md
```diff
@@ -46,6 +46,47 @@ are only shown to users handling the application, and super-users like
 the owner and reporter. In praticular, application members cannot see
 the additional user properties of each other.
 
+#### OIDC Hostname issues
```
Macroz, line 49: Since this is a rather long and very specific tip compared to the other general instructions, perhaps this could be moved to its own file or perhaps even posting this resolution as an issue (that you can close) and then creating the link here to that issue. 

I personally search first from the project issues when I have bump into problems. But I think it would be valuable to provide some hints on how to use various IdPs with REMS. A link from here would sound good. We don't have any such section yet but perhaps should.

What do you think @opqdonut ?

opqdonut: hmm I'm not sure... my first reaction would be to put stuff like this in, say, `docs/contrib/keycloak.md` or something like that

Macroz: Yeah especially if we put some further instructions. And link from here to those.

Macroz: @katpavlova How do you feel about extracting this to `keycloack.md` and writing the rest of the how to use it? It can be short and up to the point. Link to the documentation of Keycloack for the necessary parts.

katpavlova: Sure, that works for me. :)
What else would you like in the keycloak guide? Sort of a Quick Start for setting up a Keycloak IdP for REMS, followed by this documentation for hostname issues? Anything else?

Macroz: A quick start with Keycloak as how you set it up would be just great. Other developers that are experimenting would find it useful. I think we can provide one with e.g. Auth0 ourselves.

katpavlova: Done, how is this?

katpavlova: @Macroz I did as suggested, how is this?

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+    ```
+    version: "3.7"
+    services:
+    rems:
+        mem_reservation: 200m
+        mem_limit: 500m
+        ports:
+        - "127.0.0.1:3001:3000"
+        volumes:
+        - ./services/rems/simple-config.edn:/rems/config/config.edn
```
Macroz, line 22: I tried with these two additional volumes making up to these three with my directory structure:
```
        - ../rems/dev-config.edn:/rems/config/config.edn
        - ../rems/example-theme:/rems/config/example-theme
        - ../rems/test-data:/rems/test-data
```

The default config wants `ga4gh` keys and they live in `test-data` folder. Could perhaps do without them. Maybe including a complete config EDN file here makes sense? You have added the instructions later but I think it would be nice to see the end results.

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+        image: postgres:9.6
+        environment:
+        POSTGRES_USER: rems
+        POSTGRES_PASSWORD: remspassword
+        mem_reservation: 30m
+        mem_limit: 150m
+        ports:
+        - "127.0.0.1:5432:5432"
+    keycloak:
+        environment:
```
Macroz, line 36: Here I used the image from the repo so `image: jboss/keycloak:latest`. Perhaps we need to add it here too?

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+        KEYCLOAK_USER: admin
+        KEYCLOAK_PASSWORD: admin
+        ports:
+        - "3002:8080"
```
Macroz, line 40: I think the standard image prints it is accessible in port `9990` but instead I got this to work with the `8080:8080`

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+    ```
+
+2. **Preparing a Keycloak Realm**
+    1. Build and run keycloak with `docker-compose up keycloak`.
```
Macroz, line 44: I had some difficulty with these separate ups but I think there is a bit of a chicken & egg going on with Keycloak needing to be configured before REMS container can start because it goes looking for the auth configs.

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+    2. Login to keycloak at `http://keycloak:8080/auth/admin` with the admin username and password, set to `admin` and `admin` in this (very insecure) example. These can be set as environment variables, as demonstrated in the `docker-compose.yaml` example above.
+    3. Add an identity provision `realm` to Keycloak by mousing over `Master` near the top-left of the page, and clicking `Add realm`. A realm can be thought of as a single vault containing a set of users, credentials, groups, clients, security settings, etc. When authenticating with a realm, the client is only able to authenticate within that realm. Realms can be partially exported and imported by clicking `Import`/`Export` near the bottom-left of the page, although exporting users is more complicated. For this example, the realm will be named `rems-idp`.
+    4. Add a client for REMS, by navigating to `Clients` on the left side of the page and clicking `Create`. Set the `Client Protocol` to `openid-connect`. Clients are allowed to authenticate users within the Keycloak realm that they are bound to. For this example, the Client ID will be `rems-client`.
```
Macroz, line 47: Could we perhaps modify these instructions to show the correct value as not the last word but maybe in the very first sentence? Scanning through the instructions would be a bit faster I think.

docs/contrib/keycloak.md
Macroz: Maybe Keycloak can be configured completely with some initial files? Like having that export?

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+
+    :oidc-client-id "rems-client"
+    :oidc-client-secret "0ed4b9e0-5ff0-446a-ab7b-82dac465b9d5"
+    }
+    ```
+    2. Build rems, which (at the time of writing) requires:
+        1. [Installing Leiningen](https://leiningen.org/)
+        2. Building the Leiningen jar: `lein uberjar`
+        3. Building the Docker container: `docker-compose build rems`
+        4. Migrating the test data: `docker-compose run --rm -e CMD="migrate;test-data" rems`
```
Macroz, line 85: I think the alternative we could use here is to point to the [Docker Hub REMS image](https://hub.docker.com/r/cscfi/rems).

docs/contrib/keycloak.md
```diff
@@ -0,0 +1,148 @@
...
+    1. To test authentication with Keycloak, add a test User in Keycloak by navigating to `Users` and clicking `Add user`. Give them a `username` and clicking `Save`.
+    2. Navigate to the new User's `Credentials` tab and set a `password` for them. Set the `Temporary` toggle to `OFF`.
+    3. In REMS, at http://localhost:3001/ in this example, click the `Login` button to be redirected to Keycloak. 
+    4. Access the account using `varchar`/`varchar`. You should be authenticated and redirected back to REMS.
+
+6. **Exporting the Realm**
+    1. You can export your realm and save it source control, to reduce development and testing time. Simply navigate to `Export` on the bottom-left of the page, set `Export clients` to `ON`, and click `Export`.
+    2. Next time you want to add this realm to keycloak, navigate to `Add Realm` and select your `realm-export.json` file for import. The `Name` will auto-populate.
+
+## OIDC Hostname issues
```
Macroz, line 99: I think I solved these by adding Keycloak to `/etc/hosts` so that I can use `keycloak` domain name also outside Docker networking. I think this is a pretty standard solution. Also you can configure REMS `:public-url` to another  host and port.

Macroz (approved):
Thanks a lot! Here are some suggestions for you that may be improvements.

I would also add the link to here into some of our existing docs where OIDC is mentioned.

Let me know what you want to do with these and we can merge when you are ready.

docs/contrib/keycloak.md
katpavlova: Sure thing, you mean to get rid of the keycloak `Dockerfile` and just pull the image with docker-compose, right? I'm fine with either option.

docs/contrib/keycloak.md
katpavlova: Keycloak definitely needs to be up before REMS, because REMS tries to connect with it on start-up. There is not *too* much configuration that needs to be done with Keycloak once you do all of these steps once, and then export the realm with its clients. Subsequent configuration just looks like:
1. Spin up keycloak with `docker-compose up keycloak`
2. Log into the admin console
3. Import realm
4. (optional) Add a user.

You can even import/export users somehow, and probably make initialization even faster, but I haven't succeeded in doing this yet myself. I will definitely come back and add docs on how to do that once I figure it out.

From [these keycloak standalone docs](https://github.com/keycloak/keycloak-documentation/blob/master/server_admin/topics/export-import.adoc) it looks like it may be possible to do the realm import from a CLI. But since this constitutes a migration I don't think it would be run as a stage in `docker run`/`docker-compose up`.

katpavlova: If you would like a sample `realm-export.json` file for serving to REMS, I would be happy to provide one.

docs/contrib/keycloak.md
katpavlova: There's no correct or incorrect value here, you can set the client name to anything. I'm just stating what I set it to in this example so that it's easier to see where certain strings come from in the config files for REMS, docker-compose, etc.

docs/contrib/keycloak.md
katpavlova: Adding keycloak to `/etc/hosts` is common practice, but it somewhat defeats the purpose of containerizing Keycloak. It means that the developer needs to add documentation about adding the keycloak domain to `/etc/hosts` to their build instructions, and can be problematic when the user installing the app has strict Research IT.

This solution also breaks once you have multiple keycloaks, served at various ports, sometimes with varying hostnames.

I haven't tried solving this using REMS `:public-url`, thanks for the tip! I will try it out and report back.

katpavlova:
@Macroz Thank you for the thorough review! I replied to some of your comments with further questions, and once you get a chance to respond to them I will incorporate all of your suggestions.

docs/contrib/keycloak.md
Macroz: If it's possible to do this without building your own Keycloak then that is simpler, right? At least I didn't need to but I'm not so familiar with the whole setup. Do you need to do some things there (e.g. the ones I did with the `hosts`-file?)

docs/contrib/keycloak.md
Macroz: I managed with your instructions but a sample realm with the same settings as an option sounds nice. We don't need to make this perfect now, there are probably more pressing issues to work on. If I was using a lot of Keycloak myself, I would probably try to build an image that contains all the configuration too.

docs/contrib/keycloak.md
Macroz: You are right. What I meant is that perhaps there could be a "table" of settings to use, even if they are just example values. And they would then match the different steps of the instructions, and overall you get a working setup without much thinking. Then you can tune them later to what you wish. And even so I would put the example values as first just to be clear as what the target is. Maybe phrasing like this:

Next we will set up Keycloak with the following settings

| Keycloak | Setting | Value |
|-----------|----------|---------|
| Realm | Name | rems-idp |
| Client  | Client Protocol | openid-connect |
| ... | ... | ...|

1. ...
2. ...
3. **Add an identity provision `realm` to Keycloak named `rems-idp`.**
You can do that by mousing over `Master` near the top-left of the page, and clicking `Add realm`. 
A realm can be thought of as a single vault containing a set of users, credentials, groups, clients, security settings, etc. When authenticating with a realm, the client is only able to authenticate within that realm. Realms can be partially exported and imported by clicking `Import`/`Export` near the bottom-left of the page, although exporting users is more complicated.

docs/contrib/keycloak.md
Macroz: `:public-url` only solves the part of REMS however, not Keycloak. And REMS uses the URL that Keycloak reports as it is.

Maybe we can add editing `/etc/hosts` as one option, for those that do have admin priviledges on their machine. And then provide the two alternatives for others.

katpavlova:
I'm sorry for the delay in completing this PR, I have been really busy the last couple weeks. I am preoccupied at the moment but will be finishing this PR within the next 2 weeks. Thanks for your patience!

opqdonut:
No problem, we're not in a rush :relaxed:

opqdonut:
Ping @katpavlova any news on this? You can also just push whatever you have right now and we can finish the doc & merge it!

katpavlova:
Hi @opqdonut ! Thank you so much for your patience. I have implemented the changes requested by @Macroz , including adding some sample config files to help with quick-start.
I would ask you to review one more time before merging, and amend or remove sections as you see fit. For example, I added an option for modifying `/etc/hosts` near the bottom of `keycloak.md`, but given the limitations of the method, you may or may not prefer to remove it from the docs.
Please let me know if there are any further contributions needed from me on this PR. I will be available this week to implement them.

opqdonut (approved):
This is very thorough, thank you!

PR #2516, opened 2021-01-08
Fail on accessibility violation

### LINKED ISSUE #2463: Make browser test fail if accessibility violations are found

Macroz:
Closes #2463 

We'll have to see how this works in CI but here is what the error looks like.
accessibility-error

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changes since v2.15
 ### Fixes
 
 ### Additions
+- The browser tests will now fail if there are any accessibility violations. (#2463)
 
 ## v2.15 "Tapiolantie" 2021-01-08
 
--- a/test/clj/rems/browser_test_util.clj
+++ b/test/clj/rems/browser_test_util.clj
@@ -451,13 +451,18 @@
   (try
     (f)
     (finally
-      (doseq [k (->> (context-get :axe)
-                     (mapcat keys)
-                     distinct)
-              :let [filename (str (str/lower-case (name k)) ".json")
-                    content (->> (context-get :axe)
-                                 (mapcat #(get % k))
-                                 distinct
-                                 (sort-by :impact))]]
-        (spit (io/file (:accessibility-report-dir @test-context) filename)
-              (json/generate-string-pretty content))))))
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changes since v2.15
 ### Fixes
 
 ### Additions
+- The browser tests will now fail if there are any accessibility violations. (#2463)
```
opqdonut, line 16: not sure if this needs to be in the changelog since it's not user-facing at all, but I don't mind

test/clj/rems/browser_test_util.clj
```diff
@@ -451,13 +451,18 @@
...
+                      content (->> (context-get :axe)
+                                   (mapcat #(get % k))
+                                   distinct
+                                   (sort-by :impact))]]
+          (spit (io/file (:accessibility-report-dir @test-context) filename)
+                (json/generate-string-pretty content))
+          (when (and (= :violations k) (seq content))
+            (reset! violations content)))
+        (when (seq @violations)
+          (throw (Exception. (str "\n\n\nThere are accessibility violations: " (count @violations) "\n\n\n"))))))))
```
opqdonut, line 468: did you test that circle stores the artifacts even if we hit this failure?

opqdonut (approved):

test/clj/rems/browser_test_util.clj
Macroz: The artifacts are the same as before, no change.

Macroz: The spit is also done before we throw. That's as much as we can do.

Macroz: Circle stores e.g. screenshots even with run fails so it should work out with this one?

opqdonut: good points, I believe you

Macroz: Like I mentioned in Flowdock, we can't test this easily so let's just wait for a real error.

PR #2520, opened 2021-01-11
css caching adjustments, caching docs

### LINKED ISSUE #2484: HTTP caching issues

opqdonut:
follow-up for #2500, closes #2484

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -102,7 +102,7 @@
   :figwheel {:http-server-root "public"
              :server-logfile "log/figwheel_server.log"
              :nrepl-port 7002
-             :css-dirs ["target/resources/public/css"]
+             :css-dirs ["target/resources/public/css/en" "target/resources/public/css/fi" "target/resources/public/css/sv"]
              :nrepl-middleware [cider.piggieback/wrap-cljs-repl]}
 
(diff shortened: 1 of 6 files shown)


project.clj
```diff
@@ -102,7 +102,7 @@
   :figwheel {:http-server-root "public"
              :server-logfile "log/figwheel_server.log"
              :nrepl-port 7002
-             :css-dirs ["target/resources/public/css"]
+             :css-dirs ["target/resources/public/css/en" "target/resources/public/css/fi" "target/resources/public/css/sv"]
```
Macroz, line 105: Interesting that it's not recursive. `hawk` seems to support it but I think it's not used recursively here then.

Macroz (approved):

PR #2537, opened 2021-01-26
Style refactoring logo styles

marharyta:
# Checklist for author

## Reviewability
- [ ] Link to issue - no particular issue, we agreed with @Macroz I will do some styles refactoring next
- [x] Note if PR is on top of other PR - [https://github.com/CSCfi/rems/pull/2483](https://github.com/CSCfi/rems/pull/2483)
- [ ] Consider adding screenshots for ease of review - should not have any visible effect, just refactoring

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -90,9 +90,44 @@
 (defn- generate-media-queries []
   (list
    (stylesheet/at-media {:max-width (u/px 480)}
+                        [(s/descendant :.rems-table.cart :tr)
+                         {:border-bottom "none"}])
+   (stylesheet/at-media {:max-width (u/px 870)}
+                        [:.user-widget [:.icon-description {:display "none"}]])
+   (stylesheet/at-media {:min-width (u/px 480)}
+                        [:.commands {:white-space "nowrap"}])
+   (stylesheet/at-media {:prefers-reduced-motion :reduce}
+                        [:body {:scroll-behavior :auto}])))
+
+(defn- generate-logo-styles []
+  (list
+   [:.logo-menu {:height logo-height-menu
+                 :background-color (get-theme-attribute :logo-bgcolor)
+                 :width "100px"}]
+   [:.logo {:height logo-height
+            :background-color (get-theme-attribute :logo-bgcolor)
+            :width "100%"
+            :margin "0 auto"
+            :padding "0 20px"
+            :margin-bottom (u/em 1)}]
+   [(s/descendant :.logo :.img)
+    (s/descendant :.logo-menu :.img)
+    {:height "100%"
+     :background-color (get-theme-attribute :logo-bgcolor)
+     :-webkit-background-size :contain
+     :-moz-o-background-size :contain
+     :-o-background-size :contain
+     :background-size :contain
+     :background-repeat :no-repeat
+     :background-position [[:center :center]]
+     :background-origin (get-theme-attribute :logo-content-origin)}]
+   [(s/descendant :.logo :.img) {:background-image (get-logo-image context/*lang*)}]
+   [(s/descendant :.logo-menu :.img) {:background-image (get-navbar-logo context/*lang*)}]))
+
+(defn- generate-logo-media-queries []
+  (list
+   (stylesheet/at-media {:max-width (u/px 480)}
                         (list
-                         [(s/descendant :.rems-table.cart :tr)
-                          {:border-bottom "none"}]
                          [(s/descendant :.logo :.img)
                           {:background-color (get-theme-attribute :logo-bgcolor)
                            :background-image (get-logo-name-sm context/*lang*)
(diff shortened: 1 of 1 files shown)


src/cljs/rems/spa.cljs
```diff
@@ -93,6 +94,11 @@
  (fn [db _]
    (:theme db)))
 
+(rf/reg-sub
+ :user-language
+ (fn [db _]
+   (:language (:user-settings db))))
+
```
Macroz, line 101: We should rather use the existing sub `:language` (from `user_settings.cljs`). This will not work.

src/cljs/rems/navbar.cljs
```diff
@@ -60,6 +60,14 @@
               text (get-in page [:translations language :title] (text :t/missing))]
           [nav-link url text])))))
 
+(defn logo []
+  [:div {:class "logo"}
+   [:div.img]])
```
Macroz, line 65: This is not the best place for the `logo` component (`navbar.cljs`), `atoms.cljs` would perhaps be better.

src/cljs/rems/spa.cljs
```diff
@@ -332,11 +335,16 @@
...
+        theme @(rf/subscribe [:theme])
+        lang @(rf/subscribe [:user-language])]
     (lazy-load-data!)
     [:div
      [nav/navigation-widget]
-     [logo]
+     (when (or (= page-id :home)
+               (and (not ((keyword (str "navbar-logo-name-" (name lang))) theme))
+                    (not (:navbar-logo-name theme))))
+       [nav/logo])
```
Macroz, line 347: This logic is still wrong. The regular logo should be there always when configured so. It should not depend on the page id or navbar logo.

Macroz (changes requested):
Looks pretty good but there are a two issues still that need resolving before this can be merged.

src/cljs/rems/navbar.cljs
marharyta: agree

opqdonut:
did you mean to review #2483 instead @Macroz? of course if this PR is fine too it's easier to merge this one instead of first #2483 and then this...

Macroz:
> did you mean to review #2483 instead @Macroz? of course if this PR is fine too it's easier to merge this one instead of first #2483 and then this...

Yeah I didn't understand that this is a new PR on top of that old. These issues need to be resolved in the original PR and then this can be rebased and reviewed again. I don't really see the point of doing this second PR with the open issues remaining.

marharyta:
So, @opqdonut @Macroz is it good to go?

src/clj/rems/css/styles.clj
```diff
@@ -243,15 +272,15 @@
    [:.text-highlight {:color (get-theme-attribute :color3)
                       :font-weight "bold"}]))
 
-(def ^:private dashed-form-group
+(defn- dashed-form-group []
```
opqdonut, line 275: were you thinking of consistency with the `generate-foo` functions when you turned these from constants to functions? If so, should we use the `generate-` prefix for everything, or perhaps for nothing?

src/clj/rems/css/styles.clj
```diff
@@ -88,11 +88,46 @@
...
+     :-moz-o-background-size :contain
+     :-o-background-size :contain
+     :background-size :contain
+     :background-repeat :no-repeat
+     :background-position [[:center :center]]
+     :background-origin (get-theme-attribute :logo-content-origin)}]
+   [(s/descendant :.logo :.img) {:background-image (get-logo-image context/*lang*)}]
+   [(s/descendant :.logo-menu :.img) {:background-image (get-navbar-logo context/*lang*)}]))
+
+(defn- generate-logo-media-queries []
```
opqdonut, line 127: Do we need separate `generate-logo-media-queries` and `generate-logo-styles`? Could they just be one chunk since they're all related to the logos? Or is there some reason why we need to have `generate-logo-styles` early in the main stylesheet and `generate-logo-media-queries` late?

opqdonut (commented):
some questions

src/clj/rems/css/styles.clj
opqdonut: I think it makes sense to use functions for these style fragments, because they might need to look up stuff from the config/theme. Mixing constants and functions can lead to surprises when developing, so maybe let's just make everything a function. At that point, I think the `generate-` prefixes can be dropped from the names, it's just noise.

marharyta: I just remember we did have a discussion with @Macroz the other day we decided to get rid of the ```private``` definitions

src/clj/rems/css/styles.clj
marharyta: It is considered to be a good practice to separate media rules into separate mixins, in our case: functions

src/clj/rems/css/styles.clj
marharyta: I think this is a good point naming convention-wise. I can fix that @opqdonut

marharyta: Well, if we want to to drop it all together then we have to either separate then into mixins file and definitely do it in a standalone PR

Macroz: I think I remember the discussion. I was actually not suggesting then to replace these `def` with `defn`. There is no need to mark them as private here really. But also I was thinking about a case where one does `defn` and `^:private` which does not need to be so because `defn-` exists.

src/clj/rems/css/styles.clj
Macroz: I think these are two separate things. The articles you linked suggest putting the media queries into the component i.e. logo here. They suggest writing those media queries with reusable helpers and not by hand. What you are doing here is not creating a reusable media query mixing, you are simply splitting the code into two places.

Macroz: I suggest that you put the logo related stuff including styles and media queries into one function `generate-logo-styles` and remove the `generate-logo-media-queries. We can later refactor the media queries themselves.

opqdonut (approved):

marharyta:
@Macroz Cannot merge it, can you check wether it is now according to the spec?

marharyta:
I cannot merge since you have blocked it with the request changes

Issue #2531, opened 2021-01-25
/api/user-settings API should return 401 when not logged in


Macroz:
When you go to user profile page without first loggin-in you should get redirected but aren't because the user-settings API always returns 200 OK

PR #2542, opened 2021-01-28
api: add forbidden to api/user-settings

marharyta:
# Checklist for author

## Reviewability
- [x] Link to issue (#2531)

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary - TODO

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -11,6 +11,7 @@
 
     (GET "/" []
       :summary "Get user settings"
+      :roles #{:logged-in}
       :return user-settings/UserSettings
       (ok (user-settings/get-user-settings (get-user-id))))
 
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -80,10 +80,13 @@
                    (when first-time? ; to avoid an infinite loop if the settings fail to save (e.g. unsupported language in cookies)
                      [[:check-if-should-save-language!]]))})))
 
-(defn fetch-user-settings! []
+(defn fetch-user-settings! [opts]
   (fetch "/api/user-settings"
-         {:handler #(rf/dispatch-sync [:loaded-user-settings %])
-          :error-handler (flash-message/default-error-handler :top "Fetch user settings")}))
+         (merge opts
+                {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
+                (if-not :hide-error-modal?
+                  {:error-handler (flash-message/default-error-handler :top "Fetch user settings")}
+                  {:error-handler nil}))))
 
 (rf/reg-event-fx
  ::save-user-language!
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -81,23 +81,15 @@
                      [[:check-if-should-save-language!]]))})))
 
 (rf/reg-event-fx
- :flash-message-error
- (fn [_ user-settings-error]
-   ;; TODO: refactor this part
-   ;; this doesnt work
-   ;; (flash-message/default-error-handler :top "Fetch user settings")
-   ;; this works
-   (flash-message/show-default-error! :top (str "Fetch user settings")))
- {})
-
-(rf/reg-event-fx
  :loaded-user-settings-fail
  (fn [{:keys [db]} [_ user-settings-error]]
    (let [logged-in? (:user (:identity db))]
      (if-not logged-in?
-       {:db db}
-       {:db db
-        :dispatch [:flash-message-error user-settings-error]}))))
+       (if-not (= 401 (:status user-settings-error))
+         {:db db
+          :dispatch (flash-message/show-default-error! :top (str "Fetch user settings"))}
+         {:db db})
+       {:db db}))))
 
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -93,7 +93,7 @@
 
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         {:simple-error-handler? true
+         {:custom-error-handler? true
           :handler #(rf/dispatch-sync [:loaded-user-settings %])
           :error-handler #(rf/dispatch [:loaded-user-settings-fail %])}))
 
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -83,6 +83,7 @@
 (rf/reg-event-fx
  :flash-message-error
  (fn [_ user-settings-error]
+   ;; TODO: refactor this part
    ;; this doesnt work
    ;; (flash-message/default-error-handler :top "Fetch user settings")
    ;; this works
@@ -94,9 +95,9 @@
  (fn [{:keys [db]} [_ user-settings-error]]
    (let [logged-in? (:user (:identity db))]
      (if-not logged-in?
+       {:db db}
        {:db db
-        :dispatch-n [[:flash-message-error user-settings-error]]}
-       {:db db}))))
+        :dispatch [:flash-message-error user-settings-error]}))))
 
 (defn fetch-user-settings! [opts]
   (fetch "/api/user-settings"
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -87,25 +87,23 @@
    ;; this doesnt work
    ;; (flash-message/default-error-handler :top "Fetch user settings")
    ;; this works
-   (if (= 401 (:status user-settings-error))
-     (flash-message/show-default-error! :top (str "Fetch user settings" (:status user-settings-error)))
-     (flash-message/show-default-error! :top (str "Fetch user settings")))
-   {}))
+   (flash-message/show-default-error! :top (str "Fetch user settings")))
+ {})
 
 (rf/reg-event-fx
  :loaded-user-settings-fail
  (fn [{:keys [db]} [_ user-settings-error]]
    (let [logged-in? (:user (:identity db))]
      (if-not logged-in?
+       {:db db}
        {:db db
-        :dispatch [:flash-message-error user-settings-error]}
-       {:db db}))))
+        :dispatch [:flash-message-error user-settings-error]}))))
 
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         (merge {:simple-error-handler? true}
-                {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
-                {:error-handler #(rf/dispatch [:loaded-user-settings-fail %])})))
+         {:simple-error-handler? true
+          :handler #(rf/dispatch-sync [:loaded-user-settings %])
+          :error-handler #(rf/dispatch [:loaded-user-settings-fail %])}))
 
 (rf/reg-event-fx
  ::save-user-language!
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -84,12 +84,8 @@
  :loaded-user-settings-fail
  (fn [{:keys [db]} [_ user-settings-error]]
    (let [logged-in? (:user (:identity db))]
-     (if-not logged-in?
-       (if-not (= 401 (:status user-settings-error))
-         {:db db
-          :dispatch (flash-message/show-default-error! :top (str "Fetch user settings"))}
-         {:db db})
-       {:db db}))))
+     {:dispatch (when-not logged-in? (when-not (= 401 (:status user-settings-error))
+                                       (flash-message/show-default-error! :top (str "Fetch user settings"))))})))
 
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -99,9 +99,9 @@
        {:db db
         :dispatch [:flash-message-error user-settings-error]}))))
 
-(defn fetch-user-settings! [opts]
+(defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         (merge opts
+         (merge {:default-error-handler? true}
                 {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
                 {:error-handler #(rf/dispatch [:loaded-user-settings-fail %])})))
 
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -80,13 +80,32 @@
                    (when first-time? ; to avoid an infinite loop if the settings fail to save (e.g. unsupported language in cookies)
                      [[:check-if-should-save-language!]]))})))
 
+(rf/reg-event-fx
+ :flash-message-error
+ (fn [_]
+   ;; this doesnt work
+   ;; (flash-message/default-error-handler :top "Fetch user settings")
+   ;; this works
+   (flash-message/show-default-error! :top "Fetch user settings")
+   {}))
+
+(rf/reg-event-fx
+ :loaded-user-settings-fail
+ (fn [{:keys [db]} [_ user-settings-error]]
+   (let [first-time? (not (:user-settings db))
+         logged-in? (:user (:identity db))]
+      (js/console.log "first-time?" (clj->js first-time?) "logged-in?" logged-in?)
+      (if-not logged-in?
+       {:db db}
+       {:db db
+        :dispatch-n [[:flash-message-error]]}
+      ))))
+
 (defn fetch-user-settings! [opts]
   (fetch "/api/user-settings"
          (merge opts
                 {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
-                (if-not :hide-error-modal?
-                  {:error-handler (flash-message/default-error-handler :top "Fetch user settings")}
-                  {:error-handler nil}))))
+                {:error-handler #(rf/dispatch-sync [:loaded-user-settings-fail %])})))
 
 (rf/reg-event-fx
  ::save-user-language!
--- a/src/cljs/rems/user_settings.cljs
+++ b/src/cljs/rems/user_settings.cljs
@@ -87,7 +87,9 @@
    ;; this doesnt work
    ;; (flash-message/default-error-handler :top "Fetch user settings")
    ;; this works
-   (flash-message/show-default-error! :top (str "Fetch user settings" "wtf"))
+   (if (= 401 (:status user-settings-error))
+     (flash-message/show-default-error! :top (str "Fetch user settings" (:status user-settings-error)))
+     (flash-message/show-default-error! :top (str "Fetch user settings")))
    {}))
 
 (rf/reg-event-fx
@@ -95,13 +97,13 @@
  (fn [{:keys [db]} [_ user-settings-error]]
    (let [logged-in? (:user (:identity db))]
      (if-not logged-in?
-       {:db db}
        {:db db
-        :dispatch [:flash-message-error user-settings-error]}))))
+        :dispatch [:flash-message-error user-settings-error]}
+       {:db db}))))
 
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         (merge {:default-error-handler? true}
+         (merge {:simple-error-handler? true}
                 {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
                 {:error-handler #(rf/dispatch [:loaded-user-settings-fail %])})))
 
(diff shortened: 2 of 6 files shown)


marharyta:
@opqdonut @Macroz adding the forbidden path triggers infinite ridirect loop

marharyta:
After a brief discussion with  @opqdonut we came to a conclusion that solution proposed by @Macroz would be the most practical. I agree, it is good to have fallback anyways.

marharyta:
Screenshot 2021-01-29 at 9 53 01

marharyta:
Screenshot 2021-02-01 at 12 48 45
This changes triggers error on login page load due to the fact that we through error since the user is not logged in.

Macroz:
> This changes triggers error on login page load due to the fact that we through error since the user is not logged in.

Well with these changes we explicitly still show the error message while we should eat the error message completely for this case (of not logged-in).

See this part:
```
:error-handler (flash-message/default-error-handler :top "Fetch user settings")}
```

marharyta:
The latest changes seem to reflect the desired outcome: 
- no wrapper over the fetch and no error message on login page
- what I believe could be done further is checking the log-in status and then fetching with logged in status

marharyta:
update: after a brief conversation with @Macroz we decided to create a separate dispatcher event for the error case of the fetch call, I am currently working on implementing it

marharyta:
Can you confirm @Macroz this is the right direction?

marharyta:
I know we had a catch up today, yet I just wanna be extra sure

marharyta:
Also, this is not ready yet, still have to:

- [ ] double check why ```/default-error-handler```  doesn't throw error in UI
- [ ] run tests
- [ ] figure out how to mimic fetch-user settings fail when logged in

marharyta:
@opqdonut any tips on how to test?

opqdonut:
An API-level test-case should be easy to add to `rems.api.test-user-settings`. That's something we should do.

What I'm less sure about is browser testing. I guess we could add a test case for navigating to `/profile` without logging in and checking that we get forwarded to the front page. However I'm not sure this feature is important enough to warrant a slow browser test case.

Some unit tests for our `fetch` util would be nice, but we don't have any scaffolding for that kind of testing yet, and maybe t's not worth it to set that up right now.

src/cljs/rems/user_settings.cljs
```diff
@@ -81,10 +81,30 @@
...
+(rf/reg-event-fx
+ :loaded-user-settings-fail
+ (fn [{:keys [db]} [_ user-settings-error]]
+   (let [logged-in? (:user (:identity db))]
+     (if-not logged-in?
+       {:db db}
+       {:db db
+        :dispatch [:flash-message-error user-settings-error]}))))
+
+(defn fetch-user-settings! [opts]
```
Macroz, line 103: Good news, @opqdonut and I discussed this and Joel noticed that since we only use `fetch-user-settings!` for the initial fetch (and not for user settings page) we can skip these dispatch events and just ignore everything when the response code is 401. We can skip the logged-in part since this doesn't need really need the complication.

src/cljs/rems/user_settings.cljs
```diff
@@ -81,10 +81,30 @@
...
   (fetch "/api/user-settings"
-         {:handler #(rf/dispatch-sync [:loaded-user-settings %])
-          :error-handler (flash-message/default-error-handler :top "Fetch user settings")}))
+         (merge opts
+                {:handler #(rf/dispatch-sync [:loaded-user-settings %])}
+                {:error-handler #(rf/dispatch [:loaded-user-settings-fail %])})))
```
Macroz, line 107: You could do a simple check like `(if (= 401 (:status response)) ... ...)` here.

src/cljs/rems/user_settings.cljs
marharyta: I do not see the login part as a complication, but rather as a logical extension

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,32 @@
                    (when first-time? ; to avoid an infinite loop if the settings fail to save (e.g. unsupported language in cookies)
                      [[:check-if-should-save-language!]]))})))
 
+(rf/reg-event-fx
+ :flash-message-error
+ (fn [_ user-settings-error]
+   ;; TODO: refactor this part
```
opqdonut, line 86: So is there still a TODO here? The if below looks a bit weird. Why not show the status code always or never?

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,32 @@
...
+   ;; this doesnt work
+   ;; (flash-message/default-error-handler :top "Fetch user settings")
+   ;; this works
+   (if (= 401 (:status user-settings-error)) 
+      (flash-message/show-default-error! :top (str "Fetch user settings" (:status user-settings-error)))
+     (flash-message/show-default-error! :top (str "Fetch user settings")))
+   {}))
+
+(rf/reg-event-fx
+ :loaded-user-settings-fail
```
opqdonut, line 96: What's the difference between `-fail` and `-error`? Could they be combined?

You could also consider just inlining the event(s) into the `:error-handler`. We have something like that in `rems.application/save-attachment` for example.

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,32 @@
...
+     (if-not logged-in?
+       {:db db
+        :dispatch [:flash-message-error user-settings-error]}
+       {:db db}))))
+
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         {:handler #(rf/dispatch-sync [:loaded-user-settings %])
-          :error-handler (flash-message/default-error-handler :top "Fetch user settings")}))
+         (merge {:simple-error-handler? true}
```
opqdonut, line 106: no need to merge multiple single-key maps: just use a literal map

src/cljs/rems/util.cljs
```diff
@@ -88,7 +88,9 @@
    (fn [resolve reject]
      (GET url (-> (merge {:response-format :transit}
                          opts
-                         {:error-handler (wrap-default-error-handler (:error-handler opts))})
+                         (if (:simple-error-handler? opts)
```
opqdonut, line 91: Could document this option in the docstring. The current docstring is a bit weird, but a short mention about this option would fit I think.

opqdonut (changes requested):

src/cljs/rems/user_settings.cljs
marharyta: I think I just got confused with the initial 401 suggestion

src/cljs/rems/util.cljs
marharyta: did that

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,30 @@
                    (when first-time? ; to avoid an infinite loop if the settings fail to save (e.g. unsupported language in cookies)
                      [[:check-if-should-save-language!]]))})))
 
+(rf/reg-event-fx
+ :flash-message-error
```
opqdonut, line 84: now that this event is just a oneliner, I'd rather get rid of it (and inline it into `:loaded-user-settings-fail`)

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,30 @@
...
+   ;; (flash-message/default-error-handler :top "Fetch user settings")
+   ;; this works
+   (flash-message/show-default-error! :top (str "Fetch user settings")))
+ {})
+
+(rf/reg-event-fx
+ :loaded-user-settings-fail
+ (fn [{:keys [db]} [_ user-settings-error]]
+   (let [logged-in? (:user (:identity db))]
+     (if-not logged-in?
```
opqdonut, line 97: Oh right, now I remember that discussion about 401. I think the point was that it would be nice to show an error when

- the user is not logged in
- *but* the settings API fails with some other code than 401

In other words, the error should _only be hidden if_:

- the user is not logged in
- and the error was 401

src/clj/rems/api/user_settings.clj
```diff
@@ -11,6 +11,7 @@
 
     (GET "/" []
       :summary "Get user settings"
+      :roles #{:logged-in}
```
opqdonut, line 14: you could add a not logged in case to `rems.api.test-user-settings`

opqdonut (changes requested):
sorry, I got some ideas I could've also posted in the previous review

src/cljs/rems/user_settings.cljs
Macroz: A logical extension that is not useful is basically unnecessary logic and code that should be simplified i.e. not exist?

test/clj/rems/api/test_user_settings.clj
```diff
@@ -16,6 +16,13 @@
   (let [user-id (str (UUID/randomUUID))]
     (test-helpers/create-user! {:eppn user-id})
 
+    (testing "default user settings without authentification"
+      (is (= {:status 401
+              :body "unauthorized"}
+             (-> (request :get "/api/user-settings")
+                 handler
+                 assert-unauthorized))))
```
Macroz, line 24: I think there is some confusion here. This test both checks the return value for `unauthorized` and calls a function called `assert-unauthorized`. I'm sure both should not be needed.

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,22 @@
...
+       (if-not (= 401 (:status user-settings-error))
+         {:db db
+          :dispatch (flash-message/show-default-error! :top (str "Fetch user settings"))}
+         {:db db})
+       {:db db}))))
```
Macroz, line 92: How about we simplify this code to something like this?
```
   {:db db
    :dispatch (when (not= 401 (:status user-settings-error)))
                (flash-message/show-default-error! :top (str "Fetch user settings")))}
```

But then we could also inline this below like @opqdonut suggested because there is no necessary reference to `db`.

Btw. I think the description string should be localized here.

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,22 @@
...
+
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         {:handler #(rf/dispatch-sync [:loaded-user-settings %])
-          :error-handler (flash-message/default-error-handler :top "Fetch user settings")}))
+         {:simple-error-handler? true
+          :handler #(rf/dispatch-sync [:loaded-user-settings %])
+          :error-handler #(rf/dispatch [:loaded-user-settings-fail %])}))
```
Macroz, line 98: Can just inline the code here.

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,22 @@
...
+         {:db db
+          :dispatch (flash-message/show-default-error! :top (str "Fetch user settings"))}
+         {:db db})
+       {:db db}))))
+
 (defn fetch-user-settings! []
   (fetch "/api/user-settings"
-         {:handler #(rf/dispatch-sync [:loaded-user-settings %])
-          :error-handler (flash-message/default-error-handler :top "Fetch user settings")}))
+         {:simple-error-handler? true
```
Macroz, line 96: I'm not sure why this was renamed to `:simple-error-handler?` but I think it's backwards. Isn't the normal i.e. simple case that you wrap the default error handler? And this would then be the ... `:wrap-default-error-handler? false` or even `:custom-error-handler true` case. Do you see what I mean?

src/cljs/rems/util.cljs
```diff
@@ -76,7 +76,7 @@
   "Fetches data from the given url with optional map of options like #'ajax.core/GET.
 
   Has sensible defaults with error handler, JSON and keywords.
-
+  You can use :simple-error-handler? to decide weather you would like use wrapper for the error handling.
```
Macroz, line 79: The confusion about the parameter name is evident in this sentence too, isn't it? This doesn't really say what it does or what the wrapper it's talking about is. One has to read the code.

test/clj/rems/api/testing.clj
```diff
@@ -161,6 +161,11 @@
   (assert (:success body) (pr-str body))
   body)
 
+(defn assert-unauthorized [response]
+  (response-is-unauthorized? response)
+  {:status (:status response)
+   :body (read-body response)})
+
```
Macroz, line 168: This is named `assert-` yet it does no such thing wrt. to unauthorized. Copying some fields from the response makes no sense to me.

Macroz (changes requested):

src/cljs/rems/user_settings.cljs
marharyta: If you are talking about ```:flash-message-error```  then it has been addressed

src/cljs/rems/user_settings.cljs
marharyta: Yes, I agree, yet, just to note, we did not have it localized before.

src/cljs/rems/user_settings.cljs
marharyta: I would leave it like this, if you do not have a strong opinion on it.

src/cljs/rems/util.cljs
marharyta: Yes, I agree. The thing is due to so many iterations it is hard to keep the naming clear, there were a lot of discussions about this fix, first from @opqdonut then we had a call with you, then you and Joel discussed it between yourselves ... I will take a look at it with some fresh eyes and try to read through it and then maybe the naming change will be more apparent to me.

test/clj/rems/api/testing.clj
marharyta: hmm, I was referencing the logic of the other functions in the file, but I guess they are sort of all over the place. 
Ill double check. thanks!

src/cljs/rems/user_settings.cljs
marharyta: same here, need to go through the ticket again to think of the better naming

test/clj/rems/api/test_user_settings.clj
marharyta: hmm, I guess I should create something like ```read-unauthorized```

src/cljs/rems/user_settings.cljs
Macroz: One more note. ACtually from an `fx` you don't need to return the `db` if you don't touch it.

test/clj/rems/api/test_user_settings.clj
Macroz: Why does it need to do it? The fields are available in the response object as they are, aren't they? And you can check the status if you so wish. No helper required.

test/clj/rems/api/testing.clj
Macroz: They serve a specific purpose, i.e. so that the threaded code works nicely and you get a clean error if the assertion was not met. This helper is not needed for the test at all, I think.

src/cljs/rems/util.cljs
Macroz: I think the most important and also the most exciting part of a PR is once you have the solution implemented. Then you refactor (the diff and) the solution until it is actually nice. I.e. "first make it work, then make it right, ...". Only at this stage can you see what the naming actually should be, what is unnecessary, etc.

src/cljs/rems/user_settings.cljs
marharyta: I think it looks rather consistent this way.

src/cljs/rems/user_settings.cljs
marharyta: Is this a suggestion or a requirement?

Macroz: I'll let @opqdonut decide. I would definitely do this refactor, it's like a one minute thing to fix the code.

src/cljs/rems/user_settings.cljs
Macroz: Yeah that is right, the simplification is elsewhere.

test/clj/rems/api/testing.clj
marharyta: I guess I can resolve it since I deleted this code?

src/cljs/rems/user_settings.cljs
marharyta: Do you think naming is more compelling now?

src/cljs/rems/user_settings.cljs
marharyta: I would like to clarify: are you talking about the localization or the one liner?

src/cljs/rems/user_settings.cljs
Macroz: Actually I don't. I mean if I am passing `:error-handler XXX` to me it means I'm also passing a custom error handler so I don't understand why I need to pass another parameter called `:custom-error-handler` to do something else? It affects the wrapping so I would've liked to see that word in it. :shrug:

src/cljs/rems/user_settings.cljs
opqdonut: Please simplify this code a bit. The repeated `:db db` and nested ifs make the logic look a lot more complicated than it is, at least to me. At least remove the `:db` keys, and also consider combining the conditions into one if.

test/clj/rems/api/testing.clj
```diff
@@ -146,6 +146,10 @@
   (assert-response-is-ok response)
   (read-body response))
 
+(defn read-unauthorized-body [response]
```
opqdonut, line 149: This is seems like a useful little helper, but please fix the name. It doesn't have anything to do with unauthorized does it? How about a name like `status-and-body` or `read-body-and-status`?

src/cljs/rems/util.cljs
```diff
@@ -76,7 +76,8 @@
   "Fetches data from the given url with optional map of options like #'ajax.core/GET.
 
   Has sensible defaults with error handler, JSON and keywords.
-
+  You can use :custom-error-handler? to decide weather you would like use wrapper for the error handling.
```
opqdonut, line 79: Picking a good name for this parameter is hard. Let's stick with this for now.

opqdonut (approved):
Good to go after fixing the two minor points: the :loaded-user-settings-fail implementation, and the naming of read-unauthorized-body.

test/clj/rems/api/testing.clj
marharyta: Yes, that is true, I was naming it based on purpose, not the action, my bad

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,20 @@
                    (when first-time? ; to avoid an infinite loop if the settings fail to save (e.g. unsupported language in cookies)
                      [[:check-if-should-save-language!]]))})))
 
+(rf/reg-event-fx
+ :loaded-user-settings-fail
+ (fn [{:keys [db]} [_ user-settings-error]]
+   (let [logged-in? (:user (:identity db))]
+     (if-not logged-in?
```
opqdonut, line 87: an if without an else case could be a when

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,20 @@
...
+       {:db db
```
opqdonut, line 88: you could just skip the `:db db` altogether: you only need to invoke the `:db` effect if you change the db!

marharyta:
ping @opqdonut

src/cljs/rems/user_settings.cljs
```diff
@@ -80,10 +80,20 @@
...
+        :dispatch (when-not (= 401 (:status user-settings-error))
```
opqdonut, line 89: could combine this when with the above if

opqdonut (approved):
feel free to merge

some code golfing suggestions included

src/cljs/rems/user_settings.cljs
marharyta: interesting, I saw somewhere that db has to be returned every time in re-frame, but I guess I either misunderstood or that was an old resource

opqdonut: yes, a `reg-event-db` needs to return the new state of the db, but a `reg-event-fx` returns a map of effects, and all events are optional!

marharyta: ```events are optional!``` like the enthusiasm! :)

marharyta: ok, I think I did address those comments

src/cljs/rems/user_settings.cljs
marharyta: addressed

src/cljs/rems/user_settings.cljs
marharyta: yeah

test/clj/rems/api/testing.clj
marharyta: addressed

opqdonut (approved):

Issue #2540, opened 2021-01-27
Avoid using database time


opqdonut:
Due to using a mixture of database and JVM time, some tests sometimes fail when db and JVM times are not in sync. This can happen especially on MacOS, where the DB docker container is running inside a separate virtual machine.

See also: https://github.com/CSCfi/rems/blob/master/test/clj/rems/db/testing.clj#L61

- [x] minimum level: don't use database time when creating data for tests
- [x] never use database time, if it's possible
  - [x] investigate
  - [x] implement

opqdonut:
List of columns that have a default value of now() on the DB side:

- audit_log.time
- catalogue_item.start
- entitlement.start
- license_attachment.start
- organization.modified

Shouldn't be too hard to move all of these defaults to the CLJ side.

PR #2543, opened 2021-01-29
test: use JVM time when creating catalogue item, not DB time

opqdonut:
... to avoid problems when DB time is slightly ahead of JVM time. This
makes the catalogue item seem disabled since we check the item :start
against the JVM time.

for #2540

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
  - could consider an ADR about database time later

## Testing
  - however, we don't have a test that simulates a db time difference

### DIFF
--- a/test/clj/rems/api/test_over_http.clj
+++ b/test/clj/rems/api/test_over_http.clj
@@ -1,7 +1,6 @@
 (ns ^:integration rems.api.test-over-http
   "API tests that use a full HTTP server."
   (:require [clj-http.client :as http]
-            [clj-time.core]
             [clojure.test :refer :all]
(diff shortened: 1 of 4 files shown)


marharyta:
Checked with @opqdonut and it works with integration tests!
Thanks!

marharyta (approved):

test/clj/rems/api/test_over_http.clj
```diff
@@ -1,13 +1,15 @@
 (ns ^:integration rems.api.test-over-http
   "API tests that use a full HTTP server."
   (:require [clj-http.client :as http]
+            [clj-time.core]
```
opqdonut, line 4: woops will remove these unneeded requires before merge

PR #2544, opened 2021-01-29
don't use database time

### LINKED ISSUE #2540: Avoid using database time

opqdonut:
- drop all now() defaults from schema
- add (time/now) defaults to code where required

fixes #2540

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
  - should there be one?

## Testing
  - test coverage for this stuff seemed pretty good

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Changes since v2.15
 ### Changes
 - REMS no longer sends the Server: HTTP header to avoid leaking version information. (#2216)
 - Text descriptions of some events in the log were phrased better. The created event also shows the original external id. (#2614)
-- REMS no longer uses the database time when creating rows. Requires migration, but does not change visible behaviour. (#2540)
+- REMS now consistently uses the application server clock when creating timestemps. Previously, database time was used in some situations, leading to minor inconsistencies. Requires migration, but does not change visible behaviour. (#2540)
 
 ### Fixes
 - CSS files are now marked as cacheable by browsers. In v2.15 they were mistakenly marked as uncacheable. (#2484)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Changes since v2.15
 ### Changes
 - REMS no longer sends the Server: HTTP header to avoid leaking version information. (#2216)
 - Text descriptions of some events in the log were phrased better. The created event also shows the original external id. (#2614)
+- REMS no longer uses the database time when creating rows. Requires migration, but does not change visible behaviour. (#2540)
 
 ### Fixes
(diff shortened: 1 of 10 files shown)


marharyta:
I personally have not finished the review of the code, but I believe this is generally a better idea

resources/sql/queries.sql
```diff
@@ -296,9 +296,9 @@ WHERE id = :id;
 
 -- :name create-license-attachment! :insert
 INSERT INTO license_attachment
-(modifierUserId, filename, type, data)
+(modifierUserId, filename, type, data, start)
```
marharyta, line 299: are you planning to add the same code to ```organisations``` ?

opqdonut: It's there already! It turns out the default values in audit_log and organization were not used: REMS always specified the value explicitly.

marharyta: ok, so we did have this approach already? Good to refactor then!

opqdonut: Yeah it was kind of a belt-and-suspenders thing where most timestamps were created in REMS, but some were left up to the db defaults.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Changes since v2.15
 ### Changes
 - REMS no longer sends the Server: HTTP header to avoid leaking version information. (#2216)
 - Text descriptions of some events in the log were phrased better. The created event also shows the original external id. (#2614)
+- REMS no longer uses the database time when creating rows. Requires migration, but does not change visible behaviour. (#2540)
```
Macroz, line 14: Maybe would mention that it uses application server time now consistently instead.

Macroz (approved):

resources/sql/queries.sql
Macroz: The db defaults make sense for helping with the SQL console use. I think they can be left out. The middle ground is to have the defaults and always make sure you don't use the default in the app.

Macroz:
Looks good! Maybe a quick couple line ADR about the database design where we would only have this point for now?

marharyta:
Looks good to me too.

marharyta (approved):

Macroz (approved):

PR #2549, opened 2021-02-05
docs: add issue template file

marharyta:
# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up


docs/issue_template.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Input here issue description
+
+### Extra information
+
+- [ ] Login role (to reproduce, what role should I log in with)
+- [ ] Possible screenshots
+- [ ] For UI-related issues, you might include references to design or related UX/UI study
+- [ ] For the user feedback feature, you might include the name of the user (organisation) requesting a feature
+- [ ] Importance - you might mark this if issue is critical / needs to be fixed by nextv release
```
opqdonut, line 21: We've not been tracking importance with labels or anything like that yet, so I think we can drop this line for now.

docs/issue_template.md
```diff
@@ -0,0 +1,22 @@
...
+- [ ] Feature Request (requested by stakeholders)
+- [ ] Feature Proposal (possible improvements)
+
+## Issue Description
+
+### Input here issue description
+
+### Extra information
+
+- [ ] Login role (to reproduce, what role should I log in with)
```
opqdonut, line 17: Mentioning the login role can be important, but I'm not sure if _just_ the role is that useful. Could we reword this into something like

> - [ ] Steps to reproduce the bug or to illustrate the feature request (including login role)

docs/issue_template.md
```diff
@@ -0,0 +1,22 @@
+# Checklist for author
+
+Remove items that aren't applicable.
+
+## Issue Type
+
+- [ ] Bug Report
+- [ ] Feature Request (requested by stakeholders)
+- [ ] Feature Proposal (possible improvements)
```
opqdonut, line 9: I'm not sure I know what the difference between request and proposal is :)

Could we just add a list of the feature type labels we use here and instruct the user to pick one? Bug, Feature, Enhancement, Investigate?

marharyta: This was just my mind trying to think of some good naming on a Friday evening, I was thinking of more like "what the client has requested" vs "what we propose to add" but your naming conventions are way better

jaakkocsc: I guess request is a "must have", and proposal is a "would be nice to have"? I think labelling would be our task based on our investigation but I also don't feel like there's a need to distinguish between request and proposal. Given that rems users (i.e. non-technical persons) would create issues still requires training on how to create an issue anyway.

docs/issue_template.md
jaakkocsc: I agree, just a role is too specific.

docs/issue_template.md
jaakkocsc: I'm not completely sure the checkboxes actually do something here at all, except for the bug or feature request.

marharyta: From my perspective, they are more on the "checklist/reminder" side, rather then accomplishing any particular goal. like, if you are reporting a bug, it reminds you to mention what credentials to log in with, for example

Issue #1999, opened 2020-02-11
Multiple resources with different licenses in UI


jaakkocsc:
As an owner, when I go to the create catalogue item page,

if same resource exists with different licenses the UI does not distinguish them, i.e. it's impossible to know which resource to select.

marharyta:
Can you maybe give a screenshot @jaakkocsc ?

Macroz:
It's the create catalogue item page resource dropdown that displays just the name of the item?

marharyta:
<img width="369" alt="Screenshot 2021-02-03 at 10 35 09" src="https://user-images.githubusercontent.com/7344718/106721260-1a7deb80-660d-11eb-9d62-4fbf6380a30d.png">
It doesn't look like the type of licence is currently exposed in the re-frame db

marharyta:
I think the easiest would be to get licence from the ```queries.sql``` the licence and expose it into the catalogue ```get-catalogue-items```, what do you think @Macroz @opqdonut

Macroz:
> I think the easiest would be to get licence from the `queries.sql` the licence and expose it into the catalogue `get-catalogue-items`, what do you think @Macroz @opqdonut

I think the resources are fetched from the database and the licenses are joined to them in the service layer so they should be available already. You are looking at the wrong place (it's not the catalogue but the resources ...)

opqdonut:
Perhaps this helps:
![create-catalogue-item](https://user-images.githubusercontent.com/207385/106745319-97b75980-6629-11eb-8fd7-740cc3746006.png)

marharyta:
Hmm, do we have the data model definition anywhere in the docs?

Macroz:
You can have a look at it in `re-frisk` like Joel showed or from the https://rems-dev.rahtiapp.fi/swagger-ui/index.html#/resources/get_api_resources too. The relevant code can also be seen elsewhere (the localized names of the licenses).

marharyta:
Thanks a lot @Macroz

marharyta:
I think my initial idea was pretty close, though. We do fetch catalogue items from resources and we do not expose licences in that call.

Macroz:
I am confused at what you are saying, so are we talking about the same thing:
- create catalogue item admin page
- the select resource dropdown (these should show the license names)
- resources are fetched from /api/resources which joins the licenses to every resource and contains the localized name too

opqdonut:
[...]

PR #2553, opened 2021-02-09
feat: show license type in selection poc

marharyta:
Screenshot 2021-02-17 at 15 09 01

Requires to run ```lein reset``` and ```lein test-data``` to check with duplicate sources.

## Reviewability
- [x] Link to issue (#1999)

## Backwards compatibility
- [x] Events are have migrations

## Documentation

## Testing

## Follow-up

### DIFF
--- a/dev-config.edn
+++ b/dev-config.edn
@@ -3,7 +3,7 @@
  :database-url "postgresql://localhost/rems?user=rems"
  :test-database-url "postgresql://localhost/rems_test?user=rems_test"
  :search-index-path "target/search-index-dev"
- :theme-path "example-theme/theme.edn"
+ :theme-path "../rems-deploy/themes/theme-thl/theme.edn"
  :authentication :fake
  :log-authentication-details true
  ;; Open ID Connect settings
@@ -16,7 +16,7 @@
  :oidc-extra-attributes [{:attribute "nickname" :name {:en "Nickname" :fi "Lempinimi"}}
                          {:attribute "picture"}]
 
- :languages [:en :fi :sv]
+ :languages [:en]
  :public-url "http://localhost:3000/"
  :render-css-file? true
  :extra-pages [{:id "about"
--- a/dev-config.edn
+++ b/dev-config.edn
@@ -3,7 +3,7 @@
  :database-url "postgresql://localhost/rems?user=rems"
  :test-database-url "postgresql://localhost/rems_test?user=rems_test"
  :search-index-path "target/search-index-dev"
- :theme-path "../rems-deploy/themes/theme-thl/theme.edn"
+ :theme-path "example-theme/theme.edn"
  :authentication :fake
  :log-authentication-details true
  ;; Open ID Connect settings
@@ -16,7 +16,7 @@
  :oidc-extra-attributes [{:attribute "nickname" :name {:en "Nickname" :fi "Lempinimi"}}
                          {:attribute "picture"}]
 
- :languages [:en]
+ :languages [:en :fi :sv]
  :public-url "http://localhost:3000/"
  :render-css-file? true
  :extra-pages [{:id "about"
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -898,8 +898,8 @@
         res-duplicate-resource-name1 (create-resource! {:resource-ext-id "duplicate resource name"
                                                         :organization {:organization/id "hus"
                                                                        :organization/short-name {:en "HUS en"
-                                                                                    :fi "HUS fi"
-                                                                                    :sv "HUS sv"}}
+                                                                                                 :fi "HUS fi"
+                                                                                                 :sv "HUS sv"}}
                                                         :actor owner
                                                         :license-ids [license1 extra-license attachment-license]})
         res-duplicate-resource-name2 (create-resource! {:resource-ext-id "duplicate resource name"
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -868,16 +868,10 @@
 
         ;; Create resources
         res1 (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403262"
-                                :organization {:organization/id "nbn"
-                                               :organization/short-name {:en "NBN en"
-                                                                         :fi "NBN fi"
-                                                                         :sv "NBN sv"}}
+                                :organization {:organization/id "nbn"}
                                 :actor owner})
         res2 (create-resource! {:resource-ext-id "Extra Data"
-                                :organization {:organization/id "nbn"
-                                               :organization/short-name {:en "NBN en"
-                                                                         :fi "NBN fi"
-                                                                         :sv "NBN sv"}}
+                                :organization {:organization/id "nbn"}
                                 :actor owner
                                 :license-ids [license1]})
         res3 (create-resource! {:resource-ext-id "something else"
@@ -889,17 +883,11 @@
                                                   :actor organization-owner1
                                                   :license-ids [license-organization-owner]})
         res-with-extra-license (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403263"
-                                                  :organization {:organization/id "nbn"
-                                                                 :organization/short-name {:en "NBN en"
-                                                                                           :fi "NBN fi"
-                                                                                           :sv "NBN sv"}}
+                                                  :organization {:organization/id "nbn"}
                                                   :actor owner
                                                   :license-ids [extra-license attachment-license]})
         res-duplicate-resource-name1 (create-resource! {:resource-ext-id "duplicate resource name"
-                                                        :organization {:organization/id "hus"
-                                                                       :organization/short-name {:en "HUS en"
-                                                                                                 :fi "HUS fi"
-                                                                                                 :sv "HUS sv"}}
+                                                        :organization {:organization/id "hus"}
                                                         :actor owner
                                                         :license-ids [license1 extra-license attachment-license]})
         res-duplicate-resource-name2 (create-resource! {:resource-ext-id "duplicate resource name"
--- a/src/clj/rems/db/test_data.clj
+++ b/src/clj/rems/db/test_data.clj
@@ -834,6 +834,15 @@
                                    :license/link {:en "https://www.apache.org/licenses/LICENSE-2.0"
                                                   :fi "https://www.apache.org/licenses/LICENSE-2.0"
                                                   :sv "https://www.apache.org/licenses/LICENSE-2.0"}})
+        license2 (create-license! {:actor owner
+                                   :license/type :link
+                                   :organization {:organization/id "nbn"}
+                                   :license/title {:en "Demo license 2"
+                                                   :fi "Demolisenssi 2"
+                                                   :sv "Demolicens 2"}
+                                   :license/link {:en "https://fedoraproject.org/wiki/Licensing/Beerware"
+                                                  :fi "https://fedoraproject.org/wiki/Licensing/Beerware"
+                                                  :sv "https://fedoraproject.org/wiki/Licensing/Beerware"}})
         extra-license (create-license! {:actor owner
                                         :license/type :link
                                         :organization {:organization/id "nbn"}
@@ -859,10 +868,16 @@
 
         ;; Create resources
         res1 (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403262"
-                                :organization {:organization/id "nbn"}
+                                :organization {:organization/id "nbn"
+                                               :organization/short-name {:en "NBN en"
+                                                                         :fi "NBN fi"
+                                                                         :sv "NBN sv"}}
                                 :actor owner})
         res2 (create-resource! {:resource-ext-id "Extra Data"
-                                :organization {:organization/id "nbn"}
+                                :organization {:organization/id "nbn"
+                                               :organization/short-name {:en "NBN en"
+                                                                         :fi "NBN fi"
+                                                                         :sv "NBN sv"}}
                                 :actor owner
                                 :license-ids [license1]})
         res3 (create-resource! {:resource-ext-id "something else"
@@ -874,9 +889,31 @@
                                                   :actor organization-owner1
                                                   :license-ids [license-organization-owner]})
         res-with-extra-license (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403263"
-                                                  :organization {:organization/id "nbn"}
+                                                  :organization {:organization/id "nbn"
+                                                                 :organization/short-name {:en "NBN en"
+                                                                                           :fi "NBN fi"
+                                                                                           :sv "NBN sv"}}
                                                   :actor owner
                                                   :license-ids [extra-license attachment-license]})
+        res-duplicate-resource-name1 (create-resource! {:resource-ext-id "duplicate resource name"
+                                                        :organization {:organization/id "hus"
+                                                                       :organization/short-name {:en "HUS en"
+                                                                                    :fi "HUS fi"
+                                                                                    :sv "HUS sv"}}
+                                                        :actor owner
+                                                        :license-ids [license1 extra-license attachment-license]})
+        res-duplicate-resource-name2 (create-resource! {:resource-ext-id "duplicate resource name"
+                                                        :organization {:organization/id "hus"}
+                                                        :actor owner
+                                                        :license-ids [license2 extra-license attachment-license]})
+        res-duplicate-resource-name-with-long-name1 (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403263443773465837568375683683756"
+                                                                       :organization {:organization/id "hus"}
+                                                                       :actor owner
+                                                                       :license-ids [license1 extra-license attachment-license]})
+        res-duplicate-resource-name-with-long-name2 (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403263443773465837568375683683756"
+                                                                       :organization {:organization/id "hus"}
+                                                                       :actor owner
+                                                                       :license-ids [license2 extra-license attachment-license]})
 
         workflows (create-workflows! (merge users +bot-users+))
         _ (db/create-workflow-license! {:wfid (:organization-owner workflows)
--- a/src/cljs/rems/administration/create_catalogue_item.cljs
+++ b/src/cljs/rems/administration/create_catalogue_item.cljs
@@ -191,17 +191,23 @@
 (defn resource-label [r language counts]
   (let [organisation (get-in r [:organization :organization/short-name language])
         duplicate? (> (get counts (:resid r)) 1)
-        licenses? (not-empty (:licenses r))]
+        licenses? (not-empty (:licenses r))
+        localized? (if (and licenses? (not (every? nil? (mapv
+                                                         (fn [l] (get-in l [:localizations language :title]))
+                                                         (:licenses r)))))
+                     true
+                     false)]
     (str (:resid r)
          (when organisation
            (str/join [" (" (text :t.administration/org) ": " organisation ")"]))
          (when duplicate?
            (when licenses?
-             (str/join [" (" (str/lower-case (text :t.administration/licenses)) ": "
-                        (str/join ", " (mapv
-                                        (fn [l] (get-in l [:localizations language :title]))
-                                        (:licenses r)))
-                        ")"]))))))
+             (when localized?
+               (str/join [" (" (str/lower-case (text :t.administration/licenses)) ": "
+                          (str/join ", " (filter #(not (nil? %)) (mapv
+                                                                  (fn [l] (get-in l [:localizations language :title]))
+                                                                  (:licenses r))))
+                          ")"])))))))
 
 (defn- catalogue-item-resource-field []
   (let [resources @(rf/subscribe [::resources])
--- a/src/cljs/rems/administration/create_catalogue_item.cljs
+++ b/src/cljs/rems/administration/create_catalogue_item.cljs
@@ -191,7 +191,7 @@
 (defn resource-label [r language counts]
   (let [organisation (get-in r [:organization :organization/short-name language])
         duplicate? (> (get counts (:resid r)) 1)
-        licenses? (not-empty (:licenses r))]
+        licenses? (seq (:licenses r))]
     (str (:resid r)
          (when organisation
            (str " (" (text :t.administration/org) ": " organisation ")"))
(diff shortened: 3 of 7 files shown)


dev-config.edn
```diff
@@ -3,7 +3,7 @@
  :database-url "postgresql://localhost/rems?user=rems"
  :test-database-url "postgresql://localhost/rems_test?user=rems_test"
  :search-index-path "target/search-index-dev"
- :theme-path "example-theme/theme.edn"
+ :theme-path "../rems-deploy/themes/theme-thl/theme.edn"
```
opqdonut, line 6: you probably didn't mean to commit this

dev-config.edn
```diff
@@ -16,7 +16,7 @@
  :oidc-extra-attributes [{:attribute "nickname" :name {:en "Nickname" :fi "Lempinimi"}}
                          {:attribute "picture"}]
 
- :languages [:en :fi :sv]
+ :languages [:en]
```
opqdonut, line 19: or this

src/clj/rems/db/test_data.clj
```diff
@@ -859,10 +868,16 @@
 
         ;; Create resources
         res1 (create-resource! {:resource-ext-id "urn:nbn:fi:lb-201403262"
-                                :organization {:organization/id "nbn"}
+                                :organization {:organization/id "nbn"
+                                               :organization/short-name {:en "NBN en"
+                                                                         :fi "NBN fi"
+                                                                         :sv "NBN sv"}}
```
opqdonut, line 874: `create-resource!` only cares about the `organization/id`. The other members of `:organization` are ignored. You can verify this by following the definitions to `rems.api.services.resource/create-resource!` if you want.

To explain that a bit, `create-resource!` only wants a reference to an existing organization, via a parameter like `:organization {:organization/id "foo"}`. It doesn't create the organization. The `:organization/short-name` is only needed when calling `create-organization!`.

src/clj/rems/db/test_data.clj
```diff
@@ -859,10 +868,16 @@
...
                                 :actor owner})
         res2 (create-resource! {:resource-ext-id "Extra Data"
-                                :organization {:organization/id "nbn"}
+                                :organization {:organization/id "nbn"
+                                               :organization/short-name {:en "NBN en"
+                                                                         :fi "NBN fi"
+                                                                         :sv "NBN sv"}}
```
opqdonut, line 880: same here

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,13 +188,34 @@
          :item-selected? item-selected?
          :on-change #(rf/dispatch [::set-selected-workflow %])}])]))
 
+(defn resourse-dropdown [r language counts]
+  (let [organisation? (get-in r [:organization :organization/short-name (keyword language)])
```
opqdonut, line 192: I'm pretty sure the language is guaranteed to be a keyword, I think you can just replace `(keyword language)` with `language`.

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,13 +188,34 @@
          :item-selected? item-selected?
          :on-change #(rf/dispatch [::set-selected-workflow %])}])]))
 
+(defn resourse-dropdown [r language counts]
```
opqdonut, line 191: Since this is used as an `:item-label`, perhaps a name like `resource-label`? Also, it could be private.

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,13 +188,34 @@
...
+         (when organisation?
+           (str/join " " [" (org:" organisation? ")"]))
+      ;; when duplicate is found, show duplicate
+         (when duplicate?
+         ;; when licenses are found, show licences, but only if it is duplicate
+           (when licenses?
+             (str/join " "
+                       [" (licenses:"
+                        (str/join ", " (mapv
+                                        (fn [l] (:title ((keyword language) (:localizations l))))
```
opqdonut, line 206: consider: `(fn [l] (get-in l [:localizations language :title]))`

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,13 +188,34 @@
...
+
 (defn- catalogue-item-resource-field []
   (let [resources @(rf/subscribe [::resources])
+        counts (frequencies (map :resid resources))
         editing? @(rf/subscribe [::editing?])
         selected-resource @(rf/subscribe [::selected-resource])
         item-selected? #(= (:id %) (:id selected-resource))
         language @(rf/subscribe [:language])]
     [:div.form-group
+     (js/console.log "yay" (clj->js (get-in (first resources) [:organization :id])))
```
opqdonut, line 218: this can probably be left out

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+                                :enabled true
+                                :id 8
+                                :licensetype "text"
+                                :localizations {:en {:attachment-id nil
+                                                     :textcontent "Be fast."
+                                                     :title "Performance License"}}}]}
+                   {:resid "u"}
+                   {:resid "z"}]
+        counts (frequencies (map :resid resourses))]
+    (testing "sorting resources"
```
opqdonut, line 112: is this test about sorting?

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+      (is (not (empty? (map resourse-dropdown resourses))))
```
opqdonut, line 113: this test case looks a bit weird: it's calling `resourse-dropdown` with only one parameter, when it really wants 3, right?

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+      (is (= ["x (org: NBN en )"
+              "y (org: NBN en ) (licenses: Performance License, Second License )"
```
opqdonut, line 115: Could you get rid of the extra space? I mean `(org: NBN en)` instead of `(org: NBN en )` and similarly for `(licenses: ...)`. Just something that bugs me.

marharyta:
@opqdonut I could potentially add more tests, but this far I think this is the minimum that does the job

opqdonut (changes requested):
In general, a nice surgical implementation, and good tests. Thanks. Some stuff to clean up however, see my comments.

opqdonut:
> @opqdonut I could potentially add more tests, but this far I think this is the minimum that does the job

I think the current test is enough!

dev-config.edn
marharyta: oh, true, my bad

dev-config.edn
marharyta: same here

src/clj/rems/db/test_data.clj
marharyta: I ll fix this, I am pretty sure you mentioned it last time but it somehow escaped my memory

src/cljs/rems/administration/create_catalogue_item.cljs
marharyta: That could be a better way to do it, thanks!

src/cljs/rems/administration/create_catalogue_item.cljs
marharyta: yes, my bad

test/cljs/rems/administration/test_create_catalogue_item.cljs
marharyta: sure, it was bugging me too, but I wanted to very this with you

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,13 +188,34 @@
          :item-selected? item-selected?
          :on-change #(rf/dispatch [::set-selected-workflow %])}])]))
 
+(defn resourse-dropdown [r language counts]
+  (let [organisation? (get-in r [:organization :organization/short-name (keyword language)])
+        duplicate? (> (get counts (:resid r)) 1)
+        licenses? (not-empty (:licenses r))]
+    (str (:resid r)
+      ;; when organisation is found, show organisation
```
opqdonut, line 196: Oh, one more thing: I think your code is clear enough even without these comments. What do you think?

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,8 +188,24 @@
          :item-selected? item-selected?
          :on-change #(rf/dispatch [::set-selected-workflow %])}])]))
 
+(defn resource-label [r language counts]
+  (let [organisation (get-in r [:organization :organization/short-name language])
+        duplicate? (> (get counts (:resid r)) 1)
+        licenses? (not-empty (:licenses r))]
+    (str (:resid r)
+         (when organisation
+           (str/join [" (" (text :t.administration/org) ": " organisation ")"]))
```
opqdonut, line 197: `(str/join [x y z])` can be written more simply as `(str x y z)`

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,8 +188,24 @@
...
+         (when duplicate?
+           (when licenses?
+             (str/join [" (" (str/lower-case (text :t.administration/licenses)) ": "
```
opqdonut, line 200: this `str/join` can be simplified as well

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,8 +188,24 @@
...
+                        (str/join ", " (mapv
```
opqdonut, line 201: this one can't since it uses a delimiter

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
     (testing "missing form"
       (is (nil? (build-request (assoc form :form nil)
                                languages))))))
+
+(deftest resourse-dropdown-test
```
opqdonut, line 50: nit: this should probably be `resource-label-test` to match the name of the function

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+  (let [resourses [{:resid "x"
```
opqdonut, line 51: nit: resource has a C

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+                                :id 8
+                                :licensetype "text"
+                                :localizations {:en {:attachment-id nil
+                                                     :textcontent "Be fast."
+                                                     :title "Performance License"}}}]}
+                   {:resid "u"}
+                   {:resid "z"}]
+        counts (frequencies (map :resid resourses))]
+    (testing "resource-label"
+      (is (not (empty? (map #(resource-label % :en counts) resourses))))
```
opqdonut, line 113: it seems to me this test is redundant with the following one: both run the same code, but this one only checks that the result is non-empty, while the next one checks the content literally

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,82 @@
...
+      (is (= ["x ([:t.administration/org]: NBN en)"
+              "y ([:t.administration/org]: NBN en) ([:t.administration/licenses]: Performance License, Second License)"
+              "y ([:t.administration/licenses]: Performance License)"
+              "u"
+              "z"]
+             (map #(resource-label % :en counts) resourses)))
+      (is (not (= ["x (org: NBN fi)"
```
opqdonut, line 120: are you sure you want `(not (=` here?

opqdonut (changes requested):
looks good, some minor points to fix

src/cljs/rems/administration/create_catalogue_item.cljs
marharyta: oh, I didn't know that! Cool!

test/cljs/rems/administration/test_create_catalogue_item.cljs
marharyta: yeah, right, I missed it in the refactoring

src/cljs/rems/administration/create_catalogue_item.cljs
marharyta: I will not be refactoring this since it creates additional friction with testing, but overall it is a nice suggestion

marharyta: - replace nested whens with and
- remove string joins and replace with just one statement

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,8 +188,30 @@
...
+                     false)]
+    (str (:resid r)
+         (when organisation
+           (str/join [" (" (text :t.administration/org) ": " organisation ")"]))
+         (when duplicate?
+           (when licenses?
+             (when localized?
+               (str/join [" (" (str/lower-case (text :t.administration/licenses)) ": "
+                          (str/join ", " (filter #(not (nil? %)) (mapv
+                                                                  (fn [l] (get-in l [:localizations language :title]))
```
marharyta, line 208: - remove instead of filter not nil

marharyta: - license translation
- capitalize org

marharyta:
@Macroz re-request ping

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -188,8 +188,22 @@
          :item-selected? item-selected?
          :on-change #(rf/dispatch [::set-selected-workflow %])}])]))
 
+(defn resource-label [r language counts]
+  (let [organisation (get-in r [:organization :organization/short-name language])
+        duplicate? (> (get counts (:resid r)) 1)
+        licenses? (not-empty (:licenses r))]
```
Macroz, line 194: The Clojure idiom for this is `(seq (:licenses r))`

marharyta: Thanks!

marharyta:
@Macroz I believe that should do it?

marharyta:
Screenshot 2021-02-24 at 16 40 29

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -204,10 +218,7 @@
         {:id resource-dropdown-id
          :items (->> resources (filter :enabled) (remove :archived))
          :item-key :id
-         :item-label #(str (:resid %)
-                           " (org: "
-                           (get-in % [:organization :organization/short-name language])
-                           ")")
+         :item-label #(resource-label % language counts)
```
Macroz, line 221: Minor thing here, I would prefer to send the count of the one resource only here, not the counts of all resources to the `resource-label` function. The responsibilities are a bit mixed. I'm not insisting on changing this.

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,105 @@
...
+                                                     :title "Performance License"}
+                                                :fi {:attachment-id nil
+                                                     :textcontent "Be fast."
+                                                     :title "Performance License FI"}
+                                                :sv {:attachment-id nil
+                                                     :textcontent "Be fast."
+                                                     :title "Performance License SV"}}}]}
+                   {:resid "u"}
+                   {:resid "z"}]
+        counts (frequencies (map :resid resources))]
```
Macroz, line 129: And here, you would not have to do this if you sent the single count. This duplicates "complex logic".

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,105 @@
     (testing "missing form"
       (is (nil? (build-request (assoc form :form nil)
                                languages))))))
+
+(deftest resource-label-test
```
Macroz, line 50: I think this test has some problems that we could work on:
- First of all, this test is massive. We should not need to add almost 100 lines of example data to test this tiny feature. How about we reduce the resource test data to only what one call to `resource-label` actually needs? Second possibility is to refactor or extract the complex logic in another manner so that testing it becomes easier. Another possibility is writing this level of test as an integration test with real created resources. I think we should do at least the first one.
- The expected expressions contain a mix of language and unlocalized keys. I think it smells of wrong level of testing or functions. I think we can live with this for now.
- Why does the test of `resource-label` function need to use `map` and call the function to several resources? Would it not be better to e.g. test one resource at a time each in its own `is` expression with possible explanation as to what case it is?
- Why does it need to test all the different languages and why is not one enough?

Macroz (dismissed):
I think this looks pretty good now. The changes you did to the logic of `resource-label` worked out. I think we should spend a little love on the test now. It is massive! If you try the first suggestion (i.e. removing unnecessary parts of the test data and simplifying the names of things) and then streamline the actual test assertions I think it'll be good enough.

test/cljs/rems/administration/test_create_catalogue_item.cljs
marharyta: I agree with you that the test could use some clean-up. The pointer is pretty clear. I ll work on the test data a bit more and come back to you.

marharyta: "Why does it need to test all the different languages and why is not one enough? " - in the previous iterations I was not sure whether all license features were localized so adding that made sense

marharyta: "First of all, this test is massive. We should not need to add almost 100 lines of example data to test this tiny feature. How about we reduce the resource test data to only what one call to resource-label actually needs? " - done

src/cljs/rems/administration/create_catalogue_item.cljs
marharyta: I am not sure what implications it might have, maybe I can leave it like this for now?

test/cljs/rems/administration/test_create_catalogue_item.cljs
marharyta: I do not understand this comment, I am sorry

test/cljs/rems/administration/test_create_catalogue_item.cljs
marharyta: Why does the test of resource-label function need to use map and call the function to several resources? - isn't the entire purpose of the function to iterate through the selections?

test/cljs/rems/administration/test_create_catalogue_item.cljs
Macroz: Just that if you did the previous, and if you could pass the count simply to `resource-label` as a number, you would not need to duplicate this code in the test, that you must do in the real component.

test/cljs/rems/administration/test_create_catalogue_item.cljs
Macroz: First of all you are testing `resource-label` i.e. the label for a *single resource*. Therefore the test should test that. And that does not iterate through selections, right?

marharyta:
@Macroz I think I have addressed everything at this stage?

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -76,6 +76,34 @@
                    {:resid "z"}]
         counts (frequencies (map :resid resources))]
     (testing "resource-label"
+      (is (= ["y ([:t.administration/org]: NBN en) ([:t.administration/licenses]: Performance License, Second License)"
+              "y ([:t.administration/licenses]: Performance License)"]
+             (map #(resource-label % :en counts) [{:resid "y"
```
Macroz, line 81: Why still the `map` here? Shouldn't the second be its own case like the second `is` below?

marharyta: I wanted to check the case with two similar resids. 
I can add another case below.

Macroz: One `is` per one call of tested function i.e. `resource-label` is the normal case. Also now you already moved half a step into this direction but not quite there.

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -104,6 +104,19 @@
...
+                              :licenses [{:localizations {:en {:title "Performance License"}
+                                                          :fi {:title "Performance License FI"}
+                                                          :sv {:title "Performance License SV"}}}
+                                         {:localizations {:en {:title "Second License"}
+                                                          :fi {:title "Second License FI"}
+                                                          :sv {:title "Second License SV"}}}]
+                              :organization {:organization/id "nbn"
+                                             :organization/short-name {:en "NBN en"
+                                                                       :fi "NBN fi"
+                                                                       :sv "NBN sv"}}} :en counts)))
```
opqdonut, line 118: I like an explicit test like this. you could also have a literal `{"y" 2}` for `counts`. Then you could only have literal cases like this, and wouldn't need the `let` at the start of this test at all.

test/cljs/rems/administration/test_create_catalogue_item.cljs
```diff
@@ -46,3 +46,107 @@
...
+                    :licenses [{:localizations {:en {:title "Performance License"}
+                                                :fi {:title "Performance License FI"}
+                                                :sv {:title "Performance License SV"}}}]}
+                   {:resid "u"}
+                   {:resid "z"}]
+        counts (frequencies (map :resid resources))]
+    (testing "resource-label"
+      (is (= ["y ([:t.administration/org]: NBN en) ([:t.administration/licenses]: Performance License, Second License)"
+              "y ([:t.administration/licenses]: Performance License)"]
+             (map #(resource-label % :en counts) [{:resid "y"
```
Macroz, line 81: And the `map` is still here?

marharyta: I did not delete it, I added more tests cases.

marharyta:
ping @Macroz @opqdonut re-review

test/cljs/rems/administration/test_create_catalogue_item.cljs
Macroz: You marked the discussion as resolved?

marharyta: this one? no, obviously

marharyta: I can delete the test altogether

opqdonut (approved):

marharyta:
btw, @opqdonut  and @Macroz  if there is something that prevents the PR from being merged, please be a bit more explicit about it. If the fix is preventing the PR from being merged, it is not a suggestion, it is a requirement. I am completely ok with them, but please put that distinction out there so that it is clear to me.

PR #2558, opened 2021-02-11
migrations: don't recreate unique constraint on resource

### LINKED ISSUE #2557: down migrations fail when there are duplicate resources
The migration `20200204064800-drop-unique-resid.down.sql` tries to create a unique constraint on resid, but it will fail if there actually are duplicate resources.

Figure out which option is the best:
- remove the down migration: don't recreate the unique index
- make the down migration drop duplicates
- make the down migration rename duplicates

opqdonut:
... in down migrations.

This is so that we can run a `lein reset` even after adding multiple
resources with the same resid.

This is fine becuase a) we don't really support or need to rollback to
versions from 2019 and b) the code for those versions actually works
fine even without the unique constraint in the db.

fixes #2557

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Changes since v2.16
 ### Fixes
 - Searching for applications by the original REMS generated id works, even if another id has been assigned. (#2564)
 - GA4GH Visa (output by the experimental /api/permissions API) timestamps are now in seconds, instead of milliseconds. (#2554)
+- The REMS `reset` command line command now works even when you have duplicate resource ids in the database. (#2557)
 
 ### Additions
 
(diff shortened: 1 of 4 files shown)


marharyta (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Changes since v2.16
 ### Fixes
 - Searching for applications by the original REMS generated id works, even if another id has been assigned. (#2564)
 - GA4GH Visa (output by the experimental /api/permissions API) timestamps are now in seconds, instead of milliseconds. (#2554)
+- The REMS `reset` command line command now works even when you have duplicate resource ids in the database. (#2557)
```
Macroz, line 16: Perhaps mention the missing constraint if going back to 2019 versions.

Macroz: I think this line is not exactly honest about what the change was :)

opqdonut: not sure the public changelog is the right place for that, especially since we haven't documented rollbacks at all!

opqdonut: but I can do it if you'd like

Issue #2066, opened 2020-03-19
Form titles in application


Macroz:
- [x] Localized title for form, internal name is separate
- [x] Admin editor for title and name
- [x] Show localized title in application where form is
- [x] Show catalogue item title in application where form is
- [x] Migration for old titles (see e.g. localize-organizations)
- Show organization title in application where the form is? Only for handler?

opqdonut:
is this actually done?

PR #2560, opened 2021-02-12
Form title change

Macroz:
Implements #2066

titles1
titles2

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [ ] Update manual/ (if applicable) (*need to do some updating?*)

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically (*should test deprecated stuff still works?*)

## Follow-up


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Changes since v2.16
 - A new form field type "table" is now available. A table has a predefined set of columns, and applicants can fill in as many rows as they wish. **You can not roll back to an earlier release once your database contains applications with filled-in table fields. You will need to fix the database manually.** (#2551)
 
 ### Changes
+- Form titles are now deprecated. Prefer the new `internal-name` or name and `external-title` or title fields instead. The external title is shown to applicants in the heading of the form and internal name used throughout the administration. The API supports the old style for now. (#2066)
```
opqdonut, line 15: "Form titles are now deprecated" sets this off on the wrong foot I think. How about something like

- Show localized form titles to applicants. The field previously called form `title` is now `internal-name`, and there is a new `external-title` field that is a localized public title for the form. ...

resources/migrations/20210201115807-add-form-data.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE form_template ADD formdata jsonb;
```
opqdonut, line 1: formdata sounds pretty generic: could fields move inside formdata? if not, could we rename formdata to something like metadata?

test/clj/rems/test_pdf.clj
```diff
@@ -125,41 +128,41 @@
...
                   [:table {:header ["First" "Second"]}
                    ["pdf test" "pdf test"]
                    ["pdf test" "pdf test"]]]]
-                [[:paragraph pdf/label-field-style "The following field types can have a max length."]
-                 [:paragraph ""]]
-                [[:paragraph pdf/field-style "Text field with max length"]
-                 [:paragraph "pdf test"]]
-                [[:paragraph pdf/field-style "Text area with max length"]
-                 [:paragraph "pdf test"]]]]
+                 [[:paragraph pdf/field-style "Conditional field. Shown only if first option is selected above."]
```
opqdonut, line 156: you probably mixed up the order of the fields when merging this: table should go after multi-select (see test failure)

test/cljs/rems/administration/test_create_form.cljs
```diff
@@ -353,7 +367,9 @@
                                                                  :label {:en "Bacon"
                                                                          :fi "Pekonia"}}]))]
         (is (= {:organization {:organization/id "abc"}
-                :form/title "the title"
+                :form/internal-name "the name"
+                :form/external-title {:en "en external title"
+                                      :fi "fi external title"}
```
opqdonut, line 372: should there be a test case here for a missing external-title?

opqdonut (commented):
first read-through, didn't notice any glaring problems. will need to reread.

CHANGELOG.md
Macroz: Yeah I can do that. I agree it's a rather strong wording of an actual improvement.

resources/migrations/20210201115807-add-form-data.up.sql
Macroz: I would move stuff inside formdata but I don't want to do it in this PR.

Macroz: I think the gold style for PG + JSON model is to only have the id + potentially some indexed columns + at most one blob.

opqdonut: agreed! follow-up PR sounds good.

test/clj/rems/test_pdf.clj
Macroz: I guess this is the test failure, I fixed the header and then the merge has a problem.

resources/migrations/20210201115807-add-form-data.up.sql
Macroz: I put it here because it's not urgent https://github.com/CSCfi/rems/issues/2561

test/cljs/rems/administration/test_create_form.cljs
Macroz: Not sure if my test makes any sense :shrug:

opqdonut: well at least the behaviour is now explicit

src/clj/rems/api/schema.clj
```diff
@@ -203,13 +203,25 @@
          :field/private s/Bool
          (s/optional-key :field/previous-value) schema-base/FieldValue))
 
+(s/defschema FormData
+  {:form/internal-name s/Str
+   :form/external-title LocalizedString})
```
opqdonut, line 208: I think this belongs in rems.db.form, cf. rems.db.blacklist

src/clj/rems/api/schema.clj
```diff
@@ -203,13 +203,25 @@
...
+
 (s/defschema FormTemplate
   {:form/id s/Int
    :organization OrganizationOverview
-   :form/title s/Str
+   (s/optional-key :form/title) (rjs/field s/Str
+                                           {:deprecate true
+                                            :description "DEPRECATED, will disappear, use either internal name or external title as you need"})
+   :form/internal-name (rjs/field s/Str
+                                  {:description "The internal name of the form only visible to the administration."})
```
opqdonut, line 217: Thanks for the docs! You can also use `(rjs/descibe s/Str "The internal name ...")` if you don't need to set any other metadata.

src/clj/rems/db/form.clj
```diff
@@ -98,15 +112,20 @@
        (validate-fields)
        (json/generate-string)))
 
+(defn- serialize-formdata [form]
+  (json/generate-string form))
```
opqdonut, line 116: we usually validate on write as well, cf. `rems.db.blacklist/event->json`

src/cljc/rems/common/form.cljc
```diff
@@ -237,9 +250,31 @@
...
+                                :fi :t.form.validation/required}}
+         (validate-form-name-fields {} [:en :fi])))
+  (is (= {:form/external-title {:fi :t.form.validation/required}}
+         (validate-form-name-fields {:form/internal-name "foo"
+                                     :form/external-title {:en "Foo"}} [:en :fi])
+         (validate-form-name-fields {:form/internal-name "foo"
+                                     :form/external-title {:en "Foo" :fi ""}} [:en :fi])))
+  (is (= nil
+         (validate-form-name-fields {:form/internal-name "foo"
+                                     :form/external-title {:en "Foo" :fi "Foo"}} [:en :fi]))))
```
opqdonut, line 273: should probably have a case with `:form/title` here as well, right?

opqdonut (approved):
some suggestions I'd like to discuss, but otherwise this looks good to go

src/clj/rems/api/schema.clj
Macroz: I think we didn't quite finish the discussion thread about the best division of schemas so we should perhaps decide something there about this?

src/clj/rems/db/form.clj
Macroz: True, but actually here the two fields come from the API `CreateFormCommand` which is validated so there is not much to add here except double-checking. Should I do it anyway?

opqdonut: I think it's good to have both sides of the validation in place so that we don't forget it when moving the rest of the stuff to `formdata`.

src/clj/rems/api/schema.clj
opqdonut: Yeah #2562 is related to this. However I think since `FormData` only lives in the DB and isn't used in the API it can be moved to the DB namespace for now?

src/clj/rems/db/form.clj
Macroz: Yes, or it could be done there. I'll do it here :+1:

src/clj/rems/api/schema.clj
Macroz: I can do that. For this specifically, maybe the end goal is that there is no need for the `FormData` schema but we can directly use `FormTemplate` or a subset of it.

src/clj/rems/api/schema.clj
Macroz: Added a task to Ideas to refactor this. There seems to be a bunch.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Changes since v2.16
 - A new form field type "table" is now available. A table has a predefined set of columns, and applicants can fill in as many rows as they wish. **You can not roll back to an earlier release once your database contains applications with filled-in table fields. You will need to fix the database manually.** (#2551)
 
 ### Changes
+- Forms now have both an internal name as well as a localized title instead of the non-localized title. The old style form titles are now deprecated in the API and a migration copies the title to the internal name. Prefer the new `internal-name` or `external-title` fields instead. The external title is shown to applicants in the application and internal name used throughout the administration. The API supports the old style for now. (#2066)
```
opqdonut, line 15: hmm... should we add an action point here saying "you should double-check all your forms localized names after running migrations"? should we mention there are migrations in this release? I think we planned to start doing it at some point.

opqdonut (approved):

PR #2559, opened 2021-02-12
create rems.schema-base, document FieldValue

opqdonut:
for #2551, follow-up to #2556

# Checklist for author

Remove items that aren't applicable, check items that are done.

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible or completely new
- [x] Events are backwards compatible _or_ have migrations

## Documentation
- [x] API is documented and shows up in Swagger UI

## Testing
- [x] Valuable features are integration / browser / acceptance tested automatically

PR #2562, opened 2021-02-15
refactor: move organization schemas to rems.schema-base

opqdonut:
to avoid depending on rems.api.schema from rems.db.*

follow-up to #2559

# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/docs/architecture/012-layers.md
+++ b/docs/architecture/012-layers.md
@@ -136,7 +136,7 @@ when they are shared.
 
 In particular:
 
-- `rems.api.schema` should only be used from `rems.api.*` namespaces. It should be used for common schema fragments that many APIs share. Also big or importat schemas can live in `rems.api.schema` for discoverability.
+- `rems.api.schema` should only be used from `rems.api.*` namespaces. It should be used for common schema fragments that many APIs share. Also big or important schemas can live in `rems.api.schema` for discoverability.
 - `rems.api.*` namespaces that define APIs can also contain schemas for those APIs, if the schemas don't need to be shared.
 - `rems.schema-base` contains common schemas that can be used everywhere.
 - `rems.schema-base` and `rems.api.schema` shouldn't depend on any api, db or service components.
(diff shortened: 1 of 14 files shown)


docs/architecture/012-layers.md
```diff
@@ -124,3 +124,20 @@ Notable pieces of code that don't fit into the API-Services-DB layering (and don
...
+Date: 2021-02-15
+Authors: @opqdonut @Macroz
+
+This layering should also apply to schemas. We want to make it
+possible to separate DB schemas from API schemas, and make it apparent
+when they are shared.
+
+In particular:
+
+- `rems.api.schema` should only be used from `rems.api.*` namespaces. It should be used for common schema fragments that many APIs share. Also big or importat schemas can live in `rems.api.schema` for discoverability.
```
Macroz, line 139: Typo: `importat`

src/clj/rems/schema_base.clj
```diff
@@ -35,3 +54,27 @@
...
+
+(s/defschema OrganizationFull
+  (merge OrganizationOverview
+         {(s/optional-key :organization/modifier) UserWithAttributes
+          (s/optional-key :organization/last-modified) DateTime
+          (s/optional-key :organization/owners) [UserWithAttributes]
+          (s/optional-key :organization/review-emails) [{:name LocalizedString
+                                                         :email s/Str}]
+          (s/optional-key :enabled) s/Bool
+          (s/optional-key :archived) s/Bool}))
```
Macroz, line 80: Should this technically be part of organization API or here?

Macroz (approved):
Looks good, like what we decided.

src/clj/rems/schema_base.clj
opqdonut: OrganizationFull and OrganizationOverview are used in rems.db.organizations, which is why I moved them here. I guess they could also be in the db ns?

Macroz: Since Organization is the latest, it has pretty much identical DB model as API model. I guess this is fine too though I would expect that the `schema-base` would remain "slim" while the DB and API are the "fullest".

opqdonut: Yeah I'd prefer a slim `schema-base` too. I'm not that pleased with how these ended up here, but it's better than `rems.api.schema`. It makes the API sharing visible, at least.

Issue #2551, opened 2021-02-08
Simple table component


opqdonut:
Implement simple table form field type. Split out from #2404.

- configurable number of columns with configurable names
- applicant can add as many rows as they want
- question: how to store the field answers?
  - option 1: CSV as string
  - option 2: JSON string
  - option 3: allow JSON objects in `field/value`. Will require schema & backend refactors.

- [x] schema & backend
- [x] applicant UI
- [x] form editor
- [x] while implementing simple table, keep in mind possibility of adding column types (= validations)
- consider adding column types (email, text) to simple table immediately, if it is easy to do so

leftovers from first PR (#2556)
- [x] factor out the answer schema to `rems.schema-base`
- [x] think about what required/optional means for a table field
- [x] update manual
- [x] styling improvements: make add row button less wide, fix spacing of remove button
- [x] fix table diff when rows have been removed
- [x] use "require at least one row" instead of "optional" in the create form UI
- [x] pre-add a row for required table fields
- [x] require all columns to have values in all rows (even if field itself is optional)

PR #2567, opened 2021-02-15
feat: require all table cells to be non-empty

opqdonut:
for #2551, follow-up to #2556

# Checklist for author

## Reviewability
- [x] Link to issue

## Testing


opqdonut:
table-validation

src/clj/rems/form_validation.clj
```diff
@@ -4,16 +4,28 @@
             [rems.common.form :as form]
             [rems.common.util :refer [+email-regex+]]))
 
+(defn- all-columns-set? [field]
+  (let [valid-row? (fn [cells] (every? #(not (str/blank? (:value %))) cells))]
```
Macroz, line 8: `not-any?`? I think it reads slightly better here.

Macroz (approved):
I think the validation functions vs. predicates was a good call :+1:

src/clj/rems/form_validation.clj
opqdonut: yeah!

PR #2568, opened 2021-02-15
style: remove extra padding from navbar logo

marharyta:
# Checklist for author

## Reviewability

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -103,7 +103,9 @@
   (list
    [:.logo-menu {:height logo-height-menu
                  :background-color (get-theme-attribute :logo-bgcolor)
-                 :width "100px"}]
+                 :width "100px"
+                 :padding-top "0px"
+                 :padding-bottom "0px"}]
    [:.logo {:height logo-height
             :background-color (get-theme-attribute :logo-bgcolor)
             :width "100%"


src/clj/rems/css/styles.clj
```diff
@@ -35,7 +35,7 @@
       (or attr-value (recur (rest attr-names))))))
 
 (def content-width (u/px 1200))
-(def logo-height-menu (u/px 40))
+(def logo-height-menu (u/px 50))
```
opqdonut, line 38: So you both remove the padding and make it a big higher? Ok. Please add a note to the changelog so that users will know to double-check their theme when upgrading.

opqdonut (changes requested):

src/clj/rems/css/styles.clj
marharyta: That is essentially what was in the screenshot

marharyta: I can do that, sure

Macroz: Why is this increased? There is no need for it?

Macroz: Also if you increase the logo height should you not increase the menu height too or some content like LBR logo can be cropped?

marharyta: done

marharyta: @Macroz "There is no need for it?"?
I have demoed the screenshot in the rems chat

Macroz: Well here is my thinking.

I thought you said you were going to just quickly remove the unnecessary padding, but now I see you also increase the height of the logo forcing an increase in the size of the navbar. I don't see you taking into account the increased height of the navbar in the rest of the code though. Did you check that LBR theme works right after this because we have extra padding for the body so the navbar won't go over the page contents/hero logo?

I think we don't need to increase this logo height here at all, but should soncider it and any further changes with taking into account the needs to customize the logo height, customize the logo paddings, making a more responsive navbar, fixing navbar layout etc. I.e. what I suggested in for a good ticket on further tuning. For a quick improvement I'm saying we don't need this.

marharyta: Screenshot 2021-02-17 at 10 45 11

Macroz: Here is also in a picture what I mean about LBR and any theme with similar setup because the padding was not adjusted. Left is before, right is after height increase.
navbar_height_problem

marharyta: Ok, so I have double checked and the increase in menu does not seem to break anything

marharyta: Right, Ok, I will remove it

opqdonut (approved):

PR #2579, opened 2021-03-01
form validation fixes

opqdonut:
fixes #2578

# Checklist for author

Remove items that aren't applicable, check items that are done.

## Reviewability
- [x] Link to issue

## Documentation
- [ ] Update changelog if necessary
  - fixes mostly relate to features that are already in the changelog

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically
  - the validation code is at most smoke-tested in browser tests, we
  don't verify e.g. the focus logic

Issue #2578, opened 2021-03-01
misc. validation fixes


opqdonut:
follow-up to #2551 #2564 #2560 
- [x] fix "all columns must have values" validation message shown to applicant
  - [x] make the link work
  - [x] show the column name
- [x] form editor: fix the validation links for table column fields
- [x] form editor: fix JS crash when trying to save empty form (no name, no org, no title, no fields)

PR #2580, opened 2021-03-01
more consistent organization-field

opqdonut:
a follow-up to #2579, related to #2578

- show validation errors next to field in create-form
- use rems.administration.components infrastructure

# Checklist for author

## Reviewability
- [x] Link to issue

## Testing

### DIFF
--- a/src/cljs/rems/administration/components.cljs
+++ b/src/cljs/rems/administration/components.cljs
@@ -16,7 +16,10 @@
   (:require [clojure.string :as str]
             [re-frame.core :as rf]
             [rems.atoms :refer [info-field textarea]]
-            [rems.text :refer [text-format]]))
+            [rems.dropdown :as dropdown]
+            [rems.fields :as fields]
+            [rems.common.roles :as roles]
+            [rems.text :refer [text text-format]]))
 
 (defn- key-to-id [key]
   (if (number? key)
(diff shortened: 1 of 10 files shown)


src/cljs/rems/administration/components.cljs
```diff
@@ -29,7 +32,12 @@
        (str/join "-")))
 
 (defn- field-validation-message [error label]
-  [:div {:class "invalid-feedback"}
+  ;; XXX: Bootstrap's has "display: none" on .invalid-feedback by default
+  ;;      and overrides that for example when there is a sibling .form-control.is-invalid,
+  ;;      but that doesn't work with checkbox groups, dropdowns, etc., and we anyways
+  ;;      don't need the feature of hiding this div with CSS when it has no content.
+  [:div {:class "invalid-feedback"
+         :style {:display "block"}}
```
marharyta, line 40: not a fan of inline styling, I personally think this creates unnecessary "pollution" in the style definition. We can just have ```.invalid-feedback``` class in the styles file

opqdonut: Yeah I guess we could move this hack there, especially now that we have two instances of it (see fields.cljs:161)

src/cljs/rems/administration/create_workflow.cljs
```diff
@@ -15,7 +16,7 @@
             [rems.util :refer [navigate! post! put! trim-when-string]]))
 
 (defn- item-by-id [items id-key id]
-  (medley.core/find-first #(= (id-key %) id) items))
+  (medley/find-first #(= (id-key %) id) items))
```
Macroz, line 19: I think the approach we chose was to use `:refer` for Medley because it's as good as `core`.

Macroz (approved):

PR #2582, opened 2021-03-03
don't send or accept answers for invisible fields

### LINKED ISSUE #2574: hiding answer to invisible fields
Findata reports that their integrations broke because answers to invisible fields were sent by the API. This is the current behaviour of rems:

1. create a form with
    - an option field A, with options Y and N
    - a text field B, that is only visible if A is Y
2. create a catalogue item using this form, and create an application for it
3. save a draft with A=Y, B=secret
4. save a draft with A=N, the field B is hidden in the UI
5. fetch the form via the API. You get:
```
{"field/id": "A",
 "field/visible": true,
 "field/value": "N",
 ...},
{"field/id": "B",
 "field/visible": false,
 "field/value": "secret",
 ...}
```

There are multiple directions to go from here. We might even want to do all of these.

1. don't send (or actively clear) answers to invisible fields in the frontend
2. don't accept or don't store answers to invisible fields in the backend
3. strip out answers to invisible fields when fetching an application from the API
4. strip out invisible fields when fetching an application from the API

However, we should retain the current UX:

- [ ] while an applicant edits an application, don't lose answers to invisible fields, but retain them in case the field is made visible again (before saving)

opqdonut:
fixes #2574

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Changes since v2.16
   - Prefer the new `internal-name` or `external-title` fields instead.
   - The external title is shown to applicants in the application and internal name used throughout the administration.
   - The API supports the old style for now.
+- Answers to conditional fields that are not visible are no accepted by the API, nor sent by the REMS UI. (#2574)
 
 ### Fixes
 - Searching for applications by the original REMS generated id works, even if another id has been assigned. (#2564)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes since v2.16
   - Prefer the new `internal-name` or `external-title` fields instead.
   - The external title is shown to applicants in the application and internal name used throughout the administration.
   - The API supports the old style for now.
-- Answers to conditional fields that are not visible are no longer accepted by the API, nor sent by the REMS UI. (#2574)
+- Answers to conditional fields that are not visible are no longer stored by REMS. The API accepts answers for invisible fields but drops them. The UI does not send answers to invisible fields. (#2574)
 
 ### Fixes
 - Searching for applications by the original REMS generated id works, even if another id has been assigned. (#2564)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Changes since v2.16
   - Prefer the new `internal-name` or `external-title` fields instead.
   - The external title is shown to applicants in the application and internal name used throughout the administration.
   - The API supports the old style for now.
-- Answers to conditional fields that are not visible are no accepted by the API, nor sent by the REMS UI. (#2574)
+- Answers to conditional fields that are not visible are no longer accepted by the API, nor sent by the REMS UI. (#2574)
 
 ### Fixes
 - Searching for applications by the original REMS generated id works, even if another id has been assigned. (#2564)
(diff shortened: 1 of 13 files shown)


src/clj/rems/form_validation.clj
```diff
@@ -62,7 +62,7 @@
       ;; Schema validation guarantees that it's either a s/Str or
       ;; a [[{:column s/Str :value s/Str}]], and we've ruled out s/Str
       ;; in wrong-value-type-error
-      (when (not (every? row-ok? (:field/value field)))
+      (when (not (every? row-ok? value))
```
Macroz, line 66: How about using `when-not` or inverting `row-ok?` to `invalid-row?` and using `some`?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Changes since v2.16
   - Prefer the new `internal-name` or `external-title` fields instead.
   - The external title is shown to applicants in the application and internal name used throughout the administration.
   - The API supports the old style for now.
+- Answers to conditional fields that are not visible are no accepted by the API, nor sent by the REMS UI. (#2574)
```
Macroz, line 22: I think this is missing something, perhaps `are no longer`?

Macroz (approved):
Good catch of the corner cases. 

I am wondering from the API design standpoint that would it actually make more sense to keep accepting the invisible answers as is, but just ignore them i.e. never save them, rather than to validate the presence and cause validation errors? I mean now the person submitting must always know the combination of visible fields and therefore must duplicate REMS logic closely. I think the main goal was just to make sure answers that are not used are not stored which can be met with either solution. The one that does not validate may be confusing but is lenient.

src/clj/rems/form_validation.clj
opqdonut: I prefer pushing the negations as "out" as possible so that you don't need to keep tracking if something is positive or negative while reading. So I prefer `(every? row-ok? value)`. I'll switch to a `when-not` though.

opqdonut:
Good question about API design. I agree that duplicating the visible logic is a bit complicated, but on the other hand if we drop invisible answers the caller should _anyway_ duplicate the visibility logic to avoid losing data.

I'm not against switching to silently dropping invisible answers.

src/clj/rems/form_validation.clj
Macroz: Well in this case "some invalid" reads to me as "positive" all the way through both are fine for me :+1: .

test/clj/rems/test_form_validation.clj
```diff
@@ -218,6 +234,18 @@
                                            :field/visible  false
                                            :field/value    ""}]))))
 
+  (testing "invisible fields should not be filled"
```
marharyta, line 237: both tests suits have the same name? ```"invisible fields should not be filled"```

opqdonut: Heh, good catch! The first one is supposed to test `validate-fields-for-submit` and the second one `validate-fields-for-draft` but it looks like they both test `-draft`. I'll fix it.

marharyta: 👍🏻

marharyta: Otherwise, it looks alright, I think that form logic is complex in any case so suggestions to simplify hardly would make any sense coming from me

marharyta (approved):

Macroz (approved):

PR #2585, opened 2021-03-04
refactoring: create style utils namespace

marharyta:
Some minor style file refactoring. The rationale:
* Currently, the style file is rather long, and adding more styles becomes confusing
* I managed to find a couple of styles that are not applied (not tackled in this PR) that would need to be further investigated
* style utils can be a good place for functions that provide extra style benefits such as
    * get theme styles
    * style helpers that are only used once in the stylesheet


src/clj/rems/css/style_utils.clj
```diff
@@ -0,0 +1,79 @@
+(ns rems.css.style-utils
+  "CSS stylesheets are generated by garden automatically when
+  accessing the application on a browser. The garden styles can also
+  be manually compiled by calling the function `rems.css.styles/screen-css`.
+
+  For development purposes with Figwheel the styles are rendered to
+  `target/resources/public/css/:language/screen.css` whenever this ns is evaluated
+  so that Figwheel can autoreload them."
```
opqdonut, line 8: Isn't this just a repetition of the `rems.css.styles` docstring?

src/clj/rems/css/style_utils.clj
```diff
@@ -0,0 +1,79 @@
...
+            [garden.core :as g]
+            [garden.def :refer [defkeyframes]]
+            [garden.selectors :as s]
+            [garden.stylesheet :as stylesheet]
+            [garden.units :as u]
+            [medley.core :refer [map-vals remove-vals]]
+            [mount.core :as mount]
+            [rems.config :refer [env]]
+            [rems.context :as context]
+            [ring.util.response :as response]))
```
opqdonut, line 23: It looks like almost none of these requires are used. In case you want some tooling for this, the clj-kondo linter can point out unused required. An other option is the slamhound leiningen plugin, which can automatically remove unused requires from namespaces.

src/clj/rems/css/style_utils.clj
```diff
@@ -0,0 +1,79 @@
...
+(defn get-logo-name-sm [lang]
+  (resolve-image (get-theme-attribute (keyword (str "logo-name-" (name lang) "-sm")) :logo-name-sm)))
+
+(defn get-navbar-logo [lang]
+  (resolve-image (get-theme-attribute (keyword (str "navbar-logo-name-" (name lang))) :navbar-logo-name)))
+
+
+;; Fonts of the app 
+
+(defn generate-at-font-faces []
```
opqdonut, line 59: This thing kind of feels like a style, and not like a style util, so I kinda feel it belongs in rems.css.styles. It's not a big deal so it can stay here too.

src/clj/rems/css/styles.clj
```diff
@@ -20,63 +20,14 @@
             [mount.core :as mount]
             [rems.config :refer [env]]
             [rems.context :as context]
+            [rems.css.style-utils :refer [get-theme-attribute get-logo-image generate-at-font-faces get-navbar-logo get-logo-name-sm]]
```
opqdonut, line 23: I like this refactoring. If we can manage to clean up rems.css.styles so that it only contains the actual styles it would make the codebase a bit cleaner. This is a step in that direction :+1:

opqdonut (changes requested):
Please fix the ns form, otherwise good to go.

src/clj/rems/css/style_utils.clj
Macroz: This is not a style util IMHO too.

src/clj/rems/css/styles.clj
Macroz: I think this is a bit premature split but overall one util file is ok.

src/clj/rems/css/style_utils.clj
marharyta: I also thought about it, actually, but util is a rather fluid term, especially in CSS. the way I see it, it is only used once per codebase as we have just concluded on flowdock, we came to a conclusion that typogrphy is configurable to an extend, so it actually should be a util. I hope I was able to articulate myself clearly.

src/clj/rems/css/styles.clj
marharyta: LOL

src/clj/rems/css/style_utils.clj
marharyta: The slamhound leiningen plugin is an excellent suggestion!

src/clj/rems/css/style_utils.clj
Macroz: No I do not agree that we reached a conclusion that this definition should be in the utils file.

marharyta: photo_2021-03-04 11 39 45

marharyta: So what should I do then @opqdonut @Macroz ? I think me and @opqdonut are now in agreement but @Macroz does not agree?

Macroz: I would expect the issue texts updated on the board or a summary of your meeting, I guess. I was specifically reacting to this statement that in the previous chat we would have all agreed to put the font definitions into the util file. I think that is misusing the term util and it's just not the right thing to do.

marharyta: Summary of my meeting?

Macroz: You had a meeting with @opqdonut about the refactorings to the styles of which I don't know the results yet. I'm not sure if your agreement refers to that or our chat where I think we did not reach a conclusion about the fonts. I would expect the issues to be updated based on these.

opqdonut (approved):

src/clj/rems/css/style_utils.clj
opqdonut: We didn't talk about this exact issue (where to put the font-face definitions). I think it's fine here, especially since it isn't an active part of styling, more like styling infra.

marharyta:
ping @Macroz

example-theme/theme.edn
```diff
@@ -1,4 +1,4 @@
-{:font-family "sans-serif"
+{:font-family "'Lato', sans-serif"
```
Macroz, line 1: Should this be mentioned in the changelog?

test/clj/rems/test_themes.clj
```diff
@@ -35,6 +35,6 @@
 (deftest get-theme-attribute-test
   (with-redefs [config/env {:theme {:test "success"
                                     :test-color 2}}]
-    (is (= 2 (styles/get-theme-attribute :test-color)))
-    (is (= "success" (styles/get-theme-attribute :test)))
-    (is (nil? (styles/get-theme-attribute :no-such-attribute)))))
+    (is (= 2 (style-utils/get-theme-attribute :test-color)))
+    (is (= "success" (style-utils/get-theme-attribute :test)))
+    (is (nil? (style-utils/get-theme-attribute :no-such-attribute)))))
```
Macroz, line 40: Should these util tests be now in their own file / in the style utils file? Could be in a future PR.

opqdonut: This is a synthetic test namespace anyway. I mean, there's no rems.themes namespace that this tests, it tests the concept of themes. So I'm fine with them being here.

example-theme/theme.edn
marharyta: Like "new example font is Lato"?

opqdonut: Yeah, "The example-theme now uses the Lato font"

test/clj/rems/test_themes.clj
Macroz: Yeah works for me. This kind of unit test can now be moved without adding burden to the style file.

example-theme/theme.edn
marharyta: this has been addressed

opqdonut (approved):

PR #2591, opened 2021-03-08
small form-related refactorings

### LINKED ISSUE #2574: hiding answer to invisible fields

opqdonut:
we were going to hide invisible answers when fetching applications, but it turned out not to be necessary

closes #2574

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
  - It's covered by the previous entry

## Testing

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -408,18 +408,10 @@
          (localization-for :title {:localizations {:en {}
                                                    :fi {}}}))))
 
-(defn- add-default-value [field]
-  (assoc field :field/value
-         (case (:field/type field)
-           :table []
-           ;; default
-           "")))
-
 (defn- enrich-form [form get-form-template]
-  (let [form-template (get-form-template (:form/id form))
-        default-fields (map add-default-value (:form/fields form-template))
+  (let [form-template (form/add-default-values (get-form-template (:form/id form)))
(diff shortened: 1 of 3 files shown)


opqdonut:
this might not be what we want: perhaps we need to hide the fields completely. investigating...

Macroz (approved):

Issue #2453, opened 2020-11-23
Only download current field attachments via "All attachments (zip)"


jaakkocsc:
- [ ] add a `?current-field-values-only` query parameter to the `/api/application/:id/attachments` endpoint
  - when this option is enabled, the produced zip should only contain the current values of attachment fields and not previous values, or comment attachments
- [ ] use this option in the UI zip button
- [ ] consider rewording the button

opqdonut:
I had a quick look at the code and it looks like we copy the attachments one-by-one, and only copy the attachments used in field answers. So at least the code tries to do the right thing.

opqdonut:
Original bug report from Findata

> Kun hakija on kopioi vanhan hakemuksen uudeksi hakemukseksi REMSissä, uudelle hakemuslomakkeelle siirtyvät ilmeisesti myös vanhan hakemuksen liitteet. Kollegani Hanna huomasi juuri, että vaikka hakija on tällaisessa tapauksessa poistanut edellisen hakemuksen liitteet ja korvannut ne uusilla (kyseisen hakemuksen omilla), edellisen hakemuksen liitteet ovat kuitenkin jääneet jotenkin hakemukseen mukaan, sillä kun hakemuksen käsittelijä ottaa kaikki hakemuksen liitteet zippinä ulos, mukana tulevat sekä uuden että vanhan hakemuksen liitteet.

opqdonut:
Copying an application also copies the attachment, by design. However, after changing the attachment and submitting the application, the handler sees both attachments via the "show changes" feature. This is the real bug.

opqdonut:
Oh ok, that's by design too. #832 specifies:

> handler can diff changes compared to the old application

opqdonut:
There are some corner cases here that don't work quite right. For example:
- create new draft
- add attachment blue.png
- copy as new application
- change attachment in new application to red.png
- submit new application
- now handler sees blue & red.png

The line that seeds the diff for a copied application is https://github.com/CSCfi/rems/blob/de3fa7a8c63f61703c54b4ac6d647fe7ff8e5471/src/clj/rems/application/model.clj#L228

opqdonut:
Let's find out what the user requirements are for the zip feature and remove unnecessary clutter from the zip. We can leave the current "full zip" functionality in the API for archiving use.

opqdonut:
Rewrote ticket description

opqdonut:
[...]

PR #2593, opened 2021-03-09
add :attachment/occurs-in field to API

### LINKED ISSUE #2571: categories for attachments
Attachments stored in REMS should have a category attached to them (attachment to an application, attachment to a comment, attachment to a decision, etc.). This would make it easier for integrations to pick just certain attachments.

opqdonut:
closes #2571

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/schema.clj
+++ b/src/clj/rems/api/schema.clj
@@ -196,7 +196,8 @@
 (s/defschema ApplicationAttachment
   {:attachment/id s/Int
    :attachment/filename s/Str
-   :attachment/type s/Str})
+   :attachment/type s/Str
+   :attachment/occurs-in (s/maybe #{(s/enum :event :value :previous-value)})})
 
 (s/defschema BlacklistEntry
   {:blacklist/user schema-base/UserWithAttributes
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -488,6 +488,40 @@
 
              {}))))
 
+(defn- classify-attachments [application]
+  (let [from-events (for [event (:application/events application)
+                          attachment (:event/attachments event)]
+                      {(:attachment/id attachment) #{:event}})
+        from-fields (for [form (getx application :application/forms)
+                          field (getx form :form/fields)
+                          :when (= :attachment (:field/type field))
+                          k [:field/value :field/previous-value]
+                          id (form/parse-attachment-ids (get field k))]
+                      {id #{(case k
+                              :field/value :value
+                              :field/previous-value :previous-value)}})]
+    (apply merge-with set/union (concat from-events from-fields))))
+
+(deftest test-classify-attachments
+  (let [application {:application/events [{:event/type :application.event/foo
+                                           :event/attachments [{:attachment/id 1}
+                                                               {:attachment/id 3}]}
+                                          {:event/type :application.event/bar}]
+                     :application/forms [{:form/fields [{:field/type :attachment
+                                                         :field/value "5" :field/previous-value "5,1,15"}]}
+                                         {:form/fields [{:field/type :text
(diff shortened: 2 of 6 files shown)


src/clj/rems/api/schema.clj
```diff
@@ -196,7 +196,8 @@
 (s/defschema ApplicationAttachment
   {:attachment/id s/Int
    :attachment/filename s/Str
-   :attachment/type s/Str})
+   :attachment/type s/Str
+   :attachment/occurs-in (s/maybe #{(s/enum :event :value :previous-value)})})
```
Macroz, line 200: Would it make sense to also store the event ids where the attachment occurred?

src/clj/rems/application/model.clj
```diff
@@ -488,6 +488,46 @@
...
+                          attachment (:event/attachments event)]
+                      {(:attachment/id attachment) #{:event}})
+        from-fields (for [form (getx application :application/forms)
+                          field (getx form :form/fields)
+                          :when (= :attachment (:field/type field))
+                          k [:field/value :field/previous-value]
+                          id (form/parse-attachment-ids (get field k))]
+                      {id #{(case k
+                              :field/value :value
+                              :field/previous-value :previous-value)}})]
```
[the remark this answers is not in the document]
opqdonut: you mean `#{:field/value :field/previous-value :event/attachment}` instead of `#{:value :previous-value :event}`?

src/clj/rems/api/schema.clj
opqdonut: Perhaps, but that would make for a more complicated schema. Besides, you won't know what the event id number means without walking all the events, at which point you can just check where the attachment occurs anyway. I think the usecase for this is something like "grab all :value attachments and upload them to our CRM"

src/clj/rems/application/model.clj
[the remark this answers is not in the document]
Macroz: Yes just to avoid unnecessary mapping code.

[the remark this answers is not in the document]
opqdonut: I was trying to figure out what a nice simple API would look like for the consumer, but I guess I'm fine with the namespaced keywords as well.

src/clj/rems/api/schema.clj
Macroz: Do we actually know enough about the use cases to do any concrete API design? Do we need to expose this information in the API? Can't we do the download as zip even without this?

opqdonut: Yeah, we could only keep this internally, but there was an actual request for this, see #2571. I guess some integrations want to duplicate the "all attachments" functionality without going via a zip?

Macroz: Yes but with that level of specification I'm just afraid we'll miss and end up an API that keeps breaking. Like "attachment to a decision" could mean a lot of things and this does not help to answer that.

opqdonut: Yeah we discussed the granularity of classifications with Jaakko on flowdock and ended up with what's in this PR now. I don't think it makes sense to make this more fine-grained. At that point the users could just walk the application structure anyway.

Macroz: Yeah, I was just saying that perhaps we can skip the public API altogether. The issues we have don't require it in my mind. The integrators could actually use examples in the docs?

opqdonut:
need to wait for input on whether this feature is needed in the API (or only internally)

opqdonut:
let's not expose this in the API for now, but let's implement #2453

opqdonut:
see #2651 instead

PR #2595, opened 2021-03-10
refactor: simplify cljs macro requires

opqdonut:
Requiring macros directly (without `:require-macros` etc) tends to just work in cljs these days.


marharyta (approved):

src/clj/rems/guide_utils.clj
```diff
@@ -1,7 +1,6 @@
-(ns rems.guide-macros
-  "Utilities for component guide."
+(ns rems.guide-utils
```
Macroz, line 1: Why plural? The others are `-util`?

src/clj/rems/guide_utils.clj
```diff
@@ -11,21 +10,18 @@
...
 (defmacro component-info [component]
   ;; TODO we'd like to use normalize-file-path on the component
   ;; metadata at compile time, just like we do for namespaces above.
   ;; However it seems that there's no way to get the metadata at
   ;; compile time (e.g. resolve and ns-resolve return nil). Thus we
-  ;; normalize the path at runtime in rems.guide-functions/link-to-source.
+  ;; normalize the path at runtime in link-to-source.
   `(let [m# (meta (var ~component))]
-     (rems.guide-functions/render-component-info
+     (render-component-info
```
Macroz, line 24: Hmm so this works these days?

Macroz (approved):
Slightly baffling but if this is the modern way let's go for it!

I think I would rename from plural to singular to match the others.

src/clj/rems/guide_utils.clj
opqdonut: Yeah that was a leftover from the previous `guide-utils` namespace. I'll fix it!

src/clj/rems/guide_utils.clj
opqdonut: It's just normal macro behaviour where the vars get qualified with the current namespace name, which happens to be the name of the corresponding cljs namespace where these functions are defined...

Macroz: Yeah, you moved these into the same ns though not to the same file.

Issue #2466, opened 2020-12-03
Push permissions to EGA


opqdonut:
Plan
- [x] store handler api key in user secrets (new table)
- [x] show that a handler has an api key in the UI, on the profile page
  - don't send the actual api key to the frontend
  - [x] button for fetching a new api key
  - [x] button for deleting api key
- [x] when handler logs in, and there is no api key but EGA is enabled, show message
- [x] store handler of the entitlement into the entitlements table
- possibly store decider into the entitlements table?
- [x] use handler's api key to authenticate Visa push to EGA in the entitlement poller, also need to update/delete
- [x] approver bot support for entitlement push (configured through command-line)
- [x] configure test environment
- [x] More info support for "EGAD..." resources.
- [x] Handler must generate an API-Key from Profile page
- [x] Manual update for handlers
- [x] write ADR (check / link to ga4gh.md that has EGA info)

[Final specs](https://docs.google.com/document/d/1FTzUYAfV5d2a0zoDkbY9Iy_L5NbSAnHeWnmY2NIrY8M/edit)

[Specs for the permission push](https://docs.google.com/document/d/1kPeoleBkQuAhuYzr0n5BXE--eUQxyANuvwEV7JUfTfU/edit)

[Swagger for the permission push](https://ega.ebi.ac.uk:8053/ega-permissions/swagger-ui.html#/)

jaakkocsc:
There is now an endpoint which can be used for testing, swagger for it is in [here](https://ega.ebi.ac.uk:8053/ega-permissions/swagger-ui.html#/)

Macroz:
Let's wait for a while for the authentication method to resolve.

Macroz:
I updated the ticket description based on today's coffee call.

Macroz:
Some leftovers were added as new tasks. See https://github.com/CSCfi/rems/issues/2687,  https://github.com/CSCfi/rems/issues/2686 and https://github.com/CSCfi/rems/issues/2684.

Macroz:
Let's merge this after a small 2.20 release.

Macroz:
Merged.

PR #2602, opened 2021-03-18
EGA basics

Macroz:
# Checklist for author

For #2466.

Everything that is implemented so far for early comments and feedback. I think we need a set of features required for MVP demo. I implemented some of the UI too but it's not required in the first demo (where Approver Bot needs special features that are not here yet).

## Reviewability
- [x] Link to issue
- [ ] Note if related change in rems-deploy repo TBD
- [ ] Consider adding screenshots for ease of review TBD

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary TBD
- [ ] Update docs/ (if applicable) TBD
- [ ] Update manual/ (if applicable) TBD
- [ ] ADR for major architectural decisions or experiments TBD

## Testing
- [x] Valuable features are integration / browser / acceptance tested automatically (more TBD)

## Follow-up
- [ ] New tasks are created for pending or remaining tasks TBD


src/clj/rems/db/entitlements.clj
```diff
@@ -128,7 +145,10 @@
                            :resource resource-id
                            :revokedby actor
                            :end end})
-    (add-to-outbox! :remove (db/get-entitlements {:application application-id :user user-id :resource resource-id}))))
+    (let [entitlements (db/get-entitlements {:application application-id :user user-id :resource resource-id})]
+      (add-to-outbox! :remove :basic entitlements nil)
```
opqdonut, line 149: could maybe add a TODO about deprecating the :basic functionality

src/clj/rems/db/user_secrets.clj
```diff
@@ -0,0 +1,45 @@
...
+        coerce-user-secrets)))
+
+(defn get-user-secrets [user]
+  (when user
+    (json->secrets (:secrets (db/get-user-secrets {:user user})))))
+
+(defn update-user-secrets! [user new-secrets]
+  (assert user "User missing!")
+  (let [old-secrets (get-user-secrets user)
+        validated (try (validate-user-secrets new-secrets) (catch Exception _e nil))]
```
opqdonut, line 34: hmm shouldn't you validate `(merge old-secrets new-secrets)` ?

src/clj/rems/db/user_secrets.clj
```diff
@@ -0,0 +1,45 @@
...
+    (if (= (set (keys validated))
+           (set (keys new-secrets)))
```
opqdonut, line 36: is this keyset check needed? doesn't schema catch it for us?

src/clj/rems/ega.clj
```diff
@@ -0,0 +1,226 @@
...
+      :body
+      json/parse-string))
+
+(defn api-key-list
+  "List the API-Keys available.
+
+  `:access-token` - the access token
+  `:config`       - configuration of the EGA integration
+
+  NB: The token itself cannot be returned anymore."
```
opqdonut, line 66: I'm not sure what this means

src/clj/rems/ega.clj
```diff
@@ -0,0 +1,226 @@
...
+  [{:keys [access-token config]}]
+  (-> (http/get (str (:permission-server-url config) "/api_key")
+                {:oauth-token access-token
+                 :socket-timeout 2500
+                 :conn-timeout 2500})
+      :body
+      json/parse-string))
+
+(defn read-permissions
```
opqdonut, line 75: Inconsistency: This and the next function are prefixed `read-` even though it's not in the api path. Other GET requests aren't prefixed `read-`.

src/clj/rems/ega.clj
```diff
@@ -0,0 +1,226 @@
...
+  "Lists the users with permission to the dataset.
+
+  `:api-key`    - valid API-Key of the person asking
+  `:dataset-id` - id of the dataset
+  `:config`     - configuration of the EGA integration"
+  [{:keys [api-key dataset-id config]}]
+  (-> (http/get (str (:permission-server-url config) "/datasets/" dataset-id "/users")
+                {:headers {"authorization" (str "api-key " api-key)}
+                 :socket-timeout 2500
+                 :conn-timeout 2500})
```
opqdonut, line 117: should these repeating timeouts be factored out?

src/clj/rems/ega.clj
```diff
@@ -0,0 +1,226 @@
...
+      :body
+      json/parse-string))
```
opqdonut, line 119: I think you can just pass `:as :json` to `http/get`.

src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -1,9 +1,33 @@
...
 
-(defn urn? [resid]
+(defn- urn? [resid]
   (and resid (str/starts-with? resid "urn:")))
 
-(defn urn-catalogue-item-link [{:keys [resid]} {:keys [urn-organization]}]
+(defn- urn-catalogue-item-url [resid {:keys [urn-organization]}]
   (when (urn? resid)
     (str (or urn-organization "http://urn.fi/") resid)))
+
```
opqdonut, line 11: the EGA url stuff could be split out to a separate PR and just merged

src/clj/rems/db/user_settings.clj
```diff
@@ -1,21 +1,22 @@
...
   {:language (:default-language env)
    :notification-email nil})
 
 ;; TODO should this be in schema-base?
 (s/defschema UserSettings
-  {:language s/Keyword
-   :notification-email (s/maybe s/Str)})
+  {(s/optional-key :language) s/Keyword
+   (s/optional-key :notification-email) (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
```
opqdonut, line 19: This refactor could be split out to its own PR and discussed separately. I'm slightly concerned that we won't guarantee the presence of any of these keys, even though the UI expects `:language` and `:notification-email`.

opqdonut (commented):
first round of comments

src/clj/rems/ega.clj
Macroz: I mean that the API-key cannot be fetched from the API, it is only seen once when generating it.

src/clj/rems/db/user_secrets.clj
Macroz: Yeah I think it's not needed, it is legacy from user settings.

src/clj/rems/db/user_secrets.clj
Macroz: :+1: I'll rewrite that part.

src/clj/rems/db/entitlements.clj
Macroz: :+1: It's in my notes that I'll probably put in the issue or actually for this create a new one.

src/clj/rems/ega.clj
opqdonut: I like naming the functions according to the API endpoints, and I think some sort of get- / read- / fetch- prefix might work here, but it needs to be applied consistently.

src/clj/rems/ega.clj
Macroz: How about? `NB: The actual API-Key is not returned by this call. It is only returned once in `api-key-generate`.`

Macroz: Could also remove completely, it's not important.

opqdonut: Either of those is fine by me!

src/clj/rems/ega.clj
Macroz: You are right, I think I will prefix all the methods with the request-method so that they become unique. Otherwise GET permissions and DELETE permissions would be the same.

Macroz: They can also all be changed into private?

src/clj/rems/ega.clj
Macroz: Probably makes sense. I'm not sure if they are actually needed at all or what they should be. The entitlement post used to have these but I'm not sure if they make sense?

opqdonut: see also timeouts in `rems.event-notification`

src/clj/rems/db/user_settings.clj
Macroz: The API did not guarantee them before either?

opqdonut: No, the keys are required in the API response, see
- how rems.api.user-settings usess the user-settings/UserSettings schema
- https://rems-dev.rahtiapp.fi/swagger-ui/index.html#/user-settings/get_api_user_settings

Macroz: How is it possible? It used to reference `PartialUserSettings` schema where they were optional.

Macroz: Ah, I misread. Only PUT referenced them.

Macroz: Since these values come from the `default-settings` if they don't exist we could force the presence of the keys like before.

Macroz: In here the original problem is doing it differently, and skipping separate read and write models. Cleaning up in a backwards compatible way is difficult. But I could introduce separate read & write models.

opqdonut: Yeah that's why I wanted to split this out into a separate PR, we need to do this carefully. I like the schema simplification. Maybe we can have something like `(defn UserSettingsInApi (assoc UserSettings :language s/Keyword :notification-email s/Str))` instead of the `PartialUserSettings` hack.

Macroz: We can have separate DB and API schemas as the others?

opqdonut: Yeah can just do full decoupling

Macroz: I'll do that and split the small parts out.

src/clj/rems/ega.clj
Macroz: I think we'll have to see about EGA what they should eventually be. I've extracted them into local constants for now.

src/clj/rems/db/user_settings.clj
Macroz: This PR is now updated with what I think is the separate db, read and write models with backwards compatibility. I have to look at it with fresh eyes still. I'll harvest them to another PR next.

src/cljc/rems/common/catalogue_util.cljc
Macroz: https://github.com/CSCfi/rems/pull/2605

src/clj/rems/db/user_settings.clj
Macroz: https://github.com/CSCfi/rems/pull/2606

src/clj/rems/db/user_secrets.clj
```diff
@@ -31,9 +31,9 @@
 (defn update-user-secrets! [user new-secrets]
   (assert user "User missing!")
   (let [old-secrets (get-user-secrets user)
-        validated (try (validate-user-secrets new-secrets) (catch Exception _e nil))]
-    (if (= (set (keys validated))
-           (set (keys new-secrets)))
+        validated (try (validate-user-secrets (merge old-secrets new-secrets))
+                       (catch Exception _e nil))]
```
opqdonut, line 35: should probably have some sort of `{:success false}` logic here (just like in user-settings) instead of failing silently.

Macroz: The next line is the `if` that goes to `{:success false}` if we got the `nil` here?

opqdonut: Oh right, I'm blind.

Macroz:
I'm closing this draft now. Towards production code in https://github.com/CSCfi/rems/pull/2637.

PR #2605, opened 2021-03-18
Add EGA more info

Macroz:
More info support for #2466, spun from #2602.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
- [ ] Update docs/ (if applicable) (TBD after whole feature)
- [ ] Update manual/ (if applicable) (TBD after whole feature)
- [ ] ADR for major architectural decisions or experiments (TBD after whole feature)


src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -1,9 +1,33 @@
...
+    (or (get-in resource-or-item [:catalogue-item/infourl language])
+        (get-in resource-or-item [:catalogue-item/infourl default-language])
+        (get-in resource-or-item [:localizations language :infourl])
+        (get-in resource-or-item [:localizations default-language :infourl])
+        (urn-catalogue-item-url (:resource/ext-id resource-or-item) config)
+        (urn-catalogue-item-url (:resid resource-or-item) config)
+        (ega-catalogue-item-url (:resource/ext-id resource-or-item) config)
+        (ega-catalogue-item-url (:resid resource-or-item) config))))
+
+;; TODO: test here?
```
opqdonut, line 33: I'd like a short test for catalogue-item-more-info-url – it has so many cases.

src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -1,9 +1,33 @@
...
   (when (urn? resid)
     (str (or urn-organization "http://urn.fi/") resid)))
+
+;; EGA catalogue items (i.e. datasets) look like EGAD00001006673 and could be linked to e.g. https://ega-archive.org/datasets/EGAD00001006673
+
+(defn- ega-dataset? [resid]
+  (and resid (str/starts-with? resid "EGAD")))
+
+(defn- ega-catalogue-item-url [resid {:keys [enable-ega ega-organization]}]
+  (when (and enable-ega (ega-dataset? resid))
```
opqdonut, line 18: As far as I'm concerned this could be an unconditional feature. I'm fine with the feature flag too.

resources/translations/en.edn
```diff
@@ -109,7 +109,7 @@
                    :license-name "License name"
                    :licenses "Licenses"
                    :modifier "Modifier"
-                   :more-info "More info URL (optional, defaults to resource URN)"
+                   :more-info "More info URL (optional, defaults to resource URN or EGA link)"
```
opqdonut, line 112: Should we add this text if the EGA stuff is behind a feature flag? Perhaps modify the texts only when `:enable-ega` defaults to `true`?

opqdonut (changes requested):
Some questions about the feature flag, request for a test.

resources/translations/en.edn
Macroz: I guess that added complexity is not worthwhile for this. I did however want to add the `:enable-ega` flag here as in theory someone else's resource ids could start `"EGAD"`. We could perhaps remove this complexity from the localizations altogether and have the manual explain the automatic parts.

src/cljc/rems/common/catalogue_util.cljc
Macroz: URN was a bit clearer with the link style values but the values EGA wants are `"EGAD..."` and that might conflict with someone. We'll have the feature flag anyway.

src/cljc/rems/common/catalogue_util.cljc
Macroz: Yeah I can add one, these are the accumulations from the existing code gathered to one statement plus EGA.

src/cljc/rems/common/catalogue_util.cljc
opqdonut: Ok!

resources/translations/en.edn
opqdonut: So do you want to leave the description in this localization? It feels a bit weird since I imagine almost none of our current users would have `:enable-ega true` in the near future.

opqdonut (approved):

resources/translations/en.edn
Macroz: Let's remove it altogether then? For URN too?

Macroz: LBR could decide to put the URN as an extra translation for example, for others I think manual is the correct place if any.

opqdonut: Fine by me!

marharyta (approved):

PR #2606, opened 2021-03-18
EGA user settings

Macroz:
Refactors the user settings a bit to separate the DB and API read and write models. 

Then adds the EGA settings for #2466.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary (part of a greater whole that will have the log entry)

## Testing

### DIFF
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -7,10 +7,7 @@
             [schema.core :as s])
   (:import (org.joda.time DateTime)))
 
-(s/defschema GetUserSettings
-  {:language s/Keyword
-   :notification-email (s/maybe s/Str)
-   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
+(def GetUserSettings user-settings/UserSettings)
 
 (s/defschema UpdateUserSettings
   {(s/optional-key :language) s/Keyword
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -14,8 +14,7 @@
 
 (s/defschema UpdateUserSettings
   {(s/optional-key :language) s/Keyword
-   (s/optional-key :notification-email) (s/maybe s/Str)
-   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
+   (s/optional-key :notification-email) (s/maybe s/Str)})
 
 (def user-settings-api
   (context "/user-settings" []
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -3,7 +3,19 @@
             [rems.api.schema :as schema]
             [rems.db.user-settings :as user-settings]
             [rems.util :refer [getx-user-id get-user-id]]
-            [ring.util.http-response :refer :all]))
+            [ring.util.http-response :refer :all]
+            [schema.core :as s])
+  (:import (org.joda.time DateTime)))
+
+(s/defschema GetUserSettings
+  {:language s/Keyword
+   :notification-email (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
+
+(s/defschema UpdateUserSettings
+  {(s/optional-key :language) s/Keyword
+   (s/optional-key :notification-email) (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
 
 (def user-settings-api
   (context "/user-settings" []
@@ -12,12 +24,12 @@
     (GET "/" []
       :summary "Get user settings"
       :roles #{:logged-in}
-      :return user-settings/UserSettings
+      :return GetUserSettings
       (ok (user-settings/get-user-settings (get-user-id))))
 
     (PUT "/" []
       :summary "Update user settings"
       :roles #{:logged-in}
-      :body [settings user-settings/UserSettings]
+      :body [settings UpdateUserSettings]
       :return schema/SuccessResponse
       (ok (user-settings/update-user-settings! (getx-user-id) settings)))))
--- a/src/clj/rems/api/user_settings.clj
+++ b/src/clj/rems/api/user_settings.clj
@@ -18,6 +18,6 @@
     (PUT "/" []
       :summary "Update user settings"
       :roles #{:logged-in}
-      :body [settings user-settings/PartialUserSettings]
+      :body [settings user-settings/UserSettings]
       :return schema/SuccessResponse
       (ok (user-settings/update-user-settings! (getx-user-id) settings)))))
--- a/src/clj/rems/db/user_settings.clj
+++ b/src/clj/rems/db/user_settings.clj
@@ -40,24 +40,27 @@
          (when user
            (json->settings (:settings (db/get-user-settings {:user user}))))))
 
-(defn validate-new-settings [{:keys [language notification-email] :as settings}]
-  (into {} [(when (contains? (set (:languages env)) language)
-              {:language language})
-            (when (and notification-email (re-matches +email-regex+ notification-email))
-              {:notification-email notification-email})
-            (when (and (contains? settings :notification-email)
-                       (str/blank? notification-email))
-              ;; clear notification email to use identity provider's email instead
-              {:notification-email nil})
-            (when-let [ega (:ega settings)]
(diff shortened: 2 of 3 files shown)


src/clj/rems/db/user_settings.clj
```diff
@@ -25,20 +26,14 @@
...
 
 (defn- json->settings [json]
-  ;; Allows missing keys, so we don't need to write migrations
-  ;; if we add new keys to user settings. Migrations are needed if
-  ;; we remove keys or make their validation stricter.
   (when json
     (-> json
         json/parse-string
-        coerce-partial-user-settings)))
+        coerce-user-settings)))
```
opqdonut, line 36: So now if some user didn't have a `:language` or `:notification-email` this will fail. It seems that we don't populate user settings when creating a user, so this will probably break production.
```
rems=> (rems.db.users/add-user! {:userid "settings-test"})
(1)
rems=> (rems.db.core/get-user-settings {:user "settings-test"})
nil
```

One solution would be to use a PartialUserSettings-like schema for the db storage, and produce something that matches GetUserSettings (i.e. with guaranteed language and notification-email) from get-user-settings.

opqdonut (dismissed):

src/clj/rems/db/user_settings.clj
opqdonut: Oh, and in case we don't have a test that fetched the user settings of a freshly created user, we probably should have one, since it would've caught this.

opqdonut: Ah, I misread the code a bit. The `when json` handles empty user settings, but what about a case where somebody has set _just the language_ via the API. Then this code will fail due to a missing `:notification-email` key.

opqdonut: Hmm, even that seems to work since we merge `(get-user-settings user)` with the new settings. This is subtle.

src/clj/rems/db/user_settings.clj
```diff
@@ -25,20 +26,14 @@
       validate-user-settings
       json/generate-string))
 
-(s/defschema PartialUserSettings
-  (map-keys s/optional-key UserSettings))
-
-(def ^:private coerce-partial-user-settings
-  (coerce/coercer! PartialUserSettings json/coercion-matcher))
+(def ^:private coerce-user-settings
```
opqdonut, line 29: After getting rid of PartialUserSettings I think you can also remove the `(= (set (keys validated)) (set (keys new-settings)))` validation from `update-user-settings!`, right?

src/clj/rems/db/user_settings.clj
Macroz: We do have a test case for a new user in `test-user-settings`.

opqdonut: Yeah I'm pretty sure this would be safe to merge. Just needed to think through the logic.

src/clj/rems/db/user_settings.clj
Macroz: Well reading the code now. I think the key comparison is there so that all your "settings" must be valid or the entire request is considered invalid. So if you try to set with an invalid-looking email address, the entire request will fail, no matter that it passes the schema and any other properties are ok.

opqdonut: Oh right, I see. Maybe `validate-new-settings` could just return nil if any of the validations fail? And if that's logic is difficult to write neatly, you can just push the key set check inside `validate-new-settings`, right?

Macroz: I agree that it'd be more readable if the validation actaully had an easy to check failure result.

src/clj/rems/db/user_settings.clj
```diff
@@ -25,42 +26,41 @@
...
+      valid-new-settings
+      nil))) ; fail completely if there was even one error
 
 (defn update-user-settings! [user new-settings]
   (assert user "User missing!")
   (let [old-settings (get-user-settings user)
         validated (validate-new-settings new-settings)]
-    (if (= (set (keys validated))
-           (set (keys new-settings)))
+    (if validated
```
opqdonut, line 63: [golf] could use if-let, and inline old-settings

src/clj/rems/api/user_settings.clj
```diff
@@ -3,7 +3,19 @@
...
+
+(s/defschema GetUserSettings
+  {:language s/Keyword
+   :notification-email (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
+
+(s/defschema UpdateUserSettings
+  {(s/optional-key :language) s/Keyword
+   (s/optional-key :notification-email) (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
```
opqdonut, line 18: So what's the use case for setting the expiration date via the update API? I thought the expiration date would be derived from the key, and thus read-only. Or is this just a stepping stone on the way to storing the API key?

src/clj/rems/db/user_settings.clj
```diff
@@ -25,42 +26,41 @@
...
+(defn validate-new-settings [{:keys [language notification-email] :as new-settings}]
+  (let [valid-new-settings (merge (when (contains? (set (:languages env)) language)
+                                    {:language language})
+                                  (when (and notification-email (re-matches +email-regex+ notification-email))
+                                    {:notification-email notification-email})
+                                  (when (and (contains? new-settings :notification-email)
+                                             (str/blank? notification-email))
+                                    ;; clear notification email to use identity provider's email instead
+                                    {:notification-email nil})
+                                  (when-let [ega (:ega new-settings)]
```
opqdonut, line 52: should we check the `:enable-ega` config here?

src/clj/rems/api/user_settings.clj
```diff
@@ -3,7 +3,19 @@
...
             [rems.util :refer [getx-user-id get-user-id]]
-            [ring.util.http-response :refer :all]))
+            [ring.util.http-response :refer :all]
+            [schema.core :as s])
+  (:import (org.joda.time DateTime)))
+
+(s/defschema GetUserSettings
+  {:language s/Keyword
+   :notification-email (s/maybe s/Str)
+   (s/optional-key :ega) {:api-key-expiration-date DateTime}})
```
opqdonut, line 13: I know we want to be able to deduplicate the db and api models so this is nice, but since the models are the same at the moment, could this just be a `(def GetUserSettings user-settings/UserSettings)` for now?

I bet we went over this previously but I've forgotten our conversation.

opqdonut (changes requested):
questions, one suggestion

src/clj/rems/api/user_settings.clj
Macroz: Good point. It can actually be removed since it is set only from the back-end. Just need to make sure it's available for setting outside the API.

src/clj/rems/db/user_settings.clj
Macroz: Yes, this was made before that setting so it makes sense to add that check here too.

src/clj/rems/api/user_settings.clj
Macroz: If we remove it then we don't need to check for `:enable-ega` here...

src/clj/rems/db/user_settings.clj
```diff
@@ -25,42 +26,42 @@
...
+                                    {:ega (select-keys ega [:api-key-expiration-date])}))]
+    (if (= (set (keys valid-new-settings))
```
opqdonut, line 55: [golf] consider `when`

opqdonut (approved):
feel free to merge, but see my two golf comments

src/clj/rems/db/user_settings.clj
Macroz: If you agree that it's explicit enough then I'll do it.

Macroz: I kind of don't like the need to have an explicit `{}` in the beginning.

opqdonut: I think we use the when pattern for other validation-like things a lot, so it's fine. I don't mind the `if` if you want to emphasize the `nil`.

opqdonut (approved):

Issue #2608, opened 2021-03-22
Form editor improvements


opqdonut:
Follow-up to #1804

- [x] minimize fields by default
- [x] make expand field button more prominent (normal collapse component)
- [x] get rid of dashed lines (except for "add field/item/option" buttons)
- [x] make labels (e.g. "Title", "Placeholder") stand out better in form editor
- [x] don't repeat "Create form" text, instead call editor "Form editor" and preview "Form preview" (https://www.figma.com/file/XoZV4wZd8LWtDeExGjv6bL/rems?node-id=0%3A1)
- [ ] update manual with new form editor screenshots
- [x] check that browser test coverage is what we want it to be

PR #2610, opened 2021-03-23
form editor styling

opqdonut:
- collapse fields by default
- make expand/collapse button more prominent
- use less dashed lines

for #2608

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
  - the existing entry covers this
  - to be done once styling is finalized

## Testing

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -253,16 +253,19 @@
    [:.text-highlight {:color (get-theme-attribute :color3)
                       :font-weight "bold"}]))
 
-(defn- generate-dashed-form-group []
+(defn- generate-form-group []
   {:position "relative"
-   :border "2px dashed #ccc"
    :border-radius (u/rem 0.4)
    :padding (u/px 10)
    :margin-top 0
    :margin-bottom (u/px 16)})
 
+(defn- generate-dashed-form-group []
+  (assoc (generate-form-group)
+         :border "2px dashed #ccc"))
+
 (defn- generate-solid-form-group []
-  (assoc (generate-dashed-form-group)
+  (assoc (generate-form-group)
          :border "2px solid #eee"))
 
 (defn- remove-nil-vals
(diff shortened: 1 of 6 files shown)


opqdonut:
# Switching from dashed to solid groups

## Before

### Form editor
editor-dash

### Organization editor
org-dash

## After

### Form editor
editor-solid

### Organization editor
org-solid

# Form editor fields collapsed by default

editor-collapsed

# Changes in collapsible component margins

## Before

collapsible-before

## After

collapsible-after

marharyta:
@opqdonut are you sure you have included the correct last screenshots? i cannot tell how they are different.

opqdonut:
They were grabbed from different environments (localhost vs rems-dev) so there's a slight difference in fonts, but the real difference is the amount of vertical whitespace below "Show more". I'm just trying to demonstrate that my changes to rems.collapsible styling didn't break other use sites :)

marharyta:
ok! I would usually put the screenshots side by side in Figma and show it with the rulers.

marharyta (approved):
I like the changes in the PR.
the only thing I do not like is the way the styles are shuffled, but this is largely an opinion so all good to be merged.

src/clj/rems/css/styles.clj
```diff
@@ -262,12 +262,8 @@
...
 
 (defn- generate-solid-form-group []
-  {:position "relative"
-   :border "2px solid #eee"
-   :margin 0
-   :padding (u/rem 1)
-   :padding-right (u/rem 2)
-   :border-radius (u/rem 0.4)})
+  (assoc (generate-dashed-form-group)
+         :border "2px solid #eee"))
```
Macroz, line 266: I think it could be slightly more intuitive if this was written with a `(generate-group)` common helper that both variants can use (solid, dashed). However this is fine too.

Macroz (approved):

src/clj/rems/css/styles.clj
opqdonut: Why not!

Issue #2552, opened 2021-02-08
Phone number & IP address field types


opqdonut:
Add two new field types. Split out from #2404

- [x] Backend & schema
- [x] Applicant UI
- [x] Form editor
- [x] manual
- Phone number validations
  - [x] Phone number can only consist of + and numbers (spaces are accepted)
 
- IP address validations
  - [x] IP address should support ipv4 and ipv6
  - [x] Optionally reject private IP addresses (10.x.x.x etc.)
  - [x] link-local probably needs to also be excluded
  - [x] according to @Macroz "we had the discussion in grooming and the result was that ticket (i.e. there is no other validation / outside service because there can't be and but we would consider only public IP addresses so we can invalidate private ones)" 
  - [x] from @jaakkocsc "192.168.x.x
                                        172.16.x.x
                                        10.x.x.x
                                        169.254.1.0 to 169.254.254.255
                                        100.64.0.0 to 100.127.255.255
                                        224.0.0.0 to 224.0.0.255"
<img width="1195" alt="Screenshot 2021-03-25 at 11 12 05" src="https://user-images.githubusercontent.com/7344718/112448343-fad88900-8d5a-11eb-88fe-4d87fe825159.png">

marharyta:
I was actually wondering if we could split the ticket into backend and frontend and I could actually work on the schema a bit?

opqdonut:
As discussed, no need to split the ticket, but the implementation can be in multiple parts.

marharyta:
picking a regex to validate the phone number, which formats we would ideally like to support?

PR #2612, opened 2021-03-24
feat: ip address form field

marharyta:
# Checklist for author

## Reviewability
- [x] Link to issue (#2552)
- [ ] TODO: Consider adding screenshots for ease of review

## Testing

Screenshot 2021-03-24 at 17 26 46


src/clj/rems/form_validation.clj
```diff
@@ -52,6 +52,13 @@
       {:field-id (:field/id field)
        :type     :t.form.validation/invalid-phone-number})))
 
+(defn- invalid-ip-address-error [field]
+  (when (= (:field/type field) :ip-address)
+    (when-not (or (str/blank? (:field/value field))
+                  (re-matches +ip-address-regex+ (:field/value field)))
```
opqdonut, line 58: what about optionally rejecting private IP addresses?

test/clj/rems/test_form_validation.clj
```diff
@@ -393,6 +393,42 @@
...
+                                       {:field/id       "number has to start with a + prefix"
+                                        :field/type     :phone-number
+                                        :field/optional true
+                                        :field/visible  true
+                                        :field/value    "35000000000000"}
+                                       {:field/id       "valid phone number"
+                                        :field/type     :phone-number
+                                        :field/optional true
+                                        :field/visible  true
+                                        :field/value    "+358450000100"}]))))
```
opqdonut, line 430: testing wrong field type?

opqdonut (changes requested):

opqdonut:
will the frontend be in a separate PR? this looks like only the backend.

opqdonut:
nevermind that last comment, I'm blind, I missed the change to fields.cljs at first

src/clj/rems/form_validation.clj
marharyta: can I have some sort of summary of what addresses are acceptable and what are not? will making testing much easier. @jaakkocsc ?

Macroz: The summary is actually in the ticket.

opqdonut: I don't think the spec for the private ranges is in the ticket, but you can find it for example here:
https://en.wikipedia.org/wiki/Private_network#Private_IPv4_addresses

Macroz: Yes, that task is about investigating first and then doing a reasonable set.

marharyta: @Macroz 
```
 IP address should support ipv4 and ipv6
 Optionally reject private IP addresses (10.x.x.x etc.)
```

is the only thing I see in the ticket

Macroz: Yes, to me that is a summary of what needs to be done. I.e. support IPv4 and v6 (this so far has only IPv4), and investigate rejecting private IP address ranges, the details will be revealed by an investigation and we do not have the information yet. 

So the task includes the investigation. If we had exact information then it'd be in the ticket too. So this is the level of summary we can have without actually doing the investigation first. Perhaps you are asking for someone else to do the investigation?

opqdonut: In addition to the "private" ranges, it's probably good to also exclude "link-local" ranges. https://en.wikipedia.org/wiki/Private_network#Link-local_addresses

marharyta: I ll create regexes and unit tests accordingly

src/cljc/rems/common/util.cljc
```diff
@@ -19,6 +19,12 @@
   (is (= nil
          (re-matches +phone-number-regex+ "+35845000000000000000000000000000"))))
 
+;; regex from https://www.oreilly.com/library/view/regular-expressions-cookbook/9780596802837/ch07s16.html
+(def +ip-address-regex+ #"^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.↵
+                           (25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.↵
+                           (25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.↵
+                           (25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$")
```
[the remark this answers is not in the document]
marharyta: i was planning to add it in this Pr, currently working on regex testing

marharyta:
@opqdonut ready for re-review

src/cljc/rems/common/util.cljc
```diff
@@ -19,6 +19,100 @@
...
+;; 192.168.x.x - Private network, local communications within a private network
+;; 198.18.x.x - 198.19.x.x - Private network, benchmark testing of inter-network communications between two separate subnets
+;; 198.51.100.x - Documentation, Assigned as TEST-NET-2, documentation and examples.
+;; 203.0.113.x - Documentation, Assigned as TEST-NET-3, documentation and examples.
+;; 240.x.x.x–255.x.x.x - multicast
+(def +reserved-ip-address-range-regex+
+  (re-pattern
+   (str
+    "((0|10|127|224?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((100?)\\.([6-9][4-9]|1[1-2][1-7]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
```
opqdonut, line 46: This regex isn't correct, it fails to catch e.g. the second octet being between 70 and 73. Test case:

```
  (is (= "100.71.0.255"
         (first (re-matches +reserved-ip-address-range-regex+ "100.71.0.255"))))
```

src/cljc/rems/common/util.cljc
```diff
@@ -19,6 +19,100 @@
...
+    "((169?)\\.(254?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
```
opqdonut, line 47: This isn't correct either: the extra `?` characters cause ip addresses like 16.254.x.x and 169.25.x.x to be marked as reserved. Test case:

```
  (is (= nil
         (first (re-matches +reserved-ip-address-range-regex+ "16.254.1.1"))))
  (is (= nil
         (first (re-matches +reserved-ip-address-range-regex+ "169.25.1.1"))))
```

src/cljc/rems/common/util.cljc
```diff
@@ -19,6 +19,100 @@
...
+    "((172?)\\.(1[6-9]|2[0-9]|31?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((192?)\\.(0?)\\.(0|2?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((192?)\\.(168?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((198?)\\.(1[8-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((198?)\\.(51?)\\.(100?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((203?)\\.(0?)\\.(113?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((2[4-5][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)")))
```
opqdonut, line 54: The regex-based approach seems a bit fragile. Did you consider parsing the IP addresses into vectors like `[169 254 13 14]` and using code to check the ranges? Alternatively, you could consider writing some sort of util to generate a regex from a pattern like `169.254.x.x`. The only difficult case is `100.64.x.x–100.127.x.x`, and even that could be fixe by just generating patterns `100.64.x.x`, `100.65.x.x`, ..., `100.127.x.x`

opqdonut: One more thing that casme to my mind is that there should be no need to have things like `(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)` in these regexes. If a string is given to this regex, we know it has already passed `+valid-ip-address-regex+`, so we can just use patterns like `10\..*` or `169\.254\..*` here. Unless I'm missing something, that is.

src/cljc/rems/common/util.cljc
```diff
@@ -19,6 +19,100 @@
...
+;; 203.0.113.x - Documentation, Assigned as TEST-NET-3, documentation and examples.
+;; 240.x.x.x–255.x.x.x - multicast
+(def +reserved-ip-address-range-regex+
+  (re-pattern
+   (str
+    "((0|10|127|224?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((100?)\\.([6-9][4-9]|1[1-2][1-7]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((169?)\\.(254?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((172?)\\.(1[6-9]|2[0-9]|31?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
+    "((192?)\\.(0?)\\.(0|2?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$)|"
```
opqdonut, line 49: BTW, since you're using `re-matches`, the `$` is redundant: the pattern needs to match the whole string anyway.

src/clj/rems/form_validation.clj
```diff
@@ -2,7 +2,12 @@
   "Pure functions for form validation logic"
   (:require [clojure.string :as str]
             [rems.common.form :as form]
-            [rems.common.util :refer [+email-regex+ +phone-number-regex+]]))
+            [rems.common.util :refer [+email-regex+
+                                      +phone-number-regex+
+                                      +valid-ip-address-regex+
+                                      +valid-ip-address-regex-version-six+
+                                      +reserved-ip-address-range-regex+
+                                      +reserved-ip-address-range-regex-version-six+]]))
```
opqdonut, line 10: Minor point: I think we put the `+email-regex+` in `rems.common.util` so that it could be used in various places, e.g. r`ems.db.user-settings`. However the ip regexes could just live in this namespace since they'll probably not be used for anything else than form-validation.

opqdonut (changes requested):

src/cljc/rems/common/util.cljc
marharyta: hmm, I fixed that

marharyta:
since joel is now in a meeting, ping @Macroz for re-review

PR #2618, opened 2021-04-06
fix: empty read-only fields should be rendered

opqdonut:
needed to rework the info-text fix from #2576

# Checklist for author

## Reviewability
- [x] Link to issue
  - linked to original PR

## Documentation
  - the original fix is already noted there, and the bug fixed here was never released

## Testing

### DIFF
--- a/src/cljs/rems/fields.cljs
+++ b/src/cljs/rems/fields.cljs
@@ -111,7 +111,7 @@
         value (:field/value opts)
         previous-value (:field/previous-value opts)
         max-length (:field/max-length opts)
-        info-text (linkify (localized (:field/info-text opts)))
+        info-text (localized (:field/info-text opts))
         collapse-aria-label (str (text :t.create-form/collapse-aria-label) raw-title)]
     ;; TODO: simplify fieldset code
     [(if fieldset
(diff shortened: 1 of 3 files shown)


opqdonut:
# Before
empty-fields

# After
empty-fields-after

src/cljs/rems/fields.cljs
```diff
@@ -111,7 +111,7 @@
         value (:field/value opts)
         previous-value (:field/previous-value opts)
         max-length (:field/max-length opts)
-        info-text (linkify (localized (:field/info-text opts)))
```
marharyta, line 114: so, how did linkify cause this problem?

marharyta: i mean, this looks like a now links in the info text will not be heighlighted

opqdonut: I pushed the call to linkify down to line 144 to be able to check if the original text is blank or not. `(str/blank? (linkify foo))` unfortunately doesn't work since `linkify` returns a sequence of strings and not a single string.

marharyta: right, should we then try to approach it from a different angle?
like maybe have another version of the linkify function?

marharyta: i would really like to have the linkyfying functionality still, lets see where it is used

opqdonut: The linkifying functionality is still there for info texts. The call has just moved from this `let` to the use site on line 144.

src/cljs/rems/fields.cljs
```diff
@@ -136,12 +136,12 @@
...
         (text :t.form/optional)
         (text :t.form/required))
-      (when info-text
+      (when-not (str/blank? info-text)
         [info-collapse
          {:info-id (field-name opts)
           :aria-label-text collapse-aria-label
           :focus-when-collapse-opened focus-when-collapse-opened
-          :body-text info-text}])]
+          :body-text (linkify info-text)}])]
```
marharyta, line 144: Right i can see it here, thanks for pointing out

marharyta (approved):

marharyta:
Since all the questions were addressed, i approve the pr

PR #2623, opened 2021-04-07
feat: use bootstrap styling for readonly-field

### LINKED ISSUE #2621: Form fixes
Some fixes are needed for the handler/applicant view of forms

- [x] render empty readonly fields
- [x] show "No rows" when a readonly table doesn't have rows
- [ ] improve contrast of readonly input fields

opqdonut:
better contrast, more consistent

closes #2621

# Checklist for author

## Reviewability
- [x] Link to issue

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -637,10 +637,7 @@
              :margin-top 0
              :margin-bottom (u/px 16)}]
    [:div.form-control {:height :auto
-                       :white-space "pre-wrap"
-                       ;; A copy of bootstrap's .form-control:disabled
-                       :background-color "#e9ecef"
-                       :opacity 1}
+                       :white-space "pre-wrap"}
     [:&:empty {:height (u/rem 2.25)}]]
    [:.toggle-diff {:float "right"}]
    [:.diff
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -638,8 +638,9 @@
              :margin-bottom (u/px 16)}]
    [:div.form-control {:height :auto
                        :white-space "pre-wrap"
-                       :border-color "rgba(206, 212, 218, 0.2)" ; "#ced4da"
-                       :background-color "rgba(0, 0, 0, 0.01)"}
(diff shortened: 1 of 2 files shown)


opqdonut:
# Before 

application-before

catalogue-before

# After

application-after

catalogue-after

opqdonut:
firefox reports contrast ratio 6.90 AA

src/clj/rems/css/styles.clj
```diff
@@ -638,8 +638,9 @@
              :margin-bottom (u/px 16)}]
    [:div.form-control {:height :auto
                        :white-space "pre-wrap"
-                       :border-color "rgba(206, 212, 218, 0.2)" ; "#ced4da"
-                       :background-color "rgba(0, 0, 0, 0.01)"}
+                       ;; A copy of bootstrap's .form-control:disabled
+                       :background-color "#e9ecef"
+                       :opacity 1}
```
Macroz, line 643: I'm not sure this is actually so useful change as I thought. We are overriding a customization with another customization, only now we try to manually track Bootstrap styling. If we could remove some of our customizations and directly use Bootstrap classes, that'd work for me, i.e. simplifying stuff. Now we are trading one look for another with no clear use case. 

Is this trying to solve some contrast issue? Doesn't this reduce the contrast?

opqdonut: Yeah we didn't get rid of custom styles unfortunately. I would like to get bootstrap's `.form-control:disabled` style to fire, but the `:disabled` selector doesn't work on divs (even though I can mark the div as disabled). We need a div.form-control instead of an input.form-control because we want to have arbitrary html inside (e.g. lists).

We are trading one look for another, but we gain:
- consistency (see the table, and we might have other disabled inputs somewhere too)
- better contrast (between the white background and the input box)
- less custom colours

I'm not sure it's a slam dunk but it might be a net improvement.

opqdonut:
# New screenshots

## Readonly form

only-border

## Editable form

There's now a slight chance of confusing and editable form for a non-editable one, but the context (state of the application, role of the user, etc.) should make that clear I hope.

draft

Macroz (approved):
We reached consensus that this is great, let's merge it!

PR #2629, opened 2021-04-12
fix: horizontal overflow for inline info-field

opqdonut:
the style change from #2623 (issue #2621) made this overflow visible,
but it was there already before it

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
  - not necessary IMO


opqdonut:
# Before

before1
before2

# After

after1
after2

Macroz (approved):
I was trying to figure out if nesting containers is allowed or not but couldn't find a clear answers. In practice it's ok since `.container-fluid` is approximately just a flexbox with width 100%.

PR #2635, opened 2021-04-14
Add Code of Conduct

Macroz:
Let's just add this one since we do operate under CSC Code of Conduct anyway.


marharyta (approved):

marharyta:
Love and hugs! ❤️

Issue #2641, opened 2021-04-19
Upgrade to Postgres 13


opqdonut:
Currently we use Postgres 9.6 for testing and some production instances, while some production instances are running on Postgres 11. CSC wants to upgrade everything to Postgres 13 it seems.

- [x] use postgres 13 for local development & CI
- [x] use postgres 13 in docker-compose scripts
  - to be finished after next REMS release (v2.18) is out
- [x] run rems-dev.rahtiapp.fi or remsdev.kaj.pouta.csc.fi on postgres 13
  - rems-dev.rahtiapp.fi is now running on postgres 13 after, see CSCfi/rems-deploy#224
- [x] document that postgres 13 is supported, at least in CHANGELOG
- [ ] upgrade ansible scripts
- [x] upgrade rahti templates
  - [x] `postgres-novolume.yaml`
  - [x] `postgres.yaml`
- [ ] upgrade vms running in pouta
- [ ] upgrade databases running in rahti

opqdonut:
What's left in this ticket is:
- upgrade the ansible scripts - this probably doesn't make sense, we don't know of any real users of our postgres ansible scripts. We should probably just deprecate them
- upgrade rahti and pouta databases - this is ops work and needs to be planned separately

PR #2642, opened 2021-04-19
postgres 13 for local development & circleci

opqdonut:
for #2641

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
  - once we now postgres 13 works

## Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ executors:
   db:
     docker:
       - image: circleci/clojure:openjdk-11-lein-2.9.3-node-browsers
-      - image: postgres:9.6
+      - image: postgres:13
         environment:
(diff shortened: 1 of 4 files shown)


Macroz (approved):

opqdonut:
this mailing list thread seems to talk about the syntax change:

https://www.postgresql-archive.org/BUG-15237-I-got-quot-ERROR-source-for-a-multiple-column-UPDATE-item-must-be-a-sub-SELECT-or-ROW-expr-td6024672.html

relevant quote:

> (20) and row(20) are not the same thing, whereas (20,1.11) and
row(20,1.11) are the same thing.  Don't blame us, it was the SQL
committee's idea that "ROW" should be optional for the latter.

PR #2646, opened 2021-04-22
Add handler earlier

Macroz:
I noticed that the `:handler` role is added only from the applications so there needs to be an application before a person is a handler.

# Checklist for author

## Reviewability
- [ ] Link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/db/workflow.clj
+++ b/src/clj/rems/db/workflow.clj
@@ -52,3 +52,8 @@
 
 (defn join-workflow-licenses [workflow]
   (assoc workflow :licenses (get-workflow-licenses (:id workflow))))
+
+(defn get-all-workflow-roles [userid]
+  (when (some #(contains? (set (map :userid (get-in % [:workflow :handlers]))) userid)
+              (get-workflows nil))
+    #{:handler}))
--- a/src/clj/rems/middleware.clj
+++ b/src/clj/rems/middleware.clj
@@ -16,6 +16,7 @@
             [rems.db.roles :as roles]
             [rems.db.user-settings :as user-settings]
             [rems.db.users :as users]
+            [rems.db.workflow :as workflows]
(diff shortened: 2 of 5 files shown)


opqdonut (approved):

src/clj/rems/middleware.clj
```diff
@@ -73,6 +74,7 @@
                                (when context/*user*
                                  (set/union (roles/get-roles (getx-user-id))
                                             (organizations/get-all-organization-roles (getx-user-id))
+                                            (workflows/get-all-workflow-roles (getx-user-id))
                                             (applications/get-all-application-roles (getx-user-id))))
```
opqdonut, line 78: some redundancy here but I guess it won't matter, and we need to track the per-application-roles anyway for application permissions

src/clj/rems/db/workflow.clj
```diff
@@ -52,3 +52,8 @@
 
 (defn join-workflow-licenses [workflow]
   (assoc workflow :licenses (get-workflow-licenses (:id workflow))))
+
+(defn get-all-workflow-roles [userid]
+  (when (some #(contains? (set (map :userid (get-in % [:workflow :handlers]))) userid)
+              (get-workflows nil))
```
opqdonut, line 58: I wonder if we need to cache this? It does get run on every request, and there can be quite a few workflows. Oh well, I guess time will tell.

src/clj/rems/middleware.clj
Macroz: Yes I think that's fine.

src/clj/rems/db/workflow.clj
Macroz: I added a tech debt task about this concern https://github.com/CSCfi/rems/issues/2647

Issue #2639, opened 2021-04-16
Form behaviour ideas


marharyta:
Problems with the current (applicant) UX:

1. You don't know if your inputs (e.g. ip address) are valid before pressing Save, and then you need to fix all of these errors before REMS saves your work
  - Possible solution: Show validation errors to the user, but perform the save anyway. (Show errors in yellow?)
2. You need to perform an action (save/submit) to see the validation errors
  - Possible solution: Validate when user stops typing / leaves field
  - Questions:
    - What about partially filled in forms? It would be best to not show validation errors for fields that haven't been edited yet.
    - What if a user navigates away from the application and back? Do they see the validation errors, or perhaps just the state of the application as-is before they write something.
    - What to do with required fields? Should they all turn red immediately or only upon save? Or only upon submit?
  - Idea – three states for application fields:
    - untouched (not filled in at all)
    - edited but not valid (can save, can't submit)
    - filled in and completely valid (can submit)
3. You need to perform an action (save) to save your work.
  - Arguably not a problem, users are used to Save buttons?
  - Possible solution: Autosave (cache) into browser localstorage
    - With an explicit _save_ button. We wouldn't call caching "saving" to the user.
  - Possible solution: Autosave into REMS db
    - Problems:
      - Depends on 1
      - Lots of events generated to the REMS db
      
Plan:
- [x] Turn this into an ADR
- [ ] Make save succeed even with validation errors. Show validation errors in yellow (instead of red)
- What next?

Macroz:
NB: the first one is a bug #2766. The options are discussed in the ADR so we can close this issue.

PR #2648, opened 2021-04-23
doc: ADR 017: Applicant Form UX

opqdonut:
first draft

for #2639

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/docs/architecture/017-applicant-ux.md
+++ b/docs/architecture/017-applicant-ux.md
@@ -1,0 +1,66 @@
+# 017: Applicant Form UX
+
+Authors: @opqdonut, @Macroz, @marharyta
+
+This ADR aims to collect together some topics like validations in
+frontend, autosaving, etc. that are related to the Applicant's user
+experience. There have been previous discussions about all of these
+topics but no unified presentation.
+
+Based on ticket [#2639](https://github.com/CSCfi/rems/issues/2639)
+
+# Problems with the current applicant UX
+
+## A) Validation errors block saving
+
+You don't know if your inputs (e.g. ip address) are valid before
+pressing Save, and then you need to fix all of these errors before
+REMS saves your work.
+
+This can be irritating if you're working on a long application, and
+just want to save your work and go home. It's less of an issue if
+users save early & often.
+
+**Possible solution**: Show validation errors to the user, but perform
+the save anyway. Show errors in yellow?
+
+## B) Validation errors are shown late
+
+You need to perform an action (save/submit) to see the validation
+errors. Technically we should be able to run the validation code on
+the frontend at any point. There's no need to talk to the backend.
+
+**Possible solution**: Validate when user stops typing / leaves field.
+
+Questions:
+- What about partially filled in forms? It would be best to not show validation errors for fields that haven't been edited yet.
+- What if a user navigates away from the application and back? Do they see the validation errors, or perhaps just the state of the application as-is before they write something.
+- What to do with required fields? Should they all turn red immediately or only upon save? Or only upon submit?
+
+Idea: three states for application fields:
+- untouched (not filled in at all)
+- edited but not valid (can save, can't submit)
+- filled in and completely valid (can submit)
+
+## C) You need to save your work explicitly
+
+You need to click on the "Save draft" button to save your work.
+Arguably not a problem, users are used to Save buttons?
+
+**Possible solution 1**: Autosave (cache) into browser localstorage We
+wouldn't call this caching "saving" to the user, instead their work
+would automatically be preserved in their browser, so if they e.g.
+navigate away from the application page and navigate back, they'll
+still se their work-in-progress application.
+
+**Possible solution 2**: Autosave into REMS db. Also needs problem A to be fixed.
+Cons: lots of events generated to the REMS db.
+
+# TODO Decision
+
+Here's a step-by-step plan for fixing these. We might not end up doing all of the steps.
+
+1. (Problem A) Make save succeed even with validation errors. Show validation errors in yellow (instead of red).
+2. (Problem B) Run validations in the frontend after the user has stopped typing.
+3. TODO
+4. TODO
--- a/docs/architecture/017-applicant-ux.md
+++ b/docs/architecture/017-applicant-ux.md
@@ -51,7 +51,7 @@ Arguably not a problem, users are used to Save buttons?
 wouldn't call this caching "saving" to the user, instead their work
 would automatically be preserved in their browser, so if they e.g.
 navigate away from the application page and navigate back, they'll
-still se their work-in-progress application.
+still see their work-in-progress application.
 
 **Possible solution 2**: Autosave into REMS db. Also needs problem A to be fixed.
 Cons: lots of events generated to the REMS db.
--- a/docs/architecture/017-applicant-ux.md
+++ b/docs/architecture/017-applicant-ux.md
@@ -24,6 +24,17 @@ users save early & often.
 **Possible solution**: Show validation errors to the user, but perform
 the save anyway. Show errors in yellow?
 
+Historical note: historically (in the pre-SPA days) REMS used to work
+like this. This topic was more recently touched on in
+[#2117](https://github.com/CSCfi/rems/issues/2117). Interestingly
+enough, the idea idea seems to have been to only perform schema-style
+validations on drafts. However, the email content validation was
+included in the spec, so subsequent field types added their content
+validations to save-draft (see e.g.
+[#2567](https://github.com/CSCfi/rems/pull/2567),
+[#2594](https://github.com/CSCfi/rems/pull/2594))
+
+
 ## B) Validation errors are shown late
 
 You need to perform an action (save/submit) to see the validation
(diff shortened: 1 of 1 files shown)


docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+
+You don't know if your inputs (e.g. ip address) are valid before
+pressing Save, and then you need to fix all of these errors before
+REMS saves your work.
+
+This can be irritating if you're working on a long application, and
+just want to save your work and go home. It's less of an issue if
+users save early & often.
+
+**Possible solution**: Show validation errors to the user, but perform
```
Macroz, line 24: Could mention that it used to work like this?

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+
+You need to perform an action (save/submit) to see the validation
+errors. Technically we should be able to run the validation code on
+the frontend at any point. There's no need to talk to the backend.
+
+**Possible solution**: Validate when user stops typing / leaves field.
+
+Questions:
+- What about partially filled in forms? It would be best to not show validation errors for fields that haven't been edited yet.
+- What if a user navigates away from the application and back? Do they see the validation errors, or perhaps just the state of the application as-is before they write something.
```
Macroz, line 37: If you start filling field 3, will the validation error be only for field 3 or fields 1-3 or how?

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+- What to do with required fields? Should they all turn red immediately or only upon save? Or only upon submit?
+
+Idea: three states for application fields:
+- untouched (not filled in at all)
+- edited but not valid (can save, can't submit)
+- filled in and completely valid (can submit)
+
```
Macroz, line 44: I think from the point of view of the applicant, a text area can have a long answer that is still not finished. We may not be able to support this conveniently.

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+## C) You need to save your work explicitly
+
+You need to click on the "Save draft" button to save your work.
+Arguably not a problem, users are used to Save buttons?
```
Macroz, line 48: Even if we had autosave we would likely need to have the save button.

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+
+**Possible solution 1**: Autosave (cache) into browser localstorage We
+wouldn't call this caching "saving" to the user, instead their work
+would automatically be preserved in their browser, so if they e.g.
+navigate away from the application page and navigate back, they'll
+still se their work-in-progress application.
```
Macroz, line 54: typo: `se`

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+
+**Possible solution 2**: Autosave into REMS db. Also needs problem A to be fixed.
+Cons: lots of events generated to the REMS db.
```
Macroz, line 57: lots of events or a parallel but separate draft save system or event modification.

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,66 @@
...
+
+# TODO Decision
+
+Here's a step-by-step plan for fixing these. We might not end up doing all of the steps.
+
+1. (Problem A) Make save succeed even with validation errors. Show validation errors in yellow (instead of red).
+2. (Problem B) Run validations in the frontend after the user has stopped typing.
```
Macroz, line 64: Only show the validation errors for the fields the user actively changes in that "session". Nothing validated when the page is opened.

Macroz (commented):
Some discussion

docs/architecture/017-applicant-ux.md
opqdonut: I started doing that and tried to dig out the PR that changed things, but couldn't find a concrete decision to do this. I'll write my findings down.

docs/architecture/017-applicant-ux.md
opqdonut: added

docs/architecture/017-applicant-ux.md
opqdonut: added under the B heading

docs/architecture/017-applicant-ux.md
opqdonut: added

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,86 @@
...
+## B) Validation errors are shown late
+
+You need to perform an action (save/submit) to see the validation
+errors. Technically we should be able to run the validation code on
+the frontend at any point. There's no need to talk to the backend.
+
+**Possible solution**: Validate when user stops typing / leaves field.
+
+Questions:
+- What about partially filled in forms? It would be best to not show validation errors for fields that haven't been edited yet.
```
marharyta, line 47: If the field is empty, is makes sense to throw isEmpty validation on submit/save, if it is filled with data, it makes sense to validate the data

jaakkocsc (approved):

docs/architecture/017-applicant-ux.md
Macroz: Let's say if you fill two out of three fields and save, I would not like to show an error of not having filled the field three yet (because you did not proceed that far yet). Same but stronger if there exist 30 fields and you just filled the first.

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,89 @@
...
+- What if a user navigates away from the application and back? Do they see the validation errors, or perhaps just the state of the application as-is before they write something.
+- What if a user starts filling in field 3? Do they see validation errors for fields 1-3 or only 3?
+- What to do with required fields? Should they all turn red immediately or only upon save? Or only upon submit?
+
+Idea: Three states for application fields:
+- untouched (not filled in at all)
+- edited but not valid (can save, can't submit)
+- filled in and completely valid (can submit)
+
+Idea: Only show the validation errors for the fields the user actively changes in that "session". Nothing validated when the page is opened.
```
aatkin, line 60: from UX point-of-view it's not good if we validate data only on change. next time applicant comes back to the form, if they don't see the errors it would suggest that the inputs are actually valid

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,89 @@
...
+**Possible solution 1**: Autosave (cache) into browser localstorage We
+wouldn't call this caching "saving" to the user, instead their work
+would automatically be preserved in their browser, so if they e.g.
+navigate away from the application page and navigate back, they'll
+still see their work-in-progress application.
+
+**Possible solution 2**: Autosave into REMS db. Also needs problem A
+to be fixed. Cons: lots of events generated to the REMS db. Could work
+around this by implementing draft saving (or perhaps just auto-saving)
+separately from the application events.
```
aatkin, line 77: depending on what events can occur during draft phase, we could replace the current head of the event stack if the head has `eventdata->>'event/type' = 'application.event/draft-saved'`. this would leave interleaved draft-saved events depending on encasing events, but would skip a tremendous amount of extra events in case of auto-save

docs/architecture/017-applicant-ux.md
```diff
@@ -0,0 +1,89 @@
+# 017: Applicant Form UX
+
+Authors: @opqdonut, @Macroz, @marharyta
+
+This ADR aims to collect together some topics like validations in
+frontend, autosaving, etc. that are related to the Applicant's user
+experience. There have been previous discussions about all of these
+topics but no unified presentation.
```
aatkin, line 8: this document is missing some information. what kind of validations do we need? currently fields are either required or optional, which is enforced by API, and some fields have length limitation, which seems to be a hard limit.

second question, who attaches the validations to fields? would this be the owner who creates the form?

docs/architecture/017-applicant-ux.md
marharyta: This might not have been the best way to express it, what the document should say is "if we have not started typing anything in the form, then we do not show validation messages, once we do, we validate fields on change and after change", hopefully I was able to explain it well.

docs/architecture/017-applicant-ux.md
marharyta: I am sorry, I do not quite get what you are trying to do here, can you maybe provide a bit more details?

docs/architecture/017-applicant-ux.md
Macroz: Yeah, it's true that if the user comes back later, they would not initially see any errors. However you would always get them on send (with the problems due to delay)

docs/architecture/017-applicant-ux.md
Macroz: @aatkin yes that's one of the solutions. However,  we have been reluctant with modifying the event history and this would be a step that breaks that promise.

docs/architecture/017-applicant-ux.md
Macroz: There is small stuff like IP address and phone number format currently. All of them are handled in API and some would be nice to see already earlier, perhaps in data entry or so.

There is no attachment step as you mean. You can set field type or set the required, max length etc. and validation will be added.

Macroz (approved):
Time to merge, issues created.

PR #2649, opened 2021-04-26
feat: allow users to access their own /api/permissions/:user

### LINKED ISSUE #2631: Allow non-handler/owner users to query for their own permissions at the /permissions endpoint
### Feature Request
As a developer of a microservices architecture that relies on REMS for managing DAC-granted dataset-level data access permissions, I want to fetch a user's GA4GH Visa from the `/permissions` endpoint and use it to make authorization decisions in other microservices.

I am able to fetch the user's entitlements in REMS at the `/entitlements` endpoint using only a `x-rems-api-key` and the user's `x-rems-user-id` for authorization, but the `/permissions` endpoint additionally requires the user id of someone with a handler or owner role. This does not lend itself well to use in automated authorization workflows. Currently, we are needing to initialize a generic `owner` REMS user (whose ID is an environment variable) as a workaround.

### Acceptance Criteria
- [ ] Can fetch any user's permissions from the `/permissions` endpoint, such that:
    - [ ] The combination of a `x-rems-api-key` and *this user's* `x-rems-user-id` is sufficient for authorization.

opqdonut:
fixes #2631

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/permissions.clj
+++ b/src/clj/rems/api/permissions.clj
@@ -1,8 +1,11 @@
 (ns rems.api.permissions
   (:require [compojure.api.sweet :refer :all]
             [rems.api.util :as api-util] ; required for route :roles
+            [rems.auth.util :refer [throw-forbidden]]
             [rems.db.entitlements :as entitlements]
             [rems.config :as config]
+            [rems.common.roles :refer [has-roles?]]
+            [rems.util :refer [getx-user-id]]
             [ring.util.http-response :refer :all]
             [schema.core :as s]))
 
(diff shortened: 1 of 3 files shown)


src/clj/rems/api/permissions.clj
```diff
@@ -25,13 +28,22 @@
...
-         (if (empty? user)
+         (cond
+           (empty? user)
            (api-util/not-found-json-response)
+
+           (not (or (has-roles?  :handler :owner :organization-owner :reporter)
+                    (= user (getx-user-id))))
+           (throw-forbidden)
+
+           :else
```
Macroz, line 48: It's slightly unfortunate that some of these return and some throw. This expression is not quite a clean `or` or `cond`. The `not` `or` part is also rather difficult to read.

Can't see an immediate easy refactoring though.

Macroz (approved):

src/clj/rems/api/permissions.clj
opqdonut: I found it more natural to write the permission checking as `(not (allowed))`. I'll see if I can change the throw into a response.

opqdonut: We do the logging of the forbiddens in rems.api/forbidden-handler, so if I produce a forbidden response here it'll be slightly different from throwing an exception. Also, we just use throw-forbidden everywhere. So no fix for now.

PR #2651, opened 2021-04-26
feat: only include current application attachments in attachment zip

### LINKED ISSUE #2453: Only download current field attachments via "All attachments (zip)"

opqdonut:
fixes #2453, supercedes #2593

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -275,7 +275,7 @@
       :summary "Get attachments for an application as a zip file"
       :roles #{:logged-in}
       :path-params [application-id :- (describe s/Int "application id")]
-      :query-params [{all :- (describe s/Bool "Defaults to true. If set to false, will zip will only contain latest application attachments: no previous versions of attachments, and no event attachments.") true}]
+      :query-params [{all :- (describe s/Bool "Defaults to true. If set to false, the zip will only contain latest application attachments: no previous versions of attachments, and no event attachments.") true}]
       :responses {200 {}
                   404 {:schema s/Str :description "Not found"}}
       (if-let [app (applications/get-application-for-user (getx-user-id) application-id)]
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -272,13 +272,14 @@
         (api-util/not-found-json-response)))
 
     (GET "/:application-id/attachments" []
-      :summary "Get all attachments as a zip file"
+      :summary "Get attachments for an application as a zip file"
       :roles #{:logged-in}
       :path-params [application-id :- (describe s/Int "application id")]
+      :query-params [{all :- (describe s/Bool "Defaults to true. If set to false, will zip will only contain latest application attachments: no previous versions of attachments, and no event attachments.") true}]
       :responses {200 {}
                   404 {:schema s/Str :description "Not found"}}
       (if-let [app (applications/get-application-for-user (getx-user-id) application-id)]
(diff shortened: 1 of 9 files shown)


src/clj/rems/api/services/attachment.clj
```diff
@@ -43,21 +44,23 @@
       (throw-forbidden))
     (attachments/save-attachment! file user-id application-id)))
 
-(defn zip-attachments [application]
-  (let [out (ByteArrayOutputStream.)]
+(defn zip-attachments [application all?]
+  (let [classes (model/classify-attachments application)
```
opqdonut, line 48: I'm not that happy with calling model code here, but it was the simplest solution. I'd've preferred storing this information under `::attachment-classes` in the application. However, that presented a problem:

- zip-attachments needs the output of get-application-for-user because we want an application for which rems.application.model/hide-attachments has been done
- Currently, get-application-for-user is what gets otput from the API, so I can't add stuff there without it showing up in the API.
- I could add `::attachment-classes` to the output of get-application-internal (similarly to `::latest-decision-request-by-user`), but then I'd have to do the equivalent of hide-attachments here anyway

test/clj/rems/api/test_applications.clj
```diff
@@ -1897,6 +1898,12 @@
                                                            :field/title {:en "text field"
                                                                          :fi "text field"
                                                                          :sv "text field"}
+                                                           :field/optional false}
+                                                          {:field/type :attachment
+                                                           :field/id "att"
+                                                           :field/title {:en "attachment"
+                                                                         :fi "attachment"
+                                                                         :sv "attachment"}
```
Macroz, line 1916: At some point we should perhaps make all the localized values different in the tests.

src/clj/rems/api/applications.clj
```diff
@@ -272,13 +272,14 @@
         (api-util/not-found-json-response)))
 
     (GET "/:application-id/attachments" []
-      :summary "Get all attachments as a zip file"
+      :summary "Get attachments for an application as a zip file"
       :roles #{:logged-in}
       :path-params [application-id :- (describe s/Int "application id")]
+      :query-params [{all :- (describe s/Bool "Defaults to true. If set to false, will zip will only contain latest application attachments: no previous versions of attachments, and no event attachments.") true}]
```
Macroz, line 278: `will zip will`

src/clj/rems/api/services/attachment.clj
Macroz: I think overall we should have some concept of `raw` or `internal` vs. `public` or `external`. And be consistent about it. However we do have role-based authorization and data filtering so I think it can only be at the level of filtering done or not done. Also we have the db format which we mangle to get the internal format so I guess we have three. This could be something to talk about sometime that what could be the future big picture / if we could rewrite it.

I think simply reusing some code from somewhere is an acceptable solution here and some rework in the model the cleanest but more work and a future idea.

Macroz (approved):

test/clj/rems/api/test_applications.clj
opqdonut: I fixed it in this test.

PR #2652, opened 2021-04-28
test: introduce rems.browser-test-util/eventually-visible?

opqdonut:
Instead of (wait-visible :foo), try (is (eventually-visible? :foo))
for a nicer error. Instead of a hige stacktrace, you just get:

FAIL in rems.test-browser/test-guide-page (test_browser.clj:732)
expected: (btu/eventually-visible? {:tag :h2, :fn/text "Component Guide"})
  actual: (not (btu/eventually-visible? {:tag :h2, :fn/text "Component Guide"}))

# Checklist for author

## Reviewability
- [ ] Link to issue


Macroz (approved):
Good idea!

PR #2656, opened 2021-04-29
Misused email translation

marharyta:
# Checklist for author

## Reviewability
- [x] Link to issue (#2656 2434)

## Backwards compatibility

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/profile.cljs
+++ b/src/cljs/rems/profile.cljs
@@ -100,7 +100,7 @@
        {:title (text :t.profile/your-details)
         :always [:<>
                  [user/username (:user identity)]
-                 [user/attributes (:user identity)]]}]]]))
+                 [user/attributes attributes false]]}]]]))
 
 (defn guide []
   [:div
--- a/src/cljs/rems/user.cljs
+++ b/src/cljs/rems/user.cljs
@@ -14,7 +14,10 @@
     [info-field (text :t.applicant-info/name) name {:inline? true}]))
 
 (defn attributes
-  "A div with a rems.atoms/info-field for every user attribute in the given attributes."
+  "A div with a rems.atoms/info-field for every user attribute in the given attributes.
+   Accepts 
+    - attributes: map, user attributes
+    - invited-user? : boolean, if user is invited, shows different email string"
   [attributes invited-user?]
   (let [language @(rf/subscribe [:language])
         organization-by-id @(rf/subscribe [:organization-by-id])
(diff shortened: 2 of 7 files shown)


marharyta:
@opqdonut I have looked through the UI, looks like this is the only case I see for now where we can replace the Email text

marharyta:
It of coarse also depends on what policy we pick for names

opqdonut:
Yeah I guess that's all there is.

marharyta:
ping @joel and @Macroz should be all good

src/cljs/rems/profile.cljs
```diff
@@ -100,7 +100,7 @@
        {:title (text :t.profile/your-details)
         :always [:<>
                  [user/username (:user identity)]
-                 [user/attributes (:user identity)]]}]]]))
+                 [user/attributes attributes false]]}]]]))
```
opqdonut, line 103: Passing booleans always looks so opaque, but in this case I don't think we want full keyword args just for one boolean... Nothing to fix here, just thinking out loud.

src/cljs/rems/user.cljs
```diff
@@ -15,7 +15,7 @@
 
 (defn attributes
   "A div with a rems.atoms/info-field for every user attribute in the given attributes."
```
opqdonut, line 17: Please document the new parameter briefly

src/cljs/rems/user.cljs
```diff
@@ -30,7 +30,10 @@
            (when-let [mail (:notification-email attributes)]
              [info-field (text :t.applicant-info/notification-email) mail {:inline? true}])
            (when-let [mail (:email attributes)]
-             [info-field (text :t.applicant-info/email) mail {:inline? true}])
+             [info-field (text :t.applicant-info/email) mail {:inline? true}]
```
opqdonut, line 33: this line does nothing, right?

src/cljs/rems/user.cljs
```diff
@@ -53,22 +56,33 @@
...
                          :email "developer@uu.id"
                          :name "Deve Loper"
                          :notification-email "notification@example.com"
                          :organizations [{:organization/id "Testers"} {:organization/id "Users"}]
                          :address "Testikatu 1, 00100 Helsinki"
                          :researcher-status-by "so"
-                         :nickname "The Dev"}])
-   (example "invalid value for researcher status"
+                         :nickname "The Dev"}
+             true])
```
opqdonut, line 78: perhaps this case should only have `:email` and `:name` just like a realistic use case? not sure.

opqdonut (changes requested):
good to merge after fixing the missing doc and the line that does nothing

src/cljs/rems/profile.cljs
marharyta: I also thought so, but figured this is not that big of a deal.

src/cljs/rems/user.cljs
marharyta: Maybe, I probably can create another use case with "invited" user

opqdonut (approved):
:+1:

Issue #2581, opened 2021-03-02
Change applicant


jaakkocsc:
Handler should be able to change the original applicant for a submitted application.
Background: There are cases when an applicant has resigned from his/her position during the application process, and the application is thus left without ownership. A controlled way to tackle the issue would be to allow the handler to change the applicant.

opqdonut:
See also #2244

opqdonut:
We discussed this and came up with two solutions:

A) Implement a "Promote member to applicant" command.

Using this the handler could invite a new member (by email or userid), promote them, and then remove the old applicant (who is now a member.

B) Implement a "Invite applicant by email" command.

Similarly to the invite reviewer, invite decider and invite member actions.

Both approaches make sense, but we decided to _start with option A_.

jaakkocsc:
Tilastokeskus prefers method A.

opqdonut:
#2657 implements option A

opqdonut:
- [ ] email notifications

PR #2657, opened 2021-05-04
promoting member to applicant

opqdonut:
for #2581

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [x] Update manual/ (if applicable)

## Testing


src/clj/rems/application/master_workflow.clj
```diff
@@ -55,6 +56,7 @@
 
 (def ^:private handler-returned-commands
   (disj handler-all-commands
+        :application.command/promote-to-applicant ;; TODO is this right?
```
opqdonut, line 59: changing the applicant of a returned application seems a bit weird, but perhaps the application has been returned to e.g. a user who can't log in any more?

opqdonut:
make-applicant

-----

make-applicant-2

opqdonut:
do you think there should be a browser test?

test/clj/rems/api/test_applications.clj
```diff
@@ -749,6 +749,61 @@
                   "application.command/approve"]
                  (:application/permissions (get-application-for-user app-id decider)))))))))
 
+(deftest test-promote-to-applicant
+  (let [applicant "alice"
+        handler "developer"
+        app-id (test-helpers/create-application! {:actor applicant})]
+    (testing "submit application and add two members"
+      (is (= {:success true}
+             (send-command applicant {:type :application.command/submit
```
Macroz, line 758: Could introduce the `ok-command` variant of `send-command` that does the comparison?

resources/translations/en.edn
```diff
@@ -37,6 +37,7 @@
             :member-email "Email"
             :member-name "Name"
             :name-and-email-required "Name and email are required."
+            :promote-to-applicant "Make applicant"
```
Macroz, line 40: I don't like this mismatch here either i.e. promote vs. make or change. We should decide on one name and be consistent, hopefully up to the event level?

src/cljs/rems/application.cljs
```diff
@@ -654,6 +656,7 @@
         invited-members (:application/invited-members application)
         permissions (:application/permissions application)
         can-add? (contains? permissions :application.command/add-member)
+        can-promote? (contains? permissions :application.command/promote-to-applicant)
         can-remove? (contains? permissions :application.command/remove-member)
         can-invite? (contains? permissions :application.command/invite-member)
         can-uninvite? (contains? permissions :application.command/uninvite-member)]
```
Macroz, line 662: This is starting to feel excessive, combined with the need to name and pass these forwards. Maybe we could have a higher order (container) component `with-permissions` that would add to props this kind of kw-parameters? Or use lower level (container) components that do roles checking?

CHANGELOG.md
```diff
@@ -29,6 +29,8 @@ Changes since v2.17
 ### Additions
 - REMS now supports PostgreSQL version 13. (#2642)
 - Experimental GA4GH Permissions API now allows users to query their own permissions via `/api/permissions/:user`. (#2631)
+- The handler can now change the applicant of a submitted application. (#2581)
+  - This feature can be disabled by adding `:application.command/promote-to-applicant` to `:disable-commands`.
```
Macroz, line 33: Hmm which should be the default?

Macroz (approved):
Looks pretty complete. I think I would like to discuss the points I added before we merge. Like the naming of the command especially. Other things found were really small.

test/clj/rems/application/test_model.clj
```diff
@@ -1044,6 +1044,61 @@
...
+                           :application/applicant {:userid "applicant"}}
+                events (conj events new-event)
+                expected-application (merge expected-application
+                                            {:application/last-activity (DateTime. 6000)
+                                             :application/events events
+                                             :application/applicant {:userid "applicant"}
+                                             :application/members #{{:userid "member1"} {:userid "member2"}}})]
+            (is (= expected-application (recreate expected-application)))))))
+    ;; This isn't currently a planned use case for the event, just
+    ;; documenting this possibility.
```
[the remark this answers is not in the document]
opqdonut: Yeah, it is. The command handler generates a `:user-not-member` error in this case. However as per usual, we shouldn't fail to apply an event.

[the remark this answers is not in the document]
Macroz: Yeah, noticed that much later in the commits. I think the reasoning, not to limit events, is good.

src/clj/rems/application/commands.clj
```diff
@@ -74,6 +74,9 @@
   (assoc CommandBase
          :member {:name s/Str
                   :email s/Str}))
+(s/defschema PromoteToApplicantCommand
```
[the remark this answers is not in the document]
opqdonut: Yeah I wanted to name the command & event differently because I thought the event might get generated from other sources as well. Also the command is more limited in scope than the event. I'd be fine with `ChangeApplicant` or `MemberToApplicant`.

test/clj/rems/api/test_applications.clj
opqdonut: Hmm, good suggestion. One argument against is that with an `(is (= ` you get slightly nicer errors (like "expected `{:success true}` but was `{:success false :errors [{:type :forbidden}]}`)

src/clj/rems/application/commands.clj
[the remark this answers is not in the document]
Macroz: - Promote sounds to me that there can be several applicants after it. 
- Change applicant sounds like the old is not anymore (matches current implementation).
- Member to applicant could support multiple applicants as well.

Maybe change is the best match of these three ideas?

resources/translations/en.edn
opqdonut: I thought promote to applicant is a clear way of expressing the command for devlopers / API users / technical folk, but users are fine with a slightly more fuzzy "Make applicant", especially since they only see the button in context.

src/cljs/rems/application.cljs
opqdonut: I tried to do a refactoring where we'd pass `permissions` directly to `member-info`, but unfortunately we have some special cases like the applicant (can't be removed) or invited members (can't be promoted even if promote-to-applicant is in permissions).

I guess I could push the handling of these special cases inside `member-info` as it already has the `invited-user?` parameter. In that case we could just pass `permissions` down.

resources/translations/en.edn
Macroz: Sounds good, and matches the idea of having only one applicant after.

CHANGELOG.md
opqdonut: config-defaults.edn doesn't disable any commands currently, so I thought I'd follow that pattern. One alternative would be to have an explicit feature flag for this feature if we consider it experimental.

src/clj/rems/application/commands.clj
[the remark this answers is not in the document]
Macroz: MakeApplicant?

resources/translations/en.edn
Macroz: Actually on second thought it is ambivalent

src/clj/rems/application/master_workflow.clj
opqdonut: I felt like changing the applicant of a returned application would have more corner cases to test, but perhaps it's not so.

src/cljs/rems/application.cljs
opqdonut: Yeah, I'll try this refactor out and push it if it feels good.

opqdonut: The refactor came out pretty nicely I think!

src/clj/rems/application/master_workflow.clj
Macroz: At least, there is no need for a `past-applicants`?

test/clj/rems/api/test_applications.clj
Macroz: Hmm how do we manage in the other places with this?

src/cljs/rems/application.cljs
```diff
@@ -617,31 +618,43 @@
   `:element-id`         - id of the element to generate unique ids
   `:attributes`         - user attributes to display
   `:application`        - application
-  `:group?`             - specifies if a group border is rendered
-  `:can-remove?`        - can the user be removed?
-  `:accepted-licenses?` - has the member accepted the licenses?"
-  [{:keys [element-id attributes application group? can-remove? accepted-licenses? invited-user?]}]
+  `:group?`             - specifies if a group border is rendered"
+  [{:keys [element-id attributes application group?]}]
```
opqdonut, line 622: Because member-info gets the whole application, I was able to just push all the logic inside it. It might be better to split this into a view component and a container that does the logic, but then we'd be backin the world of `can-remove?` parameters.

CHANGELOG.md
Macroz: We should let @jaakkocsc decide the default (is it ok to add as default new command, and perhaps disable by hand in the environments CSC manages)

test/clj/rems/api/test_applications.clj
opqdonut: This file uses the `(is (= {:success true} ...)` a lot

src/cljs/rems/application.cljs
Macroz: I think there are some higher order components that would make sense in the future, but no need to complicate this PR. I'll add an idea task.

src/clj/rems/application/master_workflow.clj
opqdonut: Yeah, I don't think there is a need for past-applicants currently, and it's easy to add to the projection later if needed.

src/cljs/rems/application.cljs
Macroz: https://github.com/CSCfi/rems/issues/2659

test/clj/rems/api/test_applications.clj
Macroz: Potentially there could be some machinery for assertions or exception messages here like `assert-response-is-ok` or so but I don't think it makes this much better. Usually there are other fields to check so an `is map` is fine. It seems to be this file that does this a lot but I guess not enough to matter.

Macroz: Also not worth a special macro :)

src/clj/rems/application/commands.clj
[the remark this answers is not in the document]
opqdonut: Renamed to ChangeApplicant!

CHANGELOG.md
opqdonut: Discussed this on chat, agreed with Jaakko that enabling by default is ok. We'll just point people at the changelog and let them test the new feature in their demo environment before deciding whether to disable it or not. Enabling by default is simpler codewise.

Issue #2545, opened 2021-02-02
Fix media queries


marharyta:
[https://github.com/CSCfi/rems/pull/2483#discussion_r565903452](https://github.com/CSCfi/rems/pull/2483#discussion_r565903452) 
We had a discussion regarding the way menu media quesries work / how navigation looks, and we came to conclusion that they need to be tweaked

PR #2660, opened 2021-05-06
Navigation media quesries

marharyta:
# Checklist for author

## Reviewability
- [x] Link to issue  #2545 
- [ ] Note if a related change in rems-deploy repo

## Documentation


marharyta:
ping @Macroz for the review

src/clj/rems/css/styles.clj
```diff
@@ -27,6 +27,14 @@
 (def logo-height-menu (u/px 40))
 (def logo-height (u/px 150))
 (def menu-height 56)
+(def media-breakpoints {:small-screen (u/px 480)
+                        :medium-screen 870
+                        :big-screen 1080})
```
Macroz, line 32: This seems unused?

Macroz: Also one is in pixels, one in plain numbers?

src/clj/rems/css/styles.clj
```diff
@@ -71,12 +79,12 @@
 
 (defn- generate-media-queries []
   (list
-   (stylesheet/at-media {:max-width (u/px 480)}
+   (stylesheet/at-media {:max-width (:xs bootstrap-media-breakpoints)}
                         [(s/descendant :.rems-table.cart :tr)
```
Macroz, line 83: How about we consider some alternative notations for this part:
- `xs`
- `breakpoint-xs`
- `lt-xs`
- `lt-breakpoint-xs`

The last two would require putting `{:max-width ...}` and we could have `lt` and `gt` variants. Clojure allows `<` and `>` in the fn names so we could also have `breakpoint<xs` which would work for me.

I would not like to leave here any unnecessary stuff like repeating bootstrap or media words because I think they do not add value. In the breakpoint definition earlier one could mention that these come from bootstrap one way or another.

Macroz (commented):
Some comments to improve this further. I haven't clicked around yet.

src/clj/rems/css/styles.clj
marharyta: I wanted to store the current media breakpoints so that we still have some track of them untill I am sure new breakpoints work well.

src/clj/rems/css/styles.clj
Macroz: It's just a typical abbreviation for less-than i.e. `<`.

marharyta: as I mentioned on flowdock, it can be tricky, can be confused with. larger than

marharyta: What do you think about my comments from flowdock? Can we stick to just bootstrap media breakpoints?

Macroz: `gt` is usually used in this context to mean greater-than. see e.g. &lt; &gt;

Macroz (changes requested):
After the xs breakpoint, the menu items like logout are cut up until about sm-size when it gets large enough. I think the right breakpoint to use for the menu would be at least sm and not xs. 

This screenshot is with owner and it can vary depending on how many menu items there'll visible per person, per language translations and per number of extra pages configured.

cut

This problem is part of fixing the actual broken layout and not refactoring media queries.

marharyta:
yes, I agree, i can fix that soonish

Macroz (approved):
I think this does do the bare minimum fix so I'll approve.

Further considerations
- the `user-widget` label breakpoint could come from the defined breakpoint map instead of being another fixed constant
- logo and menu items could disappear at the same breakpoint (sm) instead of logo disappearing before
- there could be room to show the logo in the small navbar

marharyta:
thanks @Macroz 
just to follow up:
-logo and menu items could disappear at the same breakpoint (sm) instead of logo disappearing before - it is a big refactoring change, so I would rather do it in a separate PR
-

Macroz:
> thanks @Macroz
> 
> just to follow up:
> 
> * logo and menu items could disappear at the same breakpoint (sm) instead of logo disappearing before - it is a big refactoring change, so I would rather do it in a separate PR

Actually, it's just a one line change.

Issue #1947, opened 2020-01-27
Conditional showing of field: multiselect


Macroz:
We should support visibility of field constraint for multiselect type.

- [x] support for multiselect
- [x] OR condition (the simple choice to start with is to accept multiple values) (see #2163)
- [x] update manual
- [x] browser test

PR #2666, opened 2021-05-11
feat: some more validations for option/multiselect fields

opqdonut:
- validation option keys
  - the frontend was already disallowing spaces in keys, now we also
    do it in the api
- validate multiselect answers

related to #1947

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
  - slight bacwards compatibility break: you'll get an error if you
    edit a form that has invalid option keys unless you fix the keys

## Documentation

## Testing
  - will implement a larger browser test as part of #1947

### DIFF
--- a/src/clj/rems/form_validation.clj
+++ b/src/clj/rems/form_validation.clj
@@ -86,9 +86,8 @@
        :type     :t.form.validation/invalid-value})))
 
 (defn- multiselect-value-valid? [field]
-  (let [allowed-values (set (conj (map :key (:field/options field)) ""))]
-    (every? #(contains? allowed-values %)
-            (form/parse-multiselect-values (:field/value field)))))
+  (let [allowed? (set (conj (map :key (:field/options field)) ""))]
+    (every? allowed? (form/parse-multiselect-values (:field/value field)))))
 
 (defn- invalid-multiselect-error [field]
   (when (= (:field/type field) :multiselect)
--- a/src/cljc/rems/common/form.cljc
+++ b/src/cljc/rems/common/form.cljc
@@ -172,6 +172,7 @@
     (cond
       (str/blank? val)
       {:key :t.form.validation/required}
+
       (not= val (normalize-option-key val))
       {:key :t.form.validation/invalid-value})))
 


src/clj/rems/form_validation.clj
```diff
@@ -85,6 +85,17 @@
       {:field-id (:field/id field)
        :type     :t.form.validation/invalid-value})))
 
+(defn- multiselect-value-valid? [field]
+  (let [allowed-values (set (conj (map :key (:field/options field)) ""))]
+    (every? #(contains? allowed-values %)
```
Macroz, line 90: Wouldn't this work without `contains?` just with `(every? allowed-values ...)`.

src/clj/rems/form_validation.clj
```diff
@@ -85,6 +85,17 @@
...
+            (form/parse-multiselect-values (:field/value field)))))
+
+(defn- invalid-multiselect-error [field]
+  (when (= (:field/type field) :multiselect)
+    (when-not (multiselect-value-valid? field)
+      {:field-id (:field/id field)
+       :type     :t.form.validation/invalid-value})))
```
Macroz, line 97: Odd identation for us?

Macroz: Looks like it is not the only one.

src/cljc/rems/common/form.cljc
```diff
@@ -596,3 +596,22 @@
...
+;; space, and attachments with a comma.
+
+(defn parse-multiselect-values [string]
+  (when-not (empty? string)
+    (str/split string #" ")))
+
+(defn unparse-multiselect-values [ids]
+  (str/join " " ids))
+
+(deftest test-parse-unparse-multiselect-values
```
Macroz, line 646: Where is the canonical test that round-trips these two fns? :smile:

src/cljc/rems/common/form.cljc
```diff
@@ -159,8 +159,16 @@
            (not (<= (first max-length-range) parsed (second max-length-range)))
            :t.form.validation/invalid-value))})
 
+(defn- validate-key [option]
+  (let [val (get option :key)]
+    (cond
+      (str/blank? val)
+      {:key :t.form.validation/required}
+      (str/includes? val " ")
+      {:key :t.form.validation/invalid-value})))
```
Macroz: Maybe a blank line here to separate the cases?

Macroz (approved):
Some test errors but otherwise only small ideas.

src/clj/rems/form_validation.clj
Macroz: I've fixed them all whenever I bump into them. I agree it sometimes looks nice but is yet another exception in the formatting rules.

PR #2697, opened 2021-07-08
Fix bad option name in config-defaults.edn

talavis:
The `:file` option has the wrong name in the example configuration. It should be `:filename`. 

It's great that you report the schema in the logs, otherwise I'd never find it.


Macroz (approved):
Thanks a lot!

Issue #2674, opened 2021-05-19
allow only future dates for entitlement expiration


jaakkocsc:
Only let pick future days in the calendar thingy in UI. API can still set past times.

ossilva:
Hi, would like to contribute to this as a first issue. From what I've gathered the _better-dateinput-polyfill_ package would be used, which enables the functionality with minimal additions. What kind of testing would be desirable?

Macroz:
@ossilva Hi! 

I'm happy to have a look at a PR with this solution. If you make a first proof of concept, I can help with the other details like what kind of tests to add (likely we can add this to an existing browser test).

As far as I can see, [we try to set the `min` property already](https://github.com/CSCfi/rems/blob/master/src/cljs/rems/fields.cljs#L234) but don't likely provide any value there from above. We must make sure it is passed from the field options.

We should make it possible to set the constraint [in the Form editor](https://github.com/CSCfi/rems/blob/master/src/cljs/rems/administration/create_form.cljs#L703). This should not automatically apply to all date fields.

ossilva:
Hi @Macroz, I've created a pull request of the proof of concept. Hopefully it ticks most boxes! All feedback is appreciated and welcome. However, I will only be able to resume work on the feature in a month's time.

PR #2698, opened 2021-07-11
feat: add possibility to set min/max bounds for date-fields

ossilva:
For comments under #2674. The original feature "future dates for entitlement expiration" is is available in a separate feature branch.

both_cal

Possible issues ordred by significance:
- Date bounds only affect the frontend
- Different interpretation of date min - 00:00 vs max - 23:59 not evident
- Timezones are unaccounted for when storing in db
- No common-sense validation of date bounds is performed
- No business logic interaction with other form options
- When specifying one bound, min/max date-bound-fields occupy the same position

Parts requiring review:
- (Swedish) localization
- Accessibility properties
- Documentation

Missing:
- Examples for guide
- Tests

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
- [] Update docs/ (if applicable)

## Testing

## Follow-up


Macroz:
Hi,

Thanks, I'll have a more detailed look in August when I'm back from vacation.

One consideration that came to my mind is that @jaakkocsc probably asked for limits in the application approval and its entitlement expiration date (one can select a custom expiration date for the entitlement). This PR seems to allow setting limits to a date field in the form, which is different but valuable. See this part of the code for what I mean https://github.com/CSCfi/rems/blob/master/src/cljs/rems/actions/approve_reject.cljs#L82.

For this particular date limit feature, I think it would be helpful to be able to set past or future and not a specific concrete date necessarily. Or possibly n days in past or future. To say that this field accepts dates only in the past (e.g., when something happened) would be useful. Or only in the future (e.g., when something should happen). When an administrator creates a form, they may not have a specific date in mind.

Sorry again for the lacking specification. We tend to hash these out while implementing features or just-in-time. I'll let Jaakko clarify too.

Macroz:
> I have taken some time off side projects / OS contributions, sorry for the delay.
> w.r.t. to the PR I could add to the field:
> 
> * past/future toggle
> * input field for duration
> * radio buttons for the unit of time
>   No worries if at this time the feature isn't needed, as I used the work as a means of learning.

Also no worries about delays!

I think the past/future toggle is both simple and useful. It'd make a nice addition. No need for the duration so far. We prefer to integrate small useful PRs/features at a time anyway. 

And if you want to practice more there is similar work available for the entitlement expiration date that is part of the approval action.

Macroz:
We have implemented the other feature in #2713 but like I mentioned earlier, there is still value in the feature that is cooking up in this branch too.

ossilva:
The discussed past-future date feature is in draft PR.
https://github.com/CSCfi/rems/pull/2832

Issue #2650, opened 2021-04-26
Allow inviting handlers via email


jaakkocsc:
Current process for defining a data access committee, aka handlers in REMS, does not allow automatic provision of handlers easily; the provisor must know userid's of the handlers.

For automated provision it should be possible to invite handlers to a workflow by using their email addresses similar to inviting members to an application.

This also links to https://github.com/CSCfi/rems/issues/2244

opqdonut:
Only need to invite handlers for workflows, not for individual applications.

jaakkocsc:
Use case is like this:

A researcher from a university publishes a dataset to CSC services. Upon publishing certain metadata needs to be defined, and data access committee is one such information. Typically the DAC consists of several university's senior staff members and the publisher enters their names and email addresses into the metadata. Each and every dataset may or may not have same DAC.

An automatic process in the CSC service parses the metadata, and populates REMS with sufficient information to create a catalog item for the dataset. A workflow should also be created but at this point there's only the email address(es) of the DAC member(s) aka handler(s).

opqdonut:
Discussed some approaches:

A) Add an `invited-handlers` field to the workflow schema. Inviting a handler adds a record here and sends an email. The invitation can be accepted via a link like `rems/accept-invitation/handler/<workflow-id>/<invitation-token>`. Accessing the link will then perform an update of the workflow, which adds the user account as a handler.

B) Similar to A, but track invitations in a separate table. (Kind of similar to the outbox table.) This will make it easier to support various additional types of invitations later. Will also make it simpler to support viewing all outstanding invitations, and resending or canceling some of them. However, displaying the invited handlers when viewing a workflow will be slightly more work to implement.

C) Implement the approach from ADR 014 (TODO: link) and have separate internal and external user ids in REMS. When a workflow is edited, and a handler is invited by email, REMS creates a virtual user with just an internal id, and assigns them as a handler. An invitation email is sent to the invitee, and when they log in their external user id is associated with this internal id and they immediately are a handler of the workflow.

Problems:
- What about multiple invitations to the same email? Might need to merge users.

opqdonut:
Let's start with implementing B.

opqdonut:
@jaakkocsc says it's enough to start with an API-only implementation.

PR #2706, opened 2021-08-22
Invite handler

Macroz:
Implements #2650.

accepted

Open questions
- Does the workflow need to be active or is any ok? Perhaps it's ok, this is an admin feature anyway.
- Should a mail be sent (to the inviter) when the invitation is accepted?
- Invitation/sent is the time when it's put to outbox, which is earlier than real email sent time but should not be by much. We have no mechanism to easily set the time but one could be developed.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible or **completely new**

## Documentation
- [x] Update manual/ (if applicable)

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks

### DIFF
--- a/resources/translations/sv.edn
+++ b/resources/translations/sv.edn
@@ -2,7 +2,7 @@
  {:accept-invitation {:header "Acceptera inbjudan"
                       :title "Acceptera inbjudan"
                       :success {:workflow "Du är nu handläggaren."}
-                      :errors {:already-member {:workflow "Du redan handläggaren."}
+                      :errors {:already-member {:workflow "Du är redan handläggaren."}
                                :invalid-invitation-type "Acceptering misslyckades."
                                :invalid-token [:div [:p "Acceptering misslyckades."] [:p "Vänligen kontrollera koden för inbjudan " [:code "%1"]]]
                                :invalid-workflow "Acceptering misslyckades."}}
--- a/resources/translations/sv.edn
+++ b/resources/translations/sv.edn
@@ -360,8 +360,10 @@
           :reviewer-invited {:subject "Inbjudan att delta i behandling av ansökan %3" ; TODO check
                              :message "%1,\n\nDu är inbjuden att delta i handling av %2s ansökan %3 om tillgång till material.\n\nDu kan granska ansökningen här: %4"} ; TODO check
           :reviewer-reminder {:application "%1, skickad av %2"
-                              :message "%1,\n\nFöljande ansökningar väntar på din granskning:\n\n%2\n\nDu kan granska ansökningarna här:  %3"
-                              :subject "Ansökningar väntar på granskning"}}
+                              :message "%1,\n\nFöljande ansökningar väntar på din granskning:\n\n%2\n\nDu kan granska ansökningarna här: %3"
+                              :subject "Ansökningar väntar på granskning"}
+          :workflow-handler-invitation {:message "%1,\n\n%2 har bjudit in dig som handläggare av ansökningar till arbetsflöde %3.\n\nDu kan granska arbetsflöden här: %4"
+                                        :subject "Inbjudan att delta i handling av ansökningar"}}
   :footer "CSC – IT-centret för forskning"
   :form {:accepted-licenses "Licenserna har accepterats"
          :actions "Funktioner"
--- a/resources/translations/sv.edn
+++ b/resources/translations/sv.edn
@@ -1,5 +1,12 @@
 {:t
- {:actions {:accept-invitation "Acceptera inbjudan"
+ {:accept-invitation {:header "Acceptera inbjudan"
+                      :title "Acceptera inbjudan"
+                      :success {:workflow "Du är nu handläggaren."}
+                      :errors {:already-member {:workflow "Du redan handläggaren."}
+                               :invalid-invitation-type "Acceptering misslyckades."
+                               :invalid-token [:div [:p "Acceptering misslyckades."] [:p "Vänligen kontrollera koden för inbjudan " [:code "%1"]]]
+                               :invalid-workflow "Acceptering misslyckades."}}
+  :actions {:accept-invitation "Acceptera inbjudan"
             :accept-invitation-already-member "Du deltar redan."
             :accept-invitation-success "Du deltar nu i ansökan."
             :accept-licenses "Acceptera licenserna"
--- a/src/clj/rems/api/invitations.clj
+++ b/src/clj/rems/api/invitations.clj
@@ -2,7 +2,6 @@
   (:require [compojure.api.sweet :refer :all]
             [rems.api.services.invitation :as invitation]
             [rems.api.util :refer [not-found-json-response]] ; required for route :roles
-            [rems.application.events :as events]
             [rems.common.roles :refer [+admin-read-roles+ +admin-write-roles+]]
             [rems.schema-base :as schema-base]
             [rems.util :refer [getx-user-id]]
@@ -28,6 +27,12 @@
    (s/optional-key :invitation/invited-user) schema-base/UserWithAttributes
    :invitation/created DateTime
    (s/optional-key :invitation/sent) DateTime
+   (s/optional-key :invitation/accepted) DateTime
+   (s/optional-key :invitation/workflow) {:workflow/id s/Int}})
+
+(s/defschema AcceptInvitationResponse
+  {:success s/Bool
+   (s/optional-key :errors) [s/Any]
    (s/optional-key :invitation/workflow) {:workflow/id s/Int}})
 
 (def invitations-api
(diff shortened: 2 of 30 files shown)


jaakkocsc:
- Since this is an admin feature I do not see why it should be limited to active workflows only.
- It is very likely that workflow invitations are sent by a script which is incapable of receiving emails, so no need for that. Can an event listener be registered for this?
- Outbox time is sufficient.

Macroz:
> * It is very likely that workflow invitations are sent by a script which is incapable of receiving emails, so no need for that. Can an event listener be registered for this?

I was thinking of the owner who the script is impersonating, or in the future the owner who invites from the UI. We don't send this kind of mail from the application invites either (that the invitation was used). The result should be obvious from the workflow page in the future as well (show handlers and invited handlers).

An event listener can't be registered since this is not an event in an application and all the events we have concern applications so far. In theory everything could be considered an event and this could be added as an event. That's a good extra task that could be added to be done at some point.

resources/translations/sv.edn
```diff
@@ -1,5 +1,12 @@
 {:t
- {:actions {:accept-invitation "Acceptera inbjudan"
+ {:accept-invitation {:header "Acceptera inbjudan"
+                      :title "Acceptera inbjudan"
+                      :success {:workflow "Du är nu handläggaren."}
+                      :errors {:already-member {:workflow "Du redan handläggaren."}
```
aatkin, line 5: based on other translations I guess this should read `Du är redan handläggaren`

src/clj/rems/api/invitations.clj
```diff
@@ -0,0 +1,49 @@
...
+   (s/optional-key :errors) [s/Any]})
+
+(s/defschema InvitationResponse
+  {(s/optional-key :invitation/id) s/Int
+   :invitation/email s/Str
+   :invitation/invited-by schema-base/UserWithAttributes
+   (s/optional-key :invitation/invited-user) schema-base/UserWithAttributes
+   :invitation/created DateTime
+   (s/optional-key :invitation/sent) DateTime
+   (s/optional-key :invitation/workflow) {:workflow/id s/Int}})
```
marharyta, line 36: Should the workflow be optional? What are the use cases if we do not us the workflow?

src/clj/rems/api/services/invitation.clj
```diff
@@ -76,7 +76,7 @@
           {:success false
            :errors [{:key :t.accept-invitation.errors.already-member/workflow}]}
           (do
-            (workflow/edit-workflow! {:id (:workflow/id workflow)
+            (workflow/edit-workflow! {:id workflow-id
```
marharyta, line 79: Should we keep id references consistent? Since you refactored `user-id `to be `userid`, should we do the same to `workflow-id`?

src/clj/rems/api/invitations.clj
Macroz: If it's an invitation to something else, then there'll be something else. So far this does not replace the application invitations but an invitation to ownership has been discussed before (or any other role really).

src/clj/rems/api/services/invitation.clj
Macroz: It's like this because `userid` is legacy and we'd want to remain as consistent as possible so that we can find all the uses. We would ideally have everything follow the `xxx-id` convention or `xxx/id`. The only exception so far is the user identity (that in the db is called `eppn` because of legacy reasons too.

src/clj/rems/email/core.clj
```diff
@@ -53,6 +54,20 @@
                      (remove nil?))]
     (enqueue-email! email)))
 
+(defn- render-invitation-template [invitation]
+  (let [lang (:default-language env)] ; we don't know the preferred languages here since there is no user
+    (cond (:invitation/workflow invitation)
+          (let [workflow (workflow/get-workflow (get-in invitation [:invitation/workflow :workflow/id]))]
+            (assert workflow "Can't send invitation, missing workflow")
+            (template/workflow-handler-invitation-email lang invitation workflow)))))
```
aatkin, line 62: could this be a `when` instead of `cond`? as there is only single condition currently

marharyta: I think this was intentional since @Macroz has explicitly said in the comment an invitation could be for smth else, but good point!

src/clj/rems/email/template.clj
```diff
@@ -29,6 +29,9 @@
    (when-not (empty? (:application/description application))
      (str ", \"" (:application/description application) "\""))))
 
+(defn- format-workflow-for-email [workflow]
+  (:name workflow))
```
aatkin, line 33: this is just a purely stylistic comment as i'm not yet familiar with most of the codebase

could this function be inlined where it's used currently? or would it be too ambiguous that way? just seeing as on row 294 is similarly `(:invitation/name invitation)`

test/clj/rems/api/services/test_invitation.clj
```diff
@@ -0,0 +1,113 @@
...
+        (is (= [] (invitation/get-invitations nil))))
+
+      (testing "creating invitations"
+        (testing "without any type"
+          (is (= {:success false
+                  :errors
+                  [{:type :t.accept-invitation.errors/invalid-invitation-type :workflow-id nil}]}
+                 (invitation/create-invitation! {:userid "owner"
+                                                 :name "Dorothy Vaughan"
+                                                 :email "dorothy.vaughan@nasa.gov"}))))
```
aatkin, line 36: i guess this test refers to performing invitation without `workflow-id`? if so, perhaps this could be better reflected in the test name

src/clj/rems/email/core.clj
Macroz: Yeah, this is just preparing for the future invitation types. There are a few of these scattered around.

src/clj/rems/email/template.clj
Macroz: Yeah for this it is likely overkill. This was mimicking the one for the applications where it is more complicated. I'll simplify this. Also I'm not actually sure there is is a workflow `:name`... isn't it `:title`? :thinking:

test/clj/rems/api/services/test_invitation.clj
Macroz: My thinking is this:
- The actual test name refers to the CRUD functionality which is a collection of create, read/list, then modify, then ... that are interdependent so they work best in a single larger test function.
- Inside, we try creating an invitation without a type, which is an error.
- The only currently supported type of invitation (in this "v2" invitation PR) is a workflow handler invitation.
- There exists a "v1" invitation feature for application members and reviewers that may or may not be ported to this new one.
- In the future, with several invitation types (e.g. owner) there would also be `:userid nil` in the error. It shows you all the values it checked for the "type".
- I chose to have the test try to have an otherwise valid invitation (to a workflow) so that the only thing wrong is the missing type.

aatkin: check, thanks for the explanation! i was trying to find a `:type` in the map and maybe that confused me a little bit. :)

src/clj/rems/email/template.clj
Macroz: Yes it was supposed to be `:title` but this is now fixed.

src/clj/rems/api/invitations.clj
jaakkocsc: At this point we do not allow invites to anything else, and it is debatable whether we should allow invites to other things. But currently not.

src/clj/rems/db/users.clj
```diff
@@ -123,3 +123,7 @@
       format-user
       ;; in case user attributes were not found, return at least the userid
       (assoc :userid userid)))
+
+(defn join-user [x]
```
aatkin, line 127: what's the context behind this function name? it's used in `src/clj/rems/api/services/invitation.clj:14-15` in a function called `join-dependencies`

src/cljs/rems/accept_invitation.cljs
```diff
@@ -0,0 +1,56 @@
...
+            [rems.util :refer [navigate! post!]]))
+
+(rf/reg-event-fx
+ ::enter-page
+ (fn [{:keys [db]} [_ type token]]
+   {:db (assoc db ::type type ::token token)
+    ::accept-invitation [type token]}))
+
+(rf/reg-sub ::type (fn [db] (::type db "")))
+(rf/reg-sub ::token (fn [db] (::token db "")))
```
aatkin, line 17: do we need `type` and `token` in client db? it looks like they are only used in the `::accept-invitation` effect which gets them via event effect

Macroz: This is leftover from the other invitation page that I used as an example. Since we don't need them I'll just remove them. :+1:

src/clj/rems/db/users.clj
Macroz: I added this task so we don't forget to think about this https://github.com/CSCfi/rems/issues/2714

Issue #2701, opened 2021-07-16
Resolve DOIs


jaakkocsc:
Similar to URN resolving, REMS should resolve DOIs in Resources.

jaakkocsc:
Let's implement it like the URN resolver but after the EGA PR. See doi.org for reference.

jaakkocsc:
DOI numbering spec: https://www.doi.org/doi_handbook/2_Numbering.html

aatkin:
example DOIs:

https://doi.org/10.24340/djay-9b5tjc
https://doi.org/10.24340/x3z4-rpmfwx

PR #2707, opened 2021-08-23
feat: support DOI resolving in resources

aatkin:
Implements #2701 

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

---
### catalogue item with DOI resource
Screenshot 2021-08-24 at 17 10 01

### more info links to correct DOI resource URL
Screenshot 2021-08-24 at 17 09 09

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ have notable changes.
 
 Changes since v2.20
 
+### Additions
+- "More info" support for DOI style resources (when :enable-doi flag is set) (#2701)
+
 ## v2.20 "Vattuniemenkatu" 2021-08-24
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ have notable changes.
 Changes since v2.20
 
 ### Additions
-- "More info" support for DOI style resources (when :enable-doi flag is set) (#2701)
+- "More info" support for DOI style resources (when `:enable-doi` flag is set) (#2701)
 
 ## v2.20 "Vattuniemenkatu" 2021-08-24
 
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -300,5 +300,4 @@
  :test-database-url nil
  :accessibility-report false ; should the axe accessibility tooling be included?
  :enable-ega false ; should the EGA features be shown in the UI
- :enable-doi false ; should the DOI features be shown in the UI
- } 
+ :enable-doi false} ; should the DOI features be shown in the UI
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -299,4 +299,6 @@
  :render-css-file? false
  :test-database-url nil
  :accessibility-report false ; should the axe accessibility tooling be included?
- :enable-ega false} ; should the EGA features be shown in the UI
+ :enable-ega false ; should the EGA features be shown in the UI
+ :enable-doi false ; should the DOI features be shown in the UI
+ } 
--- a/src/cljc/rems/common/catalogue_util.cljc
+++ b/src/cljc/rems/common/catalogue_util.cljc
@@ -18,6 +18,16 @@
   (when (and enable-ega (ega-dataset? resid))
     (str (or ega-organization "https://ega-archive.org/datasets/") resid)))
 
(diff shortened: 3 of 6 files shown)


src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -18,6 +18,16 @@
   (when (and enable-ega (ega-dataset? resid))
     (str (or ega-organization "https://ega-archive.org/datasets/") resid)))
 
+;; https://www.crossref.org/blog/dois-and-matching-regular-expressions/
+;; "For the 74.9M DOIs we have seen this matches 74.4M of them. If you need to use only one pattern then use this one."
+(defn- doi? [resid]
+  (and resid (re-find #"(?i)10.\d{4,9}[-._;()/:A-Z0-9]+$" resid)))
```
aatkin, line 24: the regex is slightly modified from the one mentioned in the URL. i removed start of string character to allow matching URLs, in case URL is provided instead of resource id. this allows matching an DOI and extracting only the id, which is then later reconstructed into an URL

resources/config-defaults.edn
```diff
@@ -299,4 +299,6 @@
  :render-css-file? false
  :test-database-url nil
  :accessibility-report false ; should the axe accessibility tooling be included?
- :enable-ega false} ; should the EGA features be shown in the UI
+ :enable-ega false ; should the EGA features be shown in the UI
+ :enable-doi false ; should the DOI features be shown in the UI
+ }
```
Macroz, line 304: We mostly try to avoid the hanging parens (see the original).

CHANGELOG.md
```diff
@@ -8,9 +8,12 @@ have notable changes.
 
 Changes since v2.19
 
-## Fixes
+### Fixes
 - Show the selected but disabled resources in Change resources autocomplete (#2702, #2693)
 
+### Additions
+- "More info" support for DOI style resources (when :enable-doi flag is set) (#2701)
```
Macroz, line 15: Here in the Markdown we could use the backticks like `:enable-doi` to make the flag stand out.

src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -18,6 +18,16 @@
...
+
+(defn- doi-catalogue-item-url [resid {:keys [enable-doi doi-organization]}]
+  (when enable-doi
+    (some->> (doi? resid)
+             (str (or doi-organization "https://doi.org/")))))
```
Macroz, line 29: I think this is slightly clever code and the simple stupid version would be:
```clj
(when enable-doi
  (when-let [doi (doi? resid)]
    (str (or doi-organization "https://doi.org/") doi))
```

Macroz: For some functions, we have had a convention that a function with `?` will only return a boolean. I'm not a strong proponent of that, I am fine with the typical "truthy" but we could name `doi?` perhaps like `parse-doi` to clarify this.

src/cljc/rems/common/catalogue_util.cljc
```diff
@@ -61,6 +73,20 @@
                                                                                                           :ega-organization "https:/ega.org/"}))
         "setting custom EGA organization works"))
 
+  (testing "DOI"
+    (is (= nil (catalogue-item-more-info-url {:resource/ext-id "10.1109/5.771073"} nil {})))
```
Macroz, line 77: We could clarify here as well that without the feature flag it should not match?

Macroz (approved):
Nice work! I think this matches what I was thinking about the ticket should mean.

The review checklist is as complete as I think it should be. If we are really certain about something not being applicable we delete the row and otherwise we leave it unchecked (it reminds us to have a discussion in review if needed).

I added a few minor stylistic notes, you may or may not fix them if you think them as improvements. When you are happy with the PR then you can merge (or let's make 2.20 before that actually) or if you have more questions then just ask.

:+1:

src/cljc/rems/common/catalogue_util.cljc
aatkin: good points!

1. i stumbled upon `some->>` previously and it seemed pretty cool, but there is a bit of extra cognitive load over nested `when`:s
2. true, this function does return either `nil` or the parsed string, so it's not a good candidate for `?`

PR #2708, opened 2021-08-24
refactor: delete generate- prefix from styles

marharyta:
Could not unsee it once I saw it, a quick reafctoring job

## Reviewability
- [ ] Link to issue

## Documentation

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -35,7 +35,7 @@
 
 ;; Fonts of the app
 
-(defn generate-at-font-faces
+(defn- at-font-faces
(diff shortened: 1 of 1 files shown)


Macroz (approved):
I think there was a previous distinction between constant styles and function generated ones. Since we have moved to the direction of more use of `theme-getx` and such, and no constant `def` is left we can take advantage of a convention of `defn-` with forced function calls.

Let's wait a bit before merging this since it would break  #2698.

Macroz:
I will merge this to clean up the PRs. @ossilva this is a small change that you can merge to your branch to solve the conflict.

PR #2637, opened 2021-04-16
EGA 1

Macroz:
For #2466.

# Checklist for author

Remove items that aren't applicable, check items that are done.

## Reviewability
- [x] Link to issue
- [ ] Consider adding screenshots for ease of review (maybe next PR should be the UI?, listed in #2466)

## Backwards compatibility
- [x] API is backwards compatible or completely new
- [x] Config is backwards compatible

## Documentation
- [x] Update changelog if necessary
- [x] Update docs/ (if applicable)
- [ ] Update manual/ (if applicable) (PR once feature complete with UI, listed in #2466)
- [ ] ADR for major architectural decisions or experiments (PR later, listed in #2466)
- [x] New config options in config-defaults.edn

## Testing
- [ ] Complex logic is unit tested
- [ ] Valuable features are integration / browser / acceptance tested automatically

## Follow-up
- [x] New tasks are created for pending or remaining tasks (list in #2466)

PR #2712, opened 2021-08-31
fix: relax schema to allow empty EGA map

Macroz:
Tests seem to be failing randomly with some seeds because some tests generate users that don't have the required attributes. I think it makes sense to relax the DB schema for the EGA user settings. Existing users won't have them either without a migration. Tests should now pass with more certainty.

Original feature in #2637


aatkin (approved):
LGTM

marharyta (approved):

PR #2713, opened 2021-09-01
feat: allow only future dates in entitlement expiration date field

aatkin:
implements #2674

* add `min` value to `approve-reject-view` entitlement expiration date field as **today + 1 day**
  * this adds only date picker support, it is still possible to manually type e.g. `01.01.1989` into the field
* add check to `:application.command/approve` command handler that, if provided, `:entitlement-end` is in future (calculated from `today at 23:59:59`)

## Reviewability
- [x] Link to issue

## Documentation

## Backwards compatibility
- [x] API is **backwards compatible** or completely new

## Testing

---
Screenshot 2021-09-01 at 17 24 19


Macroz (approved):
Nice. Makes me wonder if we should actually limit the text field or API too. But that's up to @jaakkocsc

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Changes since v2.20
 
 ### Additions
 - "More info" support for DOI style resources (when `:enable-doi` flag is set) (#2701)
+- Entitlement expiration date field in application actions will now accept only future dates (#2674)
```
Macroz, line 13: This is sort of wrong isn't it? I mean calendar vs. text field / API usage? We could be very precise here but let's see if we should implement the rest of the possibilities too (see review comment).

aatkin: yes you are right - i'll update the message to mention that it concerns the front-end calendar specifically

marharyta (approved):
🔥

resources/translations/sv.edn
```diff
@@ -33,7 +33,8 @@
                      :invalid-token [:div [:p "Det gick inte att ansluta till ansökningen."] [:p "Vänligen kontrollera koden för inbjudan." [:code "%1"]]]
                      :licenses-not-accepted "Du har inte godkänt användarvillkoren."
                      :organization-does-not-exist "Organisationen \"%1\" hittades inte."
-                     :validation-errors "Var vänlig kontrollera följande:"}
+                     :validation-errors "Var vänlig kontrollera följande:"
+                     :entitlement-end-not-in-future "Utgångsdatumet för rättigheter måste vara i framtiden."}
```
Macroz, line 37: We have kept these in alphabetical order. I'm not sure it's an important concern anymore.

src/clj/rems/application/commands.clj
```diff
@@ -320,6 +321,13 @@
        (mapv (fn [license]
                {:license/id (:id license)}))))
 
+(defn- entitlement-end-not-in-future-error
+  "Checks that entitlement date time, if provided, is in the future."
+  [dt]
+  (when dt
+    (when-not (time/after? dt (time/today-at 23 59 59))
+      {:errors [{:type :t.actions.errors/entitlement-end-not-in-future}]})))
```
Macroz, line 329: Sometimes we add the value e.g. `:entitlement-end "xyz"` to the error map to have an easier time debugging or API error message. We don't always show these in the UI.

Issue #2715, opened 2021-09-07
File size limit for attachments


jaakkocsc:
There should be a limit on how large files can be added as attachments to an application.

- Configurable setting in config.edn?
- Limitation per file or per application, i.e. all files?

PR #2716, opened 2021-09-08
feat: add frontend validation to attachment file size with config

aatkin:
implements #2715 (frontend part)

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] Config is backwards `in-`compatible (breaking change mentioned in changelog)

## Documentation

## Testing

---
Screenshot 2021-09-09 at 16 21 23
---
Screenshot 2021-09-09 at 16 24 17


resources/config-defaults.edn
```diff
@@ -292,6 +292,9 @@
  ;; Show "Assign external id" action button in UI for handlers
  :enable-assign-external-id-ui false
 
+ ;; maximum size of attachment files in bytes (10485760 = 10MB)
+ :attachment-max-size 10485760
```
aatkin, line 296: this will introduce a breaking change since now uploads have basically no size limit. 10MB is a size i have seen used in many web servers and that is why it's in here, but it could also be a good discussion if this should be an optional value

src/cljs/rems/application.cljs
```diff
@@ -228,32 +228,44 @@
...
+    (if (some-> (:attachment-max-size config)
+                (< file-size))
+      (do
+        (rf/dispatch [::set-attachment-status form-id field-id :error])
+        (flash-message/show-default-error! :actions description
+                                           [:div
+                                            [:p [text :t.form/too-large-attachment]]
+                                            [:p [text :t.form/attachment-max-size]
+                                             ": "
+                                             (format-file-size (:attachment-max-size config))]]))
```
aatkin, line 244: we can extract all sorts of information from the attempted upload. would we want to display these in the error message to make it clearer what and why?

src/cljs/rems/fields.cljs
```diff
@@ -335,9 +337,15 @@
...
-                   attachment-types/allowed-extensions-string]}]]))
+       :body-text [:<>
+                   [:div
+                    [:span [text :t.form/upload-extensions]
+                     ": "
+                     attachment-types/allowed-extensions-string]]
+                   [:div
+                    [:span [text :t.form/attachment-max-size]
+                     ": "
+                     (:attachment-max-size config)]]]}]]))
```
aatkin, line 348: reminder to self: add `format-file-size` here as well

resources/config-defaults.edn
Macroz: We can put it in the release notes very visibly at least. @jaakkocsc you can decide if this is a problem or not. We can leave the size undefined / infinite but I think this is large enough not to cause problems. We could talk to the actual instance owners about this.

Macroz: We could at least make sure that `nil` is documented here (as infinite) and that it works (test it).

src/cljs/rems/application.cljs
Macroz, line 244: We can use the `text-format` here. It's slightly better than string concatenation in that the way e.g. `:` is used here can be locale specific. One can put hiccup into the localizations too!

Macroz: We can try and see. It could clarify in the case that someone sent the wrong format / file for example. However, I don't see it as necessary.

Macroz: I think the same would apply for the media type error then. I guess this could be an idea task.

src/cljs/rems/application.cljs
```diff
@@ -228,32 +228,44 @@
...
+      (post! "/api/applications/add-attachment"
+             {:url-params {:application-id application-id}
+              :body file
             ;; force saving a draft when you upload an attachment.
             ;; this ensures that the attachment is not left
             ;; dangling (with no references to it)
-            :handler (fn [response]
+              :handler (fn [response]
                        ;; no need to check (:success response) - the API can't fail at the moment
```
Macroz, line 252: I think this is not going to be true soon ...

src/cljs/rems/fields.cljs
Macroz: And this could use `text-format` too.

src/cljs/rems/util.cljs
```diff
@@ -171,3 +171,11 @@
...
            (.focus elem)
            false))))
+
+(defn format-file-size
+  [size]
+  (condp > size
+    (Math/pow 1024 2) (str (/ size 1024.0) "KB")
+    (Math/pow 1024 3) (str (/ size (Math/pow 1024 2)) "MB")
+    (Math/pow 1024 4) (str (/ size (Math/pow 1024 3)) "GB")
+    (str size "B")))
```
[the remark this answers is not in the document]
Macroz: I think the size declaration and this could decide to use 1000 for kilo because of recent changes (kilo vs. kibi). At least we should be consistent.

Macroz (commented):
Looks good so far. 

You can keep on working and I'll have a look again when you are complete. I'll base my PR on top of this and push it soonish.

src/cljs/rems/util.cljs
```diff
@@ -171,3 +172,38 @@
...
+      "1 KB" 1000
+      "10 KB" (* 1000 10)
+      "1 GB" (* 1000 1000 1000)
+      "1.55 MB" (* 1000 1000 1.55)
+      "0.01 KB" (* 1000 0.012345)
+      "0 KB" 0
+      nil -1
+      nil nil
+      nil {}
+      nil [])))
```
aatkin, line 209: this is a little extra on top of the feature. i tried to cater to the test cases so that it's relatively obvious what is happening, but if it feels unnecessary or maintenance burden, we can drop it. in API we will implement the hard restriction and FE should basically give user some actionable feedback.

Macroz: Yeah, I can see what you mean by a little extra. It's ok to me. I have previously identified this library that could be used instead https://github.com/trhura/clojure-humanize not sure how great it actually would be.

aatkin: maybe if we could utilize other functions from the library it would be better than hand-made implementation. humanized output should be localizable though

Macroz: Yes, `humanized` is lacking in some areas like localization.

src/cljs/rems/config.cljs
```diff
@@ -19,7 +19,7 @@
  (fn [db _]
    (:config db)))
 
-(defn fetch-config! []
+(defn ^:export fetch-config! []
```
aatkin, line 22: my original implementation idea was to convert current config object to `js` and return it to the caller. unfortunately doing a `js->clj` resulted in some nasty error that i wasn't able to comprehend, and exporting `fetch-config!` worked as intended (resetting client configuration).

test/clj/rems/browser_test_util.clj
```diff
@@ -500,3 +500,11 @@
...
         (when (seq @violations)
           (throw (Exception. (str "\n\n\nThere are accessibility violations: " (count @violations) "\n\n\n"))))))))
+
+(defn set-client-config
+  [config]
+  (js-execute "return window.rems.config.test_set_config_BANG_(arguments[0]);" config)
+  (fn []
+    (js-async "var args = arguments;
+               var callback = args[args.length - 1];
+               window.rems.config.fetch_config_BANG_().then(callback);")))
```
aatkin, line 510: copy-paste comment from the other line:

> my original implementation idea was to convert current config object to js and return it to the caller. unfortunately doing a js->clj resulted in some nasty error that i wasn't able to comprehend, and exporting fetch-config! worked as intended (resetting client configuration).

src/cljs/rems/application.cljs
Macroz: API breakage in #2722

src/cljs/rems/util.cljs
```diff
@@ -171,3 +172,38 @@
...
+          (str " " type)))))
+
+(deftest test-format-file-size
+  (testing "should format sizes correctly"
+    (are [expected input] (= expected (format-file-size input))
+      "1 KB" 1000
+      "10 KB" (* 1000 10)
+      "1 GB" (* 1000 1000 1000)
+      "1.55 MB" (* 1000 1000 1.55)
+      "0.01 KB" (* 1000 0.012345)
```
Macroz, line 204: We could add one test case which documents if any rounding happens (or not).

Macroz (approved):
Looks good to me. We can fine tune more after demo.

resources/config-defaults.edn
Macroz: Should we use 10 * 1000 * 1000 here?

Macroz: And we can add the comment `nil` means not limited.

src/cljs/rems/config.cljs
```diff
@@ -28,6 +28,11 @@
   (let [config @(rf/subscribe [::config])]
     (boolean (:dev config))))
 
+(defn ^:export test-set-config!
```
Macroz, line 31: Since we export this always we can just call it `set-config!`, right?

Issue #2665, opened 2021-05-11
Expiry date for drafts


jaakkocsc:
In order to

- prevent drafts piling up
- mitigate for submitting applications long since expired forms

there should be a mechanism to expire drafts. It could be a feature of an workflow.

Feature must not have an effect on submitted applications or returned applications.

opqdonut:
Previously: #2436 #2145 #1524

opqdonut:
I think we've talked in the past about moving from enabled/disable/archived to a more granular model with checkboxes like:
- visible in catalogue?
- can create new draft?
- can submit old draft?
- can resubmit returned application?

opqdonut:
Could also consider a warning like "This catalogue item has been disabled for 6 months, please consider creating a new application."

jaakkocsc:
Decision: we actually delete the events from the database for applications that have not been submitted after a certain set time limit. Check other tables too, like attachments.

aatkin:
We can create a scheduler to delete drafts periodically, and as discussed today, this scheduler should be implemented in REMS

aatkin:
reopening, discovered potential issue when `defstate` returns nil

```
:type clojure.lang.ExceptionInfo
   :message "could not stop [#'rems.application.eraser/expired-application-poller] due to"
```

aatkin:
demo notes:

- add documentation about config feature (duplicate) to `configuration.md`

Issue #2647, opened 2021-04-22
Consider caching more at the service layer


Macroz:
While implementing #2646 we noticed that perhaps we should cache the workflows so that we are not fetching them for every request from the database.

We originally had only the db and api layers and the ,ddleware often goes directly to the db level code e.g. `with-context` that populates the `context/*roles*` that needs to fetch various parts of the database.

We could
- consider adding more comprehensive caching to the service layer
- have explicit cache invalidation capability in the service layer, and reset the cache when modifying
- cache the `get-xxx` and `get-all-xxxs` types of functions
- make middleware, and perhaps other parts of REMS to rather use the service layer than directly db layer

PR #2728, opened 2021-09-16
Expired application eraser

aatkin:
implements #2665

* add new configuration options (default values here):
  * :application-expiration nil
  * takes in configuration map with form `{:application.state/{state} "P90D"}` where duration is expressed as [ISO_8601 duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)
  * multiple application-state/duration combinations can be defined in the same configuration object using application state as key: `{:application.state/draft "P90D" :application.state/closed "P7D"}`
* add new ns `rems.application.eraser`:
  * if enabled, function `remove-expired-applications!` is scheduled to run once every day (determined by server start time)
  * it searches all applications that have a defined state in config and were last updated over `[:application-expiration :application.state/{state}]` duration ago, and removes them, together with attachments
* updated .gitignore (clj-kondo)

to do: email notifications to applicant before application deletion (to be done in another ticket)

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks https://github.com/CSCfi/rems/issues/2729


src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,46 @@
...
+             :application/last-activity (time/minus (time/now) (time/days 90) (time/seconds 1))})))
+
+(defn process-expired-draft-applications! []
+  (let [removed-ids (atom [])
+        failed-ids (atom [])]
+    (doseq [app-id (->> (applications/get-all-unrestricted-applications)
+                        (filter is-expired-draft?)
+                        (map :application/id))]
+      (try
+        (applications/remove-application-data! app-id)
```
aatkin, line 33: command `:application.command/delete` would be nice due to the standard API, but under the hood it would utilize `applications/delete-application!` which performs the same deletion steps, but reloads cache, effectively after each delete. this was deemed, as discussed, not ideal, so I extracted application deletion steps into common functionality which is used here.

src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,46 @@
+(ns rems.application.cleaner
```
aatkin, line 1: name pending. suggestions are welcomed

resources/config-defaults.edn
```diff
@@ -292,10 +292,15 @@
...
 
- ;; maximum size of attachment files in bytes (10000000 = 10MB)
+ ;; Maximum size of attachment files in bytes (10000000 = 10MB).
  ;; nil = no limit
  :attachment-max-size 10000000
 
+ ;; Should draft applications be deleted after a set period of time has passed?
+ :enable-expired-draft-application-processing false
+ ;; Expiration threshold in days as number
+ :draft-application-expiration-threshold 90
```
aatkin, line 302: name(s) pending, suggestions are welcomed

src/clj/rems/application/cleaner.clj
Macroz: Terminator or Eraser?

resources/config-defaults.edn
Macroz: More future proof would be e.g. `:application-expiration {:draft "P3M"}` or so. (https://en.wikipedia.org/wiki/ISO_8601#Durations is used in `:email-retry-period` and could be nice or just `-days`)

Macroz: `:enable-expired-draft-application-processing` sounds like something can happen besides deletion. Also perhaps we can have the treshold/expiration as `nil` = never so there is no need for another flag?

src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,57 @@
...
+  (let [expiration-threshold (:draft-application-expiration-threshold env)
+        last-activity (:application/last-activity application)
+        state (:application/state application)]
+    (and (= :application.state/draft state)
+         (time/before? last-activity (time/minus (time/now) (time/days expiration-threshold))))))
+
+(deftest test-is-expired-draft
+  (with-redefs [env {:draft-application-expiration-threshold 90}]
+    (testing "should identify expired draft application"
+      (are [expected input] (= expected (is-expired-draft? input))
```
Macroz, line 21: We have so far wanted to spell out each `are` as `is` because it gives a more accurate trace including ability to jump to the correct line. We may want to consider that here too.

src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,57 @@
...
+        true {:application/state :application.state/draft
+              :application/last-activity (time/minus (time/now) (time/days 90) (time/seconds 1))}
+        false {:application/state :application.state/draft
+               :application/last-activity (time/now)}
+        false {:application/state :application.state/submitted
+               :application/last-activity (time/minus (time/now) (time/days 90) (time/seconds 1))}))))
+
+(defn process-expired-draft-applications! []
```
Macroz, line 29: We could perhaps name this `process-applications!` as the namespace tells what we are doing here. Or even `remove-expired-applications!` which would be more concrete than "process" which could mean something other than deletion. When we possibly add email sending then perhaps the former is a better name.

src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,57 @@
...
+      (try
+        (applications/remove-application-data! app-id)
+        (swap! removed-ids conj app-id)
+        (catch Throwable t
+          (log/error (str "Removing application with id " app-id " failed: ") t)
+          (swap! failed-ids conj app-id))))
+    (when (seq @failed-ids)
+      (log/info "process-expired-draft-applications! failed to remove applications" @failed-ids))
+    (when (seq @removed-ids)
+      (log/info "process-expired-draft-applications! succesfully removed applications" @removed-ids))
```
Macroz, line 44: Maybe we don't need this "complex logic"? I mean there is no harm in logging once for each application id. This would eliminate the need for the atoms. Also how about making it `log/warn` or `log/error` here because it is a problem somewhere that needs fixing?

src/clj/rems/application/cleaner.clj
```diff
@@ -0,0 +1,57 @@
...
+    (applications/reload-cache!)))
+
+(when (:enable-expired-draft-application-processing env)
+  (mount/defstate expired-draft-poller
+    :start (scheduler/start! process-expired-draft-applications! (.toStandardDuration (time/days 1)))
+    :stop (scheduler/stop! expired-draft-poller)))
```
Macroz, line 50: Hmm, is it guaranteed that `env` is loaded before we get here? I think we have rather had the `when` in the `:start` because then we can be sure the dependencies have been loaded. See e.g. https://github.com/CSCfi/rems/blob/master/src/clj/rems/css/styles.clj#L904

src/clj/rems/db/applications.clj
```diff
@@ -282,11 +283,15 @@
...
 (defn delete-application! [app-id]
   (let [application (get-application app-id)]
     (assert (= :application.state/draft (:application/state application))
             (str "Tried to delete application " app-id " which is not a draft!")))
-  (db/delete-application-attachments! {:application app-id})
-  (db/delete-application-events! {:application app-id})
-  (db/delete-application! {:application app-id})
-  (reload-cache!))
+  (remove-application-data! app-id)
+  (reload-cache!))
```
Macroz, line 297: I'm thinking that this part is slightly wrong in that it reloads the cache (too slow). We should add a task about improving this. Also is it really a `db` layer concern or `service`?

src/clj/rems/db/applications.clj
```diff
@@ -282,11 +283,15 @@
   :start (scheduler/start! reload-cache! (Duration/standardHours 1))
   :stop (scheduler/stop! all-applications-cache-reloader))
 
+(defn remove-application-data!
```
Macroz, line 286: One option for the names would be `delete-application!` and `delete-application-and-reload-cache!`. Just something completely explicit. Now they are sort of close and I wouldn't be sure what they actually do without reading the source. We could also consider adding docstrings.

src/clj/rems/db/applications.clj
```diff
@@ -282,11 +283,15 @@
...
+  [app-id]
+  (db/delete-application-attachments! {:application app-id})
+  (db/delete-application-events! {:application app-id})
+  (db/delete-application! {:application app-id}))
+
 (defn delete-application! [app-id]
   (let [application (get-application app-id)]
     (assert (= :application.state/draft (:application/state application))
             (str "Tried to delete application " app-id " which is not a draft!")))
```
Macroz, line 295: Should this assert actually be in the `remove` fn?

src/clj/rems/handler.clj
```diff
@@ -15,6 +15,7 @@
             [rems.css.styles :as styles]
             [rems.db.catalogue :as catalogue]
             [rems.email.core] ;; to enable email polling
+            [rems.application.cleaner] ;; to enable draft application clean-up job
```
Macroz, line 18: I wonder if it would make more sense to refer to these (states) in e.g. handler (state) (artificially) so that it's obvious that they are needed and this comment becomes unnecessary?

test/clj/rems/application/test_cleaner.clj
```diff
@@ -0,0 +1,80 @@
...
+                           :event/time date-time
+                           :event/actor actor
+                           :application/id app-id
+                           :application/field-values []}
+        submitted-event {:event/type :application.event/submitted
+                         :event/time date-time
+                         :event/actor actor
+                         :application/id app-id}]
+    (events/add-event! created-event)
+    (events/add-event! draft-saved-event)
```
Macroz, line 58: We could avoid these lines if we called `add-dummy-draft-application!` here? Just add one more event.

test/clj/rems/application/test_cleaner.clj
```diff
@@ -0,0 +1,80 @@
...
+    (events/add-event! submitted-event)
+    app-id))
+
+(deftest cleaner
```
Macroz, line 62: We use the `test-` prefix usually (as a convention to make it easy to search for tests).

test/clj/rems/application/test_cleaner.clj
```diff
@@ -0,0 +1,80 @@
...
+          app-id-3 (add-dummy-application! {:date-time (time/minus (time/now) (time/days 120))
+                                            :actor "alice"})
+          expired-app-id (add-dummy-draft-application! {:date-time (time/minus (time/now) (time/days 90) (time/seconds 1))
+                                                        :actor "alice"})
+          expired-app-id-2 (add-dummy-draft-application! {:date-time (time/minus (time/now) (time/days 120))
+                                                          :actor "alice"})]
+      (let [before-apps (applications/get-all-applications "alice")]
+        (is (= 5 (count before-apps)))
+        (is (every? #{app-id-1 app-id-2 app-id-3 expired-app-id expired-app-id-2}
+                    (mapv :application/id before-apps))))
```
Macroz, line 75: We could compare two sets, then it's obvious that they match exactly? Also the `=` comparison usually gives a nice diff on error.

Macroz (commented):
Looking nice!

Maybe a couple points to fix or discuss but nothing major. Good code overall, makes me wake up and notice some (old) things we can improve.

src/clj/rems/application/cleaner.clj
aatkin: i like where you are heading with this. eraser gets my vote

aatkin:
> What sort of setup creates this .clj-kondo dir? The Emacs integration doesn't seem to add it.

I believe it's created by clojure-lsp, which is used by Calva extension for VSCode: https://calva.io/linting/

resources/config-defaults.edn
aatkin: super good point, i didn't realize we had this sort of duration already expressed. :+1: one flag should be enough for this case

src/clj/rems/db/applications.clj
aatkin: i would think `service` is concerned with core business logic, while `db` is concerned with underlying implementation, such as generating correct queries or caching results?

src/clj/rems/handler.clj
aatkin: i'm not 100% sure i understood 😄  this init was a little difficult to place as i thought at first that we would have some unified place where all scheduler jobs would be initialized. do you mean perhaps that there should be a function that is called, such as `init-scheduler`?

test/clj/rems/application/test_cleaner.clj
aatkin: refactored to one function that takes in `draft?` parameter

src/clj/rems/db/applications.clj
Macroz: I'm just not sure about the caching part. If it's completely transparent and performant, it matters less how it works. But here it was too slow.

Also if it just caches the `db` representation then it belongs there. But if it would cache the "joined" full representation it should be in the `service`?

Finally, there is an mismatch here between removing one application and updating all of the cache. That feels simplistic. 

We don't have to touch this in this PR, but it feels like something we need to clarify and refactor in the future. I can add a ticket about this.

src/clj/rems/handler.clj
Macroz: Now we require the namespace to initialize the scheduler, when the handler namespace is required and mounted. That makes it mount too, automatically.

But this comment tells me it's a subtle interaction that could accidentally be removed. We don't need the scheduler var anywhere, it is sort of self-contained otherwise. We could artificially refer to it somewhere to get it started and that would be more explicit than requiring just the namespace with comment.

We can also manually `mount/start` the scheduled jobs, like in `standalone.clj`. That would also be explicit. I think it's perhaps not the correct responsibility of the route handler anyway, but rather the whole application. And it could be nice sometimes to not start them. I think we need to add the manual startup in a fixture for some tests but it would not be too much work.

This smells like another ticket to create.

aatkin: yeah i would think so too. it would make more sense to load up all scheduled jobs in the same place

src/clj/rems/application/eraser.clj
```diff
@@ -0,0 +1,53 @@
...
+    (testing "should identify expired draft application"
+      (is (true? (is-expired-draft? {:application/state :application.state/draft
+                                     :application/last-activity (time/minus (time/now) (time/days 90) (time/seconds 1))})))
+      (is (false? (is-expired-draft? {:application/state :application.state/draft
+                                      :application/last-activity (time/now)})))
+      (is (false? (is-expired-draft? {:application/state :application.state/submitted
+                                      :application/last-activity (time/minus (time/now) (time/days 90))}))))))
+
+(defn remove-expired-applications! []
+  (log/info "Start rems.application.eraser/remove-expired-applications!")
```
Macroz, line 30: Should we add `Finished ...` logging too?

Btw. sometimes I use the `#'remove-expired-applications!` in the message so it "tracks" the fn with a bit more certainty. We haven't used that pattern in REMS I think.

Something like so:
```
(log/info :start #'remove-expired-applications!)
```

src/clj/rems/application/eraser.clj
```diff
@@ -0,0 +1,53 @@
...
+  (doseq [app-id (->> (applications/get-all-unrestricted-applications)
+                      (filter is-expired-draft?)
+                      (map :application/id))]
+    (try
+      (applications/delete-application! app-id)
+      (log/info "Succesfully removed application" app-id)
+      (catch Throwable t
+        (log/error "Failed to remove application" app-id ":" t))))
```
Macroz, line 38: Looking at the `log/error` docs I noticed that I think it wants the `Throwable` as the first argument. I'm not sure this is correct everywhere in our code, so I'll add a small new ticket about checking all of the logging.

test/clj/rems/application/test_cleaner.clj
```diff
@@ -0,0 +1,59 @@
+(ns ^:integration rems.application.test-cleaner
```
Macroz, line 1: Should this be `test-eraser` (including file)?

src/clj/rems/db/applications.clj
Macroz: I added to the existing ticket https://github.com/CSCfi/rems/issues/2647

src/clj/rems/handler.clj
Macroz: New ticket is here https://github.com/CSCfi/rems/issues/2730.

src/clj/rems/application/eraser.clj
Macroz: Added an issue about this https://github.com/CSCfi/rems/issues/2731

Macroz (approved):
Well done!

Found some tiny final remarks and added the new tickets I promised. If you can review the new tickets and add any details you think are missing. 

You can think about the final comments and fix any and merge when you are happy.

resources/config-defaults.edn
```diff
@@ -292,10 +292,17 @@
...
  :enable-assign-external-id-ui false
 
- ;; maximum size of attachment files in bytes (10000000 = 10MB)
+ ;; Maximum size of attachment files in bytes (10000000 = 10MB).
  ;; nil = no limit
  :attachment-max-size 10000000
 
+ ;; Should draft applications be deleted after a set period of time has passed?
+ ;; Use map as value to indicate expired draft applications should be removed, e.g.
+ ;; {:draft "P90D"}
```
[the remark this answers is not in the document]
aatkin: yeah, that is a very nice idea and maps nicely to our data model. :)

src/clj/rems/application/eraser.clj
aatkin: nice! i was actually thinking about how to get the symbol of the executing fn in actual namespace. i tried all sorts of namespace functions but they didn't quite work out. :)

src/clj/rems/application/eraser.clj
```diff
@@ -0,0 +1,63 @@
...
+        (is (true? (is-expired? {:application/state :application.state/draft
+                                 :application/last-activity over-90d-ago})))
+        (is (false? (is-expired? {:application/state :application.state/draft
+                                  :application/last-activity over-7d-ago})))
+        (is (true? (is-expired? {:application/state :application.state/closed
+                                 :application/last-activity over-7d-ago})))
+        (is (false? (is-expired? {:application/state :application.state/closed
+                                  :application/last-activity over-1d-ago})))
+        (is (nil? (is-expired? {:application/state :application.state/rejected
+                                :application/last-activity over-90d-ago})))))))
```
aatkin, line 35: i think returning `nil` from the `is-expired?` function might be enough to indicate that the application state was not matched? it still coerces to `false` when evaluated as boolean, so net effect is the same.

src/clj/rems/application/eraser.clj
```diff
@@ -10,14 +10,15 @@
...
   [application]
-  (let [expiration-threshold (-> env :application-expiration :draft Period/parse)
+  (let [expiration-threshold (some-> env :application-expiration :application.state/draft Period/parse)
         last-activity (:application/last-activity application)
         state (:application/state application)]
-    (and (= :application.state/draft state)
-         (time/before? last-activity (time/minus (time/now) expiration-threshold)))))
+    (when expiration-threshold
+      (and (= :application.state/draft state)
+           (time/before? last-activity (time/minus (time/now) expiration-threshold))))))
```
Macroz: This looks a bit strange when the `expiration-treshold` is not in the same `and`.

src/clj/rems/application/eraser.clj
```diff
@@ -8,29 +8,37 @@
...
-      (is (false? (is-expired-draft? {:application/state :application.state/submitted
-                                      :application/last-activity (time/minus (time/now) (time/days 90))}))))))
+(deftest test-is-expired
+  (with-redefs [env {:application-expiration {:application.state/draft "P90D"
+                                              :application.state/closed "P7D"}}]
+    (testing "should identify expired application by state"
+      (let [over-90d-ago (time/minus (time/now) (time/days 90) (time/seconds 1))
+            over-7d-ago (time/minus (time/now) (time/days 7) (time/seconds 1))
+            over-1d-ago (time/minus (time/now) (time/days 1) (time/seconds 1))]
+        (is (true? (is-expired? {:application/state :application.state/draft
```
Macroz, line 26: We could save some typing and not put the `true?` here?

src/clj/rems/application/eraser.clj
```diff
@@ -8,29 +8,37 @@
...
+                                 :application/last-activity over-90d-ago})))
+        (is (false? (is-expired? {:application/state :application.state/draft
```
Macroz, line 28: And here I think we used to just put `not` but I think there is some variance.

src/clj/rems/application/eraser.clj
Macroz: We can coerce the result in `is-expired?` to be a boolean if we expect that function to be used in many places. Otherwise I'm happy with the `truthy` interpretation and returning `nil` as `false`.

Now these tests make it appear that there should be a difference between `nil` and `false` and I think perhaps we want to avoid it?

Macroz (approved):

src/clj/rems/application/eraser.clj
aatkin: yeah, you are probably right that we might want to avoid assigning this type difference. i kind of thought that semantic difference would justify separating `false` (= application state matches config but is not expired) and `nil` (application state does not match config).

i'll change the code according to the other comment so it should always evaluate to a boolean.

src/clj/rems/application/eraser.clj
Macroz: There are ways around this. 
- This kind of start/end logging is sometimes called tracing and there are I think some libraries to instrument functions with the info more automatically (kind of like aspect oriented programming examples in Java). 
  - https://github.com/clojure/tools.trace
  - https://github.com/clojure-emacs/sayid
  - lots of ...
- One can also have custom loggers / defns that would do this automatically.
- There seems to exist also https://cloojure.github.io/doc/tupelo/tupelo.misc.html#var-fn-info.

Issue #2314, opened 2020-09-10
DUO field type


opqdonut:
Implement a field type that lets the user pick from predefined DUO (Data Use Ontology) codes. There are a ton of these so some sort of search/completion if needed.

- [ ] store DUO codes in the database
  - [ ] machine readable id (`DUO:1234567`), "shorthand", english description (see https://github.com/EBISPOT/DUO/blob/master/duo.csv)
  - [ ] support for other localizations in the future?
  - [ ] populate some example codes in test-data
- [ ] api for fetching DUO codes
- [ ] new `duo` field type
  - [ ] in form editor
  - [ ] when filling in application, fetch DUO codes from API and offer a nice searchable list
  - [ ] user can pick multiple DUO codes

opqdonut:
split out from #776

jaakkocsc:
The latest released version of the DUO OWL files is always available at http://purl.obolibrary.org/obo/duo.owl and can be browsed at http://purl.obolibrary.org/obo/DUO_0000001. Documentation can be found online at https://github.com/EBISPOT/DUO.

jaakkocsc:
A feature request here: https://docs.google.com/document/d/100RDtOI-Tp_pr2AYXUi-bZ_w4Lm61ol83_knC26jBWU/edit#heading=h.o5dw7mwgnxvg

Macroz:
We have now implemented DUO as an experimental feature, a little bit different than in the task here (since we learned new things). Further tasks in #2763.

PR #2732, opened 2021-09-19
Duo codes

Macroz:
Initial data loading and database for #2314 

There is nothing visible to the users / API yet.

# Checklist for author

## Reviewability
- [x] Link to issue

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/db/duo.clj
+++ b/src/clj/rems/db/duo.clj
@@ -1,0 +1,68 @@
+(ns rems.db.duo
+  (:require [clojure.data.csv :as csv]
+            [clojure.java.io :as io]
+            [clojure.string :as str]
+            [clojure.tools.logging :as log]
+            [medley.core :refer [remove-vals]]
+            [rems.db.core :as db]
+            [rems.json :as json]
+            [rems.db.users :as users]
+            [rems.schema-base :as schema-base]
+            [schema.coerce :as coerce]
+            [clj-time.core :as time-core]
+            [schema.coerce :as coerce]
+            [schema.core :as s])
+  (:import [org.joda.time DateTime]
+           rems.DataException))
+
+(s/defschema DuoCode
+  {:id s/Str
+   (s/optional-key :shorthand) (s/maybe s/Str)
+   :label s/Str
+   :description s/Str})
+
+(def ^:private coerce-duo-code
+  (coerce/coercer! DuoCode coerce/string-coercion-matcher))
+
+(def ^:private validate-duo-code
+  (s/validator DuoCode))
+
+(defn- rows->maps [rows]
+  (mapv zipmap
+        (->> (first rows) ; header
+             (map keyword)
+             repeat)
+        (rest rows)))
+
+(defn upsert-duo-code! [duo]
+  (let [data (-> duo
+                 (dissoc :id)
+                 (->> (remove-vals str/blank?))
+                 json/generate-string)]
+    (when (:id duo)
+      (db/upsert-duo-code! {:id (:id duo)
+                            :data data}))))
+
+(defn load-duo-codes! []
+  (let [code-file (io/file "duo.csv")]
+    (log/info "Loading DUO codes from" (.getAbsolutePath code-file))
+    (when (.exists code-file)
+      (with-open [reader (io/reader code-file)]
+        (let [duos (->> reader
+                        csv/read-csv
+                        rows->maps)]
+          (doseq [duo duos]
+            (upsert-duo-code! duo))
+          (log/info "Loaded" (count duos) "codes"))))))
+
+(defn- format-duo-code [duo]
+  (coerce-duo-code (merge {:id (:id duo)}
+                          (json/parse-string (:data duo)))))
+
+(defn get-duo-codes []
+  (->> (db/get-duo-codes)
+       (mapv format-duo-code)))
+
+(comment
+  (load-duo-codes!)
+  (get-duo-codes))
--- a/src/clj/rems/db/duo.clj
+++ b/src/clj/rems/db/duo.clj
@@ -1,68 +1,144 @@
 (ns rems.db.duo
-  (:require [clojure.data.csv :as csv]
+  (:require [clj-http.client :as http]
+            [clojure.data.csv :as csv]
             [clojure.java.io :as io]
+            [clojure.set :refer [difference]]
             [clojure.string :as str]
-            [clojure.tools.logging :as log]
-            [medley.core :refer [remove-vals]]
+            [medley.core :refer [remove-vals update-existing]]
+            [rems.config]
             [rems.db.core :as db]
             [rems.json :as json]
-            [rems.db.users :as users]
-            [rems.schema-base :as schema-base]
+            [rems.schema-base :refer [LocalizedString]]
             [schema.coerce :as coerce]
-            [clj-time.core :as time-core]
-            [schema.coerce :as coerce]
-            [schema.core :as s])
-  (:import [org.joda.time DateTime]
-           rems.DataException))
+            [schema.core :as s]))
 
-(s/defschema DuoCode
+;; The CSV format has English texts only
+(s/defschema DuoCodeCsv
   {:id s/Str
    (s/optional-key :shorthand) (s/maybe s/Str)
    :label s/Str
    :description s/Str})
 
-(def ^:private coerce-duo-code
-  (coerce/coercer! DuoCode coerce/string-coercion-matcher))
+;; The DB format has localized texts
+(s/defschema DuoCodeDb
+  {:id s/Str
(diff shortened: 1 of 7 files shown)


src/clj/rems/db/duo.clj
```diff
@@ -0,0 +1,144 @@
...
+                                :bytes (.toByteArray baos)})))
+          (recur files))
+        files))))
+
+(defn fetch-duo-releases-latest
+  "Fetches the latest DUO codes from the GitHub repository.
+
+  Returns the unpacked, formatted result."
+  []
+  (let [release-response (http/get "https://api.github.com/repos/EBISPOT/DUO/releases/latest"
```
aatkin, line 80: do we plan to utilize versioning in some capacity? eg. `:tag`. or in other words, that what the game plan is now regarding them :)

src/clj/rems/db/duo.clj
```diff
@@ -0,0 +1,144 @@
...
+                    (-> duo
+                        coerce-DuoCodeCsv
+                        (update-existing :label (fn [en] {:en en}))
+                        (update-existing :description (fn [en] {:en en}))))]
+        {:tag release-tag
+         :codes codes}))))
+
+(def simple-codes
+  "Codes that can be added to a resource without any additional questions."
+  #{:DUO:0000004 :DUO:0000006 :DUO:0000011 :DUO:0000015 :DUO:0000016 :DUO:0000018 :DUO:0000019 :DUO:0000021 :DUO:0000029 :DUO:0000042 :DUO:0000043 :DUO:0000044 :DUO:0000045 :DUO:0000046})
```
aatkin, line 98: would it make sense to list these codes with the additional details as comments? like what the code is and so on

src/clj/rems/db/duo.clj
```diff
@@ -0,0 +1,144 @@
...
+        release-body (:body release-response)
+        release-tag (:tag_name release-body)
+        zip-response (http/get (:zipball_url release-body) {:as :stream})
+        csv-bytes (-> zip-response :body (java.util.zip.ZipInputStream.) (read-zip-entries #".*/duo.csv$") first :bytes)]
+    (with-open [reader (io/reader csv-bytes)]
+      (let [codes (for [duo (->> reader csv/read-csv rows->maps)]
+                    (-> duo
+                        coerce-DuoCodeCsv
+                        (update-existing :label (fn [en] {:en en}))
+                        (update-existing :description (fn [en] {:en en}))))]
```
aatkin, line 92: these utilize english translation - do we have a plan for `:fi` and `:sv`? e.g. api update or something. or maybe configuration?

aatkin (commented):
just general comments from me. feature LGTM but some core questions maybe could be addressed

src/clj/rems/db/duo.clj
```diff
@@ -0,0 +1,144 @@
...
+                            :data data}))))
+
+(defn- read-zip-entries
+  "Read the zip-file entries from the `stream` and returns those matching `re`."
+  [stream re]
+  (when stream
+    (loop [files []]
+      (if-let [entry (.getNextEntry stream)]
+        (if (re-matches re (.getName entry))
+          (let [baos (java.io.ByteArrayOutputStream.)]
```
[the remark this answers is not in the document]
Macroz: I guess that one is from Joel and this one is from me. My convention is one from Java heritage. I guess any is good.

src/clj/rems/db/duo.clj
Macroz: Well my idea is that at some point we need to say that the version has changed. Since we have no operational monitoring capability in REMS itself, and this process is now modeled as a manual process (done by developers), we could perhaps just store the version number and alert the developer if it has changed when they check the new codes.

src/clj/rems/db/duo.clj
Macroz: That's something that is in the CSV file / DB so I wouldn't really want to duplicate it. On the other hand as the process is quite manual, we could have a helper function that generates a description of all the codes to be pasted into the source code. These were manually added by me based on what the description seems to say. They need to be manually checked every time one updates the codes (even though the promise is that they will not significantly change).

src/clj/rems/db/duo.clj
Macroz: Latest thinking from me is that we'll load the texts into the database and allow the owners to localize them in the DB with some tooling (if and only if the instances will support more than English). We don't need that tooling for the demo I think, as EGA / THL would be English only. The tooling can be somewhat manual as a first try (maybe with the API) or something in the UI too if we so decide.

We can't use the translation files because we check that every key exists in the code as well as the translations. This would be a problem a) before everything is translated b) because the codes themselves and their translation keys would perhaps be dynamic they can't be checked easily (or without code generation which is an option).

I didn't want to copy the English to the other languages. I'd rather make sure the translation utilities use the default language as fallback.

src/clj/rems/db/duo.clj
aatkin: ok! so in other words fetching latest DUO codes could evolve into the direction of a CLI command? maybe we can help the administrators with some tag fetching automation, if necessary

src/clj/rems/db/duo.clj
aatkin: i'm a little worried about the manual upkeeping. however if we keep the DUO codes in source code, i guess it would be more like a bonus on top to have the descriptions there. perhaps we can link to `http://purl.obolibrary.org/obo/`? e.g. `http://purl.obolibrary.org/obo/DUO_0000007`.

src/clj/rems/db/duo.clj
aatkin: yeah, let's go with the :en for demo. API might be easiest to implement but i don't know how users want to update the translations. most translations i've seen are a really asynchronous process and the results come in every now and then from separate translation service providers.

src/clj/rems/db/duo.clj
Macroz: I think a CLI command won't be very useful here, and it's not what I'm suggesting. If any new type comes along, we must implement features to ask the right follow-up questions or handle the compatibility check correctly? So I'm suggesting we developers follow up on the releases with the commented code and make any necessary changes.

src/clj/rems/db/duo.clj
Macroz: As per the other question, we developers must follow up on the releases and update any codes / descriptions if they change. It would definitely be just a bonus. We can generate a commented version of the codes lists with e.g. the shorthand and label, and manually add any further details we want. I don't see a way of avoiding this.

src/clj/rems/db/duo.clj
Macroz: Currently they can just update the translations file as they wish but I don't want to use that. In the future they would need to translate any localisations in the db table. They already do this for all the items, resources, ... that they create themselves, in the admin editors. I think it should be a rare occurrence that new DUO codes come or the phrasing changes. But if the codes change an operator must manually insert the new localisations into the db, or we can add a DUO tab with the capability.

src/clj/rems/db/duo.clj
aatkin: as continuation to this discussion, i looked up some of the codes in the OWL and i'm not 100% sure why some codes are listed in `simple-codes` and others in `complex-codes`.

example listed as `simple code`:
```
http://purl.obolibrary.org/obo/DUO_0000021

<obo:IAO_0000115 xml:lang="en">This data use modifier indicates that the requestor must provide documentation of local IRB/ERB approval.</obo:IAO_0000115>
```

example listed as `complex-code`:
```
http://purl.obolibrary.org/obo/DUO_0000022

<obo:IAO_0000115 xml:lang="en">This data use modifier indicates that use is limited to within a specific geographic region.</obo:IAO_0000115>
```

what would be the difference between these two from the systems standpoint?

also some codes are not in either listing, like `http://purl.obolibrary.org/obo/DUO_0000017` or `http://purl.obolibrary.org/obo/DUO_0000031` for example. are these obsolete codes?

aatkin: `DUO_0000017` was actually below with the comment `; not concrete types`

Macroz: Well my current understanding is such that you would need to ask some more info for each of the complex types i.e. disease, institute, attachment of document or so. We can get back to the details of what these turn out to be when we go through the details more. I think most of these can't be automatically checked for approval anyway.

In 22, there is effort to specify a geographic area data type etc. so I take it they want this to be asked. In 21 there simply isn't anything (in the OWL). Could be that there should be a question (the text seems like it) but since they didn't model it I simply left it as simple for now.

Macroz: I'm not sure if the not concrete types can be tagged or not. The data model kind of implies to me that one should use the more specific types always.

Issue #2518, opened 2021-01-08
Nicer CLI interface


opqdonut:
- [ ] if an unrecognized CLI command is given, show the usage (instead of starting the server)
- [ ] start the server only if no CLI commands are present

aatkin:
reopening, as issue was discovered with running uberjar:

```
Unrecognized argument: run
```

aatkin:
demo notes:

- let's add `"run"` to the usage (or equivalent of starting the server)

PR #2733, opened 2021-09-21
feat: show CLI help on unrecognized command, otherwise start app

aatkin:
implements #2518 

* if no args are provided, start server app from CLI (`$ lein run`)
* if provided arg is unrecognized, show usage help (same as running `$ lein run help`)
* exit CLI with code `0` after succesful command execution

# Checklist for author

## Reviewability
- [x] Link to issue

---
Screenshot 2021-09-22 at 10 11 11

### DIFF
--- a/src/clj/rems/standalone.clj
+++ b/src/clj/rems/standalone.clj
@@ -116,102 +116,108 @@
   (let [usage #(do
                  (println "Usage:")
                  (println (:doc (meta #'-main))))]
-    (case (first args)
-      "help"
-      (usage)
-
-      ("migrate" "rollback")
+    (if-not (seq args)
+      ;; start app by default if no CLI command was supplied
+      (apply start-app args)
       (do
-        (mount/start #'rems.config/env)
-        (migrations/migrate args (select-keys env [:database-url])))
+        (case (first args)
+          "help"
+          (usage)
 
-      "reset"
-      (do
-        (println "\n\n*** Are you absolutely sure??? Reset empties the whole database and runs migrations to empty db.***\nType 'YES' to proceed")
-        (when (= "YES" (read-line))
+          ("migrate" "rollback")
           (do
-            (println "Running reset")
             (mount/start #'rems.config/env)
-            (migrations/migrate args (select-keys env [:database-url])))))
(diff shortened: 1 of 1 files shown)


src/clj/rems/standalone.clj
```diff
@@ -116,102 +116,108 @@
...
+          "validate"
+          (do
+            (mount/start #'rems.config/env #'rems.db.core/*db*)
+            (when-not (validate/validate)
+              (System/exit 2)))
+
+          ;; show usage if command is unrecognized
+          (usage))
+        ;; exit CLI after succesful command execution
+        (System/exit 0)))))
```
aatkin, line 223: the CLI app is left running indefinitely normally which I thought is a little peculiar. is there a use case that it should "stand by" after running the issued command?

src/clj/rems/standalone.clj
```diff
@@ -116,102 +116,108 @@
   (let [usage #(do
                  (println "Usage:")
                  (println (:doc (meta #'-main))))]
-    (case (first args)
-      "help"
-      (usage)
-
-      ("migrate" "rollback")
+    (if-not (seq args)
```
Macroz, line 119: We could write this as `(if (empty? args) ...)` that could be easier to read.

src/clj/rems/standalone.clj
Macroz: Definitely we can have it exit and it is better like that. This is likely just leftover behavior from years ago. Like ideally we would probably want to have a separate `start` command to start the server and get the usage from the default, but the default behavior to start the server can stay as it is backwards compatible. 

Btw. looks like the other commands mostly exit with code `1` after printing `(usage)`.

Macroz (approved):
Looks good, merge when you want.

src/clj/rems/standalone.clj
aatkin: exiting with code `1` is correct in those cases, as they seem to be an error scenario. in general i think the command should define the exit code and otherwise interpret that everything went according to plan. now that i think of it, the last catch-all `(usage)` should also exit with 1 to make it clear it's an error scenario.

PR #2737, opened 2021-09-22
fix: add run to standalone CLI parameters for uberjar

aatkin:
relates #2518 

* add `run` to supported arguments in standalone CLI to fix issue that occurred in previous PR where running uberjar with parameter `run` caused following error:
```
Unrecognized argument: run
```

# Checklist for author

## Reviewability
- [x] Link to issue

### DIFF
--- a/src/clj/rems/standalone.clj
+++ b/src/clj/rems/standalone.clj
@@ -121,6 +121,9 @@
       (apply start-app args)
       (do
         (case (first args)
+          "run"
+          (apply start-app args)
+          
           "help"
           (usage)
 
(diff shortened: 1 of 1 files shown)


Macroz (approved):
As a follow-up (PR)  we could document this happenstance in the CHANGELOG so that if someone accidentally used an unsupported argument, those are going to break with the release.

Also maybe we can check our docs to see that they use this `run` command so we slowly get rid of the default use?

Macroz:
Let's merge this immediately to get it to use.

PR #2752, opened 2021-10-20
add throttling to frontend keepalive function

aatkin:
currently keepalive function is attached to multiple event sources and triggered without restrictions, which causes a large number of function calls (and re-frame events) being triggered. keepalive needs to be called sufficiently often to make client session stay alive, so we can throttle the amount of calls made to the function.

open questions: is 10s (10000ms) too long?


aatkin:
removed re-frame state from the equation, as we can actually just rely on the `goog.functions/throttle` to make sure that `(keepalive!)` is called every minute if user is still active. this also nicely reduces the amount of code needed!

Issue #2764, opened 2021-11-02
Category DB entity


aatkin:
relates #2604 

- [x] Create `category` database table (migration)
  - `category` should contain columns for `categorydata` (jsonb) and organization id (string). `data` should support a localized title, children categories (ids of nested categories).
- [x] Add `catalogueitemdata (jsonb)` column to `catalogue_item` (migration)
  -  `catalogueitemdata` can be used to add arbitrary data, such as `:categories [...]`
- [x] Create functions to create, retrieve, update and delete category
- [x] Cache categories to memory as they will likely be loaded often

PR #2776, opened 2021-11-04
2764/category db

aatkin:
implements #2764

* create new table `category` with columns:
  * `id`, `categorydata` (jsonb), `organization`
* add `catalogueitemdata` (jsonb) column to `catalogue_item` table
* add `rems.db.category` namespace with related functions to CRUD categories
  * categories are cached to app memory
* add documentation to `development.md` regarding using migratus through CLI

`categorydata` has schema validation following keys:
* (required) `title` localized string
* (optional) `description` localized string
* (optional) `children` [{:id s/Str}]

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Follow-up
- [x] New tasks are created for pending or remaining tasks #2777

### DIFF
--- a/src/clj/rems/db/category.clj
+++ b/src/clj/rems/db/category.clj
@@ -9,7 +9,8 @@
             [medley.core :refer [assoc-some]]))
 
 (s/defschema CategoryData
-  {(s/optional-key :description) schema-base/LocalizedString
+  {:title schema-base/LocalizedString
+   (s/optional-key :description) schema-base/LocalizedString
    (s/optional-key :children) [{:id s/Int}]})
 
 (def ^:private validate-categorydata
@@ -54,10 +55,10 @@
     (reload-cache!))
   (vals @categories-cache))
 
-(defn- get-categorydata [{:keys [children description]}]
-  (-> {}
-      (assoc-some :children children)
-      (assoc-some :description description)))
+(defn- get-categorydata [{:keys [title description children]}]
+  (-> {:title title}
+      (assoc-some :description description)
+      (assoc-some :children children)))
 
 (defn- categorydata->json [category]
   (-> (get-categorydata category)
--- a/src/clj/rems/db/category.clj
+++ b/src/clj/rems/db/category.clj
@@ -1,0 +1,82 @@
+(ns rems.db.category
+  (:require [clojure.tools.logging :as log]
+            [rems.db.core :as db]
+            [rems.json :as json]
+            [rems.schema-base :as schema-base]
+            [rems.common.util :refer [build-index]]
+            [schema.coerce :as coerce]
+            [schema.core :as s]
+            [medley.core :refer [assoc-some]]))
+
+(s/defschema CategoryData
+  {(s/optional-key :description) schema-base/LocalizedString
+   (s/optional-key :children) [{:id s/Int}]})
+
+(def ^:private validate-categorydata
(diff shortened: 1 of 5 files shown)


src/clj/rems/db/category.clj
```diff
@@ -0,0 +1,82 @@
...
+            [rems.db.core :as db]
+            [rems.json :as json]
+            [rems.schema-base :as schema-base]
+            [rems.common.util :refer [build-index]]
+            [schema.coerce :as coerce]
+            [schema.core :as s]
+            [medley.core :refer [assoc-some]]))
+
+(s/defschema CategoryData
+  {(s/optional-key :description) schema-base/LocalizedString
```
aatkin, line 12: would it make more sense to have `description` as mandatory field?

also i think this is also missing `title` field, now that i'm thinking about what we discussed earlier

src/clj/rems/db/category.clj
```diff
@@ -0,0 +1,82 @@
...
+(defn update-category! [id category]
+  (let [id (:id (db/update-category! {:id id
+                                      :organization (get-in category [:organization :organization/id])
+                                      :categorydata (categorydata->json category)}))]
+    (reload-cache!)
+    id))
+
+(defn delete-category! [id]
+  (:id (db/delete-category! {:id id}))
+  (reload-cache!))
```
aatkin, line 82: this delete is now very naive and does not update any relations in categories that might have this category as a child

aatkin: also `catalogueitemdata` could contain this category id? (not implemented in this PR though)

src/clj/rems/db/category.clj
Macroz: I think optional is good, because some user may want to leave it blank and not have anything displayed.

Macroz: `:title` would be good, required, `LocalizedString`

src/clj/rems/db/category.clj
```diff
@@ -0,0 +1,82 @@
...
+          (build-index {:keys [:id] :value-fn identity}
+                       (map (comp coerce-CategoryDb format-category)
+                            (db/get-categories))))
+  (log/info :end #'reload-cache!))
+
+(defn get-category
+  "Get a single category by id"
+  [id]
+  (when (nil? @categories-cache)
+    (reload-cache!))
```
Macroz, line 47: In the DUO/Mondo features I tried to avoid this sort of code by having that one internal endpoint that does the lazy load. We can keep it like this and see if we come up with other needs in the future.

src/clj/rems/db/category.clj
Macroz: I think one OK strategy would be to make sure that everything works even though something is an unknown category. Maybe in this case we show the category in the admin UI (e.g. for the catalogue item) and allow deleting it, but we don't show it in the applicant/normal user UI (just leave it out). Someone can then a) notice that something is missing and b) rectify the situation by creating the category or removing it from the catalogue item.

Macroz (approved):
Looks good. You can fix the title and then merge. We can change things later in the subsequent parts.

src/clj/rems/db/category.clj
aatkin: yeah true! we can extract the loading if the cache needs to be warmed like this elsewhere too.

PR #2781, opened 2021-11-08
feat: load DUO and Mondo from the classpath (JAR)

### LINKED ISSUE #2778: Distribute the supported DUO and Mondo codes in the uberjar
We should distribute the supported DUO and Mondo codes in the uberjar since we don't usually want users to update them themselves.

- [ ] Package duo.edn and mondo.edn into uberjar.
- [ ] Load the EDNs from the classpath instead of file.

Leftover from #2314

Macroz:
Closes #2778 

# Checklist for author

## Reviewability
- [x] Link to issue


aatkin (approved):
:+1:

Issue #2768, opened 2021-11-02
Join category to catalogue item


aatkin:
relates #2604 

on top of #2764 

- [x] Update catalogue item CRUD API to accept `categories [{:id s/Int}]`
  - categories should be updated into `catalogueitemdata` jsonb column
- [x] Join category data to catalogue item API response
  - see `rems.api.services.catalogue/get-localized-catalogue-items`
  - maybe `:expand-categories` like `:expand-names`
- [x] Show category field in catalogue item admin UI read-only view (`/administration/catalogue-items/{id}`)
- [x] Create admin UI field to edit catalogue item category (`/administration/catalogue-items/edit/{id}`) (multiselect)

PR #2784, opened 2021-11-18
Catalogue item category API & admin UI

aatkin:
implements #2768 

* refactor categories DB & API to use `:category/id` instead of `:id`
* update categories API to return all category fields instead of a subset (simpler handling across layers)
* add `:categories` to catalogue item CRUD DB & API as `:categories [schema-base/CategoryId]`
* prevent category deletion if catalogue item depends on it (`rems.api.services.dependencies`)
* add admin UI to create/edit/view categories in catalogue item

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is **backwards compatible** or completely new

## Documentation

## Testing

## Follow-up

---
## Screenshots

**create/edit catalogue item, no categories**

Screenshot 2021-12-01 at 16 05 48
---
**create/edit catalogue item, selected categories**

Screenshot 2021-12-01 at 16 05 59
---
**view catalogue item, no categories**

Screenshot 2021-12-01 at 16 06 46
---
**view catalogue item, selected categories**

Screenshot 2021-12-01 at 16 06 30


resources/sql/queries.sql
```diff
@@ -71,16 +74,23 @@ UPDATE catalogue_item
...
-(formid, resid, wfid, organization, enabled, archived, start)
+(formid, resid, wfid, organization, enabled, archived, start, catalogueitemdata)
 VALUES (:form, :resid, :wfid, :organization,
 --~ (if (contains? params :enabled) ":enabled" "true")
 ,
 --~ (if (contains? params :archived) ":archived" "false")
 ,
 --~ (if (contains? params :start) ":start" "now()")
+,
+/*~ (if (contains? params :catalogueitemdata) */ :catalogueitemdata::jsonb /*~*/ NULL /*~ ) ~*/
```
aatkin, line 93: found this example from another row in the same file for using `NULL` as default value, which is why it looks different from the other default values above

src/clj/rems/api/services/dependencies.clj
```diff
@@ -37,12 +37,16 @@
...
+   (flatten
+    (for [cat (catalogue/get-localized-catalogue-items {:archived true :expand-catalogue-data? true})
+          dep [{:form/id (:formid cat)}
+               {:resource/id (:resource-id cat)}
+               {:workflow/id (:wfid cat)}
+               {:organization/id (:organization cat)}]]
+      (into [{:from {:catalogue-item/id (:id cat)} :to dep}]
+            (mapv (fn [category]
+                    {:from {:catalogue-item/id (:id cat)}
+                     :to (select-keys category [:category/id])}) (:categories cat)))))
```
aatkin, line 49: not sure if this is the most logical place to define dependency `from catalogue to category`, but sort of convenient because catalogue items are already iterated here

src/cljc/rems/common/util.cljc
```diff
@@ -435,4 +435,29 @@
...
+  "Replaces key `k1` with key `k2` in map `m`.
+   If `k1` does not exist in `m`, returns `m`.
+   
+   **Examples:**
+   ```clojure
+   (replace-key {:a 1} :a :b)
+   ;;=> {:b 1}
+   
+   (replace-key {:a 1} :b :c)
+   ;;=> {:a 1}
```
aatkin, line 450: i saw in a blog post that VSCode Calva plugin understands Markdown syntax, and also core Clojure functions have same syntactical highlighting for code examples, so i thought that this could be a nice idea to try out. i don't know how Emacs behaves with MD though

Screenshot 2021-12-01 at 14 47 53

src/clj/rems/api/services/dependencies.clj
aatkin, line 49: not sure if this is the most logical place to define dependency `from catalogue to category`, but sort of convenient because catalogue items are already iterated here

src/cljc/rems/common/util.cljc
Macroz: Emacs can work in a terminal so generally it does not do fancy displaying of things. CIDER doesn't seem to support this out of the box yet but at least it doesn't break anything.

src/cljs/rems/administration/catalogue_item.cljs
```diff
@@ -70,6 +71,11 @@
                         [atoms/link nil
                          (str "/administration/forms/" (:formid catalogue-item))
                          (:form-name catalogue-item)])]
+                     [inline-info-field (text :t.administration/categories)
+                      (when-let [categories (:categories catalogue-item)]
+                        (into [:<>]
+                              (->> (map #(localized (:category/title %)) categories)
+                                   (interpose [:br]))))]
```
Macroz, line 78: Or a comma separated list?

src/cljs/rems/administration/create_catalogue_item.cljs
```diff
@@ -118,22 +126,26 @@
...
-                       :infourl (map-vals :infourl localizations)}}))))))))
+              (when-let [all-categories (get-in db [::categories :data])]
+                {::form {:workflow (item-by-id workflows :id wfid)
+                         :resource (item-by-id resources :id resource-id)
+                         :form (item-by-id forms :form/id formid)
+                         :organization organization
+                         :title (map-vals :title localizations)
+                         :infourl (map-vals :infourl localizations)
+                         :categories (let [all-categories-by-id (build-index {:keys [:category/id]} all-categories)]
+                                       (mapv #(get all-categories-by-id (:category/id %)) categories))}})))))))))
```
Macroz, line 141: We could instead return the fully populated category from the "get catalogue item" API and avoid this join?

Macroz (approved):
:clap: Good stuff. I didn't see any obvious problems. A few places can perhaps be improved a little bit.

I didn't click the UI yet but I'll have a look at the whole feature later.

src/clj/rems/api/services/category.clj
```diff
@@ -14,7 +14,7 @@
   (category/get-categories))
 
 (defn- check-category-children [children]
-  (when-let [not-found (seq (remove (comp category/get-category :category/id) children))]
+  (when-let [not-found (seq (remove #(category/get-category (:category/id %)) children))]
```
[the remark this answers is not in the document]
aatkin: i think there are two points to this:
1) explicit fn is more linear to reason about, no reverse order to think about. we have talked about using "simple" syntactical constructs and while comp is often tempting, it might be off-putting if we mostly use threading macros or anonymous functions
2) performance-wise i would think that anonymous fn is faster because comp needs to dispatch additional function calls. here the difference is likely unnoticeable though

src/cljs/rems/administration/catalogue_item.cljs
aatkin: can do that too! i saw this line break separated list elsewhere and figured to use the same pattern.

for a horizontal list i would almost like to implement a "tag" component that separates the categories visually.. but it would probably look too different from other components and we would need to re-think about other read-only list components as well then.

resources/sql/queries.sql
aatkin: i think it's possible to pass `:catalogueitemdata nil` directly to the db function which would be considered ok by `contains?` as it only checks that the key exists.

in normal scenario this would not happen though, as the service layer call ensures that any value is properly jsonified

src/cljs/rems/administration/create_catalogue_item.cljs
aatkin: yeah, absolutely. i already made the change to category db enrichment and forgot to update this part of the code. :)

src/clj/rems/api/services/category.clj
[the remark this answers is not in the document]
Macroz: I've preferred some `comp` for "idiomatic" case like `(filter (comp #{val} :get-val) xs)`. I think both of the styles fit REMS codebase.

src/cljs/rems/administration/catalogue_item.cljs
Macroz: I guess it depends on the actualy values whether it should be a regular text-like list with comma separation, one row per item / <ul><li> ... or even a small table / group. I think these categories are perhaps long enough to have one row for each though there'd mostly be just one value. So not exactly like tags but I can see uses for both of these styles as common components in the future. Could be a refactor idea ticket.

Macroz (approved):

PR #2789, opened 2021-11-29
fix: reset categories cache after all tests have run

aatkin:
no issue, fixes flaky categories test

* after running all `rems.api.test-categories` tests, reset `rems.db.category/categories-cache` so that categories cache does not leak into other tests. this caused dependencies service tests to fail in flaky manner because categories were fetched from cache instead of DB, unrelated to dependencies tests
* add teardown to `rems.db.testing/caches-fixture` which can be used to rollback/reset caches after test function

# Checklist for author

## Testing

### DIFF
--- a/src/clj/rems/db/category.clj
+++ b/src/clj/rems/db/category.clj
@@ -31,6 +31,9 @@
 
 (def ^:private categories-cache (atom nil))
 
+(defn reset-cache! []
+  (reset! categories-cache nil))
+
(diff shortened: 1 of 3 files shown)


Macroz (approved):
Good find!

Should we actually put this to the common fixtures, so that it is run always the same as the rollback fixture? As some other test could end up creating categories too?

We may have other caches we need to clear too eventually.

Issue #2800, opened 2021-12-16
Catalogue tree leftovers


Macroz:
Continuation from #2796 and #2770.

- [x] bug: when an item is in multiple categories, filtering for it duplicates it more and more
- [ ] schema like in table
- [ ] share / re-use with table and implement refactorings and improvements to both (copypasta)
- [ ] owner can see disabled catalogue item in the tree (deeper than root level)
- [x] check performance of tree grouping in backend
- [x] check performance of rendering
- [ ] catalogue tree browser tests
- [x] admin UI browser tests #2802 
- [ ] refactor create catalogue item browser test to use admin UI and not test helpers
- [ ] animate opening and closing (could be difficult)
- [x] hide category if there are no items in it (yet) #2805 
- [x] the categories should be sortable by the admin to an optimal order and it should be respected while sorting #2808 
- [x] check that the category does not cause a loop when saving a category #2806 
- [x] description should show (optional) #2804 
- [ ] optional feature: count of items under a category (e.g. show "Ordinary (6)" as label)
- [ ] optional feature: show/hide category. return in API but don't show in catalogue if hidden (kind of like catalogue items).

Macroz:
Performance is OK so far.

PR #2808, opened 2021-12-20
Display order

Macroz:
Display order field that decides the ordering of the catalogue tree categories.

Tries to match it in the catalogue list too.

For #2800

edit_world_order
categories_sorted_out
new_world_order

# Checklist for author

## Reviewability
- [x] Link to issue

### DIFF
--- a/resources/translations/sv.edn
+++ b/resources/translations/sv.edn
@@ -100,7 +100,7 @@
                    :disable "Avaktivera"
                    :disabled-and-archived-explanation "Att avaktivera ett föremål gör att det döljs för användarna. Att arkivera det gör att det även döljs från administrationsvyn."
                    :display-archived "Visa arkiverade"
-                   :display-order "Visa ordning"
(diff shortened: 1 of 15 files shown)


resources/translations/sv.edn
```diff
@@ -100,6 +100,7 @@
                    :disable "Avaktivera"
                    :disabled-and-archived-explanation "Att avaktivera ett föremål gör att det döljs för användarna. Att arkivera det gör att det även döljs från administrationsvyn."
                    :display-archived "Visa arkiverade"
+                   :display-order "Visa ordning"
```
aatkin, line 103: might also be "visningsordning" - google translate seems to understand "display order" as verb + subject instead here

Macroz: True, I was thinking too much of the previous translation that actually is in a place where an action is preferable.

aatkin (approved):

PR #2816, opened 2022-01-07
Fix usernames

Macroz:
For #2818 

CLI command to rename users.

There are some additions and fixes to our db functions as required.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing
- [ ] Complex logic is unit tested (should be done while working for [ADR14](https://github.com/CSCfi/rems/blob/master/docs/architecture/014-userids.md))

## Follow-up

### DIFF
--- a/src/clj/rems/db/fix_userid.clj
+++ b/src/clj/rems/db/fix_userid.clj
@@ -1,6 +1,8 @@
 
 (ns rems.db.fix-userid
-  (:require [rems.db.api-key]
+  (:require rems.api.services.dependencies
+            [rems.db.api-key]
+            [rems.db.applications]
             [rems.db.attachments]
             [rems.db.blacklist]
             [rems.db.core]
@@ -133,11 +135,14 @@
                      (= old-userid (:revokedby old))
                      (assoc :revokedby new-userid))]
          :when (not= new old)
-         :let [params [(-> new
-                           (assoc :user (:userid old))
-                           (assoc :resource (:resourceid old))
-                           (assoc :application (:catAppId old))
-                           (assoc :id (:entitlementId old)))]]]
+         :let [params [{:user (:userid new)
+                        :resource (:resourceid new)
+                        :application (:catappid new)
+                        :approvedby (:approvedby new)
+                        :revokedby (:revokedby new)
+                        :start (:start new)
+                        :end (:end new)
+                        :id (:entitlementid new)}]]]
      (do
        (apply prn #'fix-entitlement old params)
        (when-not simulate?
@@ -208,7 +213,7 @@
                                                                      {:userid new-userid}
                                                                      %)))]
          :when (or (not= new old)
-                   (not= modifier (:organization/modifier old)))
+                   (not= modifier (get-in old [:organization/modifier :userid])))
          :let [params [modifier new]]]
      (do
        (apply prn #'fix-organization old params)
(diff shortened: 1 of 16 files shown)


src/clj/rems/standalone.clj
```diff
@@ -223,6 +225,17 @@
...
+        "rename-user"
+        (let [[_ old-userid new-userid] args]
+          (if (not (and old-userid new-userid))
+            (do (usage)
+                (System/exit 1))
+            (do (println "\n\n*** Renaming a user's identity can't easily be undone. ***\nType 'YES' to proceed or anything else to run a simulation only.")
+                (let [simulate? (not= "YES" (read-line))]
+                  (println (if simulate? "Simulating only..." "Renaming..."))
+                  (mount/start #'rems.config/env #'rems.db.core/*db*)
+                  (rems.db.fix-userid/fix-all old-userid new-userid simulate?)))))
```
aatkin, line 237: we could print a reminder line to the user to reboot the server so that caches are updated (as pointed out in the changelog)

src/clj/rems/db/fix_userid.clj
```diff
@@ -0,0 +1,346 @@
+
```
aatkin, line 1: tests would be good to write for this functionality at some point

src/clj/rems/db/fix_userid.clj
```diff
@@ -0,0 +1,346 @@
...
+                               :path-new (:path audit-log)
+                               :method-new (:method audit-log)
+                               :apikey-new (:apikey audit-log)
+                               :userid-new new-userid
+                               :status-new (:status audit-log)})]]]
+     (do
+       (apply prn #'fix-audit-log audit-log params)
+       (when-not simulate?
+         (let [result (apply rems.db.core/update-audit-log! params)]
+           (assert (= 1 (first result)) {:audit-log audit-log :params params :result result})))
```
aatkin, line 94: currently the audit log structure expects a certain format, but nevertheless the rename functionality seems like something that should be contained in audit log. what do you think?

src/clj/rems/standalone.clj
Macroz: Excellent idea!

src/clj/rems/db/fix_userid.clj
Macroz: I was thinking that we write the tests while developing the more comprehensive userid rename support i.e. https://github.com/CSCfi/rems/blob/master/docs/architecture/014-userids.md

For some of the "missing functionality" like update functions we could do it now however they are also not used anywhere else yet.

src/clj/rems/db/fix_userid.clj
Macroz: Do you mean that it should be recorded in the audit log that this script was run? I kind of support that notion but it's also limited to requests to the server in practice. Also we decided to rename the user in the audit log as well.

Macroz: We don't record database migrations, which could be one way to think about this. Also other CLI things may pass the audit log like granting roles, not sure.

src/clj/rems/db/fix_userid.clj
Macroz: So manual testing so far. Script also records everything it did.

src/clj/rems/db/fix_userid.clj
Macroz: Agreed not to add lines to the table. Also noticed that it wouldn't record in detail what it did and that's perhaps more important to save for later questions.

aatkin (approved):

Issue #2814, opened 2021-12-21
Hide invalid edit button


Macroz:
We should hide the edit button from the organization owner if the catalogue item etc. is not theirs to edit. They can access the page and get an error from the API at the moment.

- [x] admin tables
- [ ] details pages

tjb:
hello! i would be happy to take this on. I have some questions though that would need clarification. 

disclaimer i am new to this open source project & clojure but want to learn! 

After taking some time to dig into the roles within the system i noticed that there are `:organization-owner` and `:owner` roles. I am assuming that the `:owner` role is for when a catalogue item is created and designates that user as the creator or "owner" of that catalogue item (please correct me if i am wrong and misunderstanding!). 

In the current implementation of the catalogue item list all buttons are shown which are View, Edit, Enable, and Archive. This is due to a check occurring that allows these buttons to be shown if the user is has `:organization-owner` (regardless of what organization they are an owner of) or `:owner` shown here on [this line](https://github.com/CSCfi/rems/blob/c5d9037e1696bcd0b66987a793770d306dd017a7/src/cljs/rems/administration/catalogue_items.cljs#L141)

Going off the initial spec of hiding the Edit button from users who are not owner's of that catalogue item, should we also hide Enable and Archive from those users as well? If not, should _any_ organization owner have the ability to Enable or Archive a catalogue item? I would think a user can only do that iff the catalogue item belongs to the organization they are an owner of or if the user is the owner of that catalogue item. 

please correct me if i am wrong in any of my assumptions above. i appreciate the guidance and look forward to making a pull request for this.

Macroz:
Hello!

So an `:owner` is like the superuser of the entire instance. They can set it up as they like. An `:organization-owner` is a delegated owner of a part of the instance, an organization. Both are able to view and edit the various domain objects. However, the `:organization-owner` should only be able to modify what belongs in their own organization. Now we have not bothered to show the edit button with the accurate logic yet, but it would be a nice feature (of course it could end up too slow/complicated but we'll see).

As a side-note, the `:handler` user is able to view everything, but not modify. They are the people who handle applications but are not allowed to modify the domain objects (with the exception that they may need to create licenses that sometimes model custom contracts). In some organizations the same user can be an owner and a handler.

[...]

PR #2822, opened 2022-01-15
fix: #2814 hide invalid edit button

tjb:
## Reviewability
- [x] Link to issue

## Testing

## Follow-up

Hello everyone! This PR enhances the Catalogue Items page so Edit, Archive, and Disable buttons only show when the user has the appropriate role *and* is an owner of the organization. Below is a screenshot of how the UI looks with this change. 

Screen Shot 2022-01-15 at 3 18 16 PM

I do have a follow up question (left new task checkbox open above in case). Should this same logic be enforced on the `administration/resources` page? I _believe_ the rules should be the same but I am new to this project and the role system in general so any guidance is appreciated :) 

I am also fairly new to Clojure but want to become an expert in it. Please feel free to give me an constructive criticism as I want to improve and learn from others! 

I would also like to note on the creation of method `roles.cljc#show-when-correct-access`. I attempted to make this a part of `roles.cljc#show-when` however due to the variadic arg I ran into issues with creating it into a multi-arity function. The new function was my workaround. If anyone has insights on how to make this cleaner, I would really appreciate it! I was hoping the multi-arity function would have worked :(

### DIFF
--- a/src/cljc/rems/common/roles.cljc
+++ b/src/cljc/rems/common/roles.cljc
@@ -1,7 +1,8 @@
 (ns rems.common.roles
   (:refer-clojure :exclude [when])
   #?(:clj (:require [rems.context :as context])
-     :cljs (:require [re-frame.core :as rf])))
+     :cljs (:require [re-frame.core :as rf]
+                     [rems.common.util :refer [in?]])))
 
 (def +admin-write-roles+ #{:organization-owner :owner})
 (def +admin-read-roles+ #{:owner :organization-owner :handler :reporter})
@@ -32,7 +33,29 @@
             (boolean (some (set roles) context/*roles*)))
      :cljs (some (set roles) (:roles @(rf/subscribe [:identity])))))
 
+(defn is-with-organization?
+  "Determines if the user is part of an organization based on the passed in ID."
+  [organization-id]
+  #?(:cljs (let [owned-organizations @(rf/subscribe [:owned-organizations])]
+             (->> owned-organizations
+                 (map #(get-in % [:organization/id]))
+                 (in? organization-id)))))
+
+(defn has-permission?
+  "Does the current user have the proper permissions?
+   Permission is defined by having the proper role and is within an org."
+  [roles organization-id]
+  #?(:cljs
+     (and (apply has-roles? roles) (is-with-organization? organization-id))))
+
 #?(:cljs
-   (defn show-when [roles & body]
-     (clojure.core/when (apply has-roles? roles)
-       (into [:<>] body))))
+   (defn show-when
+     ([roles & body]
+      (clojure.core/when (apply has-roles? roles)
+                         (into [:<>] body)))))
+
+#?(:cljs
+   (defn show-when-correct-access
+     [roles organization-id & body]
+     (clojure.core/when (has-permission? roles organization-id)
+                        (into [:<>] body))))
--- a/src/cljc/rems/common/roles.cljc
+++ b/src/cljc/rems/common/roles.cljc
@@ -38,8 +38,8 @@
   [organization-id]
   #?(:cljs (let [owned-organizations @(rf/subscribe [:owned-organizations])]
              (->> owned-organizations
-                 (map #(get-in % [:organization/id]))
-                 (in? organization-id)))))
+                  (map #(get-in % [:organization/id]))
+                  (in? organization-id)))))
 
 (defn has-permission?
   "Does the current user have the proper permissions?
@@ -52,10 +52,10 @@
    (defn show-when
      ([roles & body]
       (clojure.core/when (apply has-roles? roles)
-                         (into [:<>] body)))))
+        (into [:<>] body)))))
 
 #?(:cljs
    (defn show-when-correct-access
      [roles organization-id & body]
      (clojure.core/when (has-permission? roles organization-id)
-                        (into [:<>] body))))
+       (into [:<>] body))))
--- a/src/cljc/rems/common/util.cljc
+++ b/src/cljc/rems/common/util.cljc
@@ -554,3 +554,8 @@
   (is (= [] (replace-key [] :a :b)))
   (is (= {:b 1} (replace-key {:a 1} :a :b)))
   (is (= {:a 1} (replace-key {:a 1} :b :c))))
+
+(defn in?
+  "Returns a true boolean value when the collection contains a element value else nil."
+  [element collection]
+  (some #(= element %) collection))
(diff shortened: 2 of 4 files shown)


Macroz:
Thanks, I'll have a detailed look next week!

Quick answer is that yes the other parts of the administration pages (resources, workflows, licenses, ...) would benefit from this mechanism as well. But let's get this one part looking nice and merged before we apply the same principle there?

I'll help you with naming, coding conventions and the code organization part. I have some clear improvements in mind.

One last thing to tackle would be also be a browser test. These are always a bit more work to do due to the nature of it all, so unless you specifically want to tackle that already, I can make one for you, and then you can quickly see the kind of code required for a browser test in the context of this kind of a pull request. We don't have to test all of the pages, as we'll be using the same mechanism, and the backend does the most important part of the access control anyway.

How does that sound?

tjb:
@Macroz yup im 100% onboard with getting this merge in first before apply it elsewhere in the app. i am also fine with you creating the browser test and adding it to this PR. i would be grateful to have an example to go off of for future issues. 

i look forward to the rest of your feedback next week!

src/cljc/rems/common/roles.cljc
```diff
@@ -32,7 +33,29 @@
             (boolean (some (set roles) context/*roles*)))
      :cljs (some (set roles) (:roles @(rf/subscribe [:identity])))))
 
+(defn is-with-organization?
```
Macroz, line 36: We could name this `is-owner-of-organization?' or so because I think organization membership is a related but different concept. We want this to refer specifically to ownership.

src/cljc/rems/common/roles.cljc
```diff
@@ -32,7 +33,29 @@
...
+  "Determines if the user is part of an organization based on the passed in ID."
+  [organization-id]
+  #?(:cljs (let [owned-organizations @(rf/subscribe [:owned-organizations])]
```
Macroz, line 39: Subscribing here, in a helper function could be slightly risky. It creates a dependency for anyone using this utility to the re-frame subscription, which is not so apparent, and may cause the "wrong component" to be re-rendered when the value changes. We could pass here the owned organizations as a parameter to make this a pure utility function. 

I think the whole `in?` we can reduce to `(contains? (set (map :organization/id owned-organizations) organization-id)` which I don't think we need a utility for. If we so like we could create a subscription for it instead.

src/cljc/rems/common/roles.cljc
```diff
@@ -32,7 +33,29 @@
...
+#?(:cljs
+   (defn show-when
+     ([roles & body]
+      (clojure.core/when (apply has-roles? roles)
+        (into [:<>] body)))))
+
 #?(:cljs
-   (defn show-when [roles & body]
-     (clojure.core/when (apply has-roles? roles)
+   (defn show-when-correct-access
```
Macroz, line 58: If we do the above then this could be `show-when-owner-of` but perhaps that could be a regular `when` clause too. I'm not sure. Let's have a look after you have digested these ideas a bit?

src/cljc/rems/common/util.cljc
```diff
@@ -554,3 +554,8 @@
   (is (= [] (replace-key [] :a :b)))
   (is (= {:b 1} (replace-key {:a 1} :a :b)))
   (is (= {:a 1} (replace-key {:a 1} :b :c))))
+
+(defn in?
```
Macroz, line 558: We could use `contains?` and regular sets so we don't have to use this helper. The sets are rather idiomatic in Clojure and more so in REMS codebase so far.

src/cljc/rems/common/roles.cljc
```diff
@@ -32,7 +33,29 @@
...
 
+(defn is-with-organization?
+  "Determines if the user is part of an organization based on the passed in ID."
+  [organization-id]
+  #?(:cljs (let [owned-organizations @(rf/subscribe [:owned-organizations])]
+             (->> owned-organizations
+                  (map #(get-in % [:organization/id]))
+                  (in? organization-id)))))
+
+(defn has-permission?
```
Macroz, line 44: Again here we could say something like `has-ownership-of?`. I believe we don't need to necessarily pass here other roles for our case (i.e., we don't need to both check for `:organization-owner` role and membership in organization owners as the former is calculated from the latter and the latter is therefore enough). Also isn't it `or` so either you are the (super) `owner` or the owner of the organization? So you don't need to be both.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -138,7 +138,7 @@
                       :sort-value (if checked? 1 2)})
            :commands {:td [:td.commands
                            [view-button (:id item)]
-                           [roles/show-when roles/+admin-write-roles+
+                           [roles/show-when-correct-access roles/+admin-write-roles+ (get-in item [:organization :organization/id])
```
Macroz, line 141: Here we could express the `or` relationship somehow. So either (super) `owner` or an owner of this particular organization. Now it's a bit ambiguous. The "framework" we have with `show-when` doesn't seem to be enough.

test/cljc/rems/common/test_util.cljc
```diff
@@ -14,3 +14,7 @@
 (deftest test-distinct-by
   (is (= [1 2]
          (sort (distinct-by even? [1 2 3 4])))))
+
+(deftest test-in?
+  (is (= true (in? 1 '(1 2 3))))
+  (is (= nil (in? 4 '(1 2 3)))))
```
Macroz, line 20: Just as a sidenote,  we have preferred returning a boolean from a predicate (ends in `?`, so `false` here). This makes it slightly easier to reason about sometimes and the API to outside world is slightly more explicit. Often it makes no difference and truthy/falsey is enough.

src/cljc/rems/common/roles.cljc
Macroz: Oh one more thing as a note we could record for the future as a task. We should probably make sure that the logic is correct wrt. enabled and disabled organizations. Perhaps organization ownership is OK to exist even for a disabled organization, and you could plausibly edit things in the disabled organization. But a disabled organization should not show up as a valid option for "end users". We could have some old bugs related to this.

Macroz (changes requested):
Ok here's my first pass of ideas. I think we need to think about the details a bit.

It's obvious that the existing `show-when` is not very composable. We have an `or` case here etc. Also naming needs a bit of thought.

Can you comment on these a bit and perhaps even try some alternative syntax for the important part i.e. the logic to show or hide the buttons? We can iterate the implementation after it looks good to us.

Macroz:
Hey @tjb,

Do you still want to keep this PR and perhaps try to finish it or shall we adopt it and finish it to get it merged?

Macroz:
I recently implemented functionality for #2814 in #3126. I'll close this PR now. If you want to implement the missing part, you are still welcome! I'll work on other features for a while.

PR #2830, opened 2022-01-19
Remove owner modifier

### LINKED ISSUE #2823: Remove owneruserid, modifieruserid
These columns are not used in all tables, and also we have the audit_log table these days. We can drop these columns.

Macroz:
Closes #2823.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up


resources/migrations/20220114122744-remove-owner-modifier.down.sql
```diff
@@ -0,0 +1,26 @@
+ALTER TABLE resource
+ADD COLUMN ownerUserId varchar(255),
+ADD COLUMN modifierUserId varchar(255);
```
aatkin, line 3: i think PSQL lowercases the column names by default, but maybe we could use the same lowercase naming as in `up` migration for consistency?

aatkin (approved):
looking good! just one tiny comment but good to merge as everything seems to work.

Issue #2821, opened 2022-01-13
Save id attribute mapping on login


aatkin:
To facilitate changing identity provider, REMS needs to provide a way to save user-id mappings so that existing users can continue to log in after identity provider change is completed.

- [ ] create a new table for user mappings with suggested columns: `attribute`, `value`, `userid`
- [ ] create functionality to save user-id mappings on OIDC login

Issue #2771, opened 2021-11-02
Rework login


Macroz:
The `:userid` should be considered an opaque internal id. While logging in, we should look for the user based on the `:oidc-userid-attributes`.

- [x] Search user based on `:oidc-userid-attributes`. The first match will be used.

See https://github.com/CSCfi/rems/blob/master/docs/architecture/014-userids.md

Issue #2377, opened 2020-10-01
Migrate db user attributes away from shibboleth format


opqdonut:
We still store attributes `eppn` (instead of `userid`), `commonName` (instead of `name`) and `mail` (instead of `email`) in the DB due to Shibboleth legacy. Switch to storing user attributes in the API format.

Split out from #1235

PR #2834, opened 2022-01-21
WIP 2821/user id mappings

aatkin:
relates #2821 

# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up


test/clj/rems/test_browser.clj
```diff
@@ -691,6 +691,8 @@
       (is (btu/eventually-visible? {:tag :h1 :fn/has-text "test-handling"})))
     (testing "handler should see the applicant info"
       (btu/scroll-and-click :applicant-info-collapse-more-link)
+      (Thread/sleep 1000) ;; figure out what to wait for
+      (btu/screenshot "after-opening-applicant-info.png")
```
aatkin, line 695: something caused a race condition in CI, as the screenshot without `(Thread/sleep 1000)` showed a partially rendered form. i managed to replicate the test breakage locally by running the entire browser test suite, which co-incidentally broke also two other tests. these tests run completely fine in isolation, and apparently CI has enough beef to run them without failing as well.

src/clj/rems/auth/oidc.clj
```diff
@@ -43,6 +45,33 @@
...
+    (user-mappings/create-user-mapping! {:from (name attr)
+                                         :from-value userid
+                                         :to-value (get-userid (apply dissoc id-data mapped-attrs))})))
+
+(defn should-map-userid?
+  "Determine if user-id should be mapped to `user_mappings` table.
+   user-id is mappable if:
+   - user-id is not already mapped,
+   - Token contains at least one attribute defined in `:oidc-mapped-userid-attributes` env variable, and
+   - Token contains at least one attribute not defined in `:oidc-mapped-userid-attributes` env variable."
```
aatkin, line 66: this is currently the only function with doc string. i'm not sure if it's useful though? the related code itself is quite short.

src/clj/rems/auth/fake_login.clj
```diff
@@ -1,23 +1,38 @@
...
+            [rems.config :refer [env]]
+            [rems.db.test-data-users :refer [+fake-user-data+ +demo-user-data+]]
+            [rems.common.util :refer [assoc-some-in]]
+            [rems.ga4gh :as ga4gh]
             [ring.util.response :refer [redirect]]))
 
+(defn get-fake-login-users []
+  (condp = (:fake-authentication-data env)
+    :test +fake-user-data+
+    :demo +demo-user-data+))
```
aatkin, line 15: this function allows easy users substitution from outside if needed (see tests for example in `test/clj/rems/api/test_user_settings.clj`)

test/clj/rems/testing_util.clj
```diff
@@ -54,3 +54,8 @@
                                         (organizations/get-all-organization-roles ~user)
                                         (applications/get-all-application-roles ~user))]
      ~@body))
+
+(defmacro with-fake-login-users [users & body]
+  `(with-redefs [rems.auth.fake-login/get-fake-login-users (constantly ~users)]
+     ~@body))
+
```
aatkin, line 61: i need to pay a little bit attention to the empty new lines, as for some reason my VSCode stopped correctly adding new lines at the end of the file. could be a broken plugin.

Screenshot 2022-01-25 at 17 55 15

aatkin:
for some reason, my editor (VSCode) stopped properly appending new line at EOF. might be a broken plugin. i'll try to add these manually as i notice.

src/clj/rems/auth/fake_login.clj
```diff
@@ -1,23 +1,38 @@
...
+        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))]
+    (when (oidc/should-map-userid? id-data) (oidc/create-user-mapping! id-data))
+    (-> (redirect "/redirect")
+        (assoc :session session)
+        (assoc-in [:session :access-token] (str "access-token-" username))
+        (assoc-in [:session :identity] (merge {:eppn (or (oidc/get-mapped-userid id-data) (oidc/get-userid id-data))
+                                               :commonName (some id-data [:name :unique_name :family_name :commonName])
+                                               :mail (some id-data [:email :mail])}
+                                              extra-attributes))
+        (assoc-some-in [:session :identity :researcher-status-by] (ga4gh/passport->researcher-status-by id-data)))))
```
aatkin, line 35: this feature is mostly similar to OIDC functions but maybe it could also be extracted into common functionality?

src/clj/rems/db/test_data.clj
```diff
@@ -23,8 +23,9 @@
 ;;; generate test data
 
 (defn- create-users-and-roles! [users attrs]
-  (doseq [attr (vals attrs)]
-    (create-user! attr))
+  (doseq [user (vals users)]
+    (when-let [data (get attrs user)]
+      (create-user! data)))
```
aatkin, line 28: one consequence of this change is that, for example, `+fake-user-data+` can contain entries that do not exist in `+fake-users+`. in the scope of this PR `elixir-alice` is a user that only needs to exist as token data because they are mapped into a existing user.

src/clj/rems/db/user_mappings.clj
```diff
@@ -0,0 +1,46 @@
...
+            [rems.db.core :as db]
+            [rems.json :as json]
+            [medley.core :refer [deep-merge]]
+            [schema.coerce :as coerce]
+            [schema.core :as s]))
+
+(s/defschema UserMappingsDataDb
+  {:from s/Str
+   :from-value s/Str
+   :to-value s/Str})
```
aatkin, line 12: the namings are quite ad-hoc. maybe we can come up with better ones? in my head, `from` is the value/attribute we want to use to get the target (`to-value`).

test/clj/rems/testing_util.clj
```diff
@@ -54,3 +54,8 @@
                                         (organizations/get-all-organization-roles ~user)
                                         (applications/get-all-application-roles ~user))]
      ~@body))
+
+(defmacro with-fake-login-users [users & body]
+  `(with-redefs [rems.auth.fake-login/get-fake-login-users (constantly ~users)]
+     ~@body))
```
aatkin, line 60: i'm not happy to use a macro, but this seemed like a plausible solution to succinctly swap in desired login data wherever needed. we have some other similar wrappers but in the current form the composing aesthetic is a bit XML-like.

src/clj/rems/db/test_data.clj
Macroz: We could also have one user that does not map to an existing user. Since that'll be the case often and we could make sure a new user is created fine.

src/clj/rems/db/test_data_users.clj
```diff
@@ -39,7 +38,8 @@
    "organization-owner1" {:eppn "organization-owner1" :mail "organization-owner1@example.com" :commonName "Organization Owner 1"}
    "organization-owner2" {:eppn "organization-owner2" :mail "organization-owner2@example.com" :commonName "Organization Owner 2" :organizations [{:organization/id "organization2"}]}
    "owner" {:eppn "owner" :mail "owner@example.com" :commonName "Owner"}
-   "reporter" {:eppn "reporter" :mail "reporter@example.com" :commonName "Reporter"}})
+   "reporter" {:eppn "reporter" :mail "reporter@example.com" :commonName "Reporter"}
+   "elixir-alice" {:eppn "alice" :elixirId "elixir-alice" :mail "alice@elixir-europe.org" :commonName "Alice Applicant" :organizations [{:organization/id "default"}] :nickname "In Wonderland" :researcher-status-by "so"}})
```
Macroz, line 42: So what I was saying is that we probably can't control that `"elixirId"` comes from IdP. Likely it will be named `"sub"`. That's why we need the rename, if we internally want to have it be something else.

src/clj/rems/db/user_mappings.clj
```diff
@@ -0,0 +1,46 @@
...
+
+(def ^:private coerce-UserMappingsDataDb
+  (coerce/coercer! UserMappingsDataDb coerce/string-coercion-matcher))
+
+(defn- format-user-mappings [user-mappings]
+  (let [usermappingsdata (-> (json/parse-string (:usermappingsdata user-mappings))
+                             coerce-UserMappingsDataDb)
+        {:keys [from from-value to-value]} usermappingsdata]
+    {from {from-value to-value}}))
```
Macroz, line 21: Maybe the mapping object should be just what was in the database. And then the index structure we cache can be different? So return the coerced `usermappingsdata` as is here? Just seems a bit odd to have it like this.

src/clj/rems/db/user_mappings.clj
```diff
@@ -0,0 +1,46 @@
...
+(def ^:private user-mappings-cache (atom nil))
+
+(defn reset-cache! []
+  (reset! user-mappings-cache nil))
+
+(defn reload-cache! []
+  (log/info :start #'reload-cache!)
+  (let [user-mappings (->> (db/get-user-mappings)
+                           (map format-user-mappings)
+                           (reduce deep-merge {}))]
```
Macroz, line 33: Here I'm thinking that we could `(index-by [:from :from-value] mappings)`.

resources/config-defaults.edn
```diff
@@ -52,6 +56,14 @@
...
  ;; order and the first non-empty one used.
  :oidc-userid-attribute "sub"
+ ;; Which id-token attributes to use for user-id mapping. Must be a
+ ;; list of attributes ["elixirId"], which is searched in order
+ ;; and the first non-empty one is used. user-id mapping is created
+ ;; when OIDC token contains one mapped attribute and one non-mapped
+ ;; attribute (see :oidc-userid-attribute configuration). Mapped
+ ;; user-id is fetched and used for subsequent logins.
+ ;; nil => mapping disabled
+ :oidc-mapped-userid-attributes nil
```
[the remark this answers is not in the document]
Macroz: NB: I noticed that `:name` is used for the extra attributes as the localized name, so maybe `:alias`, `:rename` or so would work better and is something like Clojure requires.

test/clj/rems/testing_util.clj
Macroz: Since they don't come from the database, then I guess the `with-redefs` is the way to go. I guess we use it mostly plain but this wrapper seems OK. Maybe it can be more generic once we put the OIDC/login refactoring in so that it can be used no matter what the login method was (and tests can do fun stuff).

Macroz (commented):
Good stuff! I wrote here some food for thought first before doing anything.

resources/migrations/20220120121117-user-mappings.up.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE user_mappings (
+  id serial NOT NULL PRIMARY KEY,
+  usermappingsdata jsonb
```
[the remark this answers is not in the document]
aatkin: yeah true - i was thinking that since we have already used the blob elsewhere that it would be a natural fit here too. but OTOH there is nothing else than the blob so it could be either this way, or as you wrote, 3 column table with indexing. the DB version could be a better fit here

src/clj/rems/auth/fake_login.clj
aatkin: extracted identity attributes into shared functions, this should be the "beef" in the feature scope

src/clj/rems/db/test_data.clj
aatkin: added extra user into `+fake-user-data+` that doesn't map into any existing user :+1:

src/clj/rems/db/user_mappings.clj
aatkin: removed cache in favor of indexed db table (and query)

src/clj/rems/db/user_mappings.clj
aatkin: removed cache in favor of indexed db table (and query)

src/clj/rems/db/user_mappings.clj
aatkin: renamed these, good suggestions :+1:

resources/config-defaults.edn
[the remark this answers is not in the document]
aatkin: removed the extra `:oidc-mapped-userid-attributes` and renamed existing `:oidc-userid-attribute` -> `:oidc-userid-attributes` and also refactored the format how attributes are specified:
```clojure
:oidc-userid-attributes ["sub" "atr"] ;; before
:oidc-userid-attributes [{:attribute "sub"} {:attribute "atr"}] ;; after
```

resources/migrations/20220120121117-user-mappings.up.sql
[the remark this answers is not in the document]
aatkin: used indexed table instead of cache solution, which i think simplifies the solution and makes it more readable

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +27,89 @@
...
+        (let [userid (get-userid id-data)]
+          (users/add-user-raw! userid id-data)
+          userid))))
+
+(defn save-user-mappings! [id-data userid]
+  (let [attrs (->> (getx env :oidc-userid-attributes)
+                   (map #(or (:rename %) (:attribute %)))
+                   (map keyword))]
+    (doseq [[attr value] (select-keys id-data attrs)
+            :when (not= value userid)]
```
aatkin, line 96: not sure about if the `:when` is needed - i'm thinking that those mappings are somewhat redundant:

```clojure
{:userid "123"
 :ext-id-attribute "xyzzy"
 :ext-id-value "123"}
```

src/clj/rems/middleware.clj
```diff
@@ -63,8 +63,8 @@
   [handler]
   (fn [request]
     (binding [context/*user* (keywordize-keys (:identity request))]
-      (when context/*user*
-        (users/add-user-raw! (get-user-id) context/*user*))
+      ;; (when context/*user*
+      ;;   (users/add-user-raw! (get-user-id) context/*user*))
```
aatkin, line 67: this probably should be removed now that user is created on actual login and not on every request?

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +27,89 @@
...
+      (user-mappings/create-user-mapping! {:userid userid
+                                           :ext-id-attribute (name attr)
+                                           :ext-id-value value}))))
+
+(defn get-user-attributes [id-data]
+  (let [identity-base {:eppn (get-userid id-data)
+                       ;; need to maintain a fallback list of name attributes since identity
+                       ;; providers differ in what they give us
+                       :commonName (some id-data [:name :unique_name :family_name])
```
Macroz, line 105: Let's make this configurable also. It seems strange to have these hard-coded.

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +27,89 @@
...
+                       :mail (:email id-data)}
+        extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))]
```
Macroz, line 107: This could in theory support rename, and the email etc. could be handled through something? We can pull these out as an improvement ticket.

src/clj/rems/middleware.clj
Macroz: Yeah, I think so.

Macroz (commented):
Looks pretty good. I'll do these tunings and then we can see what we merge.

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +27,89 @@
...
+      (user-mappings/create-user-mapping! {:userid userid
+                                           :ext-id-attribute (name attr)
+                                           :ext-id-value value}))))
+
+(defn get-user-attributes [id-data]
+  (let [identity-base {:eppn (get-userid id-data)
+                       ;; need to maintain a fallback list of name attributes since identity
+                       ;; providers differ in what they give us
+                       :commonName (some id-data [:name :unique_name :family_name])
+                       :mail (:email id-data)}
```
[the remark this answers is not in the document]
Macroz: I added `oidc-email-attributes`.

src/clj/rems/auth/oidc.clj
Macroz: I think the motivation to include these plain mappings would be to store that the `userid` was from `xyzzy` sometime.

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +26,71 @@
...
+(defn save-user-mappings! [id-data userid]
+  (let [attrs (get-userid-attributes id-data)]
+    (doseq [[attr value] attrs
+            :when (not= value userid)]
+      (user-mappings/create-user-mapping! {:userid userid
+                                           :ext-id-attribute (name attr)
+                                           :ext-id-value value}))))
+
+(defn get-user-attributes [id-data]
+  (let [identity-base {:eppn (get-userid id-data)
```
aatkin, line 82: `:eppn` seems a little redundant here (looking at fake-login usage) but we can refactor it with #2771 if needed

src/clj/rems/auth/oidc.clj
```diff
@@ -23,25 +26,71 @@
...
+                       ;; need to maintain a fallback list of name attributes since identity
+                       ;; providers differ in what they give us
+                       :commonName (some id-data (:oidc-name-attributes env))
+                       :mail (some id-data (:oidc-email-attributes env))}
```
aatkin, line 86: would it make sense to document these attribute mappings in an API test as well?

src/clj/rems/auth/oidc.clj
Macroz: Or here https://github.com/CSCfi/rems/issues/2377

src/clj/rems/auth/oidc.clj
Macroz: Yes, we should add new tests for these. Something I noticed while doing these is that the fake login at the moment does not store the user attributes into the database and we should minimize the differences so it should do so.

Macroz: Tests should be added now.

PR #2848, opened 2022-02-22
ADR 019: DUO

aatkin:
relates #2763 

# Checklist for author

Remove items that aren't applicable, check items that are done.

## Reviewability
- [x] Link to issue

## Documentation
- [x] ADR for major architectural decisions or experiments

## Follow-up
- [ ] New tasks are created for pending or remaining tasks

Issue #2791, opened 2021-12-07
DUO checker for handler


jaakkocsc:
Compare the DUO codes of the applicant and the dataset and if they match, give a "thumbs up" indication for the handler. This is supposed to give a brief information for the handler that the application matches the dataset's intended use.

Consider having approver-bot compatibility.

Issue #2763, opened 2021-11-02
DUO release leftovers


aatkin:
Continuation from #2314 

These items need to be covered before DUO feature can be released:

- [ ] Update CHANGELOG
- [ ] Write up ADR on DUO, see [rems/docs/architecture](https://github.com/CSCfi/rems/tree/master/docs/architecture)
- [ ] Implement browser tests to cover primary user stories

These items are nice-to-haves:

- [ ] Refactor DUO code restriction `:type` - currently restriction type is used to access both restriction field type (e.g. date field or number input or basic input) and restriction translation, but for example `publication moratorium` uses restriction type `:date` which is generic field type

PR #2854, opened 2022-03-04
DUO: UI & API compatibility checking

aatkin:
relates #2791 

includes both back-end and front-end changes that implement DUO codes in application (applicant/handler views) and resource creation

**schema updates:**
- `api/schema/Application`: optional `:application/duo` key, implements `:duo/codes` and `:duo/matches`
- `application/commands/SaveDraftCommand`: optional `:duo-codes`
- `application/events/DraftSavedEvent`: optional `:application/duo-codes`

**application model updates:**
- new enrichment: `enrich-application-duo-matches`

**other changes:**
- `rems.ext.duo/check-duo-code` implements DUO code compatiblity check
- updated `resources/mondo.edn` with mondo code parents (big diff)
- added mondo code parents querying to satisfy DUO code "DUO:0000007" (DS) checking, with new dependency `com.stuartsierra.dependency`
- moved two DUO codes to `ext/duo/abstract-codes` set: `DUO:0000006` (HMB) and `DUO:0000042` (GRU)
- add guide component for `dropdown/async-dropdown`

**front-end changes:**
- add common `duo-info-field` and `duo-field` components in `rems.administration.duo` namespace that should be extensible and unify all DUO fields in application/resource
- create `rems.atoms/expander` component that implements simple chevron expandable block

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible (schema additions are optional)

## Documentation
- [ ] Update changelog if necessary #2763 
- [ ] Components are added to guide page (follow up once features are accepted. for now dropdown and expander have been added)
- [ ] ADR for major architectural decisions or experiments #2848 

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically (follow up once features are accepted)

## Follow-up

---
# screenshots

## resource creation
Screenshot 2022-03-30 at 17 48 45

## resource view
Screenshot 2022-03-30 at 17 48 20

## application, applicant view
Screenshot 2022-03-30 at 17 46 42

## application, handler view
Screenshot 2022-03-30 at 17 47 25


src/cljs/rems/atoms.cljs
```diff
@@ -173,6 +174,34 @@
...
+   
+   Pass a map of options with the following keys:
+   * `id` unique id for expanded content
+   * `content` content which is displayed in expanded state
+   * `expanded?` initial expanded state
+   * `title` content which is always displayed together with animated chevron"
+  [{:keys [id content expanded? title] :or {expanded? false}}]
+  (let [expanded (reagent/atom expanded?)
+        id (escape-element-id id)]
+    (fn []
```
Macroz, line 188: Shouldn't we repeat the parameters here?

src/cljs/rems/util.cljs
```diff
@@ -134,18 +134,15 @@
...
-           splitted))))
+  (for [substring (some-> s
+                          (str/replace link-regex #(str "\t" %1 "\t"))
+                          (str/split "\t"))
+        :let [url (when (re-matches link-regex substring)
+                    (if-not (str/starts-with? substring "http")
+                      (str "http://" substring) substring))]]
+    (if url
+      ^{:key (random-uuid)} [:a {:target :_blank :href url} substring]
+      substring)))
```
Macroz, line 145: So what changed here? I can see just adding `:key` but that could be avoided in the caller?

src/cljs/rems/util.cljs
```diff
@@ -211,4 +208,19 @@
...
+        (str/replace #"\s" "")
+        (str/replace #"[^A-Za-z0-9\-_]" "_")
+        (str/replace-first #"^[0-9]" #(str "id_" %)))))
+
+(deftest test-escape-element-id
+  (testing "should replace element special characters with underscore"
+    (is (nil? (escape-element-id nil)))
+    (is (= "elementid-123" (escape-element-id "element id-123 ")))
+    (is (= "element_id_123" (escape-element-id "element.id#123")))
+    (is (= "id_123-element" (escape-element-id " 123-element")))))
```
Macroz, line 225: I don't quite understand the escaping logic here. Should we rather escape ` ` with `_` and not trim?

src/cljs/rems/application.cljs
```diff
@@ -125,8 +129,18 @@
...
-    (assoc db ::edit-application {:field-values field-values
+                          (build-index {:keys [:form :field] :value-fn :value}))
+        duo-codes (->> (for [duo (-> application :application/duo :duo/codes)]
+                         (update-existing duo
+                                          :restrictions
+                                          (partial build-index {:keys [:type]
+                                                                :value-fn (fn [{:keys [type values]}]
+                                                                            (case type
+                                                                              :mondo values
+                                                                              (-> values first :value)))})))
```
Macroz, line 140: Why do we need to process the codes here? Couldn't the application contain all the relevant info when it is loaded without the need to do extra fetching or processing?

src/cljs/rems/application.cljs
```diff
@@ -846,7 +895,104 @@
...
+(rf/reg-sub ::duo-form
+            :<- [::edit-application]
+            (fn [edit-application] (:duo-codes edit-application)))
+(rf/reg-event-db ::set-duo-form-code
+                 (fn [db [_ keys value]] (assoc-in db (concat [::edit-application :duo-codes] keys) value)))
+
+(defn- find-duo-more-info [duo]
+  (flatten
+   (for [resource (:application/resources @(rf/subscribe [::application]))]
+     (for [res-duo (filter :more-info (-> resource :resource/duo :duo/codes))
```
Macroz, line 907: Same smell about `flatten` here.

src/cljs/rems/application.cljs
```diff
@@ -855,11 +1001,15 @@
     [:div.col-lg-8
      [application-state application config highlight-request-id]
      [:div.mt-3 [applicants-info application]]
+     (when (:enable-duo config)
+       (if (= userid (-> application :application/applicant :userid))
```
Macroz, line 1005: What about members? What kind of DUO view shoudl they have?

Macroz (commented):
Looks pretty solid. There are some small refactoring opportunities. I'll have to click around a bit.

src/cljs/rems/util.cljs
aatkin: how would you avoid `:key` in the caller side without sacrificing readability? the common use case looks like this
```clojure
(linkify raw-title)
```

src/cljs/rems/util.cljs
aatkin: `-` is a valid id character, stated by at least MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
> Technically, in [HTML5](https://developer.mozilla.org/en-US/docs/Glossary/HTML5), the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters. However, to avoid inadvertent errors, only [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters, digits, '_', and '-' should be used and the value for an id attribute should start with a letter.

src/cljs/rems/util.cljs
Macroz: Something like this? `(into [:span] (linkify raw-title))`

Macroz: I feel like we should perhaps improve the contract of the function to return hiccup. Using random UUIDs feels like a smell.

src/cljs/rems/util.cljs
aatkin: refactored the function so it replaces whitespace with underscore instead of trim

src/cljs/rems/util.cljs
aatkin: maybe we could do the `(into)` dance in the utility function itself then?

src/cljs/rems/util.cljs
aatkin: looking at `rems.test-util/test-linkify` the original usage seems to have been more akin to string than hiccup:
```clojure
(testing "retain original string"
      (is (= (apply str (linkify "a b c") "a b c"))))
```

src/cljs/rems/util.cljs
aatkin: looking at `rems.test-util/test-linkify` the original usage seems to have been more akin to string than hiccup:
```clojure
(testing "retain original string"
      (is (= (apply str (linkify "a b c") "a b c"))))
```

Macroz: We are free to fix it :+1:. Creating links in the middle of a string doesn't work so it should be a kind of a component and every place of use could just be fixed one way or another. But how about we create a new issue out of it?

src/clj/rems/application/model.clj
```diff
@@ -464,10 +466,40 @@
...
+                                              :duo/restrictions (:restrictions dataset-code)}]
+               nil)}))
+
+(defn- enrich-application-duo-matches [application]
+  (if-not (:enable-duo rems.config/env)
+    application
+    (let [duos (->> application :application/duo :duo/codes)
+          matches (flatten
+                   (for [resource (:application/resources application)]
+                     (for [res-duo (-> resource :resource/duo :duo/codes)
```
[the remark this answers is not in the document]
aatkin: ahh. i kind of suspected this should be possible but i didn't try it. should be a lot better for sure!

src/cljs/rems/application.cljs
aatkin: i don't know. what does a member do? is member that person who is invited to an application?

src/cljs/rems/atoms.cljs
aatkin: yes :+1:  minus the `expanded?` state which is only needed for the initial state

src/cljs/rems/administration/duo.cljs
```diff
@@ -1,65 +1,181 @@
...
+          :label restriction-label}])
+
+      nil)))
+
+(defn duo-field [duo & [opts]]
+  [:<>
+   [:h3
+    [duo-valid-icon (duo-validation-summary (:duo/statuses opts))]
+    (str (:shorthand duo) " – " (localized (:label duo)))]
+   [:pre (get duo :id "unknown code")]
```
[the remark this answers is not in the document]
aatkin: "unknown code" is not localized in the original back end code either. we can leave it out altogether. all duo codes should have a label so this should be a corner case

src/cljs/rems/administration/duo.cljs
```diff
@@ -1,65 +1,181 @@
...
+       ^{:key (str (:id duo) (:resource/id info))}
+       [duo-more-info info]))
+   (for [restriction (:restrictions duo)]
+     ^{:key (key restriction)}
+     [duo-restriction-field {:duo/id (:id duo)
+                             :context (:context opts)
+                             :duo/restriction {:type (key restriction)
+                                               :values (val restriction)}}])
+   (for [error (:duo/errors opts)]
+     ^{:key (random-uuid)}
```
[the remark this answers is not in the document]
aatkin: what do you suggest instead?

src/cljs/rems/util.cljs
aatkin: i reverted the change. feel free to create the issue

src/cljs/rems/administration/create_resource.cljs
```diff
@@ -36,39 +33,42 @@
...
 
 (defn- valid-request? [request]
   (and (not (str/blank? (get-in request [:organization :organization/id])))
-       (not (str/blank? (:resid request)))))
+       (not (str/blank? (:resid request)))
+       (every? true? (for [restriction (mapcat :restrictions (-> request :resource/duo :duo/codes))]
+                       (case (:type restriction)
+                         :mondo (some? (not-empty (:values restriction)))
+                         (some->> (not-empty (:values restriction))
+                                  (every? (comp not-empty :value))))))))
```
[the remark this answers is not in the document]
aatkin: how does it feel complicated? what could we do differently in your opinion?

src/cljs/rems/administration/duo.cljs
```diff
@@ -1,65 +1,181 @@
...
+        [:li (str (text :t.applications/resource) ": " (localized (:catalogue-item/title error)))]
+        (for [mondo (:duo/restrictions error)]
+          ^{:key (:id mondo)}
+          [:li (str label ": " (:id mondo) " - " (:label mondo))])]])
+
+    :t.duo.validation/needs-validation
+    [:div.alert.alert-warning
+     [:p (text (:type error))]
+     [:ul
+      [:li (str (text :t.applications/resource) ": " (localized (:catalogue-item/title error)))]
```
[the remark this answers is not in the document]
aatkin: added new translation keys `:t.label`, `:t.label-long` and `:t.label-dash` for these

src/clj/rems/application/model.clj
```diff
@@ -464,10 +466,40 @@
...
                      :catalogue-item/expired (:expired item)
                      :catalogue-item/archived (:archived item)}
-                    (assoc-some-in [:resource/duo :duo/codes] (seq (enrich-duo-codes duo-codes)))))))
+                    (assoc-some-in [:resource/duo :duo/codes] (seq (duo/enrich-duo-codes duo-codes)))))))
        (sort-by :catalogue-item/id)
        vec))
 
+(defn- validate-duo-match [dataset-code query-code resource]
+  (let [valid (duo/check-duo-code dataset-code query-code)]
+    {:valid valid
```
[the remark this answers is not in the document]
aatkin: changed to `:validity`

src/clj/rems/ext/duo.clj
```diff
@@ -70,13 +72,11 @@
...
-(defn- enrich-duo-code
-  "Adds convenience attributes to DUO codes."
-  [duo]
-  (merge duo
-         (get complex-codes (:id duo))))
+;; http://purl.obolibrary.org/obo/DUO_0000001 data use permission
+;; http://purl.obolibrary.org/obo/DUO_0000006 health or medical or biomedical research
+;; http://purl.obolibrary.org/obo/DUO_0000017 data use modifier
+;; http://purl.obolibrary.org/obo/DUO_0000042 general research use
+(def abstract-codes #{"DUO:0000001" "DUO:0000006" "DUO:0000017" "DUO:0000042"}) ; not concrete types can't be used as tag
```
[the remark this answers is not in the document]
aatkin: let's get back to this in weekly then

src/cljs/rems/application.cljs
aatkin: i'm not sure what you mean - this block transforms application duo codes into a form that is usable by the application editor. in this case restriction data is indexed so that it is easier to refer to by front end components.

src/clj/rems/ext/mondo.clj
```diff
@@ -85,42 +100,71 @@
...
+
+(defn get-mondo-parents
+  [code]
+  (if (:enable-duo rems.config/env)
+    (do
+      (ensure-codes-are-loaded)
+      (->> code
+           strip-mondo-prefix
+           (dep/transitive-dependencies @codes-dag)
+           (map (partial str "MONDO:"))
```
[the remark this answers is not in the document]
aatkin: can you elaborate how you would want this code to be changed?

src/cljs/rems/application.cljs
aatkin: added check for `:member` role

src/clj/rems/ext/mondo.clj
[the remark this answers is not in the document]
Macroz: There's now two places that add `"MONDO:"` prefix to a string. One of them is extracted to a function `external-format`. So I'm just wondering could we have one place where the internal vs. external format is handled that works for both of these cases?

src/cljs/rems/administration/duo.cljs
```diff
@@ -1,65 +1,181 @@
...
+      (doall
+       (for [restriction (:duo/restrictions error)
+             :let [label (text (get duo-restriction-label (:type restriction)))]]
+         (for [{:keys [value]} (:values restriction)]
+           ^{:key (random-uuid)}
+           [:li (str label ": " value)])))]]
```
Macroz: Btw. this can also be written with just one `for` without nesting?

src/cljs/rems/util.cljs
Macroz: :+1: Created https://github.com/CSCfi/rems/issues/2899

src/cljs/rems/application.cljs
Macroz: Ok, I guess it's fine. Just looked like a big expression needed to do it.

src/cljs/rems/administration/duo.cljs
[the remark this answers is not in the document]
aatkin: i'm not sure about this - check out https://reactjs.org/docs/lists-and-keys.html

> A “key” is a special string attribute you need to include when creating lists of elements.

src/cljs/rems/application.cljs
aatkin: extracted the main gist into separate function. i guess it looks more neat now.

src/clj/rems/ext/mondo.clj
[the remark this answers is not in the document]
aatkin: extracted into `add-mondo-prefix` function

src/cljs/rems/administration/create_resource.cljs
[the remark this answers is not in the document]
aatkin: refactored as per suggestion

src/cljs/rems/administration/duo.cljs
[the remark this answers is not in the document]
Macroz: Yeah the reason `into` works is actually that in React's eyes it's just `[:div first second ... nth]` after the hiccup is expanded, i.e. there is no list anymore. If the items change a lot, and there is a natural `:key` then it definitely makes sense to put the `:key` to get the optimizations related to it. But this avoids having to deal with `index` or UUID which are not any better for this case.

Macroz (approved):
Looks great now. We'll do some manual testing still anyway.

PR #2862, opened 2022-03-16
User rename mapping

Macroz:
The user rename now works with mappings. This was a missing feature. After #2858.

If a user already exists (i.e., there is a user with two identities), the user mapping code will not use the old identity. Then you can use the `rename "new id" "old id"` trick to get rid of the new user, fix the data and effectively remove the wrong mapping. This happened in dev with "elixir-alice" since it existed already. And would have fixed the case (tested manually).

# Checklist for author

## Reviewability
- [ ] Link to issue

## Documentation
- [ ] Update changelog if necessary (I guess nothing is needed as this is kind of expected to be supported)

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically (user rename is not tested as it is difficult)

## Follow-up

### DIFF
--- a/test/clj/rems/api/test_users.clj
+++ b/test/clj/rems/api/test_users.clj
@@ -143,7 +143,7 @@
                  (users/format-user (:identity (middleware/get-session cookie)))))))
 
       (testing "log in elixir-alice and create user mapping"
-        (is (nil? (user-mappings/get-user-mappings "elixirId" "elixir-alice")) "user mapping should not exist")
+        (is (nil? (user-mappings/get-user-mappings {:ext-id-attribute "elixirId" :ext-id-value "elixir-alice"})) "user mapping should not exist")
         (let [cookie (login-with-cookies "elixir-alice")]
           (assert-can-make-a-request! cookie)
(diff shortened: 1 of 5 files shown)


test/clj/rems/api/test_users.clj
```diff
@@ -143,17 +143,17 @@
...
                    {:userid "alice" :name "Elixir Alice" :email "alice@elixir-europe.org"}}
                  (set (api-call :get "/api/users/active" nil
                                 +test-api-key+ "owner"))))
           (is (= [{:userid "alice"
-                   :extidvalue "elixir-alice"
-                   :extidattribute "elixirId"}]
-                 (user-mappings/get-user-mappings "elixirId" "elixir-alice")))
+                   :ext-id-value "elixir-alice"
+                   :ext-id-attribute "elixirId"}]
+                 (user-mappings/get-user-mappings {:ext-id-attribute "elixirId" :ext-id-avlue "elixir-alice"})))
```
Macroz, line 156: Must fix the typo here.

Macroz: Fixed in #2867

aatkin (approved):

PR #2864, opened 2022-03-17
fix: deduplicate forms

### LINKED ISSUE #2853: Using the same form for a workflow and catalogue item duplicates it
If we have the same form in the workflow and catalogue item, it will be duplicated in the application, but answering each question answers the same question in the other form.

For regular catalogue items, if the same form exists twice, it will only be shown once. This should be the immediate fix for workflow forms too.

In the future, we may want to consider if a duplicated form is once per resource or once per application. For example if the same questions need to be asked but are specific to each resource. This currently requires duplicating the same looking form for each resource and makes later management difficult. This could be a form option.

Macroz:
Fixes #2853.

Workflow forms were concatenated to catalogue item forms so
duplicates could happen. Now they are deduplicated after
concatenation. Also added some tests of this feature.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -1,10 +1,8 @@
 (ns rems.application.commands
   (:require [clojure.test :refer [deftest is testing]]
-            [com.rpl.specter :refer [ALL transform]]
-            [medley.core :refer [assoc-some]]
+            [medley.core :refer [assoc-some distinct-by]]
             [rems.common.application-util :as application-util]
             [rems.common.form :as form]
-            [rems.common.util :refer [build-index]]
             [rems.form-validation :as form-validation]
             [rems.permissions :as permissions]
             [rems.schema-base :as schema-base]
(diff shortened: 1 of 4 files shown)


src/clj/rems/application/commands.clj
```diff
@@ -299,13 +297,14 @@
...
-(defn- build-forms-list [catalogue-item-ids {:keys [get-catalogue-item]}]
+(defn- build-forms-list [workflow catalogue-item-ids {:keys [get-catalogue-item]}]
   (->> catalogue-item-ids
        (mapv get-catalogue-item)
        (mapv :formid)
        (remove nil?)
-       (distinct)
-       (mapv (fn [form-id] {:form/id form-id}))))
+       (mapv (fn [form-id] {:form/id form-id}))
+       (concat (get-in workflow [:workflow :forms])) ; NB: workflow forms end up first
```
aatkin, line 306: is ordering significant for these forms or could we sort by id?

aatkin (approved):
looks good to me!

src/clj/rems/application/commands.clj
Macroz: There is no definite rule. The workflow forms were originally put as first, as I think they are likely to contain "common stuff". Otherwise the order of the forms should be kept stable, so one can control the order in which they appear. The order of the catalogue items / resources determines the form order mostly.

Ordering by form title or other condition could be sensible too.

Issue #2828, opened 2022-01-19
Organization owners could be allowed to edit their own organization owners


Macroz:
Currently we don't allow organization owners to edit organization ownership. The owner role is the only one allowed to do so. This could be changed so that the organization owner can edit the owners of the organizations they own. This would be slightly better alignment with the delegated ownership model.

Macroz:
See also #2850.

aatkin:
User should not be able to edit organization if they are not this organization's owner.

PR #2868, opened 2022-03-19
fix: owner as organization owner

### LINKED ISSUE #2850: Owner as organization owner
Making the owner user an organization owner leads to a problem wrt. editing organizations because we don't allow organization owners edit them and the code seems to check that access right first (denying a rightful "super" owner, who is an organization owner too, of editing organization owners).

We have some similar inconsistencies in role handling, and this happens sometimes when new users of REMS are trying out various features. E.g. using the "super" owner to apply and handle applications.

- [ ] make sure owner can always edit organization owners, even if they are an organization owner too

Macroz:
Fixes #2850.

Owner wasn't able to edit organization ownership when they were also
an organization owner themselves.

owner1
owner2

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/api/services/organizations.clj
+++ b/src/clj/rems/api/services/organizations.clj
@@ -2,6 +2,7 @@
   (:require [clojure.set :as set]
             [medley.core :refer [assoc-some find-first remove-keys]]
             [rems.api.services.dependencies :as dependencies]
+            [rems.context :as context]
             [rems.db.applications :as applications]
             [rems.db.core :as db]
             [rems.db.organizations :as organizations]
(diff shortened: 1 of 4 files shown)


src/clj/rems/api/services/organizations.clj
```diff
@@ -60,9 +61,10 @@
   (organizations/update-organization! (:organization/id org)
                                       (fn [db-organization]
                                         (let [organization-owners (set (map :userid (:organization/owners db-organization)))
-                                              organization-owner? (contains? organization-owners userid)]
+                                              organization-owner? (contains? organization-owners userid)
+                                              owner? (contains? context/*roles* :owner)]
                                           (merge db-organization
-                                                 (remove-keys (if organization-owner?
+                                                 (remove-keys (if (and organization-owner? (not owner?))
```
aatkin, line 67: would this be simpler to express as `(if-not owner? ...)`? as in only owner role can update owners but roles can also include other roles (like organization owner)

aatkin (approved):
looks good to me :+1:

src/clj/rems/api/services/organizations.clj
Macroz: Hmm, I guess it would. Let's however change this in the issue #2828, where we need to check whether it's an owner or organization owner of the same organization.

PR #2890, opened 2022-03-21
refactor: replace update-present with update-existing from medley.core

### LINKED ISSUE #2866: Remove update-present
As noticed in #2859 we can remove `update-present` because it's the same as http://weavejester.github.io/medley/medley.core.html#var-update-existing

Macroz:
Closes #2866 

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary (doesn't need to as it is internal)


aatkin (approved):
nice clean-up of unused imports as well :+1:

PR #2893, opened 2022-03-24
chore: create issue templates

Macroz:
We don't strictly need these for our own use, but we could have some just because they are recommended by GitHub.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Create a report to help us improve
 title: ''
-labels: BUG
+labels: 'Needs Triage'
 assignees: ''
 
 ---
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,29 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: BUG
+assignees: ''
+
+---
+
+**Describe the bug or behavior**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,29 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: BUG
```
aatkin, line 5: would it be worth it to have a label like "not verified" to indicate that the bug hasn't been checked yet by a core team developer? as the tickets don't move automatically to project and "user feedback" lane this would make them a bit more obvious to spot

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature
```
aatkin, line 5: and here "feature request" or such?

aatkin (approved):
nice! :+1:

.github/ISSUE_TEMPLATE/bug-report.md
Macroz: Yeah it depends what we want to do. We could discuss this on Monday. Ideally they would be put directly to User Feedback or so.

.github/ISSUE_TEMPLATE/feature-request.md
Macroz: Maybe we need the one label that is "incoming", because accurate classification (bug, feature, enhancement, ...) can only happen by us.

.github/ISSUE_TEMPLATE/bug-report.md
Macroz: I created `Needs Triage`

.github/ISSUE_TEMPLATE/feature-request.md
Macroz: I created `Needs Triage`

PR #2894, opened 2022-03-25
chore: bump dependencies

Macroz:
Update many dependencies.

React / frontend was left out and likely requires transitioning to
Shadow-CLJS before it can be done. For example, latest ClojureScript
is not possible to use because `lein-cljsbuild` does not work with
them (some Google Closure change).

Only very minor patches were required in code.


Macroz:
> nice! seems to work on my machine. on first run i got some odd error which went away on second run. a bunch of these appear in the console but they don't have a source so maybe related to a dependency?
> 
> ```shell
> WARNING: update-vals already refers to: #'clojure.core/update-vals in namespace: clojure.tools.analyzer.utils, being replaced by: #'clojure.tools.analyzer.utils/update-vals
> ```

These must be related to Clojure 1.11, which introduces `update-vals`. Not everything has been updated yet.

PR #2897, opened 2022-03-29
feat: more aggressive scaling for previous applications

### LINKED ISSUE #2855: Optimize user previous applications to fit small screens
When a handler opens an application with previous applications, the table is sometimes too wide to fit the screen and will be below the floating actions menu. We can likely remove the applicant column and see if it's enough. But check what happens with members.

image

Macroz:
Introduces more aggressive scaling to fix #2855 in Finnish and Swedish translations.

aggressive

### DIFF
--- a/src/clj/rems/css/styles.clj
+++ b/src/clj/rems/css/styles.clj
@@ -29,7 +29,8 @@
 (def bootstrap-media-breakpoints {:xs (u/px 576)
                                   :sm (u/px 768)
                                   :md (u/px 992)
-                                  :lg (u/px 1200)})
+                                  :lg (u/px 1200)
+                                  :xl (u/px 1600)})
 
 
 ;; Fonts of the app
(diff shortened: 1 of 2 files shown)


src/clj/rems/css/styles.clj
```diff
@@ -844,6 +845,12 @@
                        :padding "0.2rem"
                        :border-radius "0.25rem"}]]
 
+   ;; special case for previous applications
+   [:#previous-applications-except-current [:.resource {:overflow :hidden
+                                                        :text-overflow :ellipsis
+                                                        :white-space :nowrap
+                                                        :max-width "10em"}]]
```
aatkin, line 852: should we use `garden.units` everywhere for consistency? or switch to writing units as here

Macroz: I don't have a clear answer at the moment. I used to write using the units, but mostly it seems like a waste of time. We can consider unifying how we write them when we restyle the UI.

PR #2898, opened 2022-03-29
Harmonize names

Macroz:
In a recent planning we discussed harmonizing the column names and the view/edit field names in the admin pages.


aatkin:
should workflow and resource also have a column called title? at least form has a field called title, so i guess at least it should be harmonized? not sure about resource though

Screenshot 2022-04-01 at 16 16 17
Screenshot 2022-04-01 at 16 16 23

Macroz:
> should form and resource also have a column called title? at least form has a field called title, so i guess at least it should be harmonized? not sure about resource though
> 
> Screenshot 2022-04-01 at 16 16 17 Screenshot 2022-04-01 at 16 16 23

The harmonization is that the column that is in the details is also on the table. For workflows there is no title so `Name` it is. For resources, `resid` is called `Resource` so that's what we have to use?

aatkin:
> The harmonization is that the column that is in the details is also on the table. For workflows there is no title so `Name` it is. For resources, `resid` is called `Resource` so that's what we have to use?

I was talking about `Forms` here. Sorry for the confusion - the first message contained `workflow` but i typo'ed it there and had to edit it out.

aatkin:
Looks like I was mistaken there - Form uses `name` as discussed and is thus harmonized already.

aatkin (approved):
:+1:

PR #2902, opened 2022-04-01
Catalogue tree leftovers

### LINKED ISSUE #2800: Catalogue tree leftovers

Macroz:
Closes #2800 

Work in progress:
- [x] schema
- [x] simple browser tests
- [x] browser test refactoring
- [x] refactor create catalogue item browser test to use admin UI and not test helpers
- [x] browser test with category hierarchy
- [x] owner can see disabled catalogue item in the tree
- [ ] ~share / re-use with table~ (there is not much to share, the implementations may diverge)
- [ ] ~animate opening and closing~ (as the tree nodes are flattened to rows, there is no simple way to animate a bunch of rows)

The tested structure is like this:
categories

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

## Follow-up


test/clj/rems/test_browser.clj
```diff
@@ -2613,24 +2619,98 @@
...
       (btu/accept-alert)
+
       (is (btu/eventually-visible? :categories))
-      (is (= #{"Edited title (EN)"}
+      (is (= #{(str (btu/context-getx :category-name) " Edited")}
              (->> (set (slurp-categories-by-title))
-                  (intersection #{"Edited title (EN)" "E2E Ancestor category (EN)"})))))))
+                  (intersection #{(str (btu/context-getx :category-name) " Edited") (str (btu/context-getx :category-name) " Ancestor")})))))))
+
+(deftest test-catalogue-tree
```
aatkin, line 2646: would you see value in testing with `check-axe` in this test? application tests don't open subcategories and i think they use table rather than catalogue tree.

aatkin (approved):
LGTM :+1:

test/clj/rems/test_browser.clj
Macroz: Yeah since it is meant for the applicant it makes sense to have it covered. I'll add a separate task though, because I want to check the state of the check too.

PR #2903, opened 2022-04-04
fix: zoom

Macroz:
Fixes https://github.com/CSCfi/rems/issues/2855

The zoom attribute isn't supported in Firefox. We can use font-size for a
similar effect. Trying to use transform scale runs into trouble with
not being so responsive as it does not fill available space.

zoom-to-size

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation


aatkin (approved):
nice! also works in safari and firefox :+1:

Issue #2297, opened 2020-09-03
Generate new demo data


jaakkocsc:
Come up with some meaningful demo data that can be used for demonstrating
  * [] basic application handling
  * [] support for multi-organisation handling
  * [] auto-approve
  * [] decider workflow

jaakkocsc:
Likely four catalog items split into two categories, utilising two different workflows (normal, decider).

PR #2909, opened 2022-04-07
Improve fake login page

### LINKED ISSUE #2896: Improve the fake login page
Improve the fake login page by describing what the users are.

Since we split the fake IdP data from the test data, we can easily add a description for the user that can be shown in the fake login page. We can also group the users so that simple users (for simple use cases) are first and special cases later.

Macroz:
Close #2896

This feature should make it more obvious what kind of users exist when trying out REMS. There is a further ticket that should be done to actually clean up the users more #2297, perhaps rename etc.

users
demo-users

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks

### DIFF
--- a/src/clj/rems/auth/fake_login.clj
+++ b/src/clj/rems/auth/fake_login.clj
@@ -52,7 +52,8 @@
 (defn- user-selection [username]
   (let [url (url (login-url) {:username username})]
     [:div.user.m-3 {:onclick (str "window.location.href='" url "';")}
-     [:a.btn.btn-primary {:href url :style "width: 12rem;"} username]]))
+     [:a.btn.btn-primary {:href url :style "width: 12rem; overflow: hidden; text-overflow: ellipsis; white-space: nowrap"}
+      username]]))
 
 (defn- fake-login-screen [{session :session :as req}]
   (if-let [username (-> req :params :username)]
--- a/src/clj/rems/auth/fake_login.clj
+++ b/src/clj/rems/auth/fake_login.clj
@@ -71,6 +71,9 @@
                                       [:div
                                        [:h2.mx-3 group]
 
+                                       ;; XXX: this could be refactored so that
+                                       ;; this isn't a special case but an attribute of
+                                       ;; the group data
                                        (when (= "Special Users" group)
                                          [:div.ml-3
                                           [:p "Special users with special cases. Not for daily use!"]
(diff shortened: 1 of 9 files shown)


src/clj/rems/auth/fake_login.clj
```diff
@@ -42,25 +51,38 @@
 
 (defn- user-selection [username]
   (let [url (url (login-url) {:username username})]
-    [:div.user.mb-3.mr-3.flex-fill.d-flex {:onclick (str "window.location.href='" url "';")}
-     [:a.btn.btn-primary.flex-fill {:href url} username]]))
+    [:div.user.m-3 {:onclick (str "window.location.href='" url "';")}
+     [:a.btn.btn-primary {:href url :style "width: 12rem; overflow: hidden; text-overflow: ellipsis; white-space: nowrap"}
```
aatkin, line 55: i noticed that we have a few places where the text truncation is used, and coincidentally bootstrap has a class that implements it: `text-truncate` https://getbootstrap.com/docs/4.6/utilities/text/#text-wrapping-and-overflow

src/clj/rems/auth/fake_login.clj
```diff
@@ -42,25 +51,38 @@
...
+                                    (for [{:keys [group users]} (get-fake-user-descriptions)
+                                          :let [id (escape-element-id group)]]
+                                      [:div
+                                       [:h2.mx-3 group]
+
+                                       (when (= "Special Users" group)
+                                         [:div.ml-3
+                                          [:p "Special users with special cases. Not for daily use!"]
+                                          [:button#show-special-users.btn.btn-secondary {:data-toggle "collapse" :data-target (str "#" id)}
+                                           "Show"]])
```
aatkin, line 78: perhaps this could be configured in the group data as well?
```clojure
{:group "Special Users"
 :description "Special users with special cases. Not for daily use!"
 :collapsed true
 :users [...]}
```

aatkin (approved):
looks good! :+1: only minor improvement suggestions

src/clj/rems/auth/fake_login.clj
Macroz: Since there is only one exception with the collapse, I find that it would just add a bunch of generic code that isn't that useful, that actually complicates the code for the screen and the data. Let's do it if we find another use for it.

Macroz: Added a comment about this possible refactoring.

Issue #2762, opened 2021-11-01
Update Dockerfile to use JDK17


aatkin:
relates #2726 

REMS has been tested in development and built in CI with JDK17 and leiningen version 2.9.7

- [ ] Mention also in the release notes

PR #2913, opened 2022-04-08
feat: update Dockerfile to use JDK17

aatkin:
relates #2762 

Update Dockerfile to use JDK17 image. `openjdk:17-jre-slim` was not available, but based on following SO discussion `eclipse-temurin` builds could work instead: https://stackoverflow.com/questions/69525199/openjdk-java-17-docker-image

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,16 @@
 # For documentation see docs/installing-upgrading.md
 
-FROM openjdk:11-jre-slim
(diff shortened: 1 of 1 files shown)


Macroz (approved):
We could also try creating our own slimmed image using `jlink`. That seems to be the recommended practice these days.

Issue #2906, opened 2022-04-05
Notify user about expiring drafts


jaakkocsc:
Enhancement for #2665

Macroz:
See also https://github.com/CSCfi/rems/issues/2729

PR #2919, opened 2022-04-21
Send notification emails before application expiration

aatkin:
relates #2906 

- change application expiration config to accept map per application state with keys `#{:delete-after :reminder-before}` using ISO-8601 durations
- create new command `:application.command/send-expiration-notifications` that creates event `:application.event/expiration-notifications-sent`
- create ns `rems.application.expirer-bot` which implements generating application delete and expiritaion notifications send command logic. expirer-bot is triggered by external event from `rems.application.eraser/process-applications!`
- user `:expirer` role must be granted to `expirer-bot` user prior to enabling the feature. the role is granted permissions to delete and send expiration notifications in application creation

This PR changes the logic of application eraser so that user `expirer-bot` executes commands to delete and send expiration notification emails. `expirer-bot` is triggered externally by eraser process, but it's behaviour is configurable. Sending reminder emails is optional, and it is configured with ISO-8601 duration, defined as time period before expiration. Only one reminder email is sent before expiration. If application is edited before expiration, both timers are reset.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [ ] Config is backwards compatible (expiration notification change is breaking)

## Documentation

## Testing


aatkin:
quickly discussed over rocket chat about using bots instead of the db table - will look into this so there wouldn't be need for a migration

docs/bots.md
```diff
@@ -55,6 +55,34 @@ Example of creating the bot user with the API.
...
+application members about impending expiration. Expirations can be configured
+by application state for both application removal and sending reminder email.
+
+Unlike other bots, the Expirer bot is designed to be triggered by an external
+event, such as a periodically ran process. The bot is supplied an application
+and it returns appropriate command description, if applicable. Responsibility
+of the command execution is left to the caller. A scheduler is implemented in
+REMS, enabled by config option `:expiration` to periodically call the Expirer
+bot.
+
```
Macroz, line 72: Should we mention around here that, if the person does something, the count starts from zero and another expiration notification is potentially sent later?

resources/config-defaults.edn
```diff
@@ -338,13 +338,18 @@
...
  :attachment-max-size 10000000
 
- ;; Should applications be deleted after a set period of time has passed?
- ;; Use map with key as required application state, and value as ISO-8601 duration:
- ;; {:application.state/draft "P90D" :application.state/closed "P7D"}
+ ;; Should applications be deleted after having been inactive for a set period of time?
+ ;; Use map with key as required application state, and value as a map:
+ ;;   {:application.state/draft {:delete-after "P90D" :reminder "P7D"}}
+ ;;   :delete-after - delete application after period of time since last activity
+ ;;   :reminder - (optional) send reminder about expiration to application members (as period of time before expiration)
```
Macroz, line 345: Isn't this called `:reminder-before` now?

Btw. this design limits us to only one reminder about expiration. But we can expand it later.

src/clj/rems/application/expirer_bot.clj
```diff
@@ -0,0 +1,85 @@
...
+        (is (not (expire-application {:application/state :application.state/draft
+                                      :application/last-activity over-7d-ago})))
+        (is (some? (expire-application {:application/state :application.state/closed
+                                        :application/last-activity over-7d-ago})))
+        (is (not (expire-application {:application/state :application.state/closed
+                                      :application/last-activity over-1d-ago})))
+        (is (not (expire-application {:application/state :application.state/rejected
+                                      :application/last-activity over-90d-ago})))))))
+
+(defn- reminder-time [application]
```
Macroz, line 56: Should there be also a `text-reminder-time` ans there is for `expire-application`?

src/clj/rems/application/eraser.clj
```diff
@@ -1,65 +1,43 @@
...
+    (log/warn (str "Cannot process applications, because user " expirer-bot/bot-userid " does not exist")
+              #'process-applications!))
   (applications/reload-cache!)
-  (log/info :finish #'remove-expired-applications!))
+  (log/info :finish #'process-applications!))
 
 (mount/defstate expired-application-poller
   :start (when (:application-expiration env)
-           (scheduler/start! "expired-application-poller" remove-expired-applications! (.toStandardDuration (time/days 1))))
+           (scheduler/start! "expired-application-poller"
```
Macroz, line 26: Should we rename to `eraser-poller` to be consistent with the ns? Or is this perhaps still clearer?

src/clj/rems/application/master_workflow.clj
```diff
@@ -124,6 +127,7 @@
   [application event]
   (-> application
       (permissions/give-role-to-users :applicant [(:event/actor event)])
+      (permissions/give-role-to-users :expirer [expirer-bot/bot-userid])
```
Macroz, line 130: I guess it doesn't make sense to make this dependent on the config, as that would add dependency to this pure model to the config, as well as complicate the tests. It's sort of slightly confusing that even if expire isn't confirued, or `expirer-bot` used doesn't exist, it is added.

src/clj/rems/application/model.clj
```diff
@@ -257,6 +257,10 @@
   [application _event]
   application)
 
+(defmethod application-base-view :application.event/expiration-notifications-sent
+  [application _event]
+  application)
```
Macroz, line 262: We could record the time when they were sent to the model, but we don't seem to need it now, because we can show it from the event.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ NB: The login has changed to allow more configurable user identity and other att
 - The `:oidc-userid-attribute` config has been renamed to `:oidc-userid-attributes` and has new options to allow internally renaming an attribute from IdP to REMS db. (#2771, #2821)
 - Users are required a name and email from the IdP to be allowed in. (#2889)
 - User attributes have been renamed internally. If you directly accessed the database, please note that `eppn -> userid`, `commonName -> name` and `mail -> email`. (#2377)
+- Application expiration notification can now be configured to send reminder email to applicant and members before expiration. This requires a change to the configuration option `:application-expiration`. (#2906)
```
Macroz, line 21: We should perhaps add here a note, that customized email templates should be checked for the new expiry template, if this feature is used?

test/clj/rems/application/test_eraser.clj
```diff
@@ -1,58 +1,235 @@
...
+                                              :date-time (time/minus now (time/days 90) (time/seconds 1))
+                                              :actor "alice"})
+          outbox-emails (atom [])
+          log-messages (atom [])]
+
+      (testing "does not process applications when expirer-bot user does not exist"
+        (is (= #{draft old-submitted expired-draft} (set (get-all-application-ids "alice"))))
+        (with-redefs [env {:application-expiration {:application.state/draft {:delete-after "P90D"}}}
+                      outbox/put! (fn [email] (swap! outbox-emails conj email))
+                      log/log* (log*-mock log-messages)]
```
Macroz, line 92: Should we move this to a higher level `with-redefs` so that we don't have to repeat them in every test?

test/clj/rems/application/test_eraser.clj
```diff
@@ -1,58 +1,235 @@
...
+                         :outbox/email {:subject (str "Your unsubmitted application " draft-expires-in-6d " will be deleted soon")
+                                        :body (str "Dear member,\n\n"
+                                                   "Your unsubmitted application has been inactive since 2021-10-09 and it will be deleted after 2022-01-07, if it is not edited.\n\n"
+                                                   "You can view and edit the application at localhost/application/" draft-expires-in-6d)
+                                        :to-user "member"}
+                         :outbox/type :email}]
+                       @outbox-emails))
+
+                (testing "processing applications again should not send new notifications"
+                  (eraser/process-applications!)
```
Macroz, line 207: Could reset `outbox-emails` and check it's empty after this one?

test/clj/rems/application/test_eraser.clj
```diff
@@ -1,58 +1,235 @@
...
+                  (is (empty? (expiration-notification-events draft-expires-in-8d)))
+                  (is (empty? (expiration-notification-events submitted)))
+
+                  (testing "expiration notifications events have not changed for draft expiring in 6 days"
+                    (let [events (expiration-notification-events draft-expires-in-6d)]
+                      (is (= 1 (count events)))
+                      (is (time/equal? (time/plus now (time/days 6)) (:expires-on (first events))))
+                      (is (time/equal? (time/minus now (time/days 84)) (:last-activity (first events))))
+                      (is (time/equal? now (:event-time (first events))))))
+                  (is (= [{:outbox/deadline (time/date-time 2022)
```
Macroz, line 220: And then perhaps we can skip testing for this one (it's a lot of duplicate code).

Macroz (approved):
Looks very clear now!

There's a couple points that we can improve, but mostly was just thinking aloud.

src/clj/rems/application/master_workflow.clj
aatkin: good point. `rems.application.model/enrich-super-users` implements this already for `:reporter` role, so i moved the functionality there and added documentation about granting `:expirer` role manually to the Expirer bot.

Macroz (approved):
Looks good, browser test didn't pass because of flaky test so I'll merge this when it goes through.

Issue #1373, opened 2019-06-18
After "Change resources", Terms of Use is not updated


okahilak:
When applying for a resource, if you change the resource you're applying for by pressing "Change resources", the Terms of Use is not updated.

aatkin:
Original issue does not seem to exist anymore and may have been fixed sometime in the past. An edge-case may exist, but it has not been documented. Moving the ticket back to ideas from backlog.

PR #2925, opened 2022-05-06
feat: empty accepted licenses after resources changed event

aatkin:
relates #1373 

- `:application/accepted-licenses` is now emptied after processing `:application.event/resources-changed`

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -2,7 +2,7 @@
   (:require [clojure.set :as set]
             [clojure.test :refer [deftest is testing]]
             [com.rpl.specter :refer [ALL select transform]]
-            [medley.core :refer [find-first map-vals update-existing update-existing-in]]
+            [medley.core :refer [assoc-some find-first map-vals update-existing update-existing-in]]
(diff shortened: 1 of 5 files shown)


src/clj/rems/application/model.clj
```diff
@@ -222,7 +222,8 @@
       (assoc :application/forms (vec (:application/forms event)))
       (assoc :application/resources (vec (:application/resources event)))
       (assoc :application/licenses (map #(select-keys % [:license/id])
-                                        (:application/licenses event)))))
+                                        (:application/licenses event)))
+      (assoc-some :application/accepted-licenses {})))
```
Macroz, line 226: I think we used to keep track of which licenses you have accepted, so that you would basically only need to accept some new licences. While that is not much of an improvement (to see that yes there are new licenses and must press the same approve licenses -button), in case the licenses required happen to be the same, you don't have to accept again. Have we lost this capability?

aatkin:
after some discussion in weekly, decided to close as the feature seems to be working as intended after all.

Issue #2838, opened 2022-02-01
Change "Application: success"


Macroz:
When creating a new application, we show the standard type of status message with the text "Application: success". Sometimes this is confusing as this is only the start of the application process. We should rather hide the message, or make it more descriptive. For example, it could say "Application creation: success", "Application: saved" or "Application: started".

PR #2926, opened 2022-05-06
feat: remove flash message from successful new application page

aatkin:
relates #2838 

- removed flash message `"Application: Success"` from new application page on successful application create, as the message was deemed unnecessary and rather confusing

# Checklist for author

## Reviewability
- [x] Link to issue


Macroz (approved):
How about we just hide the whole message in case it succeeds?

Macroz:
Changelog entry?

Issue #2835, opened 2022-01-24
Removing blacklisting with bad res_id returns 500


juholehtonen:
`/api/blacklist/remove` returns 500, if 'resource id' is wrong.

Trace:
```
2022-01-24 09:04:08,873 [qtp2068615081-21 3659397fef7a6b4607e76a7e81d8f52c4917edd7@elixir-europe.org POST /api/blacklist/remove] DEBUG o.e.jetty.server.HttpConnection - releaseRequestBuffer HttpConnection@6a1f41a4::SocketChannelEndPoint@6f1fadd6{l=/10.128.10.183:3001,r=/10.128.2.1:56060,OPEN,fill=-,flush=-,to=23/200000}{io=0/0,kio=0,kro=1}->HttpConnection@6a1f41a4[p=HttpParser{s=END,203 of 203},g=HttpGenerator@5290d9b3{s=START}]=>HttpChannelOverHttp@17c7b020{s=HttpChannelState@5dd21e36{s=HANDLING rs=BLOCKING os=OPEN is=IDLE awp=false se=false i=true al=0},r=1,c=false/false,a=HANDLING,uri=//rems-bona-fide-test.rahtiapp.fi/api/blacklist/remove,age=22}
2022-01-24 09:04:08,882 [qtp2068615081-21 3659397fef7a6b4607e76a7e81d8f52c4917edd7@elixir-europe.org POST /api/blacklist/remove] ERROR compojure.api.exception - resource doesn't exist
java.lang.IllegalArgumentException: resource doesn't exist
	at rems.db.blacklist$check_foreign_keys.invokeStatic(blacklist.clj:45)
	at rems.db.blacklist$check_foreign_keys.invoke(blacklist.clj:41)
	at rems.db.blacklist$add_event_BANG_.invokeStatic(blacklist.clj:50)
	at rems.db.blacklist$add_event_BANG_.invoke(blacklist.clj:49)
	at rems.api.services.blacklist$remove_user_from_blacklist_BANG_.invokeStatic(blacklist.clj:19)
	at rems.api.services.blacklist$remove_user_from_blacklist_BANG_.invoke(blacklist.clj:18)
	at rems.api.blacklist$fn__32228.invokeStatic(blacklist.clj:76)
	at rems.api.blacklist$fn__32228.invoke(blacklist.clj:71)
	at compojure.core$wrap_response$fn__5391.invoke(core.clj:158)
[...]

PR #2927, opened 2022-05-06
WIP fix: return 404 from blacklist api

aatkin:
relates #2835

- moved blacklist entry validation from `db` to `api` level, and return `(not-found-json-response)` if user or resource is not valid when adding or removing blacklist entry

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible - add/remove returns 404 instead of 500 if user/resource not found

## Testing


src/clj/rems/api/blacklist.clj
```diff
@@ -26,6 +27,11 @@
          :blacklist/added-by schema-base/UserWithAttributes
          :blacklist/added-at DateTime))
 
+(defn user-or-resource-not-found-error [command]
+  (when-not (and (users/user-exists? (get-in command [:blacklist/user :userid]))
+                 (resource/ext-id-exists? (get-in command [:blacklist/resource :resource/ext-id])))
+    (not-found-json-response)))
```
Macroz, line 33: Is it not found or invalid? The resource in question is blacklist, and a parameter in the "command" is invalid. That would suggest 400 to me. If the URL is invalid, i.e. main resource missing, then 404?

src/clj/rems/api/blacklist.clj
```diff
@@ -54,13 +60,14 @@
...
       (let [userid (user-mappings/find-userid (getx-in command [:blacklist/user :userid]))
             command (assoc-in command [:blacklist/user :userid] userid)]
-        (blacklist/add-user-to-blacklist! (getx-user-id) command)
-        (doseq [cmd (rejecter-bot/reject-all-applications-by userid)]
-          (let [result (command/command! cmd)]
-            (when (:errors result)
-              (log/error "Failure when running rejecter bot commands:" {:cmd cmd :result result})))))
-
-      (ok {:success true}))
+        (or (user-or-resource-not-found-error command)
```
Macroz, line 63: I think this would read slightly better, and like the other places where use `or`-pattern, if the error would be split into `invalid-user-error` and `invalid-resource-error`. (or `not-found-*-error`). No need to cram them together into one function?

src/clj/rems/api/blacklist.clj
aatkin: mayhaps HTTP 422? looking at this SO thread https://stackoverflow.com/questions/42143115/which-status-code-is-correct-404-or-400-and-when-to-use-either-of-these

Issue #2948, opened 2022-05-30
Better warning for old attributes


Macroz:
This message is confusing as it shows the old style / old value and not the new.
```
Exception in thread "main" clojure.lang.ExceptionInfo: Please migrate to :oidc-userid-attributes {:oidc-userid-attribute "sub"}
	at rems.validate$validate.invokeStatic(validate.clj:56)
	at rems.validate$validate.invoke(validate.clj:53)
	at rems.standalone$start_app.invokeStatic(standalone.clj:83)
	at rems.standalone$start_app.doInvoke(standalone.clj:76)
	at clojure.lang.RestFn.applyTo(RestFn.java:137)
	at clojure.core$apply.invokeStatic(core.clj:667)
	at clojure.core$apply.invoke(core.clj:662)
	at rems.standalone$_main.invokeStatic(standalone.clj:141)
	at rems.standalone$_main.doInvoke(standalone.clj:100)
	at clojure.lang.RestFn.applyTo(RestFn.java:137)
	at rems.standalone.main(Unknown Source)

```

PR #2949, opened 2022-05-30
Improve OIDC config validation message

aatkin:
relates #2948 

- add current and expected configuration to the validation exception when starting REMS with config key `:oidc-userid-attribute`

# Checklist for author

## Reviewability
- [x] Link to issue

### DIFF
--- a/src/clj/rems/validate.clj
+++ b/src/clj/rems/validate.clj
@@ -52,8 +52,10 @@
 
 (defn validate []
   (log/info "Validating configuration")
-  (when (:oidc-userid-attribute env)
-    (throw (ex-info "Please migrate to :oidc-userid-attributes" (select-keys env [:oidc-userid-attribute]))))
+  (when-some [oidc-attribute (:oidc-userid-attribute env)]
+    (throw (ex-info "Please migrate to :oidc-userid-attributes"
+                    {:from (select-keys env [:oidc-userid-attribute])
+                     :to {:oidc-userid-attributes [{:attribute oidc-attribute}]}})))
 
   (log/info "Validating data")
   (try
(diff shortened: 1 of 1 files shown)


src/clj/rems/validate.clj
```diff
@@ -52,8 +52,10 @@
 
 (defn validate []
   (log/info "Validating configuration")
-  (when (:oidc-userid-attribute env)
-    (throw (ex-info "Please migrate to :oidc-userid-attributes" (select-keys env [:oidc-userid-attribute]))))
+  (when-some [oidc-attribute (:oidc-userid-attribute env)]
+    (throw (ex-info "Please migrate to :oidc-userid-attributes"
+                    {:from (select-keys env [:oidc-userid-attribute])
+                     :to {:oidc-userid-attributes [{:attribute oidc-attribute}]}})))
```
Macroz, line 58: I believe it could have been a sequence (as per https://github.com/CSCfi/rems/commit/1c83d3d1780bb8d6158e44561f8665ef4d1c3959) so maybe add a loop?

aatkin: ah thanks! i completely forgot and just looked at the single attribute as in the issue screenshot.

src/clj/rems/validate.clj
```diff
@@ -52,10 +52,13 @@
 
 (defn validate []
   (log/info "Validating configuration")
-  (when-some [oidc-attribute (:oidc-userid-attribute env)]
+  (when-let [old-attributes (some-> env :oidc-userid-attribute list flatten)]
```
Macroz, line 55: I don't think I have ever seen this use of `list flatten` :sweat_smile:. So it handles singular and sequential version.

Macroz (approved):

Issue #2755, opened 2021-10-20
Use Shadow-CLJS to build the front-end


Macroz:
Using Shadow-CLJS to build the front-end would help use npm dependencies such as latest `react-select`.

We could keep Leiningen for the back-end / packaging.

This also means removing Figwheel (classic) as Shadow will replace it.

- [x] read and update docs (development.md, ...)
- [x] check circle build
- [x] make new release
- [x] decide between re-frisk and re-frame-10x (enabled re-frisk)
- [x] replace doo with shadow-cljs or kaocha (replaced with shadow-cljs + karma)
- [x] separate build for dev with less optimizations or at least source maps and pseudo names (can use `--debug`)

Macroz:
We should finish this before any UI refresh.

PR #2950, opened 2022-05-31
Into shadow

Macroz:
Closes all of #2755 except the new release.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing
- All the current tests pass
- Manual testing seems to show it works

## Follow-up


src/cljs/rems/spa.cljs
```diff
@@ -592,16 +592,16 @@
       (set-location! url))
     (rf/dispatch [:landing-page-redirect!])))
 
-(secretary/defroute "*" []
-  (rf/dispatch [:set-active-page :not-found]))
+;; (secretary/defroute "*" []
+;;   (rf/dispatch [:set-active-page :not-found]))
```
aatkin, line 596: is this code still needed?

Macroz:
Noticed a couple things while developing on top of this:
- Figwheel is still mentioned in `styles.clj` (about rendering CSS into files)
- We don't actually need the Shadow `:dev-http` because we can serve everything from the backend, and live reload still works.
- also source maps might be nice in prod too (default is off in shadow) but not sure

Issue #2871, opened 2022-03-20
Applicant's view of application is complicated


Macroz:
The applicant's view of application (esp. draft) is complicated – actual form fields are buried under 1 page of metadata.

We can consider hiding and simplifying it.

For applicants, members (maybe also for reviewer / decider):
- [x] In the "State" block, show only the three phase diagram open, others upon expanding.
- [x] In the members block, show only the names of applicants as a list (e.g. "Alice Applicant, Bob Applicant"). other details upon expanding

PR #2951, opened 2022-05-31
feat: slim application state UI to non-handler users

aatkin:
relates #2871 

- as non-handler user, in application "State" block show only the three phase diagram open, others upon expanding
- as non-handler user, in application members block show only the names of applicants as a list (e.g. "Alice Applicant, Bob Applicant"), other details upon expanding

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks #2961

## Screenshots
as applicant, collapsed state and member blocks
Screenshot 2022-06-07 at 12 02 02

---
expanded
Screenshot 2022-06-07 at 11 16 26

Screenshot 2022-06-07 at 12 03 41


src/cljs/rems/collapsible.cljs
```diff
@@ -14,11 +15,14 @@
...
-(defn- hide [id]
+(defn- hide [id callback]
   (.collapse (js/$ (str "#" id)) "hide")
   (.. (js/$ (str "." id "-less"))
       (collapse "hide")
-      (one "hidden.bs.collapse" (fn [_] (. (js/$ (str "." id "-more")) collapse "show"))))
+      (one "hidden.bs.collapse" (fn [_]
+                                  (. (js/$ (str "." id "-more")) collapse "show")
+                                  (when callback
+                                    (callback)))))
```
aatkin, line 25: this allows bootstrap element to communicate state back to reagent component so that UI can render "hidden/show" elements in correct order

Macroz: We may also want to consider a new implementation of the collapsible with Bootstrap 5 or so. Maybe CSS animations + React implementation is better.

src/cljs/rems/application.cljs
```diff
@@ -646,9 +647,35 @@
     (for [event (:application/events application)]
       (update event :event/attachments (partial mapv (comp attachments-by-id :attachment/id))))))
 
-(defn- application-state [application config highlight-request-id]
+(defn- render-state-fields [application config events]
```
Macroz, line 650: I kind of preferred the older naming convention of `application-state` component instead of a "render function". Would `application-state-details` communicate that this is the details that we have initially hidden? Same logic for the others.

src/cljs/rems/collapsible.cljs
```diff
@@ -56,34 +60,40 @@
...
-        (when top-less-button? show-less)
-        [:div.collapse {:id id
-                        :class (when open? "show")
-                        :tab-index "-1"}
-         content-hideable]
-        show-more
-        (when-not (false? bottom-less-button?) show-less)])
-     content-footer]))
+(defn- block [{:keys [open?]}]
+  (let [show? (r/atom open?)] ; track internal open/closed status
```
Macroz, line 66: This kind of duplicates what was previously handled by the classes/ids. Would be good to refactor this into the atom completely (with CSS animation for opening/closing). Doesn't have to happen now, but we can make sure we have an issue for it.

Macroz (approved):
Looks good to me. Let's make sure we have issue for the eventual refactoring.

src/cljs/rems/collapsible.cljs
aatkin: yeah, currently bootstrap is handling the animation. i'll create a ticket for this

aatkin: #2961

Issue #2928, opened 2022-05-09
Fix misalignment of licenses


Macroz:
![Screenshot_2022-05-09_17-20-04](https://user-images.githubusercontent.com/823661/167430494-29f3e98c-e5c7-4590-828d-101597f6c720.png)

PR #2962, opened 2022-06-10
fix indentation of not accepted licenses

aatkin:
relates #2928

`rems.atoms/empty-symbol` uses `fa-stack` to represent empty state. however `fa-stack` is not an actual icon, but rather an mechanism to compose a new icon out of existing ones, and it has much larger width than default icons which have a varying width by themselves. 

- implement `rems.atoms/make-empty-symbol` that can create a new "empty symbol" out of any given element with `opacity: 0` which takes up the exact same space

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Screenshots
before
Screenshot 2022-06-10 at 10 51 06

---
after
Screenshot 2022-06-10 at 11 57 41


Macroz (approved):
I guess we could also add symbols an optional props map that could have the opacity or so. Let's merge this one anyway.

Issue #2767, opened 2021-11-02
Autosave in the background


Macroz:
Autosave in the background by replacing/patching the latest draft-saved event. If the latest event is not a draft-saved then create a new one. This may require some event handling logic to change.

See https://github.com/CSCfi/rems/blob/master/docs/architecture/017-applicant-ux.md.

- [x] automatic saving in the background
- [x] compacting consecutive saves into one

Macroz:
The feature is now ready to experiment with and test. A few leftover ideas and tasks for changing into default have been recorded to #3037.

PR #2972, opened 2022-06-22
Autosave

Macroz:
Implements autosave from #2767 

Autosave happens "on change", when a second has passed from last activity.

# Notes

- [x] When user session has expired, autosave does not actually autosave after the login redirect. 
  - OK, session is not likely to expire with keep alives and while you are editing the autosave is triggered after 1s so there shouldn't be anything significant lost
- [x] Should the event description be changed "Alice saved a draft." -> "Alice updated the application."
  - Let's do it after we clean up the regular save draft and this is the default behavior (#3037).
- [x] Move attachment save feedback to actions box without the scrolling. We could make it use the autosave or at least imitate the behavior (view does not jump away).
- [x] Attachment remove does not trigger autosave.
- [x] Clean up the flash message code. There are many similar functions. (separate issue #3038)
- [x] Browser test both autosave and regular save

# Screenshots

autosave1
autosave2
autosave3

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] Events are backwards compatible _or_ have migrations (events are the same)
- [x] Config is backwards compatible (autosave is disabled by default except in dev)

## Documentation
- [x] ADR for major architectural decisions or experiments (existing ADR 017)

## Testing

## Follow-up


aatkin:
DUO field is not autosaved currently (although it is still under works, the feature could be implemented at least in initial capacity)

aatkin:
> > DUO field is not autosaved currently (although it is still under works, the feature could be implemented at least in initial capacity)
> 
> Hmm how's that possible since they both call the very same function to do the save? Or do you mean those fields do not trigger an autosave? Yes, I see, because they don't use the same form components.

I actually meant the DUO code adding/removing did not trigger autosave - I didn't test the answer fields, but they should too trigger the save.

src/cljs/rems/application.cljs
```diff
@@ -224,11 +225,38 @@
...
+                      (apply (handle-validations! description application {:on-success #(do (rf/dispatch [::set-autosaving false])
+                                                                                            (flash-message/show-quiet-success! :actions-form-flash [text :t.form/autosave-confirmed]))
+                                                                           :default-success? false
+                                                                           :focus? false})
+                             [response])))
+       {:db (-> db
+                (assoc ::autosaving true)
+                (assoc-in [::edit-application :validation :errors] nil)
+                (assoc-in [::edit-application :validation :warnings] nil))})
+     {})))
```
aatkin, line 258: one idea for shortening the function, if fetching application is not critical to perform prior to validations

```clojure
(->> (handle-validations! description
                          application
                          {:on-success (fn []
                                         (rf/dispatch [::set-autosaving false])
                                         (flash-message/show-quiet-success! :actions-form-flash
                                                                            [text :t.form/autosave-confirmed])
                                         ;; equal to (rf/dispatch [::fetch-application (:application/id application) false])
                                         (reload! (:application/id application) false))
                           :default-success? false
                           :focus? false})
     (save-draft! description application edit-application))
```

test/clj/rems/test_browser.clj
```diff
@@ -507,13 +511,19 @@
...
 
-        (testing "save draft succesfully"
-          (btu/scroll-and-click :save)
-          (is (btu/eventually-visible? :status-success)))
+        (testing "saved manually or automatically"
+          (if (:enable-autosave env)
+            (is (btu/eventually-visible? {:id :status-success :fn/text "Application is saved."}))
+            (do
+              (btu/scroll-and-click :save)
+              (is (btu/eventually-visible? {:id :status-success})))))
```
Macroz, line 519: Let's do both. We can set the autosave on and off, and try that both work.

src/cljs/rems/application.cljs
Macroz: This is now included in #3037

aatkin (approved):
LGTM! :+1:

Issue #2974, opened 2022-07-18
Readonly checkbox background is incorrectly rendered


aatkin:
> There is a gray line in the middle of the readonly-checkbox. Please remove.

PR #2977, opened 2022-07-19
Fix read-only checkbox visuals

aatkin:
# Checklist for author

relates #2974

- Readonly checkbox had incorrectly rendered background. This was due to `.readonly-checkbox` class being applied to `<i>` element, which had 1px height and thus rendered only single pixel height gray line instead of gray background. Quick fix to this was to add encompassing `<span>` element with the readonly-checkbox class.

## Reviewability
- [x] Link to issue

## Documentation

---
### before
Screenshot 2022-07-19 at 15 00 38

### after
Screenshot 2022-07-19 at 15 00 13

### DIFF
--- a/src/cljs/rems/atoms.cljs
+++ b/src/cljs/rems/atoms.cljs
@@ -87,7 +87,8 @@
               :aria-checked value
               :aria-label (if value
                             (text :t.form/checkbox-checked)
-                            (text :t.form/checkbox-unchecked))}
+                            (text :t.form/checkbox-unchecked))
+              :aria-readonly (nil? on-change)}
        on-change (assoc :on-click wrapped-on-change
                         :on-key-press #(when (= (.-key %) " ")
                                          (wrapped-on-change %))))]))
--- a/src/cljs/rems/atoms.cljs
+++ b/src/cljs/rems/atoms.cljs
@@ -82,7 +82,7 @@
                               (on-change value)))]
     [:i.far.fa-lg
      {:id id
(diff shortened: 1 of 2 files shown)


src/cljs/rems/atoms.cljs
```diff
@@ -78,18 +78,19 @@
...
+              :class [:checkbox class (if value :fa-check-square :fa-square)]
+              :tabIndex 0
+              :role :checkbox
+              :aria-checked value
+              :aria-label (if value
+                            (text :t.form/checkbox-checked)
+                            (text :t.form/checkbox-unchecked))}
+       on-change (assoc :on-click wrapped-on-change
+                        :on-key-press #(when (= (.-key %) " ")
+                                         (wrapped-on-change %))))]))
```
Macroz, line 94: Should we `let` the `wrapped-on-change` here if it's not used elsewhere?

Macroz (approved):

src/cljs/rems/atoms.cljs
Macroz: If you like, you can fix this later, I'll merge to get the fix in.

aatkin: did you mean binding inside the conditional expression on line 92? the `cond->` tries to macroexpand the expression so that cannot work

Macroz: Yeah, right, that would make it rather difficult to thread in.

Issue #2869, opened 2022-03-20
Extra styles


Macroz:
Add a simple mechanism of configuring extra styles like translations and pages. We need to investigate to make sure important styles can be overridden.

Specifically custom fonts need good support though this workaround works for both https://github.com/CSCfi/rems/issues/2586#issuecomment-900866366

Another specific request is to affect font styling line letter spacing in https://github.com/CSCfi/rems/issues/1467

Macroz:
We discovered we are still missing the Docker (Compose) support for this.

PR #2987, opened 2022-08-16
Extra stylesheets

aatkin:
relates #2869 

custom fonts and custom stylesheets have been a topic every now and then. would this solution work?

- added new config option `:extra-stylesheets` that works similarly to `:extra-scripts`, but includes css-files instead of javascript

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Changes since v2.27
 - Validate organization when adding or editing it. (#2964)
 - Consecutive save events are compacted into one. This does not affect old save events. This is turned off by default, until the whole autosave feature is finished. (#2767)
 - Application licenses are now rendered alphabetically in UI and PDF render. (#2979)
-- Custom stylesheets can now be added using `:extra-stylesheets` configuration option. This can be used to include custom fonts, for example.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Changes since v2.27
 - Validate organization when adding or editing it. (#2964)
 - Consecutive save events are compacted into one. This does not affect old save events. This is turned off by default, until the whole autosave feature is finished. (#2767)
 - Application licenses are now rendered alphabetically in UI and PDF render. (#2979)
+- Custom stylesheets can now be added using `:extra-stylesheets` configuration option. This can be used to include custom fonts, for example.
```
Macroz, line 27: Let's add the ticket number here #2869

Macroz (approved):
Looks good, like I thought it'd look like.

Perhaps we could add some tests to extra scripts + stylesheets, something like we have for extra translations (and extra pages). Ideally we would have a test that checks that a font can be changed (like using the browser). If it's tricky then perhaps we could move the Lato etc. into the default extra stylesheet so they we could see that they work.

PR #3053, opened 2022-10-04
Require user to be owner or in organization owners to allow organization edit

### LINKED ISSUE #2828: Organization owners could be allowed to edit their own organization owners

aatkin:
closes #2828 

Previously any user with `:organization-owner` role could edit any organization in admin view. Editing organization owners was restricted in #3043, but it was deemed that user should not be able to edit organization at all if they are not in that organizations owner list. User must now be either `:owner` or `:organization-owner` of the organization they wish to edit. Organization edit button is now hidden for users that do not meet these requirements.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

Leftover task #3061


src/clj/rems/api/services/organizations.clj
```diff
@@ -57,19 +54,12 @@
...
-   (:organization/id org)
-   (fn [db-organization]
-     (let [organization-owners (set (map :userid (:organization/owners db-organization)))
-           organization-owner? (contains? organization-owners userid)
-           owner? (contains? context/*roles* :owner)]
-       (merge db-organization
-              (cond-> org
-                true (dissoc :organization/id)
-                (not (or organization-owner? owner?)) (dissoc :organization/owners))))))
+(defn edit-organization! [{:organization/keys [id] :as org}]
```
Macroz, line 57: We have tried to avoid destructuring by namespaced keys since it makes it difficult to grep for the uses. This is not always followed in the test utils but perhaps should.

src/clj/rems/db/test_data_helpers.clj
```diff
@@ -80,7 +81,7 @@
     (:organization/id result)))
 
 (defn ensure-default-organization! []
-  (when-not (organizations/get-organization-raw {:organization/id "default"})
+  (when-not (rems.db.organizations/get-organization-by-id-raw "default")
```
Macroz, line 84: Good change. Now we don't have dependency from `db` layer to `service`.

src/clj/rems/api/services/util.clj
```diff
@@ -1,6 +1,6 @@
 (ns rems.api.services.util
   (:require [clojure.test :refer [deftest is testing]]
-            [rems.api.services.organizations :as organizations]
+            [rems.db.organizations]
```
Macroz, line 3: Good change. Now we don't have dependency from `db` layer to `service`.

src/cljs/rems/administration/organization.cljs
```diff
@@ -61,35 +61,39 @@
...
+                [review-emails-field (:organization/review-emails organization)]
+                [inline-info-field (text :t.administration/active) [readonly-checkbox {:value (status-flags/active? organization)}]]]}]
+     (let [id (:organization/id organization)]
+       [:div.col.commands
+        [administration/back-button "/administration/organizations"]
+        (when org-owner?
+          [edit-button id])
+        [roles/show-when roles/+admin-write-roles+ ;; TODO doesn't match the API roles exactly
+         [status-flags/enabled-toggle organization #(rf/dispatch [:rems.administration.organizations/set-organization-enabled %1 %2 [::enter-page id]])]
+         [status-flags/archived-toggle organization #(rf/dispatch [:rems.administration.organizations/set-organization-archived %1 %2 [::enter-page id]])]]])]))
```
Macroz, line 96: Should we replace this with the check for the `org-owner?` for the other buttons as well? Since it is the same condition, isn't it?

Macroz (approved):
Looks just like we discussed. 

Should we add the other organization "modification buttons" with the same ownership check?

src/cljs/rems/administration/organization.cljs
aatkin: :+1: it's almost the same condition: `(def roles/+admin-write-roles+ #{:organization-owner :owner})` is not as strict about what organization owner means since user can be organization owner in another organization. the meaning is preserved as i think it was unintended previously that _any_ org owner could edit any organization.

aatkin: modified this to check that user must strictly be `#{:owner}` as organization owners currently cannot use enable/archive actions, as discussed

PR #3068, opened 2022-10-17
Delete orphan attachments

### LINKED ISSUE #3041: Remove orphan attachments when sending application
We can remove orphan attachments when sending the application. Since these attachments are from the applicant, that's a good time to clean them away.

Macroz:
Closes #3041

Creates a new process manager that cleans an application on submission of unused attachments. An unused attachment is something that is not used in a field value (new or old) or an event (attachment). This can happen if an attachment is uploaded but the application is not saved before submission (or something is uploaded in its place before saving).

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/application/process_managers.clj
+++ b/src/clj/rems/application/process_managers.clj
@@ -1,0 +1,44 @@
+(ns rems.application.process-managers
+  "Miscellaneous process managers."
+  (:require [clojure.set :refer [difference]]
+            [rems.api.services.attachment :as attachment]
+            [rems.api.services.blacklist :as blacklist]
+            [rems.common.application-util :as application-util]
+            [rems.db.applications :as applications]
+            [rems.db.attachments :as attachments]))
+
+(defn revokes-to-blacklist
+  "Revokation causes the users to be blacklisted."
+  [new-events]
+  (doseq [event new-events
+          :when (= :application.event/revoked (:event/type event))
+          :let [application (applications/get-application-internal (:application/id event))]
+          user (application-util/applicant-and-members application)
+          resource (:application/resources application)]
+    (blacklist/add-user-to-blacklist! (:event/actor event)
+                                      {:blacklist/user {:userid (:userid user)}
+                                       :blacklist/resource {:resource/ext-id (:resource/ext-id resource)}
+                                       :comment (:application/comment event)})))
+
+(defn delete-applications
+  "The deleted event causes a side-effect that completely deletes the application."
+  [new-events]
+  (doseq [event new-events]
+    (when (= :application.event/deleted (:event/type event))
+      (applications/delete-application-and-reload-cache! (:application/id event))))
+  [])
+
+(defn delete-orphan-attachments [application-id]
+  (let [application (applications/get-application-internal application-id)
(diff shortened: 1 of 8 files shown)


test/clj/rems/application/test_process_managers.clj
```diff
@@ -0,0 +1,190 @@
...
+                                     handler
+                                     read-ok-body)))
+
+                          ;; NB: attachment 3 should have been cleaned
+
+                          (is (= [{:attachment/id attachment-id1 :attachment/filename "attachment1.txt" :attachment/type "text/plain"}
+                                  {:attachment/id attachment-id2 :attachment/filename "attachment2.txt" :attachment/type "text/plain"}]
+                                 (:application/attachments (applications/get-application-internal app-id))
+                                 (attachments/get-attachments-for-application app-id))
+                              "attachment1 and attachment2 are saved, but not attachment3"))))
```
aatkin, line 124: should we also verify that the attachment has been removed from attachments table?

src/clj/rems/application/process_managers.clj
```diff
@@ -0,0 +1,44 @@
...
+        attachments-in-use (rems.api.services.attachment/get-attachments-in-use application)
+        all-attachments (set (map :attachment/id (:application/attachments application)))]
+    (doseq [attachment-id (difference all-attachments attachments-in-use)]
+      (attachments/delete-attachment! attachment-id))))
+
+(defn delete-orphan-attachments-on-submit
+  "When an application is submitted, we delete its unused attachments, if any."
+  [new-events]
+  (doseq [event new-events]
+    (when (= :application.event/submitted (:event/type event))
```
aatkin, line 43: hypothetically remark and review can both generate attachments that can linger around, not sure if that would ever happen but if it did, then maybe `:application.event/closed` could also trigger delete orphan attachments?

src/clj/rems/application/process_managers.clj
```diff
@@ -0,0 +1,44 @@
+(ns rems.application.process-managers
```
aatkin, line 1: some inconsistent namespace referring here:

```clojure
;; alias on line 20
(blacklist/add-user-to-blacklist! ...)

;; full ns on line 34
(rems.api.services.attachment/get-attachments-in-use application)
``` 

would it make sense to always refer by full namespace when using functions outside this namespace? (with allowed exceptions, like `clojure.core`). there's verbosity included but reading and using is simplified quite a bit.

src/clj/rems/application/process_managers.clj
Macroz: Could be most state changes actually, because a lot of them have the comment that can include an attachment. I think we have to still decide if we want to a) check this in every state change b) have a background job.

test/clj/rems/application/test_process_managers.clj
Macroz: Now that I look at it, that's exactly what `get-attachments-for-application` does, isn't it?

src/clj/rems/application/process_managers.clj
Macroz: Added this concern to https://github.com/CSCfi/rems/issues/3039

Maybe it's an easier solution than the background job.

Issue #2491, opened 2020-12-14
Ability for handler to redact attachments.


opqdonut:
Sometimes the wrong attachment gets added (e.g. #2488) and needs to be redacted. We could add an API (and perhaps a UI) that would let the owner redact (i.e. zero the contents of) attachments. Redaction could also add an event to the application events, e.g. "Handler redacted attachment X".

- [x] implement an API to redact an attachment (like the other commands in `api/applications.clj`)
  - [x] clear the attachment contents in DB, possibly in process manager
  - [x] add new event of `attachment-redacted` with optional comment
- [ ] implement action in UI for redacting an attachment similar to review action with checkboxes to filenames
- [x] show event and optional comment for handlers always, and to applicants if redacting an application attachment
  - If the event where the attachment came from is shown to the applicant, then the redaction should be shown too. Otherwise not.
- [x] Show related events could show the event from where the redacted attachment came from.
- [x] Attachment names should be unique (seem to be already e.g. `foo (1).png`, `foo (2).png`)

First PR could be the API with the UI later.

Macroz:
Could be that it should be the handler, in case the handler has added the document.

PR #3076, opened 2022-11-01
Attachment redaction API

aatkin:
relates #2491 (partial implementation)

# API for attachment redaction
- allow `:handler`, `:decider` and `:reviewer` to redact attachments whenever these roles have the permission to add attachments via another command
- process manager handles redaction, by zeroing file contents of given attachment. filename is masked in application model via enrichment and replaced with keyword `:filename/redacted`, which is returned from the API. localization support has been added so that REMS client can render the redacted filename localized.
- redacted attachments are not available for download via client, but they can still be downloaded using the API.
- `:redacted-attachments` event property describes both redacted attachments and events from where the redacted attachments originate from. this can be used for highlighting events in UI. it is relatively similar to `:request-id` but can contain multiple references.
- highlighting mechanism has been refactored to support different highlighting scenarios. currently there are two: `request-id` and redacted attachment events. adding more scenarios in the future should be relatively straightforward.

```clojure
;; :application.command/redact-attachments
(s/defschema RedactAttachmentsCommand
  (assoc CommandWithComment
         :redacted-attachments [CommandAttachment]
         :public s/Bool))
```

```clojure
;; :application.event/attachments-redacted
(s/defschema AttachmentsRedactedEvent
  (assoc EventWithComment ; includes comment and attachments
         :event/type (s/enum :application.event/attachments-redacted)
         :event/redacted-attachments [(merge EventAttachment {:event/id s/Int})]
         :application/public s/Bool))
``` 

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is completely new

## Documentation
- [ ] Update changelog if necessary (maybe could include with separate PR that implements UI?)

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks #3081 

---

## Screenshots

application
Screenshot 2022-12-08 at 18 10 11
Screenshot 2022-12-09 at 10 29 32

pdf
Screenshot 2022-12-09 at 10 28 04


src/clj/rems/application/commands.clj
```diff
@@ -413,6 +421,33 @@
                                           (when-let [end (:entitlement-end cmd)]
                                             {:entitlement/end end})))))
 
+(defn- empty-redacted-attachments-error [cmd]
+  (when (empty? (:redacted-attachments cmd))
+    {:errors [{:type :empty-redacted-attachments}]}))
+
+(defn- find-original-attachment-event-ids
```
Macroz, line 428: Should we use this information to allow the decider, reviewer to only redact attachments they added, but allow handler to redact any attachment?

src/cljs/rems/application.cljs
```diff
@@ -647,11 +659,12 @@
...
      [:div.col-sm-10
       [:div.col-form-label.event-description [:b event-text]
-       (when request-id
+       (when (or request-id attachments-from)
          [:div.float-right
           [:a {:href "#"
                :on-click (fn [e]
-                           (rf/dispatch [::highlight-request-id request-id])
+                           (rf/dispatch [::highlight-request-ids [request-id]])
+                           (rf/dispatch [::highlight-event-ids (map :event/id attachments-from)])
```
Macroz, line 667: Can we later refactor and combine these two? Could add a separate task.

test/clj/rems/api/test_applications.clj
```diff
@@ -1700,6 +1723,39 @@
...
+        (is (= ["handler-public-remark.txt"
+                "reviewer-review.txt"
+                "kissat-koiria-gfffögkfhjd.txt"]
+               (get-application-attachments application-id reviewer-id)))
+        (let [replace-id (add-attachment reviewer-id (file "reviewer-private-remark.txt"))]
+          (is (= {:success true} (send-command reviewer-id
+                                               {:type :application.command/redact-attachments
+                                                :application-id application-id
+                                                :comment "accidental upload, redacting"
+                                                :public false
```
Macroz, line 1743: Should we test one with `:public true`?

test/clj/rems/application/test_commands.clj
```diff
@@ -2122,3 +2128,64 @@
...
+        injections {:valid-user? #{applicant-user-id handler-user-id "reviewer"}
+                    :find-userid identity
+                    :get-attachment-metadata
+                    {attachment-id-1 {:application/id (:application/id application)
+                                      :attachment/id attachment-id-1
+                                      :attachment/user handler-user-id}
+                     attachment-id-2 {:application/id (:application/id application)
+                                      :attachment/id attachment-id-2
+                                      :attachment/user handler-user-id}}}]
+    (testing "handler can redact attachment"
```
Macroz, line 2149: We could perhaps test the case where `:redacted-attachments` is empty because it's new code and only in use here.

Macroz (approved):
Looks good. 

I left some comments to consider going forward.

resources/sql/queries.sql
```diff
@@ -309,6 +309,11 @@ UPDATE attachment
 SET (appId, userId, filename, type) = (:application, :user, :filename, :type)
 WHERE id = :id;
 
+-- :name redact-attachment! :!
+UPDATE attachment
+SET (data, filename, type) = (decode('', 'hex'), 'redacted.txt', 'text/plain')
```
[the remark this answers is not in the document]
aatkin: refactored the implementation so that filenames are preserved in db and instead enriched in model via `:application/attachments`. this needed a small fix to attachments zip mechanism to use `:application/attachments` metadata, for which i updated tests as well.

resources/translations/fi.edn
```diff
@@ -180,6 +180,7 @@
                  :entitlement-end "Pääsyoikeudet loppuvat %1."
                  :events {:applicant-changed "%1 vaihtoi hakijan henkilöksi %2."
                           :approved "%1 hyväksyi hakemuksen."
+                          :attachments-redacted "%1 poisti liitetiedostoja."
```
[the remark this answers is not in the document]
aatkin: added conditional translation to the event so that new attachments display a text indicating that redacted attachments were replaced

test/clj/rems/api/test_applications.clj
```diff
@@ -1764,15 +1781,238 @@
...
+                                      handler))))
+    (testing "invite reviewer"
+      (is (= {:success true} (send-command handler-id
+                                           {:type :application.command/request-review
+                                            :application-id application-id
+                                            :reviewers [reviewer-id]
+                                            :comment "please"}))))
+    (testing "invite decider"
+      (is (= {:success true} (send-command handler-id
+                                           {:type :application.command/request-review
```
Macroz, line 1829: Was this supposed to be `request-decision`?

src/clj/rems/application/model.clj
```diff
@@ -667,6 +673,19 @@
...
+       (filter (comp #{:application.event/attachments-redacted} :event/type))
+       (mapcat :application/redacted-attachments)
+       (map :attachment/id)
+       set))
+
+(defn- mask-redacted-attachments [attachments redacted-ids]
+  (for [attachment attachments
+        :let [redacted? (some #{(:attachment/id attachment)} redacted-ids)]]
+    (assoc-some attachment :attachment/filename (when redacted?
+                                                  "redacted"))))
```
Macroz, line 687: Is the value `"redacted"` going to show up in the UI? I guess we must translate it in that case in the UI. :thinking:

Macroz (approved):
Looks like this part is almost done. I would try to nail the event before we release this or perhaps even in this PR.  We can add to the issue as follow-up.

src/clj/rems/application/commands.clj
```diff
@@ -408,36 +431,62 @@
...
+  "Finds events from which the redacted attachments originate from,
+   and returns those events ids."
+  [redacted-attachments application]
+  (let [events (:application/events application)
+        redacted-ids (set (map :attachment/id redacted-attachments))]
+    (vec (for [event events
+               :let [event-attachment-ids (set (map :attachment/id (:event/attachments event)))]
+               :when (seq (clojure.set/intersection redacted-ids event-attachment-ids))]
+           (select-keys event [:event/id])))))
```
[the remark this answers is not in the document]
aatkin: hmm, i see what you mean. it makes more sense from the model side of view to include redacted attachment id with the originating event id. however the context seeps unfortunately into the highlighter mechanism, which is dependent on event data. one solution could be to add `:application/request-ids` which accepts a list of related ids, and map the request ids into redacted event. however this does not take into account events that have no generated request ids, for example remarked event.

src/clj/rems/application/model.clj
```diff
@@ -691,6 +710,7 @@
       (update :application/events (partial mapv #(enrich-event % get-user get-catalogue-item)))
       (assoc :application/applicant (get-user (get-in application [:application/applicant :userid])))
       (assoc :application/attachments (get-attachments-for-application (getx application :application/id)))
+      (update :application/attachments mask-redacted-attachments (get-redacted-attachment-ids application))
```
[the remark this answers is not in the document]
aatkin: agreed, this looks like better location for applying privacy filters.

src/clj/rems/application/model.clj
```diff
@@ -748,8 +769,8 @@
   (->> events
        (remove (comp sensitive-events
                      :event/type))
-       (remove #(and (= :application.event/remarked
-                        (:event/type %))
+       (remove #(and (some #{(:event/type %)} [:application.event/remarked
+                                               :application.event/attachments-redacted])
                      (not (:application/public %))))))
```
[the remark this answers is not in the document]
aatkin: agree, renaming requires migration work which seems outside the scope of this PR. the naming is confusing so refactoring is a good idea. i'll create issue to track the work.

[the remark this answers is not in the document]
aatkin: created #3087

src/clj/rems/application/commands.clj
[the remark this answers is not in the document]
Macroz: I think the highlighter very likely should do different things depending on what the event is. 
- So for a review request or response, it could go through all the events and pick the ones that have the same request id. 
- For the redaction event, it should highlight the event where the attachment came from. 

I think we should have all this data available in the UI, so we can do it in the UI. There are also only "few" events ever (like 100 or so at max) so it should be workable? I don't think we need to pack this stuff somehow into the events or the read model?

test/clj/rems/api/test_applications.clj
aatkin: yeah, copy-paste error here. good catch :)

src/clj/rems/application/model.clj
aatkin: yes, that was the original intention. localization is a good idea for the client. we also need to make sure that the original filename is visible only to the user who owns the attachment. this requires some refactoring, as a few places have already relied solely in the attachment db data instead of application model.

src/clj/rems/application/commands.clj
```diff
@@ -442,24 +442,23 @@
...
   (or (empty-redacted-attachments-error cmd)
       (add-comment-and-attachments cmd application injections
                                    {:event/type :application.event/attachments-redacted
-                                    :application/redacted-attachments (vec (:redacted-attachments cmd))
-                                    :application/attachments-from (find-original-attachment-event-ids
-                                                                   (:redacted-attachments cmd)
-                                                                   application)
+                                    :application/redacted-attachments (find-original-attachment-event-ids
+                                                                       cmd
+                                                                       application)
```
aatkin, line 461: could rename to `find-redacted-attachments`?

src/clj/rems/application/events.clj
```diff
@@ -120,8 +120,8 @@
 (s/defschema AttachmentsRedactedEvent
   (assoc EventWithComment
          :event/type (s/enum :application.event/attachments-redacted)
-         :application/redacted-attachments [EventAttachment]
-         :application/attachments-from [{:event/id s/Int}]
+         :application/redacted-attachments [(merge EventAttachment
+                                                   {:event/id s/Int})]
```
aatkin, line 124: could def as `EventId` or so?

src/cljs/rems/application.cljs
```diff
@@ -171,41 +171,42 @@
...
-                                      (:event/id highlightable-event)))]
-      (when (< 1 (count highlightable-events))
-        highlightable-events))
-
-    nil))
+(defn- set-highlightables-by-request-id [events]
+  (let [related-events (->> events
+                            (filter :application/request-id)
+                            (group-by :application/request-id)
+                            (filter (comp #(< 1 %) count second))
```
aatkin, line 178: filter-vals instead?

src/cljs/rems/application.cljs
```diff
@@ -171,41 +171,42 @@
...
+                            (mapcat (fn [event]
+                                      (->> (map :event/id (:application/redacted-attachments event))
+                                           (interleave (repeat (:event/id event))))))
+                            (partition 2))] ; pairs of [event redacted-event]
+    (for [event events
+          :let [event-id (:event/id event)
+                highlightable-event-ids (->> related-events
+                                             (find-first #(contains? (set %) event-id)))]]
+      (update event :highlight-event-ids #(set (concat (or % [])
+                                                       highlightable-event-ids))))))
```
aatkin, line 200: filter-vals instead? (then can drop into)

use `for` instead of mapcat/partition

src/cljs/rems/application.cljs
```diff
@@ -171,41 +171,42 @@
...
    be highlighted within the UI."
   [db]
-  (let [set-highlightables (fn [events event]
-                             (->> (find-highlightable-event-ids event events)
-                                  (assoc event :highlight-event-ids)))]
-    (update-in db [::application :data :application/events]
-               #(map (partial set-highlightables %) %))))
+  (update-in db [::application :data :application/events]
+             (comp set-highlightables-by-request-id
+                   set-highlightables-by-redacted-attachments)))
```
aatkin, line 209: consider writing unit test for highlightables?

src/clj/rems/api/services/attachment.clj
```diff
@@ -63,7 +63,9 @@
         ;; failsafe in case we have duplicate filenames in old
         ;; applications
         (try
-          (.putNextEntry zip (ZipEntry. (getx metadata :attachment/filename)))
+          (.putNextEntry zip (ZipEntry. (let [filename (getx metadata :attachment/filename)]
+                                          (if (= :filename/redacted filename)
+                                            "redacted" filename))))
```
aatkin, line 68: consider separating if branches with newline

Issue #3017, opened 2022-09-08
Report REMS technical layers


Macroz:
Automatically create a report that contains a list of REMS namespaces, functions grouped into technical layers (API, service, DB, SQL).

Macroz:
Preview of the report. 
- This one has colors R G B for `API -> service -> DB`. Ext is in cyan.
- Edges in red are "disallowed", and don't affect layout.
- `rems.api.services` has been renamed `rems.service` to see if it'd make things clearer.
- `rems.db.test-data` has been moved to `rems.service.test-data` as that seems like a better place.
![layers](https://user-images.githubusercontent.com/823661/189383245-66df2099-b360-43d5-b272-7dea6131b07d.png)

PR #3094, opened 2022-11-30
Rename rems.db.test-data namespace to rems.service.test-data

aatkin:
relates #3017

- rename `rems.db.test-data` ns to `rems.service.test-data` and update references from other namespaces
- update `rems.service.test-data` ns docstring

# Checklist for author

## Reviewability
- [x] Link to issue

### DIFF
--- a/src/clj/rems/service/test_data.clj
+++ b/src/clj/rems/service/test_data.clj
@@ -1,4 +1,4 @@
-(ns rems.db.test-data
+(ns rems.service.test-data
   "Populating the database with nice test data."
   (:require [clj-time.core :as time]
             [clojure.string :as str]
--- a/src/clj/rems/service/test_data.clj
+++ b/src/clj/rems/service/test_data.clj
@@ -1,5 +1,7 @@
 (ns rems.service.test-data
-  "Populating the database with nice test data."
+  "Generate and populate database with usable test data. Contains multiple high-level
+   test data helper functions. Separate functions are provided to generate complete
+   test and demo data, which create the same data, but with different users."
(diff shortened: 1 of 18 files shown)


src/clj/rems/service/test_data.clj
```diff
@@ -1,5 +1,7 @@
-(ns rems.db.test-data
-  "Populating the database with nice test data."
+(ns rems.service.test-data
+  "Generate and populate database with usable test data. Contains multiple high-level
+   test data helper functions. Separate functions are provided to generate complete
+   test and demo data, which create the same data, but with different users."
```
aatkin, line 4: judging both `(create-test-data!)` and `(create-demo-data!)` the user data seemed like key difference to me, and maybe good to point out?

test/clj/rems/test_layers.clj
```diff
@@ -26,11 +26,6 @@
            (str/starts-with? s "rems.service.")
            (str/starts-with? s "rems.api"))))
 
-(defn rename-ns [s]
-  ;; TODO implement these namespace renames
-  (-> s
-      (str/replace "rems.db.test-data" "rems.service.test-data")))
-
```
Macroz, line 33: We could still keep the function for future renames as we progress with the architecture tool. Although, we can of course bring it back from git history.

Macroz (approved):

test/clj/rems/test_layers.clj
aatkin: let's keep the function for now, it's easier to modify when present in the code (easy to skip by adding reader tags).

PR #3095, opened 2022-12-01
Move /entitlement.csv to entitlements API

aatkin:
relates #3017 

NB: /entitlements.csv is deprecated and redirects to /api/entitlements/export-csv , functionality should remain the same as before when no query parameters are supplied.

- remove namespace rems.entitlements (contained /entitlements.csv route)
  - create route /api/entitlements/export-csv
  - redirect /entitlements.csv to /api/entitlements/export-csv (added deprecation notice to changelog)
- add service layer to entitlements and permissions, contains db/service/api level refactoring
  - add exception middleware for 501 not implemented response code (permissions api)
- entitlements api improvements:
  - use route :roles to validate user in /api/entitlements/export-csv
  - unify entitlements json and export-csv route query parameters (user/resource/expired)
  - allow `separator` selection in export-csv
  - test export-csv in entitlements api (instead of db mocks)

exception middleware could be modularized somehow. compojure-api has `restructure-params` mechanism (used by roles) but i encountered some weird behaviour in tests and decided to not proceed further there. practically this would be combination of middleware and swagger responses so that single assertion adds a possible response to the chain, like 501 not implemented etc.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible and new, should be deprecated at a later point in time

## Documentation

## Follow-up


src/clj/rems/handler.clj
```diff
@@ -61,7 +60,10 @@
     (redirect "/redirect"))
 
   (GET "/favicon.ico" []
-    (redirect "/img/favicon.ico")))
+    (redirect "/img/favicon.ico"))
+
+  (GET "/entitlements.csv" [] ; DEPRECATED: legacy url redirect 
+    (redirect "/api/entitlements/csv")))
```
aatkin, line 66: i'll add reminder ticket about deprecation if this is a viable solution. redirect is quite dependent on clients following the location response header.

test/clj/rems/test_db.clj
```diff
@@ -65,52 +63,6 @@
   (is (= #{:logged-in} (roles/get-roles "juho"))) ;; default role
   (is (thrown? RuntimeException (roles/add-role! "pekka" :unknown-role))))
 
-(deftest test-get-entitlements-for-export
```
Macroz, line 68: So is there any test for getting actual CSV content out?

Macroz (approved):
Looks good. 

I think one improvement I'd like to see is a test that fetches a CSV and checks that it actually is data in the CSV format.

test/clj/rems/test_db.clj
aatkin: yes, `rems.api.test-entitlements` namespace contains API tests that fetch entitlements from the `/csv` route. the `test-db` namespace is redundant, if we can also test the API and different roles. i can test the other scenarios here as well, once i get them implemented in application code.

src/clj/rems/api/entitlements.clj
```diff
@@ -20,4 +21,14 @@
...
+                                     :resource-ext-id resource
+                                     :expired expired})))
+
+    (GET "/csv" []
+      :summary "Return entitlements as CSV"
+      :roles #{:handler :reporter}
+      :produces ["text/csv"]
+      :responses {200 {:schema s/Str}}
+      (-> (ok (get-entitlements-for-csv-export))
+          (response/content-type "text/csv")))))
```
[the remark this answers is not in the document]
aatkin: idea sounds good, but as you said, the format is currently different to the base route entitlements, and roles are different too (csv export only allows handler and reporter roles). we could change csv export to work with the same attributes and roles. we could figure out proper `format` from request headers and/or query params. documenting the API with swagger doesn't work very well, as swagger 2.0 doesn't seem to support `x/oneOf`, so we cannot document different schemas for same response code.

maybe best way forward would be to refactor underlying code so that both JSON and CSV routes use same functions and differ only in response format? if/when we make the move to a later version of swagger, then we can use the multiple schemas responses with relative ease.

Issue #3028, opened 2022-09-08
Support configuring attributes from token and user_info


Macroz:
We could support random attributes also from user_info, like email could come in the auth token or user_info.

PR #3096, opened 2022-12-09
feat: support configuring attributes from token and user_info

flowdnb:
relates #3028

Support configuring attributes from token and user_info.

## Reviewability
- [x] Link to issue

## Documentation

### DIFF
--- a/src/clj/rems/auth/oidc.clj
+++ b/src/clj/rems/auth/oidc.clj
@@ -90,15 +90,15 @@
     (users/add-user-raw! userid user)
     user))
 
-(defn- get-user-attributes [id-data user-info]
+(defn- get-user-attributes [id-data researcher-status]
   ;; TODO all attributes could support :rename
   (let [userid (or (find-user id-data) (get-new-userid id-data))
-        _ (assert userid (when (:log-authentication-details env) {:id-data id-data :user-info user-info}))
+        _ (assert userid (when (:log-authentication-details env) {:id-data id-data :researcher-status researcher-status}))
         identity-base {:userid userid
                        :name (some (comp id-data keyword) (:oidc-name-attributes env))
                        :email (some (comp id-data keyword) (:oidc-email-attributes env))}
         extra-attributes (select-keys id-data (map (comp keyword :attribute) (:oidc-extra-attributes env)))
-        user-info-attributes (select-keys user-info [:researcher-status-by])]
+        user-info-attributes (select-keys researcher-status [:researcher-status-by])]
     (merge identity-base extra-attributes user-info-attributes)))
 
 ;; XXX: consider joining with rems.db.users/invalid-user?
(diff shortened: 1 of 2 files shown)


src/clj/rems/auth/oidc.clj
```diff
@@ -162,9 +162,10 @@
                 user-info (when-let [url (:userinfo_endpoint oidc-configuration)]
                             (-> (http/get url {:headers {"Authorization" (str "Bearer " access-token)}})
                                 :body
-                                json/parse-string
-                                ga4gh/passport->researcher-status-by))
-                user (find-or-create-user! id-data user-info)]
+                                json/parse-string))
+                researcher-status (ga4gh/passport->researcher-status-by user-info)
+                id-data (merge id-data user-info)
```
Macroz, line 167: Perhaps we should merge `researcher-status` to here too? Then we could pass just the `id-data` which could perhaps be renamed to `user-data`?

Macroz: Alternatively we can merge the parsed `researcher-status` to `user-info` and pass both `id-data` and the full `user-info` to the `find-or-create-user!`.

src/cljs/rems/user.cljs
```diff
@@ -43,7 +43,7 @@
           (for [[k v] other-attributes]
             (let [title (or (localized (get-in extra-attributes [(name k) :name]))
                             k)]
-              [info-field title v {:inline? true}])))))
+              [info-field title (str v) {:inline? true}])))))
```
Macroz, line 46: What sort of values do we get here? Should we build a bit smarter printing here?

src/clj/rems/auth/oidc.clj
```diff
@@ -164,10 +164,10 @@
                                 :body
                                 json/parse-string))
                 researcher-status (ga4gh/passport->researcher-status-by user-info)
-                id-data (merge id-data user-info)
-                user (find-or-create-user! id-data researcher-status)]
+                user-data (merge id-data user-info researcher-status)
+                user (find-or-create-user! user-data)]
             (when (:log-authentication-details env)
-              (log/info "logged in" id-data user-info user))
+              (log/info "logged in" user-data user))
```
Macroz, line 170: Should we keep the printing of the separate structures here, as it would show the situation before the merge, so you can see where each value came from? This is only for debugging IdP issues so it could be useful to know what happens in that level of detail.

Macroz:
Thanks a lot for the PR and the changes look really good too. I added a couple questions. I can accept the PR and continue improvements myself (like on the printing side) after Xmas holidays.

src/cljs/rems/user.cljs
flowdnb: Ohne Titel 2e
The attached screenshot was created with the suggested string conversion. It shows _email_verified_ as a boolean value, resulting in an empty form field before, while voperson_external_affilition is an array, resulting in the user profile page not being displayed in the browser at all.

src/clj/rems/auth/oidc.clj
flowdnb: I agree with this.

src/cljs/rems/user.cljs
Macroz: Thanks, I can add a rudimentary checkbox and array support for this (for example separated by `,`).

Macroz (approved):
Looks fine now. 

I have created the issue https://github.com/CSCfi/rems/issues/3103 to track progress on the better formatting of user attributes.

Macroz:
@flowdnb I have implemented the first version of better formatting here https://github.com/CSCfi/rems/pull/3104 if you are interested to have a look.

PR #3118, opened 2023-02-10
Disable email

### LINKED ISSUE #3116: Option to silence mails to handlers from work other handlers do or likely silence all handling mails from handlers

### LINKED ISSUE #3117: Don't try to send mail if translation used in mail is set to empty
Don't try to send mail if translation used in mail is set to empty. That could then be used to silence some mails.

Macroz:
Close #3117 
Close #3116 

Adds two ways of disabling emails. One for handlers completely, one per email type.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ Changes since v2.31
 ### Additions
 - User attributes can now be retrieved from ID token and user_info endpoint. (#3028)
 - User attributes are now formatted better (a boolean becomes a checkbox, etc.) (#3103)
-- All handler, reviewer, decider emails can be disabled with `:enable-handler-emails false`. This does not include the invitation emails, which will always be sent. (#3116)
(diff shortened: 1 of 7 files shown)


aatkin (approved):
LGTM! One thing to notice is that email API (sending handler/reviewer reminders) is not suppressed by this change. Maybe this could be reflected in the changelog message, or be included as a change in API. Now changelog states that all handler emails can be disabled, which is technically not the case.

Macroz:
> LGTM! One thing to notice is that email API (sending handler/reviewer reminders) is not suppressed by this change. Maybe this could be reflected in the changelog message, or be included as a change in API. Now changelog states that all handler emails can be disabled, which is technically not the case.

I'll improve the text. Reminder mails are not sent unless configured, so they can be disabled with their own config. I think it could make sense to keep the reminders while having other emails disabled still so the current combination is perhaps good.

PR #3129, opened 2023-03-03
Refactor test data form fields

aatkin:
Extract test data form fields into separate `(def)`s, and make example field groups more explicit (e.g. conditional field with select).

# Checklist for author

## Follow-up

### DIFF
--- a/src/clj/rems/service/test_data.clj
+++ b/src/clj/rems/service/test_data.clj
@@ -187,32 +187,32 @@
                  :sv "IP-adress"}})
 
 (def conditional-field-example
-  [(-> option-field (merge {:field/id "option"
-                            :field/title {:en "Option list. Choose the first option to reveal a new field."
-                                          :fi "Valintalista. Valitse ensimmäinen vaihtoehto paljastaaksesi uuden kentän."
-                                          :sv "Lista. Välj det första alternativet för att visa ett nytt fält."}
-                            :field/optional true}))
-   (-> text-field (merge {:field/title {:en "Conditional field. Shown only if first option is selected above."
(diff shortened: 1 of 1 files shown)


src/clj/rems/service/test_data.clj
```diff
@@ -63,156 +63,172 @@
...
+           header-field
+           (-> date-field (merge {:field/optional true}))
+           (-> email-field (merge {:field/optional true}))
+           (-> attachment-field (merge {:field/optional true}))]
+          conditional-field-example ; array of fields
+          [(-> multiselect-field (merge {:field/optional true}))
+           (-> table-field (merge {:field/optional true}))]
+          max-length-field-example ; array of fields
+          [(-> phone-number-field (merge {:field/optional true}))
+           (-> ip-address-field (merge {:field/optional true}))]))
```
Macroz, line 231: We could also use `assoc` which would likely make this a tiny bit simpler.

Macroz: Also a plain `merge` without `->` would work. There's a bunch of other places too.

Macroz (approved):

src/clj/rems/service/test_data.clj
```diff
@@ -359,25 +369,26 @@
...
-    (test-helpers/command! {:type :application.command/invite-member
-                            :application-id appid2
-                            :actor applicant
-                            :member {:name "John Smith" :email "john.smith@example.org"}})
-    (test-helpers/command! {:type :application.command/submit
-                            :application-id appid2
-                            :actor applicant})
-    (doseq [member members]
-      (test-helpers/command! {:type :application.command/add-member
+;; XXX: can this be removed?
```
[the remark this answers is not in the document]
aatkin: Agree, #3135 includes a change to add members into test data.

PR #3140, opened 2023-03-28
Don't always send application data to notification target

Macroz:
Sometimes the application can be big, and the event notification target doesn't need it. With `:send-application? false` one can disable the sending.

On top of #3139.

# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -157,6 +157,7 @@
  ;;   :event-types (optional) - an array of event types to send. A missing value means "send everything".
  ;;   :timeout (optional) - timeout for the PUT in seconds. Defaults to 60s.
  ;;   :headers (optional) - a map of additional HTTP headers to send.
+ ;;   :send-application? (optional) - should application be sent in `:event/application` (defaults to true)
  ;;
  ;; See also: docs/event-notification.md
  ;;
--- a/resources/config-defaults.edn
+++ b/resources/config-defaults.edn
@@ -164,7 +164,11 @@
  ;; Example:
  ;;
  ;; :event-notification-targets [{:url "http://events/everything"}
- ;;                              {:url "http://events/filtered"
+ ;;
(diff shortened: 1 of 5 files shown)


resources/config-defaults.edn
```diff
@@ -157,6 +157,7 @@
  ;;   :event-types (optional) - an array of event types to send. A missing value means "send everything".
  ;;   :timeout (optional) - timeout for the PUT in seconds. Defaults to 60s.
  ;;   :headers (optional) - a map of additional HTTP headers to send.
+ ;;   :send-application? (optional) - should application be sent in `:event/application` (defaults to true)
```
aatkin, line 160: example below could also include the new key

src/clj/rems/event_notification.clj
```diff
@@ -19,6 +19,9 @@
             "to" (:url target))
   (try
     (let [timeout-ms (* 1000 (get target :timeout default-timeout))
+          body (if (:send-application? target true)
+                 body
+                 (dissoc body :event/application))
```
aatkin, line 24: should we do this check in `(notification-body)` already and not fetch the application in the first place? is there a reason we should fetch the application, copy it into outbox table and not use it?

Macroz: Originally I remembered that the logging used it so I delayed it later. 

But as for sending, the notification can go to multiple targets. It could be that one of the targets needs the application, so it would need to be changed to be lazily fetched so as not to fetch it multiple times. If we had a consistent cache then it could be gotten from there, but it could also return a different application for different targets.

Macroz: Ok so I fixed it with a `delay`. It cleaned everything up a bit, I think.

aatkin (approved):
LGTM! :+1: (also TIL `(delay)` function)

PR #3150, opened 2023-04-27
CI improvements

aatkin:
- browser tests now take advantage of parallelization in CircleCI (reducing total runtime from 18min to 9min)
- refactored CI jobs ordering and config syntax
- removed `re-frisk` dev dependency
- flaky test fixes (or attempts)

---

motivation: browser tests are currently the slowest running tests at well over 18min, which almost equals total runtime for any test run due to concurrency, so improving browser tests performance directly affects total runtime.

approach:  use [CircleCI parallelism and test splitting](https://circleci.com/docs/parallelism-faster-jobs/#introduction) to run browser tests in concurrently running identical containers
- create custom plugin for Kaocha which reads in all test ids from test plan and asks CircleCI to split test ids during CI execution
- each parallel container receives different test id set, which custom plugin uses to modify Kaocha test plan before test run
- Kaocha continues testing in parallel containers, and CircleCI puts together results from all containers.

notes:
- early measurements increasing `resource_class` to `medium+ / large` decreased runtime in browser tests (measurements only from actual testing, not setup) by 8s and 85s respectively. this could be different now with shadow build and npm dependencies moved to earlier CI job. with `large` CPU usage hovers around 30% for each container.
- some browser tests are far slower than others, for example:
  * `rems.test-browser/test-form-editor 270.40408 seconds`
  * `rems.test-browser/test-create-catalogue-item 156.55687 seconds`
  * `rems.test-browser/test-organizations 59.61294 seconds`
  * `rems.test-browser/test-conditional-field 45.53360 seconds`
  * `rems.test-browser/test-edit-catalogue-item 16.03691 seconds`

# Checklist for author

## Reviewability
- [ ] Link to issue

## Follow-up

Screenshot 2023-05-04 at 15 18 38
Screenshot 2023-05-04 at 15 16 08

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -87,10 +87,11 @@
 
   :clean-targets ["target"]
 
-  :aliases {"shadow-build" ["shell" "sh" "-c" "npm install && npx shadow-cljs compile app"]
-            "shadow-release" ["shell" "sh" "-c" "npm install && npx shadow-cljs release app"]
-            "shadow-test" ["shell" "sh" "-c" "npm install --include=dev && npx shadow-cljs compile cljs-test && ./node_modules/karma/bin/karma start"]
-            "shadow-watch" ["shell" "sh" "-c" "npm install --include=dev && npx shadow-cljs watch app"]
+  :aliases {"npm-deps" ["shell" "sh" "-c" "npm install --save=false --package-lock=true"] ; npm ci always removes node_modules by design, which is unnecessary and time consuming
+            "shadow-build" ["do" ["npm-deps"] ["shell" "sh" "-c" "npx shadow-cljs compile app --config-merge '{:build-options {:cache-level :jars}}'"]]
+            "shadow-release" ["do" ["npm-deps"] ["shell" "sh" "-c" "npx shadow-cljs release app"]]
+            "shadow-test" ["do" ["npm-deps"] ["shell" "sh" "-c" "npx shadow-cljs compile cljs-test && npx karma start"]]
+            "shadow-watch" ["do" ["npm-deps"] ["shell" "sh" "-c" "npx shadow-cljs watch app"]]
             "kaocha" ["with-profile" "test" "run" "-m" "kaocha.runner"]
             "browsertests" ["do" ["shadow-build"] ["kaocha" "browser"]]
             "alltests" ["do" ["shadow-build"] ["kaocha"] ["shadow-test"]]
--- a/test/clj/rems/browser_test_util.clj
+++ b/test/clj/rems/browser_test_util.clj
@@ -113,11 +113,13 @@
   (enable-downloads! (get-driver)))
 
 (defn refresh-driver!
-  "Refreshes an existing driver and cleans up."
+  "Refreshes an existing driver, cleans up and sets default values."
   []
   (assert (get-driver) "must have initialized driver already!")
   ;; start with a clean slate
-  (et/delete-cookies (get-driver)))
+  (et/delete-cookies (get-driver))
+  ;; big enough to show the whole page in the screenshots
+  (et/set-window-size (get-driver) 1400 7000))
 
 (defn fixture-init-driver
   "Executes a test running a fresh driver except when in development."
--- a/test/clj/rems/browser_test_util.clj
+++ b/test/clj/rems/browser_test_util.clj
@@ -1,4 +1,4 @@
-(ns ^:browser rems.browser-test-util
+(ns rems.browser-test-util
   "Browser test utils.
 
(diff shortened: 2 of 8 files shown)


aatkin:
for some reason parallel run shows 92 tests instead of 23. sum of 4 * 23 = 92 so could be that some part of machinery is confused.

aatkin:
> for some reason parallel run shows 92 tests instead of 23. sum of 4 * 23 = 92 so could be that some part of machinery is confused.

this should be fixed in latest version, probably junit-xml counted all tests even when they were marked skipped in test-plan.

project.clj
```diff
@@ -87,10 +87,11 @@
 
   :clean-targets ["target"]
 
-  :aliases {"shadow-build" ["shell" "sh" "-c" "npm install && npx shadow-cljs compile app"]
-            "shadow-release" ["shell" "sh" "-c" "npm install && npx shadow-cljs release app"]
-            "shadow-test" ["shell" "sh" "-c" "npm install --include=dev && npx shadow-cljs compile cljs-test && ./node_modules/karma/bin/karma start"]
-            "shadow-watch" ["shell" "sh" "-c" "npm install --include=dev && npx shadow-cljs watch app"]
+  :aliases {"npm-deps" ["shell" "sh" "-c" "npm install --save=false --package-lock=true"] ; npm ci always removes node_modules by design, which is unnecessary and time consuming
+            "shadow-build" ["do" ["npm-deps"] ["shell" "sh" "-c" "npx shadow-cljs compile app --config-merge '{:build-options {:cache-level :jars}}'"]]
```
aatkin, line 91: `{:build-options {:cache-level :jars}}` described more here: https://shadow-cljs.github.io/docs/UsersGuide.html#_compiler_cache

here so we can save only dependencies from shadow-cljs build cache

test/clj/rems/circle_ci.clj
```diff
@@ -0,0 +1,49 @@
...
+                       (remove :kaocha.testable/skip)
+                       (filter (comp #{:kaocha.type/var} :kaocha.testable/type))
+                       (map :kaocha.testable/id)))))
+
+(defn split-test-ids [test-ids]
+  (->> (sh/sh "circleci" "tests" "split" "--split-by=timings"
+              :in (str/join "\n" test-ids))
+       :out
+       (str/split-lines)
+       (into #{} (map #(keyword (subs % 1)))))) ; str/join turns :a/b into ":a/b", but keyword prepends ":", so remove first
```
Macroz: I don't understand this explanation here.

test/clj/rems/browser_test_util.clj
```diff
@@ -543,7 +543,7 @@
   Returns the test report.
 
   Ignores:
-  - all divs of body except #app and a re-frisk unnamed div
+  - all divs of body except #app
```
Macroz, line 546: Here only the comment is removed but the selector is not touched? Perhaps the `'body > div:not(#app)'` could be left out? Well maybe we do want to keep it since all our stuff is inside `#app`.

test/clj/rems/circle_ci.clj
```diff
@@ -0,0 +1,49 @@
+(ns rems.circle-ci
```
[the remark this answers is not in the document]
aatkin: changed name to `rems.kaocha-circleci-plugin` - i think that explains itself a little better now, and is more greppable as well.

test/clj/rems/circle_ci.clj
aatkin: this has been changed now, but original intent was to explain that the CLI command returns list of keywords as strings like `":rems.test-browser.test-new-application\n:rems.test-browser.test-handling"`, which needs some preprocessing:
```clojure
(keyword ":rems.test-browser.test-handling") ;=> ::rems.test-browser.test-handling
(keyword "rems.test-browser.test-handling") ;=> :rems.test-browser.test-handling
```

test/clj/rems/browser_test_util.clj
aatkin: let's keep the selector for now.

test/clj/rems/test_browser.clj
```diff
@@ -47,8 +47,13 @@
...
 ;;; common functionality
 
+(defn set-default-window-size
+  "Set browser window size big enough to show the whole page in the screenshots"
+  []
+  (btu/set-window-size 1400 7000))
+
 (defn login-as [username]
-  (btu/set-window-size 1400 7000) ; big enough to show the whole page in the screenshots
+  (set-default-window-size)
```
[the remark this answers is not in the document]
aatkin: yeah, that is a better solution, and less repetitive!

Issue #3131, opened 2023-03-13
Hide "invite members" from certain catalogue items


meericsc:
Add an option to workflows that allows you to disable invite members feature.

For some applications it doesn't make sense for applicants to be able to invite others to join their application.  

`:disable-commands` hides invite members for the whole REMS instance.

Needed before summer.

PR #3153, opened 2023-05-08
Disable commands by workflow

aatkin:
implements #3131

Workflow now has optional field `:disable-commands` which disables application commands dynamically. Workflow can disable commands by:
- application state (when state is any OR matches selected states), and
- user role (when role is any OR matches selected roles).

User role and application state rules are combined with AND, so both must match.

Example use cases include
- prevent applicants from closing a returned application
- prevent applicants from inviting members to application

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is completely new (`api/applications/commands`)

## Documentation
- [ ] Components are added to guide page (at present, guide page does not exist for workflow admin pages)
- [ ] Update manual/ (if applicable) (`owner.md` could use a description of the feature?)

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks #3163 

# Screenshots

## 1. admin create/edit workflow when no disable command rules
Screenshot 2023-05-30 at 18 17 19

## 2. admin create/edit workflow when disable command rules
Screenshot 2023-05-30 at 18 17 55
Screenshot 2023-05-30 at 18 18 45

## 3. admin workflow view when disable command rules
Screenshot 2023-05-30 at 18 18 56

### DIFF
--- a/src/clj/rems/application/model.clj
+++ b/src/clj/rems/application/model.clj
@@ -672,31 +672,27 @@
              form/enrich-form-field-visible
              application))
 
-(defn- get-roles-for-permission [disable-command]
-  (let [roles (:when/role disable-command)
-        all-roles (list nil)]
-    (if (seq roles)
-      roles
-      all-roles)))
+(defn- get-permission-roles [disable-command]
+  (if-some [roles (seq (:when/role disable-command))]
+    roles
+    (list nil))) ; nil means for every role
 
-(defn- get-workflow-disable-commands [application get-workflow]
-  (let [workflow (-> application
-                     (get-in [:application/workflow :workflow/id])
-                     (get-workflow))]
-    (for [disable-command (get-in workflow [:workflow :disable-commands])
-          :let [states (set (:when/state disable-command))]
-          :when (or (empty? states)
-                    (contains? states (:application/state application)))
-          role (get-roles-for-permission disable-command)]
-      {:permission (:command disable-command)
-       :role role})))
+(defn- is-disabled-for-application [application disable-command]
+  (let [states (:when/state disable-command)]
+    (or (empty? states)
+        (contains? (set states) (:application/state application)))))
 
 (defn enrich-disable-commands [application get-config get-workflow]
-  (permissions/blacklist application
-                         (permissions/compile-rules
-                          (concat (for [command (:disable-commands (get-config))]
-                                    {:permission command})
-                                  (get-workflow-disable-commands application get-workflow)))))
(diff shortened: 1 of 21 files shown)


src/clj/rems/application/model.clj
```diff
@@ -682,10 +672,18 @@
...
              application))
 
+(defn- get-disable-command [application command]
+  (cond
+    (map? command) (for [[cmd states] command
+                         :when (contains? (set states) (:application/state application))]
+                     cmd)
+    :else [command]))
+
 (defn- enrich-disable-commands [application get-config]
```
Macroz: There's something a bit unclear with how this was setup a long time ago. There exists `application-permissions-for-workflow-view` which probably doesn't have the full (injected) application state to do the decisions there so this `enrich-disable-commands` is added here. I guess not much can be done here for this change, but some refactoring opportunity for later.

src/clj/rems/application/model.clj
```diff
@@ -682,10 +672,18 @@
              form/enrich-form-field-visible
              application))
 
+(defn- get-disable-command [application command]
+  (cond
+    (map? command) (for [[cmd states] command
```
Macroz, line 677: We could normalize the `:disabled-commands` while loading to avoid this `map?`?

src/clj/rems/config.clj
```diff
@@ -42,16 +44,20 @@
   (-> config
       (update :email-retry-period #(Period/parse %))
       (update-existing :extra-pages (partial mapv #(update-existing % :roles set)))
-      (update :disable-commands (partial mapv keyword))))
+      (update-existing :disable-commands #(->> (transform [ALL string?] keyword %)
+                                               (transform [ALL map? MAP-KEYS] keyword)
+                                               (transform [ALL map? MAP-VALS ALL] keyword)))))
```
Macroz, line 49: We could walk the `:disabled-commands` with a simple `for` since it's just a seq and do any necessary coercion and normalization here (without having to use Specter). If we normalized every item here, then the other code may become simpler.

Or this level of coercion could use Schema instead. This could provide some detailed validation message.

src/clj/rems/application/model.clj
aatkin: Since full workflow body is not available here, we cannot use dynamic workflow for changing user permissions. `(application-view)` itself declares to be pure function, so it seems that design would need to be updated.

(Change which this comment originally refers to has been removed, but comment is still relevant)

Macroz (approved):

Issue #3130, opened 2023-03-13
"Anonymous" workflow handling


meericsc:
Hide the name of the organisation/person who submitted the dataset and processed the application from the applicant. Also from emails. 

~Instead we could show "DAC or REMS has approved the application" or something similar.~ Show the text "Handler" localized.

BP hopes to have this for a demo in 19th of June. 

- [x] workflow option "hide handler name"
  - show example
- [x] hides user details of `:handler`, `:reviewer`, `:decider` (and `:past-*`)
- [x] hide in API if role is applicant or member (replace attributes with blanks or so)
- [x] hide from emails to applicants or members
- [x] check PDF also

PR #3165, opened 2023-06-16
Anonymize handling workflow

aatkin:
relates #3130 

Add new option :anonymize-handling to workflow. When enabled, `handling users` are anonymized `for applying users` in:
- application events and attachments
- event emails.

Anonymized user is replaced with "rems-handler" and name/email is set to nil. Other attributes are omitted.

Events contain :event/actor (userid) and :event/actor-attributes, attachments contain :attachment/user. Certain event emails to applying users contain event actor that may be a handling user.

Anonymized userid can be localized in user interface and event emails with :t.roles/anonymized-handler, which uses :t.roles/handler. For example the english default translation is "Handler", and can be safely overridden in extra translations without overriding :t.roles/handler.

Other notable changes:
- rems.application.model/censor-user now returns only `[:userid :name :email]` instead of `[:userid :name :email :organizations :notification-email]`
- updated lambdaisland/kaocha to `1.86.1355` (originally thought something was broken with earlier version, but was actually a case of PEBKAC. however newer version did not break anything so updated anyway)
- workflow type field now has label in admin page.

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility
- [x] API is backwards compatible

## Documentation
- [ ] Update manual/ (if applicable) (owner manual perhaps?) #3197

## Testing

### Admin

Screenshot 2023-10-13 at 14 58 14

Screenshot 2023-09-19 at 23 57 07

### Application as handler
Screenshot 2023-09-20 at 0 07 58

### PDF as handler
Screenshot 2023-09-20 at 0 02 31

### Application as applicant
Screenshot 2023-09-20 at 0 07 29

### PDF as applicant
Screenshot 2023-09-20 at 0 02 10


Macroz (approved):
Based on our review together looks good 👍

PR #3181, opened 2023-08-29
feat: support named arguments in localization

aatkin:
relates #3183 

- add tempura support for map arguments with custom resource compiler `rems.tempura/resource-compiler`
- use `rems.tempura/tr` as "base" function for `rems.text` (`rems.tempura` is shared code namespace)
- removed `rems.context/*tempura*` with goal to share more code between clj/cljs

Localization strings may use either map or vector argument syntax, but not both. `rems.tempura/resource-compiler` can switch between map argument and vector arguments when both are provided (e.g. from `rems.text/text-format`):

```clojure
(text-format :k {:arg1 :val1 :arg 2 :val2} :val1 :val2)
;;=> works for {:k "%1 %2"}
;;=> and {:k "%:arg1% %:arg2%"}
```

Custom resource compiler works as so:
1. When translation fn is called, check provided arguments (vector).
2. If arguments contain map as first element, check if resource (localization string) contains vector indexes ("%1" etc).
2.1. If yes, drop first argument `(rest arguments)` and use resource as-is.
2.2. If not, walk through resource and replace any keyword arguments `"%:kw%"` with equivalent vector index argument `"%1"`. Use replaced resource string, and map argument map to ordered argument indexes (creating vector argument list): `(mapv {:kw "%1"} [:kw]) ;=> ["%1"]`
3. Call tempura default resource compiler with returned resource and argument vector.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation
- [ ] ADR for major architectural decisions or experiments (experiment for sure, but maybe not architectural decision?)

## Testing

## Follow-up
- [ ] New tasks are created for pending or remaining tasks (tasks maybe for updating email templates to use new translation keys?)


Macroz (approved):
I think we can try this out.

Macroz:
One thought for a follow-up task, we could try upgrading tempura since we haven't been able to, due to some backwards incompatibility.

Issue #3184, opened 2023-09-04
audit log: store relevant id and change


hallundbaek:
**Is your feature request related to a problem? Please describe.**
I would like to be able to precisely determine what entity was created/changed when querying the `audit_log` table, as well as have the ability to see exactly what was created/changed in that entity.

**Describe the solution you'd like**
I would like the `audit_log` table to have a reference to the table where the change was made, and the id of the relevant entity. I would also like to be able to store the request data, so I can precisely determine the state of an entity at a given time.

**Additional context**
While not really a viable path at this point, it seems like REMS and [datomic](https://www.datomic.com/) would've been a match made in heaven wrt. provenance and auditability.

Macroz:
Hi,

I'm not sure what kind of a use case are you thinking here. So first I would have to ask why would do you feel a need to track DB changes or a very detailed audit log?

Here are some thoughts:
- The audit log may be useful to store in the DB, or maybe not. It adds nothing compared to the application log, everything is available there too, and more. We don't have a specific use case in mind, so it has not been developed further. We have considered that maybe we will completely remove the audit log and simply fetch the application logs and store them for later. We also don't have any UI features to look a the audit log table yet.
- There is definitely room for adding some parameters of the request, or response to the audit log. But that will bloat the size of the audit log a lot. In some special REMS, where there is a handful of applications each year, but they are each very important, something like this could be considered. Also when there is a need to quickly debug what is going on, but that is just a technical issue for developers, and we have other means. Consider that the execution of each command could touch several database tables, the request and response can be tens of kilobyes without considering attachments. If we also log each DB query and response, that would add even more bloat. While it is technically possible, does it make sense?
- The state of the application is what it is, and it is event-sourced, so the events completely define what the state is. Not all of REMS uses event sourcing, but at least this most important part does.
- Many asynchronous processes change DB state, for example eventually sending email, bots doing automated decisions. They can't be easily tracked to some request, or the tracking doesn't make sense.
- Having a temporal database does not completely cover all the features of event sourcing because it tracks very low level information. It is necessary to also store the commands that change state, in a natural business process level (i.e., submit application, approve application).
[...]

PR #3187, opened 2023-09-15
Extended logging

hallundbaek:
Implements #3184

Though I wish I'd found a more elegant solution, this does the job quite handily, while ensuring backwards compatibility. But a more experience clojure developer probably has a solution that circumvents the need for adding this to every relevant route.

I tried a general approach at the logging middleware, but I couldn't read request data, as it is a read-once bytestream on the HttpRequest. And since the result is not in the response, but loaded in a subsequent request, it cannot be pulled from that either.

I started looking into the Compojure wrappers, but my cursory look didn't yield a more elegant solution.

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

### DIFF
--- a/dev-config.edn
+++ b/dev-config.edn
@@ -92,7 +92,7 @@
  :enable-catalogue-tree true
  :enable-save-compaction true
  :enable-autosave true
- :extended-logging true
+ :enable-extended-logging true
 
  ;; let's use plugins in dev
  :oidc-require-name false
--- a/dev-config.edn
+++ b/dev-config.edn
@@ -92,6 +92,17 @@
  :enable-catalogue-tree true
  :enable-save-compaction true
  :enable-autosave true
+ :extended-logging-routes ["/api/catalogue-items/create"
+                           "/api/catalogue-items/edit"
+                           "/api/categories/create"
+                           "/api/categories/edit"
+                           "/api/forms/create"
+                           "/api/forms/edit"
+                           "/api/users/create"
+                           "/api/users/edit"
+                           "/api/user-settings/edit"
+                           "/api/workflows/create"
+                           "/api/workflows/edit"]
 
  ;; let's use plugins in dev
  :oidc-require-name false
--- a/dev-config.edn
+++ b/dev-config.edn
@@ -92,17 +92,7 @@
  :enable-catalogue-tree true
  :enable-save-compaction true
  :enable-autosave true
- :extended-logging-routes ["/api/catalogue-items/create"
-                           "/api/catalogue-items/edit"
-                           "/api/categories/create"
-                           "/api/categories/edit"
-                           "/api/forms/create"
-                           "/api/forms/edit"
-                           "/api/users/create"
-                           "/api/users/edit"
-                           "/api/user-settings/edit"
-                           "/api/workflows/create"
-                           "/api/workflows/edit"]
+ :extended-logging true
 
  ;; let's use plugins in dev
  :oidc-require-name false
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -183,7 +183,7 @@ REMS uses [Logback](https://logback.qos.ch/) for logging. By default everything 
 
 ### Extended Logging
 
-The config option `:extended-logging` can be enabled to additionally log the content of entity mutations.
+The config option `:enable-extended-logging` can be toggled to additionally log the content of entity mutations.
 
 These logs enable auditors to unequivocally determine the state of a given entity at a given time. Which may aid with various regulatory requirements.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -183,29 +183,14 @@ REMS uses [Logback](https://logback.qos.ch/) for logging. By default everything 
 
 ### Extended Logging
 
-The config option `:extended-logging` is for routes that should additionally log the content of entity mutations.
-
-Logging is at `INFO` level and the log messages are prefixed with `extended-logging (:uri request)` e.g. `extended-logging /api/forms/edit`.
+The config option `:extended-logging` can be enabled to additionally log the content of entity mutations.
 
 These logs enable auditors to unequivocally determine the state of a given entity at a given time. Which may aid with various regulatory requirements.
 
-Logging occurs after request data has been coerced to a schema, but before the mutation is persisted.
-as such, the logged information is only authoritative about the state of the respective entity iff the request was successful.
+Logging is at `INFO` level and the log messages are prefixed with `extended-logging (:uri request)` e.g. `extended-logging /api/forms/edit`.
 
-Only routes that mutate an existing entity benefit from extended logging, the possible routes are as follows:
-```
- ["/api/catalogue-items/create"
-  "/api/catalogue-items/edit"
-  "/api/categories/create"
-  "/api/categories/edit"
-  "/api/forms/create"
-  "/api/forms/edit"
-  "/api/users/create"
-  "/api/users/edit"
-  "/api/user-settings/edit"
-  "/api/workflows/create"
-  "/api/workflows/edit"]
-```
+Logging occurs before the mutation is persisted. As such, the logged information is only authoritative about the state of the respective entity iff the request was successful.
+
 ## Application expiration scheduler
 
 REMS can be configured to delete applications after a set period of time has passed since last activity. Expiration can be defined for application states with ISO-8601 duration formatting, and optionally email notification can be configured to be sent to applicant and members before deletion. Application expiration scheduler is disabled by default. See `:application-expiration` in [config-defaults.edn](https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn) for details.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -181,6 +181,31 @@ REMS uses [Logback](https://logback.qos.ch/) for logging. By default everything 
 
     java -Dlogback.configurationFile=logback-prod.xml -jar rems.jar run
 
+### Extended Logging
+
+The config option `:extended-logging` is for routes that should additionally log the content of entity mutations.
+
+Logging is at `INFO` level and the log messages are prefixed with `extended-logging (:uri request)` e.g. `extended-logging /api/forms/edit`.
+
+These logs enable auditors to unequivocally determine the state of a given entity at a given time. Which may aid with various regulatory requirements.
+
+Logging occurs after request data has been coerced to a schema, but before the mutation is persisted.
+as such, the logged information is only authoritative about the state of the respective entity iff the request was successful.
+
+Only routes that mutate an existing entity benefit from extended logging, the possible routes are as follows:
+```
+ ["/api/catalogue-items/create"
+  "/api/catalogue-items/edit"
+  "/api/categories/create"
+  "/api/categories/edit"
+  "/api/forms/create"
+  "/api/forms/edit"
+  "/api/users/create"
+  "/api/users/edit"
+  "/api/user-settings/edit"
+  "/api/workflows/create"
+  "/api/workflows/edit"]
+```
 ## Application expiration scheduler
 
 REMS can be configured to delete applications after a set period of time has passed since last activity. Expiration can be defined for application states with ISO-8601 duration formatting, and optionally email notification can be configured to be sent to applicant and members before deletion. Application expiration scheduler is disabled by default. See `:application-expiration` in [config-defaults.edn](https://github.com/CSCfi/rems/blob/master/resources/config-defaults.edn) for details.
--- a/src/clj/rems/api/catalogue_items.clj
+++ b/src/clj/rems/api/catalogue_items.clj
@@ -107,7 +107,7 @@
       :roles +admin-write-roles+
       :body [command CreateCatalogueItemCommand]
       :return CreateCatalogueItemResponse
-      (extended-logging request command)
+      (extended-logging request)
       (ok (catalogue/create-catalogue-item! command)))
 
     (PUT "/edit" request
@@ -115,7 +115,7 @@
       :roles +admin-write-roles+
       :body [command EditCatalogueItemCommand]
       :return schema/SuccessResponse
-      (extended-logging request command)
+      (extended-logging request)
       (if (nil? (catalogue/get-localized-catalogue-item (:id command)))
         (not-found-json-response)
         (ok (catalogue/edit-catalogue-item! command))))
--- a/src/clj/rems/api/catalogue_items.clj
+++ b/src/clj/rems/api/catalogue_items.clj
@@ -3,7 +3,7 @@
             [compojure.api.sweet :refer :all]
             [rems.api.schema :as schema]
             [rems.service.catalogue :as catalogue]
-            [rems.api.util :refer [not-found-json-response check-user]] ; required for route :roles
+            [rems.api.util :refer [not-found-json-response check-user extended-logging]] ; required for route :roles
             [rems.common.roles :refer [+admin-write-roles+]]
             [rems.common.util :refer [apply-filters]]
             [rems.schema-base :as schema-base]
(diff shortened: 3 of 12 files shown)


hallundbaek:
Oh, and I was not able to run `lein alltests`, as there was something getting jacked up with the postgresql connection, that I didn't have the time to fix.

```0/23Execution error (PSQLException) at org.postgresql.core.v3.ConnectionFactoryImpl/lambda$doAuthentication$4 (ConnectionFactoryImpl.java:804).
The server requested SCRAM-based authentication, but no password was provided.
```

Macroz:
> Oh, and I was not able to run `lein alltests`, as there was something getting jacked up with the postgresql connection, that I didn't have the time to fix.
> 
> ```
> 0/23Execution error (PSQLException) at org.postgresql.core.v3.ConnectionFactoryImpl/lambda$doAuthentication$4 (ConnectionFactoryImpl.java:804).
> The server requested SCRAM-based authentication, but no password was provided.
> ```

If `lein kaocha integration` works, then basically DB tests run. Otherwise it may be a problem with your PG setup. Perhaps there is a change in a new version or so.

Tests pass in the CI environment so there isn't an actual problem.

src/clj/rems/api/util.clj
```diff
@@ -49,3 +51,7 @@
   (let [body (json/generate-string {:error (or message "unprocessable entity")})]
     (-> (http-response/unprocessable-entity body)
         (http-response/content-type "application/json"))))
+
+(defn extended-logging [request mutation]
+  (-> (when (some (partial = (:uri request)) (:extended-logging-routes env))
```
Macroz, line 56: There may be some cases where this fails. What happens when there query parameters?

If exact string matching is required, it would be useful to use a set. But I think a more realistic version would either
- support regex here
- support matching the request method too or
- automatically log all modifications (`POST`, `PUT`, `DELETE`) in the whole of the API, with one simple config.

docs/configuration.md
```diff
@@ -181,6 +181,31 @@ REMS uses [Logback](https://logback.qos.ch/) for logging. By default everything
 
     java -Dlogback.configurationFile=logback-prod.xml -jar rems.jar run
 
+### Extended Logging
+
+The config option `:extended-logging` is for routes that should additionally log the content of entity mutations.
+
+Logging is at `INFO` level and the log messages are prefixed with `extended-logging (:uri request)` e.g. `extended-logging /api/forms/edit`.
```
Macroz, line 188: Should the logging happen to the audit_log in DB, to the application log, or a separate audit log file (potentially replacing audit log in DB)?

hallundbaek: wrt our previous conversation on issue #3184, I got the impression that the audit log and its functionality wasn't a widely used feature. 
Furthermore, for our needs, it is a requirement that services relay logs to a separate logging service. In which case, using application logs simplifies integration greatly, as such it would be ideal if they could provide an authoritative source for audits.
I would imagine that other organisations will have similar requirements. But I have no basis for that statement other than a hunch.

src/clj/rems/api/util.clj
hallundbaek: I agree.

I reckon the need for logging all modifications probably would be a more common scenario than only logging a specific set. And the less common scenario can easily be accommodated by filtering on an application log level.

src/clj/rems/api/catalogue_items.clj
```diff
@@ -102,18 +102,20 @@
         (ok it)
         (not-found-json-response)))
 
-    (POST "/create" []
+    (POST "/create" request
       :summary "Create a new catalogue item"
       :roles +admin-write-roles+
       :body [command CreateCatalogueItemCommand]
       :return CreateCatalogueItemResponse
+      (extended-logging request command)
```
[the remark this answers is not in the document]
hallundbaek: Grand! I'm happy that my prediction that, a more experienced Clojure developer could find a neater solution, was fulfilled!

Side note: I assumed that `:params` related to url query params, and glossed over it. Maybe it would improve readability to avoid a name with an overlap with a closely related domain, maybe `:vars`?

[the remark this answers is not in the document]
hallundbaek: Unfortunately, I do not get any content with a more general strategy

```
2023-09-20 13:05:00,669 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.middleware - req > :put /api/forms/edit
2023-09-20 13:05:00,669 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.middleware - wrap-entry-logging {}
2023-09-20 13:05:00,682 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.middleware - > :put /api/forms/edit lang: :en user: {:userid owner, :name Owner, :email owner@example.com}  roles: #{:logged-in :owner}
2023-09-20 13:05:00,682 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.middleware - wrap-context-logging {}
2023-09-20 13:05:00,690 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.api.util - extended-logging /api/forms/edit {:organization {:organization/id hus}, :form/internal-name dsdsssee, :form/external-title {:en sd, :fi sdsd, :sv sdsd}, :form/fields [], :form/id 14}
2023-09-20 13:05:00,700 [jetty-183 rqc:1 owner oQdVKBYp] INFO  rems.middleware - req < :put /api/forms/edit 200 30ms
```

Where the first `wrap-context-logging`/`wrap-entry-logging` correspond to the following two additional lines
```diff clojure
(defn wrap-context-logging
  "Logs detailed information about a request just before
  the main business code is run.

  Here we have the most context about the user and request available."
  [handler]
  (fn [request]
    (let [uri (str (:uri request)
                   (when-let [q (:query-string request)]
                     (str "?" q)))
          log? (not (re-matches silence-logging-regex uri))]
      (when log?
        (log/info ">" (:request-method request) uri
                  "lang:" context/*lang*
                  "user:" context/*user*
                  (if (:uses-valid-api-key? request)
                    "api-key"
                    "")
                  "roles:" context/*roles*)

        (log/debug "session" (pr-str (:session request)))
+       (log/info "wrap-context-logging" (pr-str (:params request)))
        (when (seq (:form-params request))
          (log/debug "form params" (pr-str (:form-params request)))))

      (handler request))))

(defn wrap-entry-logging
  "Logs every request into an entry `req >` and exit `req <` log.

  This happens in the very beginning of the request processing, so
  not a lot of context is available yet.

  This is important to show a request as soon as it's possible, so if it
  starts to wait on middleware DB requests or such, it still appears
  in the logs."
  [handler]
  (fn [request]
    (let [uri (str (:uri request)
                   (when-let [q (:query-string request)]
                     (str "?" q)))
          start (System/nanoTime)
          log? (not (re-matches silence-logging-regex uri))
          userid (or (get-in request [:session :identity :userid])
                     (:userid context/*user*)
                     (auth/get-api-userid request))]
      (try
        (swap! request-count inc)

        (with-mdc {:userid userid
                   :request-count @request-count}
          (when log?
            (log/info "req >" (:request-method request) uri)
+           (log/info "wrap-entry-logging" (pr-str (:params request)))
            (when (seq (:form-params request))
              (log/debug "form params" (pr-str (:form-params request)))))

          (let [response (handler request)]
            (when log?
              (log/info "req <" (:request-method request) uri (:status response)
                        (str (int (/ (- (System/nanoTime) start) 1000000)) "ms")
                        (or (get-in response [:headers "Location"]) "")))
            response))
        (finally
          (swap! request-count dec))))))
```

Do you have a suggestion for a better place for this?

src/clj/rems/api/util.clj
hallundbaek: resolved in [317d982](https://github.com/CSCfi/rems/pull/3187/commits/317d982bca638aefb178e89193cfc4ff414227e5)

src/clj/rems/api/catalogue_items.clj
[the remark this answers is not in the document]
Macroz: I can have a look at the approach and confirm that it should work. The thing with the middleware wrappers is that they must be placed in the correct order, so at least if we want to use the `:params`, then we must place it after the standard params wrapper has done its job. Maybe there is something else too.

docs/configuration.md
Macroz: Audit log as a feature is used, i.e. the logs go to the database, in the REMS production instances. It's just that the use is such that not many features have been built on top of it (like browsing in the UI). Also nobody else has requested to log the request / response, so it has not been implemented.

It is just a forward-looking idea to implement an option to have audit log as a separate file, instead of the DB. To me it sounds like you would want the whole audit log as a clean file, and not bother with the DB table at all. And possibly it doesn't make sense to log everything always to the regular log file. There is going to be a lot to optimize in the logs if a compact log file is to be stored for a long time.

We would normally work out a rough design for the idea, that would include potential features to be implemented first and some left as optional. Not all REMS production instances are interested in this logging, so a simple set of features to choose from would be nice.

We can usually accept a partial feature easier if it's useful, a part of the whole set and by default is not configured to be used everywhere. Not every (improvement) idea that is discovered in a review typically is implemented. It could be a follow-up task or an idea for later. Just something to think about (the whole).

dev-config.edn
```diff
@@ -92,17 +92,7 @@
...
-                           "/api/categories/create"
-                           "/api/categories/edit"
-                           "/api/forms/create"
-                           "/api/forms/edit"
-                           "/api/users/create"
-                           "/api/users/edit"
-                           "/api/user-settings/edit"
-                           "/api/workflows/create"
-                           "/api/workflows/edit"]
+ :extended-logging true
```
Macroz, line 95: For consistency we could name this `:enable-extended-logging` or consider alternatives like `:audit-log-request-and-response` or `:enable-audit-logging-of-request-and-response`.

src/clj/rems/api/catalogue_items.clj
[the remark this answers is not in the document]
Macroz: Another problem that may happen is that one needs to restart the server for changes in middleware to take effect.

But I guess the main problem is that the request body contains most parameters and in a typical request (when they are not actually as HTTP form data) it's not available through params. So the solution should log both params and the body.

Then, about reading the body twice. In another project I implemented my own params reader middleware (just patching the code of the default one), because that was the only place that consumed the body from the stream, or complained about it being read twice. It's something to think about here. It depends on which middleware here actually wants to read the body, and how easily it can be changed.

Likely it's a little bit more involved to read the body separately to log it, but I'm sure it can be made to work.

docs/configuration.md
hallundbaek: That makes sense.

I reckon it might be useful in the future, if users want the audit log to be able to provide an authoritative source on the state of an entity at a given time.

Though you could also argue that the audit log is for high-level auditing, but if you want to investigate a specific incident, then you would also need to go over the application logs. At that point, it would probably bloat the audit log beyond its initial use-case.

For our specific use case, with this feature implemented, we are able to comply with all our logging requirements (and effectively our auditability requirements), by monitoring the application log and routing relevant log lines to a separate logging service. So, the audit log will probably not see a lot of use, whether it is a db or a clean file.

dev-config.edn
hallundbaek: Definitely should be prefixed with `enable`.

But I think the other two alternatives are not descriptive of what it is meant to do. It is not necessary to log at this level for all requests, as most are persisted in other ways that result in immutable entities, as such, there is always an audit trail that points to a specific user who committed a specific change.

The entities can be divided into:
1. Immutable entities that cannot be changed, but support copy and modify.
2. Mutable entities that can be changed.

Extended logging is only relevant for the second group.

src/clj/rems/api/catalogue_items.clj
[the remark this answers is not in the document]
hallundbaek: > I can have a look at the approach and confirm that it should work. The thing with the middleware wrappers is that they must be placed in the correct order, so at least if we want to use the `:params`, then we must place it after the standard params wrapper has done its job. Maybe there is something else too.

Thank you, that would be a great help. This is my first time writing Clojure, so my understanding is still quite limited.

dev-config.edn
hallundbaek: naming addressed in [33a93e8](https://github.com/CSCfi/rems/pull/3187/commits/33a93e8622f6b11d1c6556a90d6eb5ecec357e82)

src/clj/rems/api/catalogue_items.clj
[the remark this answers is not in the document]
hallundbaek: Though I am wondering, as described in [this comment](https://github.com/CSCfi/rems/pull/3187#discussion_r1333003179), whether it is desirable to do extended logging for all requests, which I don't think it is.

And in that case, there is going to be some code that is specific to those routes somewhere anyways.

The other option is to write generalized code that catches the mutable entity cases, but I think that is going to add unnecessary complexity, as opposed to being careful when new entities and create/edit routes are added, whether or not to add the extended logging functionality as well.

hallundbaek:
I've no idea why the test failed, it doesn't seem related to the change in any way?

larsnaesbye:
You're right, it doesn't seem to be related to this change. Could the tests be rerun to see if this was due to an unrelated change from elsewhere?

Macroz:
> I've no idea why the test failed, it doesn't seem related to the change in any way?

Hi,

Unfortunately there exist a couple tests that are flaky (mostly some browser tests). This intergration test is such that sometimes the order of execution for the tests is such that it fails. I will rerun the failed tests and we'll see then.

Macroz:
> > I've no idea why the test failed, it doesn't seem related to the change in any way?
> 
> Hi,
> 
> Unfortunately there exist a couple tests that are flaky (mostly some browser tests). This intergration test is such that sometimes the order of execution for the tests is such that it fails. I will rerun the failed tests and we'll see then.

There was also some temporary CI environment hiccups.

Now the code passes the tests just fine.

Is there something specific that I could help you with? Maybe you could summarize what you think needs to be done and what you feel comfortable with doing and what you'd like me to handle instead?

hallundbaek:
Thanks for reaching out @Macroz, as it stands, this MR is fine for our needs.

Though there might be a more elegant solution to be found in the middleware, which was discussed in the thread of this comment: https://github.com/CSCfi/rems/pull/3187#discussion_r1334041227 so if you want to give that a stab, it would make it a bit neater.

But, a solution there should easily be backwards compatible with the current implementation, so that could be a future issue.

Macroz:
> Thanks for reaching out Macroz, as it stands, this PR is fine for our needs.

Alright, that's good. I can adopt this PR in the near future and finish it with what I consider to be useful additions to make it more generally applicable.

hallundbaek:
> > Thanks for reaching out Macroz, as it stands, this PR is fine for our needs.
> 
> Alright, that's good. I can adopt this PR in the near future and finish it with what I consider to be useful additions to make it more generally applicable.

Should I read this as:

1. This PR will stay open until you get to it, or
2. You will adopt it as is and make useful additions in a separate PR.

In any case it would be helpful if you either resolve or elaborate the open reviews in this PR, as the ball is in the reviewers court in all of them.

src/clj/rems/api/catalogue_items.clj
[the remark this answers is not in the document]
Macroz: I will change this later so that every endpoint is caught automatically. For example I don't see applications (attachments), blacklist or email here, which I would like to be included for completeness. It's easy to forget to add this feature in the future too.

Macroz:
> > > Thanks for reaching out Macroz, as it stands, this PR is fine for our needs.
> > 
> > 
> > Alright, that's good. I can adopt this PR in the near future and finish it with what I consider to be useful additions to make it more generally applicable.
> 
> Should I read this as:
> 
> 1. This PR will stay open until you get to it, or
> 2. You will adopt it as is and make useful additions in a separate PR.
> 
> In any case it would be helpful if you either resolve or elaborate the open reviews in this PR, as the ball is in the reviewers court in all of them.

We had a discussion and I think it's OK to merge this now and do the additions in another PR. We'll try to get it to the next release.

Macroz (approved):
Looks good to include. I will continue with a couple improvements and a test next.

PR #3188, opened 2023-09-20
Two corrections to Danish i18n

larsnaesbye:
Simple corrections to Danish i18n, less formal and more modern.


Macroz (approved):
Thanks, we'll have some new translations soon, at least for the next release, and I think we will set them as `nil` in the files so they should be easy to find.

larsnaesbye:
Super. I will remember to look for `nil` in the next revision.

Macroz:
> @Macroz I notice there are already some new ones for voting - do you think they will be the only ones before next release?

There will be some more, related to table paging and other new features still coming.

PR #3189, opened 2023-09-28
feat: allow handler to redact handler attachments

aatkin:
relates #3190

Allow handler to redact also other handler attachments. Refactored `model/allowed-redact-roles` and `model/apply-attachment-permissions`, and amended api test with test case where handler redacts another handlers attachment.

Other changes
- handling users may now see which attachments were redacted in event/redacted-attachments

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up

# Screenshots

## Handler

Screenshot 2023-10-05 at 19 15 31

Screenshot 2023-10-05 at 19 19 23

## Applicant

Screenshot 2023-10-05 at 19 16 18

Screenshot 2023-10-05 at 19 18 53


src/cljc/rems/common/application_util.cljc
```diff
@@ -69,27 +70,37 @@
...
     (cond
-      (nil? (get-in attachment [:attachment/event :event/id]))
+      (or (:attachment/redacted attachment)
+          (not event-id))
       false
 
-      (some roles (:attachment/redact-roles attachment))
-      true
+      (not= userid (get-in attachment [:attachment/user :userid]))
+      (some? (not-empty (clojure.set/intersection (set roles)
```
Macroz, line 81: As it seems explicit here that we aim to return a boolean always, we could name this function like normal predicates `can-redact-attachment?`. Otherwise we could just use `seq` here.

Macroz: I think there's also some other possibilities for golfing for simplicity in the whole `cond`.

Macroz: An idea
```clj
(defn can-redact-attachment? [attachment roles userid]
  (let [event-id (get-in attachment [:attachment/event :event/id])
        allowed-roles (:attachment/redact-roles attachment)]
    (cond
      (:attachment/redacted attachment) false

      (not event-id) false

      (= userid (get-in attachment [:attachment/user :userid]))
      true

      (empty? (clojure.set/intersection (set roles)
                                        (set allowed-roles)))
      false

      :else true)))```

**Review by Macroz — APPROVED** [review]

File: src/cljc/rems/common/application_util.cljc (hunk 1 above)
Comment (aatkin) [reply src/cljc/rems/common/application_util.cljc:81]: yeah, that could work!

i like `cond` because it makes multiple cases more readable, but sometimes i'm not sure which way reads better, explicit return values for each case or function call returning value. like here, `(not (empty? ...))` is discouraged by clj-kondo (warning) so i tend to not use it, but it is more readable than `(some? (not-empty ...))` when only boolean is required.

Comment (Macroz) [reply src/cljc/rems/common/application_util.cljc]: Yeah, I like `cond` too, just have to organize it so that the clauses are simple and make sense.

File: src/cljc/rems/common/application_util.cljc [hunk 2]
```diff
@@ -70,19 +70,22 @@
...
 (defn can-redact-attachment [attachment roles userid]
-  (let [event-id (get-in attachment [:attachment/event :event/id])
+  (let [already-redacted (:attachment/redacted attachment)
+        event-id (get-in attachment [:attachment/event :event/id])
         allowed-roles (:attachment/redact-roles attachment)]
     (cond
-      (or (:attachment/redacted attachment)
-          (not event-id))
-      false
+      already-redacted false
```
Macroz, line 77: This is nice naming 👍

Issue #3204, opened 2023-10-20
Restrict action spamming


aatkin:
Currently users can unrestrictedly click actions, which perform potentially long-running requests. REMS tries to process these requests concurrently, which (at this time) may lead to process hang. Spam-clicking an action is likely a symptom of unresponsive UI.

Add "double-click listener" which prevents action from being re-submitted before request has returned:
- [x] `rems.application/actions-form`
- [x] `rems.administration.blacklist` (add/remove)

Other implementation concerns:
- [ ] Indicate action processing to user #3210

PR #3205, opened 2023-10-23
Limit repeated action clicking

aatkin:
implements #3204

rems.atoms/rate-limited-button wraps click handler and allows only 1 click every 2 seconds. `perform-action-button` wraps this and additionally subscribes to `rems.spa/any-pending-request?` to wait until any put/post requests are done. This should prevent repeated clicking which creates duplicate requests to API

`rems.actions.*` [accept-licenses add-licenses add-member approve-reject assign-external-id change-applicant change-resources close decide delete invite-decider-reviewer invite-member redact-attachments remove-member request-decision request-review return-action review revoke vote]
`rems.administration.blacklist`
`rems.application.*` [save-draft submit copy-as-new]

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Follow-up

### DIFF
--- a/src/cljs/rems/application.cljs
+++ b/src/cljs/rems/application.cljs
@@ -2,11 +2,11 @@
   (:require [clojure.string :as str]
             [clojure.set :refer [union]]
             [goog.string]
+            [reagent.core :as r]
             [re-frame.core :as rf]
             [medley.core :refer [find-first update-existing]]
             [cljs-time.core :as time-core]
             [rems.actions.accept-licenses :refer [accept-licenses-action-button]]
-            [rems.actions.components :refer [button-wrapper]]
             [rems.actions.add-licenses :refer [add-licenses-action-button add-licenses-form]]
             [rems.actions.add-member :refer [add-member-action-button add-member-form]]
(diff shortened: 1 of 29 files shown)


src/cljs/rems/application.cljs
```diff
@@ -509,20 +506,26 @@
...
                               :on-click #(rf/dispatch [::save-application [text :t.form/save]])}])
 
 (defn- submit-button []
   [atoms/rate-limited-button {:id "submit"
                               :text (text :t.form/submit)
                               :class :btn-primary
-                              :disabled @(rf/subscribe [:rems.spa/pending-request ::submit])
+                              :disabled @(rf/subscribe [:rems.spa/any-pending-request #{"/api/applications/save-draft"
+                                                                                        "/api/applications/submit"
+                                                                                        "/api/applications/copy-as-new"}])
```
Macroz, line 520: This is OK but perhaps a bit fragile when we add new commands. Would it actually be simpler if we would always just wait for one thing? So that every action blocks the others? Maybe for application? 🤔

Macroz (approved):
I think this is good enough to merge (and perhaps click around in dev).

Macroz (approved):

PR #3219, opened 2023-11-08
fix: do not require form in catalogue item dependencies

### LINKED ISSUE #3217: Enable unarchiving catalogue items without form
If you archive a catalogue item that doesn't have a form and you cannot unarchive it again. Unarchiving catalogue items without a form should also be possible. 

To test:
1. Create catalogue item without adding a form directly to it
2. Archive it
3. Try to unarchive it

aatkin:
fixes #3217 

Form can be nil because it is optional, but it should not fail unarchive operation on catalogue item.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

### DIFF
--- a/src/clj/rems/service/dependencies.clj
+++ b/src/clj/rems/service/dependencies.clj
@@ -37,16 +37,17 @@
               [(:organization res)])]
      {:from {:resource/id (:id res)} :to dep})
 
-   (flatten
-    (for [cat (catalogue/get-localized-catalogue-items {:archived true :expand-catalogue-data? true})
-          dep [{:form/id (:formid cat)}
-               {:resource/id (:resource-id cat)}
-               {:workflow/id (:wfid cat)}
-               {:organization/id (:organization cat)}]]
-      (into [{:from {:catalogue-item/id (:id cat)} :to dep}]
-            (mapv (fn [category]
-                    {:from {:catalogue-item/id (:id cat)}
-                     :to (select-keys category [:category/id])}) (:categories cat)))))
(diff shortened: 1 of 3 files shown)


src/clj/rems/service/dependencies.clj
```diff
@@ -37,16 +37,17 @@
...
+           (for [cat (catalogue/get-localized-catalogue-items {:archived true :expand-catalogue-data? true})
+                 :let [deps (filter some? [(when (:formid cat) ; form is not required
+                                             {:from {:catalogue-item/id (:id cat)} :to {:form/id (:formid cat)}})
+                                           {:from {:catalogue-item/id (:id cat)} :to {:resource/id (:resource-id cat)}}
+                                           {:from {:catalogue-item/id (:id cat)} :to {:workflow/id (:wfid cat)}}
+                                           {:from {:catalogue-item/id (:id cat)} :to {:organization/id (:organization cat)}}])]]
+             (into deps
+                   (map (fn [category]
+                          {:from {:catalogue-item/id (:id cat)}
+                           :to (select-keys category [:category/id])}) (:categories cat)))))
```
Macroz, line 50: Something like this could have worked too (uses more features of `for`). We should have a look at the whole namespace later if it can be simplified and cached better in the future.
```clj
(for [cat ...
      dep (concat [{:form/id ...}
                   ...]
                  (map (fn [category] ...)))
      :when (some? (val (first dep)))]
  ...)
```

Macroz (approved):

PR #3221, opened 2023-11-08
Wider action link click area in dropdown

### LINKED ISSUE #3167: Dropdown buttons only work when clicking the link text
On Administration tab, the links under Modify-button only react if you click on the link text. If you click outside the text, the dropdown closes.

image

It would be easier for the user if also the outer edge of the button would be clickable.

aatkin:
fixes #3167 

Bootstrap examples use dropdown-item class directly in anchor link, and that is now implemented in `atoms/action-link`.

`atoms/action-link` seems to lose underlining in this process, but it is only used in the dropdown (other action-links found by grep reference `actions/components`), so maybe it is not needed?

https://github.com/CSCfi/rems/assets/794087/edc511ae-3122-4cd7-b9b1-ef2f47b3b1c2

# Checklist for author

## Reviewability
- [x] Link to issue
- [x] Consider adding screenshots for ease of review (quicktime recording)

## Documentation

### DIFF
--- a/src/cljs/rems/atoms.cljs
+++ b/src/cljs/rems/atoms.cljs
@@ -324,7 +324,7 @@
          :label (:label action)
          :href (:url action)
          :on-click (:on-click action)
-         :class (str "dropdown-item btn btn-link py-2 " (:class action))}])
+         :class (str/trim (str "btn btn-link " (:class action)))}])
 
 (defn edit-action
   "Standard edit action helper."
@@ -365,7 +365,7 @@
         label]
        (into [:div.dropdown-menu.dropdown-menu-right]
              (for [action actions]
-               [action-link action]))])))
+               [action-link (update action :class #(str "dropdown-item py-2 " %))]))])))
 
 (defn guide []
   (let [state (r/atom false)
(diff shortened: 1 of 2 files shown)


src/cljs/rems/atoms.cljs
```diff
@@ -324,7 +324,7 @@
          :label (:label action)
          :href (:url action)
          :on-click (:on-click action)
-         :class (str "btn btn-link " (:class action))}])
+         :class (str "dropdown-item btn btn-link py-2 " (:class action))}])
```
Macroz, line 327: Maybe it would be better to have the `dropdown-item` class passed from the use of this component so that `action-link` itself could be used outside of dropdowns? So could be passed in the `:class`.

Macroz (approved):

src/cljs/rems/atoms.cljs
aatkin: Yeah, that is better for re-usability. I was initially unsure where this atom was used, and since `:class` property can receive either string or list, I went for "simpler" solution so as to not make a decision about it.

PR #3232, opened 2023-12-04
Small fixes

### LINKED ISSUE #2880: Copy another organization's form
Can't "copy as new" on another organization's form as organization owner

### LINKED ISSUE #3230: Unify display of org in dropdowns
Most dropdowns in administration e.g. License dropdown shows the organization of each license, but at least Create Workflow dropdown for Form doesn't. It's difficult to select a form with the same name but different organization.

Macroz:
Close #2880 
Close #3230 

- Show organization in create workflow forms dropdown (was shown everwhere else)
- Use localized "org" text. Could be further improved by localizing `:` and parens too.
- Fixes warning in license view about lazy deref and missing key prop.
- Fixes warning from `:loading?` attribute being passed to HTML where it has invalid meaning.

image

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/cljs/rems/administration/components.cljs
+++ b/src/cljs/rems/administration/components.cljs
@@ -288,17 +288,33 @@
             [inline-info-field (to-label lang) value]))))
 
 (defn organization-field [context {:keys [keys readonly on-change]}]
-  (let [label (text :t.administration/organization)
-        organizations @(rf/subscribe [:owned-organizations])
+  (let [id (keys-to-id keys)
+        label (text :t.administration/organization)
+        owned-organizations @(rf/subscribe [:owned-organizations])
+        valid-organizations (->> owned-organizations (filter :enabled) (remove :archived))
+        disallowed (roles/disallow-setting-organization? @(rf/subscribe [:roles]))
         language @(rf/subscribe [:language])
         form @(rf/subscribe [(:get-form context)])
-        value (get-in form keys)
+        potential-value (get-in form keys)
         on-change (or on-change (fn [_]))
+        wrapped-on-change #(let [new-value %]
+                             (rf/dispatch [(:update-form context) keys new-value])
+                             (on-change new-value))
+
+        ;; if item was copied then this org could be something old
+        ;; where we have no access to so reset here
+        value (if (or readonly
+                      disallowed
+                      (contains? (set (mapv :organization/id valid-organizations))
+                                 (:organization/id potential-value)))
+                potential-value
+
+                ;; not accessible, reset
+                (wrapped-on-change nil))
+
         form-errors (when (:get-form-errors context)
                       @(rf/subscribe [(:get-form-errors context)]))
-        id (keys-to-id keys)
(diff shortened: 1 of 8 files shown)


src/cljs/rems/administration/components.cljs
```diff
@@ -288,31 +288,45 @@
...
+        ;; if item was copied then this org could be something old
+        ;; where we have no access to so reset here
+        value (if (or readonly
+                      disallowed
+                      (contains? (set (mapv :organization/id valid-organizations))
+                                 (:organization/id potential-value)))
+                potential-value
+
+                ;; not accessible, reset
+                (wrapped-on-change nil))
```
aatkin, line 313: reflecting on earlier discussion: i noticed that copy as new does fetch form template from API, and it got me thinking that would it make sense to include extra param and have API remove the disallowed organization? this could make UI code be less complected wrt. disallowed organization, although extra parameter smells a lot like backend for frontend. existing API would be unconcerned with it though.

related implementation thoughts:
```clojure
;; new function in rems.service.form
(defn get-form-for-copy [form]
  (let [organization (rems.db.organizations/get-organization-by-id-raw
                      (get-in form [:organization :organization/id]))]
    (if (rems.service.util/may-edit-organization? organization)
      form
      (dissoc form :organization))))

;; amend rems.api.forms with query-params
(GET "/:form-id" []
  :summary "Get form by id"
  :roles +admin-read-roles+
  :path-params [form-id :- (describe s/Int "form-id")]
  :query-params [{copy :- (describe s/Bool "whether to format return value suitable for copy-as-new action") false}]
  :return schema/FormTemplate
  (if-let [form (form/get-form-template form-id)]
    (ok (cond-> form
          copy form/get-form-for-copy))
    (not-found-json-response)))

;; add query-param in rems.administration.create-form/enter-page event-fx fetcher
:dispatch-n [(when form-id [::form {:copy true}])] 
```

aatkin (approved):

src/cljs/rems/administration/components.cljs
Macroz: I think there could be two different API endpoints, one for copy like `/forms/123/copy`, and choose which one to call.

aatkin: OK, sounds reasonable. here i don't see added value in moving more logic to fetcher, compared to doing filtering in organization component.

having a common way to filter domain objects in client, such that it is easy to present only e.g. own organization forms could be good for maintainability. this could be achieved by refactoring entities to use same patterns like `[:organization :organization/id]`. enriching API return values with ACL could provide similar benefits, however it is quite a bit heavier to implement.

Macroz: We should try to unify the domain objects a bit first. And perhaps this filter can try first `:organization/id` and then `:organization :organization/id` somewhat automatically. That's basically what the pattern could be after unifying the domain objects a bit better.

PR #3233, opened 2023-12-05
Optimize more

Macroz:
- Catalogue page has been sped up with optimizations. The API supports not joining organization data if not required (`join-organization=false`).
- Adding a user to blacklist does not reload the full cache anymore. This should make it faster. Also the user is directed to Applications page after delete, not to Catalogue.
- Deleting a (draft) application is now faster because it does not reload the full cache, only update it.
- Editing workflows (e.g. handlers) should now be faster because it does not reload the full cache, only update it.

# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changes since v2.34.2
 - Catalogue page has been sped up with optimizations. The API supports not joining organization data if not required (`join-organization=false`).
 - Adding a user to blacklist does not reload the full cache anymore. This should make it faster. Also the user is directed to Applications page after delete, not to Catalogue.
 - Deleting a (draft) application is now faster because it does not reload the full cache, only update it.
+- Editing workflows (e.g. handlers) should now be faster because it does not reload the full cache, only update it.
 
 ### Fixes
 - Big improvements in performance from small improvements in how applications and events are processed and cached. REMS should be able to handle a 100k event DB without breaking a sweat. (#2783)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Changes since v2.34.2
 
 **NB: This release removes the experimental application PDF export API. The non-experimental PDF export API is preferred instead. (#3098)**
 
+### Changes
+- Catalogue page has been sped up with optimizations. The API supports not joining organization data if not required (`join-organization=false`).
+- Adding a user to blacklist does not reload the full cache anymore. This should make it faster. Also the user is directed to Applications page after delete, not to Catalogue.
+- Deleting a (draft) application is now faster because it does not reload the full cache, only update it.
+
 ### Fixes
 - Big improvements in performance from small improvements in how applications and events are processed and cached. REMS should be able to handle a 100k event DB without breaking a sweat. (#2783)
 - Administration dropdown buttons should now respond to clicks more widely, and not only by directly clicking text. (#3167)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,6 @@ Changes since v2.34.2
 
 ### Changes
 - Application expiration now logs more, and more often. (#3225)
-
-### Changes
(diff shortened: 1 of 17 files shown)


src/cljs/rems/catalogue.cljs
```diff
@@ -166,18 +166,24 @@
...
+
+      (when (:enable-catalogue-tree config)
+        (if @(rf/subscribe [::full-catalogue :fetching?])
+          [spinner/big]
+          [catalogue-tree]))
+
+      (when (:enable-catalogue-table config)
+        (if @(rf/subscribe [::full-catalogue-tree :fetching?])
+          [spinner/big]
+          [catalogue-table]))]]))
```
aatkin, line 189: faster rendering is :+1:

with larger data, even though catalogue now loads a lot faster, there seems to be a rendering gap: if catalogue-table is enabled, it widens the layout once filled with data.

something that could be looked into as later improvement, maybe with bootstrap update

CHANGELOG.md
```diff
@@ -20,6 +20,12 @@ Changes since v2.34.2
 ### Changes
 - Application expiration now logs more, and more often. (#3225)
 
+### Changes
+- Catalogue page has been sped up with optimizations. The API supports not joining organization data if not required (`join-organization=false`).
+- Adding a user to blacklist does not reload the full cache anymore. This should make it faster. Also the user is directed to Applications page after delete, not to Catalogue.
+- Deleting a (draft) application is now faster because it does not reload the full cache, only update it.
+- Editing workflows (e.g. handlers) should now be faster because it does not reload the full cache, only update it.
```
aatkin, line 27: these headings could be merged

aatkin (approved):
good stuff!

src/clj/rems/service/catalogue.clj
```diff
@@ -31,22 +31,20 @@
...
 (defn get-localized-catalogue-items [& [query-params]]
   (->> (catalogue/get-localized-catalogue-items (or query-params {}))
-       (mapv join-dependencies)))
+       (mapv #(join-dependencies % query-params))))
 
 (defn get-localized-catalogue-item [id]
-  (->> (catalogue/get-localized-catalogue-item id)
-       join-dependencies))
+  (-> (catalogue/get-localized-catalogue-item id)
+      (join-dependencies nil)))
```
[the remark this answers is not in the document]
Macroz: That's something to consider in the future. I guess the joining at least won't go exactly like this, but I think we do need to support handing down a random set of options from API.

src/cljs/rems/catalogue.cljs
```diff
@@ -17,15 +17,15 @@
...
                    [::full-catalogue-tree])
-                 (when (roles/is-logged-in? (get-in db [:identity :roles])) [::draft-applications])
-                 [:rems.table/reset]]}))
+                 (when (:enable-catalogue-table (:config db))
+                   [::full-catalogue])]}))
 
-(fetcher/reg-fetcher ::full-catalogue "/api/catalogue")
-(fetcher/reg-fetcher ::full-catalogue-tree "/api/catalogue/tree" {:result :roots})
+(fetcher/reg-fetcher ::full-catalogue "/api/catalogue?join-organization=false")
+(fetcher/reg-fetcher ::full-catalogue-tree "/api/catalogue/tree?join-organization=false" {:result :roots})
```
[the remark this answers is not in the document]
Macroz: Yeah, that's true. I didn't remember it at the time and it's pretty natural to put them in the URL elsewhere. Some day we should rewrite the fetcher to make it a bit clearer.

src/cljs/rems/catalogue.cljs
Macroz: Currently all the pages suffer from the same issue of layout shift. The root cause is that I removed the fixed / max width some time ago. This seemed to be the only way to have the tables in the admin side to fill only as much space as they need and be centered.

PR #3254, opened 2024-01-16
Fix shadow tests

aatkin:
Fixed a bunch of warnings that occur during cljs tests due to uninitialized client state:

```
LOG: 'Testing rems.administration.test-create-catalogue-item'
ERROR: Error: Doesn't support name:
Error: Doesn't support name:
    at Object.cljs$core$name [as name] (http://localhost:9876/base/cljs-test.js?1160024ef03ca4aff729e52f86d481b08f219405:38752:8)
    at Object.taoensso$tempura$impl$expand_locale [as expand_locale] (http://localhost:9876/base/cljs-test.js?1160024ef03ca4aff729e52f86d481b08f219405:117127:100)

LOG: 'Testing rems.administration.test-create-form'
ERROR: 're-frame: no', ':event', 'handler registered for:', {ns: null, name: 'set-roles', fqn: 'set-roles', _hash: -181107226, cljs$lang$protocol_mask$partition0$: 2153775105, cljs$lang$protocol_mask$partition1$: 4096}

LOG: 'Testing rems.test-fields'
ERROR: Error: Doesn't support name:
Error: Doesn't support name:
    at Object.cljs$core$name [as name] (http://localhost:9876/base/cljs-test.js?062c36d3d855ea93d4a41254cc1d91ccc720099a:38752:8)
    at Object.taoensso$tempura$impl$expand_locale [as expand_locale] (http://localhost:9876/base/cljs-test.js?062c36d3d855ea93d4a41254cc1d91ccc720099a:117127:100)
```

Other changes include:
- Updated shadow-cljs, karma and karma-chrome-launcher to latest. Also bumped `medley` in shadow-cljs to latest (fixes cljs.core/abs shadowing)
- Npm aliases for commonly used shadow-cljs scripts, like build/watch. These are used by leiningen aliases (for example in CI). May be useful if starting up JVM feels too slow.
- `lein shadow-test` now starts karma with `--single-run` CLI flag, and `karma.conf.js` no longer has `singleRun: true`. This was a source of confusion because you could not make karma run in watcher mode, but it would also not tell you why.

NB! It looks like [karma is deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) since 27/04/2023.

# Checklist for author

## Documentation

## Testing


test/cljs/rems/testing.cljs
```diff
@@ -1,9 +1,18 @@
...
 
-(defn isolate-re-frame-state [f]
+(defn- clear-rf-state! []
+  #_(rf/clear-cofx)
+  #_(rf/clear-event)
+  #_(rf/clear-global-interceptor)
+  #_(rf/clear-sub)
+  (rf/clear-subscription-cache!))
+
+(defn init-client-fixture [f]
```
Macroz, line 11: Maybe this name could still be something else. I mean it talks about some client but I'm asking what client? Maybe `init-spa-fixture` or something related to that?

test/cljs/rems/test_util.cljs
```diff
@@ -22,7 +25,8 @@
 (deftest test-linkify
   (let [link [:a {:target :_blank :href "http://www.abc.com"} "http://www.abc.com"]]
     (testing "retain original string"
-      (is (= (apply str (linkify "a b c") "a b c"))))
+      (is (= (linkify "a b c")
+             ["a b c"])))
```
Macroz, line 29: Is the expected and actual value reversed in this test? I guess they should be in reverse order?

aatkin: True, even though other tests in same file specify expected/actual as elsewhere. Fixed

Macroz (approved):

Issue #3105, opened 2023-01-16
Optimize form editor


Macroz:
The form editor fields subscribe to the whole form. We should add more detailed subscription to improve rendering performance.

Possibly React/Reagent update could help too.

Issue #2368, opened 2020-09-29
Analyze browser console warnings


Macroz:
Walk through the application and have a look at what the browser console shows as warnings and errors. Create issues of each or fix right away if easy.

- [ ] create issues or fix normal applicant pages
- [ ] create issues or fix handling pages
- [ ] create issues or fix admin pages

marharyta:
<img width="1440" alt="Screenshot 2021-01-27 at 19 04 53" src="https://user-images.githubusercontent.com/7344718/106026483-8bd01280-60d2-11eb-94ad-f11ab0cfc17a.png">
The only thing I am getting on my Chrome

Macroz:
I think you must be running on your local machine as devtools would not be in the release build.

Macroz:
There are plenty more if you just walk through all the screens ...
![warnings](https://user-images.githubusercontent.com/823661/106035047-4f091900-60dc-11eb-94a1-13f114a3c403.png)

aatkin:
https://cljdoc.org/d/etaoin/etaoin/1.0.40/api/etaoin.api#get-logs

PR #3269, opened 2024-03-04
Update re-frame

aatkin:
relates #3105 (front-end update)

After update (https://github.com/day8/re-frame/issues/740), re-frame warns about subscriptions created outside of reactive context.

Document title did not seem to set correctly before, which should be fixed and has browser tests now.

- updated `package-lock.json` to lockfileVersion 3, and `package.json` dependencies are pinned by exact version
  - ~**NB:** `follow-redirects` is downgraded to 1.15.4 (updated by dependabot in https://github.com/CSCfi/rems/pull/3274). `npm ls follow-redirects` traces back to karma test runner.~ edit: `follow-redirects` is now pinned to 1.15.6 in dev dependencies instead to keep dependabot changes.
- updated re-frame from [1.2.0](https://day8.github.io/re-frame/releases/2021/#120-2021-02-25) → [1.4.3](https://day8.github.io/re-frame/releases/2021/#120-2021-02-25) and fixed several warnings "re-frame: Subscribe was called outside of a reactive context":
  - `rems.dropdown`: label functions that call e.g. `(rems.text/text ...)`
  - ~`rems.table`: subscription functions that call e.g. `(rems.text/text ...)`~ (edit: will be changed later)
  - `rems.flash-message`: use reagent.core/with-let instead of reagent.core/create-class, due to issues with reactive context
  - `rems.atoms/document-title`: use reagent.core/with-let instead of reagent.core/create-class, due to issues with reactive context
  - user focus "grab" and lazy-load-data in `rems.spa` are implemented using reagent.core/track!. at first this was only for focus grab which was splitted from main-content that used reagent.core/create-class, but the pattern seemed to fit hook-like helpers
- rems main app is now rendered using `rems.spa/app`, that separates some setup code from page layout
- extracted `rems.theme` namespace (cljs)
- added [react strict mode](https://react.dev/reference/react/StrictMode) to development build. this has subtle effect of "double rendering", but should help catching issues earlier.

# Checklist for author

## Reviewability
- [x] Link to issue

## Testing


src/cljs/rems/application.cljs
```diff
@@ -1178,35 +1177,37 @@
                (when-let [missing-duos (seq (unmatched-duos application-duo-matches))]
                  [:div.alert.alert-warning
                   [:p (text :t.applications.duos.validation/missing-duo-codes)]
-                  [:ul (for [duo missing-duos
-                             :let [label (localized (:duo/label duo))]]
-                         ^{:key (:duo/id duo)}
-                         [:li (text-format :t.label/dash
-                                           (:duo/shorthand duo) label)])]])
+                  (into [:ul]
+                        (for [duo missing-duos]
```
Macroz, line 1181: This is not a very satisfying solution. In some projects I've made some custom `for` loops for components to clean up the code. Another solution would be to, instead of `:ul` to have our own `list` component that would accept the sequence (or multiple sequences to get rid of extra `concat` etc.) and do the needful.

E.g., `[List (for [duo missing-duos] ...)]`

We could consider doing such cleanup after we merge the open PRs and do a release.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -115,48 +116,48 @@
...
-                            :sort-value (if checked? 1 2)})
-                 :commands {:display-value [:div.commands
-                                            [view-button (:id item)]
-                                            [modify-item-dropdown item]]}})))))
+ (fn [_db-sub]
+   (r/reaction
+    (let [language @(rf/subscribe [:language])
+          org-ids @(rf/subscribe [:rems.administration.administration/displayed-organization-ids])
+          filter-by-displayed-organizations (partial administration/filter-by-displayed-organization org-ids #(get-in % [:organization :organization/id]))]
+      (doall
```
Macroz, line 127: This feels like a step backwards. Should we just separate the `text` as function and `text` as component? Or if we used the text from a global var, we would not have to do this?

Also everything used to "just work" and https://github.com/day8/re-frame/pull/754 would just fix this?

src/cljs/rems/flash_message.cljs
```diff
@@ -1,9 +1,9 @@
 (ns rems.flash-message
   (:require [clojure.test :refer [deftest is]]
-            [reagent.core :as reagent]
+            [clojure.walk]
```
Macroz, line 3: What is `clojure.walk` required for?

src/cljs/rems/spa.cljs
```diff
@@ -130,6 +126,7 @@
  (fn [db [_ path]]
    (assoc db :path path)))
 
+;; XXX: unused event?
```
Macroz, line 129: I think so.

src/cljs/rems/spa.cljs
```diff
@@ -341,61 +333,41 @@
...
+  (b/cond
+    :when page-id
+    :let [content (get pages page-id)]
+
+    (some? content)
+    [:main#main-content.container-fluid {:class (str "page-" (name page-id))}
+     [missing-email-warning]
+     [content]]
+
+    (do ; implementation error
```
Macroz, line 360: Could be nice to keep using `:else` like in regular `cond` as it is a bit more explicit? Not sure.

src/cljs/rems/spa.cljs
```diff
@@ -648,9 +620,39 @@
...
+     #(when-let [element (or (.querySelector js/document "h1")
+                             (.querySelector js/document "#main-content"))]
+        (focus/focus element)
+        (rf/dispatch-sync [::focus-grabbed])))))
+
+(defn- app []
+  (r/with-let [tracked (mapv r/track! [lazy-load-data! grab-focus!])]
+
+    (if (config/dev-environment?)
+      [react-strict-mode [page]]
```
Macroz, line 646: Are there a lot of warnings? Should browser tests fail if there are any/more?

Would actually be nice to fail any browser tests with any seen warnings (or maybe ignore some difficult to fix ones).

Macroz (commented):
Looks pretty good to me. I think we need to think a little bit if we should still decide not to do the raw reaction. Stuff just used to work and the workaround is pretty awful. There seems to be a fix to re-frame in the pipe although in danger of being forgotten.

Maybe it would make sense to do the "raw global var" instead for translations. The mechanism is very central to REMS but using it through the subscriptions is not necessary. If we want to try that, we could do everything here except using the raw reactions, and keep them as is even with the warnings.

src/cljs/rems/flash_message.cljs
aatkin: It has been used but not required previously in this same file:

https://github.com/CSCfi/rems/blob/58520dfc83ac7e1e7f09b09ad3e415cd6898ee87/src/cljs/rems/flash_message.cljs#L127-L132

src/cljs/rems/spa.cljs
aatkin: Almost none actually, i have only seen one from `komponentit.autosize/textarea` for using (deprecated) react-dom/findDOMNode instead of ref. `react-transition-group` (not in this PR) also triggers that warning under strict mode if DOM element ref is not manually handled.

It could be good to test what warnings/how many are produced. #2368 already exists for tracking this, too.

aatkin: React 18 slightly changes component mounting/unmounting in strict mode: https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-strict-mode

Although this looks more related to using react/useState, which our components are not using (dropdown via react-select might be).

src/cljs/rems/administration/duo.cljs
```diff
@@ -118,10 +118,12 @@
          [:label.administration-field-label {:for "mondos-dropdown"} restriction-label]
          [dropdown/async-dropdown
           {:id "mondos-dropdown"
+           :items mondos
            :item-key :id
-           :item-label #(text-format :t.label/dash (:id %) (:label %))
+           ;; due to async loading, item label cannot be set ahead of time, but label function also cannot use subscriptions.
+           ;; TODO: investigate solutions
```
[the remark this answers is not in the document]
Macroz: Could the backend do the localization for the search results?

src/cljs/rems/theme.cljs
```diff
@@ -0,0 +1,12 @@
+(ns rems.theme
+  (:require [re-frame.core :as rf]))
+
+(rf/reg-sub :rems.theme/theme (fn [db] (::theme db)))
+(rf/reg-event-db :rems.theme/loaded-theme (fn [db [_ theme]] (assoc db ::theme theme)))
```
Macroz, line 5: This could use `::theme`, I think that's OK in its own file.

The reason we had `:theme` before was just that it's kind of a common and standard sub/event for the entire app so it has a "global identity". Now everyone must use the specific namespaced name, which is OK too, just that it is a change.

Macroz (approved):
Ok let's get this merged and we can find any problems with manual testing.

PR #3285, opened 2024-04-18
Extract globals from re-frame state

aatkin:
# Checklist for author

## Reviewability
- [ ] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up


Macroz (approved):
This looks pretty much ready. I think we should consider the spa, app & globals and possibly decide to have just two.

Also we are using direct atom references, but if it reads badly we could use helper functions instead. As long as everything is found with `rems.globals` then we can change in the future.

Issue #3283, opened 2024-04-15
Investigate performance differences in environments.


Macroz:
rems-dev is faster than FD test REMS (about 3x so 100ms vs. 300ms). Could the difference be due to:
- email being sent for real (to all handlers)
- one handler email is wrong and gives out an error
- many more handlers (10-26)
- larger forms (128 fields)
- event notification integration

Production is maybe even slower (for full cache reload at least, seems not for single events).

Macroz:
Once #3295 and #3305 are merged and a new release deployed to testing we'll see if this can be closed already.

Macroz:
If email address is wrong the retrying is stopped. This was in https://github.com/CSCfi/rems/commit/f1d1f01cca223cf63d3db01615ceee76efacbfb5.

PR #3305, opened 2024-05-21
Many handlers perf

Macroz:
Tries to fix the many handlers perf problems from #3283 

There's two changes here of which one is more subtle.

The original logic from years ago was such that all users rights were
updated whenever there was any change. This was fast for a while.

Last year, updates were implemented to only touch users that were
involved in the applications of the events. This was much faster.
There is however a flaw in this version.

Consider a case where one application was changed but there are 100
potential handlers in the organization. All the rights of these 100
handlers were updated by considering all their applications. Handlers
tend to be members of all applications so that could be 100 times the
application count of updates, e.g. 300K for 3000 applications. This
adds up to a few hundred milliseconds, which is too slow for us now,
,specially after adding more performance test data (10 handlers in
production so 100 handlers in performance test data).

We can avoid this amount of changes by recording how many times you have a
role instead of only whether you have the role. Then we remove the
role once per member of the old application and add 1 per member of new
application. This will total 100 changes for 100 users in one
application, which is as good as it can be. We could also record which
application gives which role, but that is slightly more memory
consuming for no additional gain.

The other subtler change is to cache only ids of applications that a person
has, not the personalized applications. 100 handlers with 3000
applications could result in 300K applications being cached.
Clojure's persistent data structures do help us here and in practice
that would not happen. However, we can avoid this by doing the
personalization in the functions that return results instead and
have less memory use for some runtime cost. We do store the enriched
applications, so this personalization is only in-memory code, so it
should be fast.

These changes seem to be fast for the handled applications case, so it
should be fast enough for the other cases.

Further changes can be done in a new PR also, after we get all the improvement PRs integrated.

Main changes:
- adds 100 handlers to perf test data
- fail outbox job immediately on address error
- faster applications cache updates
- fetch only active handlers from `/handled` API

Potential future tasks:
- fetch only active handlers from other "overview" APIs
- low level cache for more (user) tables
- investigate real performance (in dev)

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary (TBD)

## Testing

## Follow-up
- [ ] New tasks are created for pending or remaining tasks (TBD)

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -114,6 +114,7 @@
 
    :project/dev {:dependencies [[binaryage/devtools "1.0.7"]
                                 [com.clojure-goes-fast/clj-memory-meter "0.3.0"]
+                                [com.clojure-goes-fast/clj-async-profiler "1.2.0"]
                                 [criterium "0.4.6"]
                                 [lambdaisland/kaocha "1.87.1366"]
                                 [lambdaisland/kaocha-junit-xml "1.17.101"]
--- a/src/clj/rems/db/applications.clj
+++ b/src/clj/rems/db/applications.clj
@@ -6,7 +6,7 @@
             [clojure.test :refer [deftest is]]
             [clojure.tools.logging :as log]
             [conman.core :as conman]
-            [medley.core :refer [distinct-by map-vals]]
+            [medley.core :refer [distinct-by filter-vals map-vals]]
             [mount.core :as mount]
             [rems.application.events-cache :as events-cache]
             [rems.application.model :as model]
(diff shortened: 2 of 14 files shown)


project.clj
```diff
@@ -115,6 +115,7 @@
    :project/dev {:dependencies [[binaryage/devtools "1.0.7"]
                                 [com.clojure-goes-fast/clj-async-profiler "1.2.0"] ; also check extra :jvm-opts https://github.com/clojure-goes-fast/clj-async-profiler?tab=readme-ov-file#tuning-for-better-accuracy
                                 [com.clojure-goes-fast/clj-memory-meter "0.3.0"]
+                                [com.clojure-goes-fast/clj-async-profiler "1.2.0"]
```
aatkin, line 118: this is now extra, since the dependency was added previously (two rows above)

aatkin (approved):

src/clj/rems/db/applications.clj
```diff
@@ -307,7 +348,7 @@
 
     {::raw-apps new-raw-apps
      ::enriched-apps new-enriched-apps
-     ::apps-by-user new-personalized-apps-by-user
+     ::app-ids-by-user new-app-ids-by-user
      ::roles-by-user new-roles-by-user
      ::users-by-role new-users-by-role}))
```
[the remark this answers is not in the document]
Macroz: See #3309

project.clj
Macroz: Yeah, automatic merge conflict.

PR #3308, opened 2024-05-23
Faster translations 2

aatkin:
on top of #3305

- memoized resource compiler
- optimize missing tr key case (a lot faster)
- added more perf test namespace tools 

Performance tests that find every translation key in dict for every language, then runs according function for every key:
- if translation contains vector arguments, `text-format` with generated vector values
- if translation contains map arguments, `text-format` with single map with generated values
- else, `text` (no arguments)

mean value is extracted from performance test after cache warmup

no memoize, no fn optimization
---
| total translations | mean         | empty cache size | warm cache size | final cache size |
| ------------------ | ------------ | ---------------- | --------------- | ---------------- |
| 1885               | 10,148342 ms | 88 B             | 890,6 KiB       | 890,6 KiB        |

no memoize, fn optimization
---
| total translations | mean        | empty cache size | warm cache size | final cache size |
| ------------------ | ----------- | ---------------- | --------------- | ---------------- |
| 1885               | 4,606966 ms | 88 B             | 890,6 KiB       | 890,6 KiB        |

memoize, fn optimization
---
| total translations | mean        | empty cache size | warm cache size | final cache size |
| ------------------ | ----------- | ---------------- | --------------- | ---------------- |
| 1885               | 2,722741 ms | 88 B             | 1,2 MiB         | 1,2 MiB          |

# Checklist for author

## Reviewability


src/cljc/rems/text.cljc
```diff
@@ -82,11 +81,13 @@
   "Return the tempura translation for a given key. Additional fallback
   keys can be given."
   [& ks]
-  #?(:clj (tr (conj (vec ks) (text-format :t/missing (vec ks))))
+  #?(:clj (or (tr ks) ; optimization: fetch missing translation only when needed
+              (text-format :t/missing ks))
```
aatkin, line 85: this change affects `text` function, not `text-no-fallback`. github diff looks slightly confusing

aatkin:
converting to draft, checking potential issue with the performance test

PR #3322, opened 2024-08-21
Fix broken tests

aatkin:
Fixes issue with browser tests not passing:
- `open-collapsible` helper no longer uses class name in selector
- `add-form-field!` helper uses less explicit id management (again)

Other changes:
- removed unused refresh driver fixture (actually was removed already before, due to stability issues)
- added flag to disable chrome from asking default search (quite annoying when running non-headless)
- removed custom implementation of `fill-human` and replaced with etaoin.api/fill-human. tuning the algorithm was quite time consuming and it wouldn't prevent tests from flaking anyway
- added wait and extra selector path to dropdown select in `test-invite-reviewer` (perhaps flaky due too fast clicking)
- updated `etaoin` to latest, introduced some new API

### DIFF
--- a/test/clj/rems/browser_test_util.clj
+++ b/test/clj/rems/browser_test_util.clj
@@ -447,41 +447,10 @@
   (wait-for-idle)
   (no-timeout? #(apply wait-invisible args)))
 
-;; TODO our input fields process every character through re-frame.
-;; Etaoin's fill-human almost works, but very rarely loses characters,
-;; probably due to the lack of a _minimum_ delay between keypresses.
-;; This is a reimplementation.
-(def +character-delay+ 0.01)
-(def +typo-probability+ 0.05)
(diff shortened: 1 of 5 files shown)


test/clj/rems/browser_test_util.clj
```diff
@@ -174,21 +159,6 @@
         (log/warn e "WebDriver failed to start, retrying...")
         (run)))))
 
-(defn refresh-driver-fixture
```
Macroz, line 176: The reason for using this previously was that we can re-use the same browser as opening a new browser window / driver is much slower.

Seems like it was removed in https://github.com/CSCfi/rems/commit/dbe1dc49ed35f10e1806d4cb0564da1481da3f4b

Macroz (approved):
More stability = good

PR #3324, opened 2024-08-22
Many handlers perf2.1

Macroz:
Replaces #3313 

Continues work in #3305 

Includes:
- Makes sure handled applications list is fetched lazily so correct amount of work is done when a limit is given (e.g. only 50).
- Logs test data creation speed to be able to easily follow it. This gives an indication of general REMS performance.
- Emails for one application were sent one at a time. E.g., with 12 handlers there would 12 separate calls to save the email into the outbox. Now this is done with a single call, which helps the 100 handler "worst-case scenario". Tests needed updating to use the new function.
- Replaced `mapcat` with other code to make it faster (micro-optimization).
- The main problem for the performance of the `/api/applications/handled` is actually the verification of the return Schema. If there are thousands of applications, there is a lot of data and half of the time is just spent verifying that it matches our Schema. We could investigate if we can leave the checking out in some cases (e.g., when we know there are a lot of applications). See [issue](https://github.com/orgs/CSCfi/projects/13/views/1?pane=issue&itemId=75940840)

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
- [ ] Update changelog if necessary  (no need for a new one as the last improvement is already mentioned)

## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks (see [issue](https://github.com/orgs/CSCfi/projects/13/views/1?pane=issue&itemId=75940840))

### DIFF
--- a/src/clj/rems/api/applications.clj
+++ b/src/clj/rems/api/applications.clj
@@ -78,7 +78,7 @@
 (defn- overview-only-active-handlers [app]
   (update-in app
              [:application/workflow :workflow.dynamic/handlers]
-             #(filterv :handler/active? %)))
+             #(filter :handler/active? %)))
 
 ;; Api implementation
 
@@ -135,6 +135,14 @@
     (merge {:success true}
            (commands/validate-application application (:field-values request)))))
 
+(defn- get-handled-applications [{:keys [query only-active-handlers limit]}]
+  (time
+  (cond->> (todos/get-handled-todos (getx-user-id))
+    only-active-handlers (map overview-only-active-handlers)
+    query (filter-with-search query)
+    true (sort-by last-activity >)
+    limit (take limit))))
+
 (def my-applications-api
   (context "/my-applications" []
     :tags ["applications"]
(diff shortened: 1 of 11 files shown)


src/clj/rems/api/applications.clj
```diff
@@ -176,15 +184,14 @@
     (GET "/handled" []
       :summary "Get all applications that the current user no more needs to act on."
       :roles #{:logged-in}
+      ;; XXX: checking this schema can be mighty slow (thousands of applications to check individually)
```
aatkin, line 187: there seems to be a newer version of compojure-api released: https://github.com/metosin/compojure-api/blob/master/CHANGELOG.md#200-alpha33-2024-04-30

though testing locally, e.g. resources api with performance data is still as slow as with older version.

aatkin (approved):
going ever faster! :+1:

src/clj/rems/email/core.clj
```diff
@@ -30,35 +30,43 @@
...
-                              (template/reviewer-reminder-email lang reviewer apps))))
-                     (remove nil?))]
-    (enqueue-email! email)))
+  (let [deadline (-> (time/now) (.plus ^Period (:email-retry-period env)))]
+    (->> (for [email (->> (applications/get-users-with-role :reviewer)
+                          (map users/get-user)
+                          (map (fn [reviewer]
+                                 (let [lang (:language (user-settings/get-user-settings (:userid reviewer)))
+                                       apps (->> (todos/get-todos (:userid reviewer))
+                                                 (map #(= :waiting-for-your-review (:application/todo %))))]
```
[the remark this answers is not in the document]
Macroz: I can see if I can refactor this. `keep` would work here but there's more.

I think the tests are in `rems.api.test-email`, e.g., `test-send-reviewer-reminder`.

src/clj/rems/api/applications.clj
Macroz: We should keep an eye on it an upgrade once there is a good "2.0" or so.

PR #3337, opened 2024-10-28
feat: keyboard accessibility for catalogue tree

aatkin:
implements #3336 

Enables keyboard accessibility for following use:
- Tree items can be "tabbed" into (pressing tab key, or shift+tab) which enables rest of the keyboard accessibility
- Tree can be navigated vertically using the following keys: `[up arrow, down arrow, home, end]`
  - `home` and `end` jump to first/last tree item respectively
- Tree items can be opened/closed using the following keys: `[left arrow, right arrow, enter]`
- Catalogue items can be added to cart, removed from cart, and applied for using the following keys: `[enter]`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tree_role

ARIA tree "widget" is implemented like so:
- `<table aria-label aria-labelledby aria-keyshortcuts>`
  - screen reader seems to always include table when tabbing into treeitems, so it contains extra context
- `<thead role="presentation">`
  - contextual role is removed so screen reader does not repeat row,cell every time. this is included only because of "misusing" table
- `<tbody role="tree">`
  - contains the list of treeitems
- `<tr role="treeitem" aria-level aria-expanded aria-label aria-describedby aria-selected="false">`
  - each is either expandable (category) or node (resource)
  - tabindex is set for treeitems so they can be tabbed into
  - aria-describedby refers to element in category that may contain description
  - aria-selected="false" is included so screen reader does not repeat "selected" every time, when tree rows are not selectable

Because the underlying implementation is `<table>`, it's quite difficult to not make screen reader repeat a lot of table-related information, without removing usability of the assistive tools. For example `role="presentation"` on table somehow prevents screen reader from speaking "row expanded/collapsed" event. Table also has rules regarding element nesting, which React oversees vigorously.

## Other fixes:
- Replaced catalogue tree `h3` headers with div + `.tree-header` class that imitates Bootstrap header styles. Should look the same, but doesn't conflict with document semantics. Font size is still generated with `fs-depth-n` util.

https://github.com/user-attachments/assets/09cce04b-f842-48c2-a427-a03d981f05a6

# Checklist for author

## Reviewability
- [x] Link to issue
- [x] Consider adding video for ease of review

## Documentation
- [x] Components are added to guide page (updated existing and tested that they work, too)

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically (maybe enough that axe + current tests pass?)

## Follow-up


Macroz (approved):
I think we could perhaps merge this as it's better than the existing version. We have to create a new issue to improve using NVDA still. But it could be its own issue.

aatkin:
> I think we could perhaps merge this as it's better than the existing version. We have to create a new issue to improve using NVDA still. But it could be its own issue.

Created issue #3352 for continuation

PR #3340, opened 2024-11-04
Separate request count logging

Macroz:
The logging of the simultaneous request count (`rqc:`) is now done for the request and response log lines only, and separately. Previously the same value was used for all logging in a request, which could be confusing as the actual request count can change between the lines.

# Checklist for author

## Reviewability

## Documentation


aatkin (approved):
:+1:

PR #3343, opened 2024-11-05
User simulator

aatkin:
Implements simple tool that generates user traffic to target REMS.

Creates headless browser for each simulated user, who login and proceed to navigate into random applications (that belong to the user). Thread daemon attempts to keep all users working, in case the webdriver crashes or something surprising happens.

Some limitations:
- only supports fake login users at the time
- fixed user actions (viewing random application)
- no load test statistics
- no burst traffic

# Checklist for author

## Documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ have notable changes.
 
 Changes since v2.38
 
+### Additions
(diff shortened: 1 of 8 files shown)


Macroz (approved):
Looks good :+1: 

We can work on top of this.

Possibly we could add it to CHANGELOG too, as an experimental tool for developers.
And we could add the running instructions around here somewhere https://github.com/CSCfi/rems/blob/master/docs/development.md#running-tests

PR #3344, opened 2024-11-07
Fixes for blacklist

aatkin:
Fixes erratic behaviour in blacklist and subsequent applications reloading.

- Blacklist events are now sorted before event reducer. 
  - Previous behaviour caused events to be processed in semi-random order after generating enough events (>=10 or so).
- Blacklisting now correctly handles application cache update. 
  - Previously applications reload never occurred, because the code used wrong cache path for user-ids.

# Checklist for author

## Reviewability

## Documentation
- [ ] Update changelog if necessary (fixes issue in unreleased feature)

## Testing

### DIFF
--- a/src/clj/rems/db/blacklist.clj
+++ b/src/clj/rems/db/blacklist.clj
@@ -53,6 +53,7 @@
 
 (defn get-events [& [params]]
   (->> (vals (cache/entries! blacklist-event-cache))
+       (sort-by :event/id)
        (apply-filters params)))
 
 (defn add-event! [event]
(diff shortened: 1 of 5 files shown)


src/clj/rems/db/blacklist.clj
```diff
@@ -53,6 +53,7 @@
 
 (defn get-events [& [params]]
   (->> (vals (cache/entries! blacklist-event-cache))
+       (sort-by :event/id)
```
Macroz, line 56: `rems.db.events` uses `sorted-map` to maintain the ordering

Macroz (approved):

src/clj/rems/db/blacklist.clj
aatkin: Good point, let's use the same here as well

PR #3346, opened 2024-11-12
Friendlier cache in development mode

aatkin:
Reloading namespaces with caches causes some friction when cache is mistakenly treated as duplicate.

- only log warning **in dev mode** when cache is created using existing cache id

# Checklist for author

## Reviewability

## Testing


Macroz (approved):
Thanks!

PR #3353, opened 2024-11-21
Fix hooks

### LINKED ISSUE #3351: hooks no longer seem to work as doc'ed
**Describe the bug or behavior**
I'm trying to implement a hook in `js` as described here [hooks.md](https://github.com/CSCfi/rems/blob/master/docs/hooks.md#hooks)

**To Reproduce**
1. Create the hook file `foo.js`
2. Reference it in config via `:extra-scripts` 
3. Start services, open the JS console in the browser, and perform any UI action in REMS
4. `foo.js` throws an error stating that `rems.hooks` is undefined:

image of JS console showing a red cross with text Uncaught TypeError: Cannot read properties of undefined (reading 'hooks)

**Expected behavior**
This used to work, prior to d436777, which removed the inline JS initialization of the hooks.

If I revert back to 2f3321e and build, the exact hook example file in the [documentation](https://github.com/CSCfi/rems/blob/master/docs/hooks.md#hooks) works as expected.
So I'm not sure if this is a bug in the documentation because hooks actually work a different way now, or a bug in the current hook implementation.

**Environment**
 - Version of REMS: v2.37
 - OS: Fedora Linux fc40
 - Browser: Chrome 131.0.6778.33

Macroz:
Close #3351 

Hook initialization overwrites the functions set in extra scripts. We can implement them slightly differently to avoid this problem. Also we can add a test.

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing


aatkin (approved):
looking good :+1:

Issue #1412, opened 2019-07-18
Adding members shows users already belonging to the application in the list


hukka:
Better not to even show those users who are already in, than let the user try to add them and then show an error.

tjb:
hey everyone! i am new to learning clojure & want to participate in Hacktoberfest. is it ok if i take on this issue?

Macroz:
Yes definitely!

opqdonut:
@tjb Do you need help getting started? `README.md` and `docs/development.md` are good places to start. Ping us if you're stuck with something.

tjb:
@opqdonut honestly the documentation is *really* good! i was able to get the app running very quickly. 

i am having a little issue of making the REPL work with both the backend and frontend using [Calva](https://marketplace.visualstudio.com/items?itemName=betterthantomorrow.calva) but that is ok.

I believe what will work best for this situation is create a new endpoint that the dropdown fetches from that will filter out any applicants based on application id. this allows us to make the backend the "source of truth" so to speak. I will have additional time to work on this this weekend and hopefully get a PR up by Sunday.  

an additional question i am not sure where to ask or if it is ok to log an issue for: i believe the endpoint should be paginated as well. with the test data, this is an extremely large list of users that takes a while for a return response and then populating the UI. i think that may be out of scope for this and would love to make an issue around it if that seems appropriate to everyone!

Macroz:
The existing members are also available in the app db under `:rems.application/application :data :application/members`. So we could filter the dropdown also in the UI? This would probably be the simplest solution to the original issue.

As far as the second question goes. It's fine to file an idea issue about pagination! 

So far we are expecting (and have perf test data for) about 1000 items, also users, so we have not considered doing pagination in the backend because the amount is rather small. However now that I'm trying it out the performance is quite poor and I notice at least two things. It is transfering 92kB of data which sounds ok (roughly 100 bytes per user including name and email at least). The second and real issue is probably the rendering of the 1000 items into the `react-select` dropdown. Maybe it has poorer performance than it should. Could be that it causes also unnecessary re-rendering of the page. Good ideas about that are always also welcome!

Lastly, perhaps doing the filtering in the backend would optimize here. We use a different approach i.e. a Lucene index for searching applications on the Actions page and it's also potentially something to use for the backend of a dropdown.

hukka:
> The second and real issue is probably the rendering of the 1000 items into the react-select dropdown.

[...]

PR #3391, opened 2025-10-31
Filter dropdown members

vuorioma:
filter out existing members from the 'Add member' dropdown menu

Related issue: https://github.com/CSCfi/rems/issues/1412

~~Minor followup issue: If the user removes a member while the dropdown is displayed, the removed member does not instantly appear in the search results.~~ Fixed

### DIFF
--- a/src/cljs/rems/actions/add_member.cljs
+++ b/src/cljs/rems/actions/add_member.cljs
@@ -26,7 +26,8 @@
   (let [application (get-in db [:rems.application/application :data])
         applicant-id (get-in application [:application/applicant :userid])
         members (get-in application [:application/members])
-        existing-ids (into #{applicant-id} (map :userid members))]
+        existing-ids (into #{applicant-id} 
+                           (map :userid members))]
     (->> ::potential-members db
          (remove (comp existing-ids :userid))
          (map atoms/enrich-user))))


src/cljs/rems/actions/add_member.cljs
```diff
@@ -26,9 +26,17 @@
...
-   (assoc db
-          ::potential-members (set (map atoms/enrich-user members))
-          ::selected-member nil)))
+   (let [application (-> db :rems.application/application :data)
+         applicant (-> application :application/applicant)
+         existing-ids (set (concat 
+                            [(:userid applicant)] 
+                            (map :userid (:application/members application))))
+         filtered (->> (set members)
+                       (filter #(not (contains? existing-ids (:userid %))))
```
Macroz: Something to consider here:

`(filter #(not (contains? existing-ids (:userid %))))` 

could be written simpler as 

`(remove #(contains? existing-ids (:userid %)))`

or even like

`(remove (comp existing-ids :userid))`

The use of `remove` removed the need of negation. And using `(comp set ...)` is an example of the search idiom in Clojure which is also quite common in REMS codebase.

src/cljs/rems/actions/add_member.cljs
```diff
@@ -22,12 +22,23 @@
...
+    (let [application (get-in db [:rems.application/application :data])
+          applicant (get-in application [:application/applicant])
+          members (get-in application [:application/members])
+          existing-ids (set (concat [(:userid applicant)]
+                                    (map :userid members)))]
+      (->> ::potential-members db
+           (remove (comp existing-ids :userid))
+           (map atoms/enrich-user))))
+
+(rf/reg-sub ::potential-members (fn [db _] (potential-members db)))
```
Macroz: We could remove this extraneous inline `fn` if the only thing is calling `potential-members`, or inline `potential-members` here. Something like this?
```
(rf/reg-sub ::potential-members potential-members)
```

Macroz (approved):
This is now good to merge. 

If you want to continue with a test, then you can do it here, or in another PR or we can leave it and continue on something else.

PR #3396, opened 2025-11-05
fix: hide edit buttons when organization owner is not allowed to modify the resource

pikariop:
https://github.com/CSCfi/rems/issues/2814

Logged in as "Organization Owner 2"
Screenshot 2025-11-05 at 15 53 49
Screenshot 2025-11-05 at 16 34 40

Screenshot 2025-11-05 at 16 10 19
Screenshot 2025-11-05 at 15 46 38
Screenshot 2025-11-05 at 16 46 24

Screenshot 2025-11-05 at 15 48 46

# Checklist for author

Remove items that aren't applicable, check items that are done.

## Reviewability
- [x] Link to issue
- [x] Consider adding screenshots for ease of review

## Documentation
- [x] Update changelog if necessary

## Testing
- [ ] Valuable features are integration / browser / acceptance tested automatically

PR #3400, opened 2025-11-13
test: when editing an organization without being the owner, edit button is hidden

pikariop:
Tests for pr https://github.com/CSCfi/rems/pull/3396
Fix https://github.com/CSCfi/rems/issues/2814

Also, on the catalogue items admin page, fix a slight discontinuity in previous implementation by checking user's permission to edit the selected items before enabling the "Update catalogue item" button

# Checklist for author

## Reviewability

## Testing

### DIFF
--- a/src/cljs/rems/administration/catalogue_items.cljs
+++ b/src/cljs/rems/administration/catalogue_items.cljs
@@ -88,7 +88,8 @@
    {:id :update-catalogue-item
     :class "btn-primary"
     :disabled (when (or (empty? items)
-                        (some (complement roles/can-modify-organization-item?) items)) :disabled)
+                        (not-every? roles/can-modify-organization-item? items))
+                :disabled)
     :on-click (fn []
                 (rf/dispatch [:rems.administration.update-catalogue-item/enter-page items])
                 (navigate! "/administration/catalogue-items/update-catalogue-item"))


vuorioma:
The feature works as intended on my end and the browser tests look ok to me.
However I cannot run the browser tests locally right now, so i'm trusting CircleCI to check  they pass.

Out of scope for this pull request:  I discovered that the user can still manually go to the edit page (e.g. http://localhost:3000/administration/workflows/edit/7) and click the 'Save' button which takes the user to a 'Forbidden' page. We probably want to reject access to the edit page instead?

Macroz:
> Out of scope for this pull request: I discovered that the user can still manually go to the edit page (e.g. http://localhost:3000/administration/workflows/edit/7) and click the 'Save' button which takes the user to a 'Forbidden' page. We probably want to reject access to the edit page instead?

Since the users don't typically type URLs when they want to edit something, I find that this probably doesn't add enough value to warrant programming it. All sort of hacking is already handled by the API backend protections. The features in the UI are just to guide the user to the usually correct place and make the job easy to do. By hiding a command that makes no sense for that user is typically enough.

src/cljs/rems/administration/catalogue_items.cljs
```diff
@@ -87,7 +87,8 @@
   [atoms/rate-limited-action-button
    {:id :update-catalogue-item
     :class "btn-primary"
-    :disabled (when (empty? items) :disabled)
+    :disabled (when (or (empty? items)
+                        (some (complement roles/can-modify-organization-item?) items)) :disabled)
```
Macroz: Could be easier to read with `not-every?`?

Issue #3100, opened 2023-01-10
Fix invite applicant link for owner/handler


jaakkocsc:
- Accept-invitation for role logged in users (currently everyone-else, see https://github.com/CSCfi/rems/blob/master/docs/application-permissions.md)
- See master_workflow.clj

PR #3410, opened 2025-12-15
Allow handler to accept invitation to application

pikariop:
Fix: #3100

Allow a handler to join an application as a member, and be added as a member. If desired, the owner can toggle back the old behaviour from a given workflow's settings ("Disable actions").

# Checklist for author

## Reviewability
- [x] Link to issue

## Documentation

## Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ Changes since v2.38.1
   - Attachment upload error is indicated by small icon next to the upload button (detailed error is still shown in top of the page).
   - Pending attachment upload is indicated by spinner next to the upload button.
 - Edit buttons on the administration detail pages are hidden when the user is not permitted to edit, like they are hidden on the list pages already (#2814)
-- Handler can now join an application as a member, and be added as a member. (#3100) 
+- Handler can now join an application as a member, and be added as a member. This is to facilitate demos in certain situations rather than a suggested way of using REMS in production. The earlier behaviour where handlers are not permitted to accept invitations can be switched back on under Workflow settings. (#3100)
 
 ### Fixes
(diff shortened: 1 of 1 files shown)


test/clj/rems/application/test_commands.clj
```diff
@@ -1295,17 +1295,18 @@
...
+                :event/actor user
+                :application/id app-id
+                :invitation/token "very-secure"}
+               (ok-command {:type :application.command/accept-invitation
+                            :actor user
+                            :token "very-secure"}
+                           (build-application-view [dummy-created-event
+                                                    dummy-member-invited-event]))))))
 
     (testing "can't join if they are already a member"
```
Macroz: Could be nice to check what happens for a reviewer or decider to specify their behavior explicitly. They are already members of the application so should get the error.

pikariop: Good suggestion. It seems like these users are seen as `everyone-else`, and are able to join when the application is in draft state. I'll add them to the user vector to be explicit and slurp in the other test under a similar `doseq` as well. 

```clojure
(rems.permissions/user-roles (build-application-view [dummy-created-event
                                                      dummy-member-invited-event])
                             reviewer-user-id)
;; => #{:everyone-else}

(rems.permissions/user-roles (build-application-view [dummy-created-event
                                                      dummy-member-invited-event])
                             decider-user-id)
;; => #{:everyone-else}
```

The way `r.a.commands/already-joined-error` works with these changes is that it checks if the user already has any of the roles associated with the invitation, so someone like the decider could technically be invited to join as a member first, then as the application is being reviewed, get another invitation to join in their actual role (decider) and they're only given an error when they attempt to do that twice for a given role. I'll add a test for that in another pr.

Issue #2622, opened 2021-04-07
Collaborating on an application


opqdonut:
An end-user at Findata reported that they'd like to be able to collaborate on an application with other people. At the moment Findata doesn't even allow inviting members, but even if they did, the members don't have write access to the application. 

Also, currently it's pretty much impossible to export the form template as an applicant (without using API). If we had something like that, the applicant could import the questions into e.g. Google Docs and then collaborate there, and finally copy-paste all the answers back to REMS.

Macroz:
Full support for multi-user editing would likely require a lot of work. Even partial support requires some new capabilities like sharing application editing state between multiple users. This is not impossible but definitely a major feature.

Macroz:
I will be working on a proof-of-concept for multi-user editing, so that applicant and members could edit the application together, even at the same time. This will not completely solve this case, if Findata does not allow members in applications.

Macroz:
Proof of concept done and seems to be easy to do. There could be a separate issue that creates the "back channel" from server to browser first. Waiting for time to continue.

PR #3430, opened 2026-02-24
Multiplayer

Macroz:
Mostly working draft for #2622 

Some open issues:
- File upload is broken, needs investigation.
- There should be a config whether the applicant only can send the application or can members also.
- Config should exist whether members are allowed to edit or just view.
- Test in browser for multiple people editing and seeing the changes.
- Test API security
- Test the long poll functionality to make sure it keeps working.
- Use the same keepalive tactic we have in REMS arleady with the subscription to stop it when the user is not actually doing anything (optimization).
- Long polling will cause the REMS browser app to reconnect and re-authorize automatically but keepalive would fix this?
- Same user has many tabs open. Should it have one or many names displayed?
- Check the UI terminology i.e., Users vs. Applicants vs. ?
- If reloading the application, it takes one long poll cycle to get the other clients. We could load these immediately.
- There is some internal naming that could perhaps be improved with `form-id` vs. `form/id`, and `field-id` vs. `field/id`.
- Manual mention.
- Overall, implementation could also include a refactoring of keepalive together with this subscription mechanism so that there exists only one "backchannel" to communicate with REMS:
- The design uses long-polling because it is fool-proof and simple. A WebSocket API or SSE is possible too, but maybe not worthwhile at the moment. At least the risks with this API style are minimal.

Screenshot_2026-02-24_18-56-16

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up


vuorioma:
Closing this to clean up the PR list, we may reopen it in the future.

PR #3455, opened 2026-04-23
Reapply "Merge pull request #3429 from CSCfi/external-frontend-url"

### LINKED ISSUE #3402: In emails exchanges - Allow redirection of Applicants from REMS to GDI User Portal
**Is your feature request related to a problem? Please describe.**
When integrating GDI User Portal with REMS, we decided to make REMS only visible in the back-office, by the DAC members, so we enable infrastructure teams to apply extra security measures - e.g. REMS behind firewall.

That means then that any kind of interaction between Applicant and the system is via User Portal only. However, the emails sent by REMS are pointing out to REMS directly, breaking the first principle.

**Describe the solution you'd like**
Allow operators to configure REMS email template, in an way that is flexible enough to redirect applicants to the User Portal, or the front-office.

**Describe alternatives you've considered**
There are two options:
- Implement template engines for emails - such as Thymeleaf, FreeMarker. And document what are the variables available.
- Make the URL string patern configurable - e.g. `FRONT_OFFICE_URL_PATTERN="https://portal.dev.gdi.lu/applications/%s"`

**Additional context**
The first option is more robust and enables other features, like content formatting.

pikariop:
This reverts commit 8cc54b7655864c74c2c81e89d54f7fbbc3a88d47.

Fix #3402 

# Checklist for author

## Reviewability
- [X] Link to issue

## Backwards compatibility

## Documentation

## Testing

## Follow-up

### DIFF
--- a/src/clj/rems/email/template.clj
+++ b/src/clj/rems/email/template.clj
@@ -14,12 +14,14 @@
 ;; TODO list of resources?
 ;; TODO use real name when addressing user?
 
+(defn base-url [] (or (not-empty (:external-frontend-url env)) (:public-url env)))
+
 ;; move this to a util namespace if its needed somewhere else
 (defn- link-to-application [application-id]
-  (str (:public-url env) "application/" application-id))
+  (str (base-url) "application/" application-id))
 
 (defn- invitation-link [token]
-  (str (:public-url env) "accept-invitation?token=" token))
+  (str (base-url) "accept-invitation?token=" token))
 
 (defn- format-application-for-email [application]
   (str
@@ -301,7 +303,7 @@
                                      (text-format-map :t.email.handler-reminder/application
                                                       params
                                                       [:application-id :applicant]))
-            params {:actions-url (str (:public-url env) "actions")
+            params {:actions-url (str (base-url) "actions")
                     :applications (str/join "\n" formatted-applications)
                     :recipient (application-util/get-member-name handler)}]
         {:to-user (:userid handler)
@@ -319,7 +321,7 @@
                                      (text-format-map :t.email.reviewer-reminder/application
                                                       params
                                                       [:application-id :applicant]))
-            params {:actions-url (str (:public-url env) "actions")
+            params {:actions-url (str (base-url) "actions")
                     :applications (str/join "\n" formatted-applications)
                     :recipient (application-util/get-member-name reviewer)}]
         {:to-user (:userid reviewer)
--- a/src/clj/rems/email/template.clj
+++ b/src/clj/rems/email/template.clj
@@ -23,7 +23,7 @@
 (defn- link-to-application [application-id use-external-frontend-url]
   (str (base-url use-external-frontend-url) "application/" application-id))
 
-(defn- invitation-link [token use-external-frontend-url] 
+(defn- invitation-link [token use-external-frontend-url]
   (str (base-url use-external-frontend-url) "accept-invitation?token=" token))
 
 (defn- format-application-for-email [application]
@@ -73,10 +73,10 @@
                        (when (and body-text
                                   (not (str/blank? (text-no-fallback body-text))))
                          (let [event (apply-event-privacy event application (:userid recipient))
-                               application-id (:application/id application) 
+                               application-id (:application/id application)
                                params {:applicant (application-util/get-applicant-name application)
                                        :application-id (format-application-for-email application)
-                                       :application-url (link-to-application 
+                                       :application-url (link-to-application
                                                          application-id
                                                          (application-util/is-applying-user? application (:userid recipient)))
                                        :catalogue-items (resources-for-email application)
@@ -218,7 +218,7 @@
     (let [invited-user (:application/member event)
           params {:applicant (application-util/get-applicant-name application)
                   :application-id (format-application-for-email application)
-                  :invitation-url (invitation-link (:invitation/token event) 
+                  :invitation-url (invitation-link (:invitation/token event)
                                                    (application-util/is-applying-user? application (:userid invited-user)))
                   :recipient (application-util/get-member-name invited-user)}]
       [{:to (:email invited-user)
@@ -236,7 +236,7 @@
   (let [invited-user (:application/reviewer event)
         params {:applicant (application-util/get-applicant-name application)
                 :application-id (format-application-for-email application)
-                :invitation-url (invitation-link (:invitation/token event) 
+                :invitation-url (invitation-link (:invitation/token event)
                                                  (application-util/is-applying-user? application (:userid invited-user)))
                 :recipient (application-util/get-member-name invited-user)}]
     (with-language (:default-language env)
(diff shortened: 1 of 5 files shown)


src/clj/rems/email/template.clj
```diff
@@ -61,16 +66,19 @@
 (defmethod event-to-emails :default [_event _application]
   [])
 
-(defn- emails-to-recipients [recipients event application subject-text body-text]
+(defn- recipients-to-emails [recipients event application subject-text body-text]
```
pikariop: Minor suggestion. This does seem like a better name if we look at the signature and the return value, `[input]-to-[output]`, and it's a bit less ambiguous at the call site (`event-to-emails`) but I think the original intention of "emails addressed to these recipients" could perhaps be more clear to the future reader, so we could consider adding a verb (which is always nice to have in a function, as opposed to a value where the noun form makes more sense)

```suggestion
(defn- create-emails-to-recipients [recipients event application subject-text body-text]
```

which I think also puts emphasis on desired outcome ("an event of this type means we'll create a message to these groups of people") rather than the activity we're performing to achieve that ("iterate over `recipients` to produce emails addressed to them")

src/clj/rems/email/template.clj
```diff
@@ -14,12 +14,17 @@
 ;; TODO list of resources?
 ;; TODO use real name when addressing user?
 
+(defn base-url [use-external-frontend-url]
+  (if use-external-frontend-url
+    (or (not-empty (:external-frontend-url env)) (:public-url env))
+    (:public-url env)))
```
pikariop: I think it could be a bit tidier if we'd add an arity to the function so that we don't need to change the call sites where we only want the existing functionality

```suggestion
(defn base-url 
  ([] (base-url false))
  ([use-external-url]
   (or (when use-external-url
         (:external-frontend-url env))
       (:public-url env))))
```
For style points, the `not-empty` can be left out because we aren't dealing with collections (a string is a sequence of chars but if the value is empty we're just doing a `(not-empty nil)` when that nil is sufficient already), and to leverage nil-punning further, the `if` can become a `when` inside the `or`

test/clj/rems/email/test_template.clj
```diff
@@ -167,6 +167,20 @@
             :body "Dear Amber Assistant,\n\nAlice Applicant has submitted a new application 2001/3, \"Application title\" to access resource(s) en title 11, en title 21.\n\nYou can review the application at http://example.com/application/7"}
            (email-to "assistant" mails)))))
 
+(deftest test-external-frontend-url
+  (with-redefs [rems.config/env (assoc rems.config/env :external-frontend-url "http://external-frontend-url.com/")]
```
pikariop, line 188: Two test cases I'd like to see here: 

with the external frontend url configured, when calling the other functions which use `base-url`, does the resulting url remain unchanged from `:public-url`
 
without the external frontend url configured, when calling all these functions, does the resulting url remain unchanged from `:public-url`

These would in my view document the full "contract" of the new functionality and alert us if we accidentally change the implementation to not fulfill the contract

src/clj/rems/email/template.clj
vuorioma: What if the value is an empty string?
`(or "" "hello")`  evaluates to empty string, while
`(or (not-empty "") "hello")` evaluates to "hello".

vuorioma: I think this suggestion is definitely more idiomatic, but the original is more readable, at least for me.

test/clj/rems/email/test_template.clj
vuorioma: Could you specify which functions should have a test? 

I could add similar tests for each possible type of event/role, but that seems a bit too much.

src/clj/rems/email/template.clj
vuorioma: Agreed. Could be more explicit.

src/clj/rems/email/template.clj
pikariop: I think we're already doing some kind of validation [here in `rems.config`](https://github.com/CSCfi/rems/blob/external-frontend-url-pt-2/src/clj/rems/config.clj#L88C1-L94C77)

Generally, `clojure.string/blank?` would be a better tool if we're worried about empty strings since it also checks for any whitespace, even the non-printing kind

Actually, I think the REMS-preferred way to look up configuration values would be `getx`. Perhaps this could be a straightforward way to pick the desired key and more readable than my previous suggestion
```clojure
(defn base-url 
  ([] (base-url false))
  ([use-external-url]
   (getx env
         (if use-external-url 
           :external-frontend-url 
           :public-url))))
```

test/clj/rems/email/test_template.clj
pikariop: Sorry, I mean methods of `event-to-emails`. Perhaps we could leverage the original implementation (which seems to cover these events and outcomes already) and for each test, add another `testing` clause where in the `with-redefs` we check only the email body and if it contains the external frontend url or not.

So `test-submitted` for example would become
```clojure
(deftest test-submitted
  (let [mails (emails created-events submit-event)]
    (is (= #{"applicant" "assistant" "handler"} (email-recipients mails)))
    (is (= {:to-user "applicant"
            :subject "Your application 2001/3, \"Application title\" has been submitted"
            :body "Dear Alice Applicant,\n\nYour application 2001/3, \"Application title\" has been submitted. You will be notified by email when the application has been handled.\n\nYou can view the application at http://example.com/application/7"}
           (email-to "applicant" mails)))
    (is (= {:to-user "assistant"
            :subject "(2001/3, \"Application title\") A new application has been submitted"
            :body "Dear Amber Assistant,\n\nAlice Applicant has submitted a new application 2001/3, \"Application title\" to access resource(s) en title 11, en title 21.\n\nYou can review the application at http://example.com/application/7"}
           (email-to "assistant" mails))))
  (testing "with external frontend url"
    (with-redefs [rems.config/env (assoc rems.config/env :external-frontend-url "http://external-frontend-url.com/")]
      (let [mails (emails created-events submit-event)]
        (is (= "Dear Alice Applicant,\n\nYour application 2001/3, \"Application title\" has been submitted. You will be notified by email when the application has been handled.\n\nYou can view the application at http://external-frontend-url.com/application/7"
               (:body (email-to "applicant" mails))))
        (is (= "Dear Amber Assistant,\n\nAlice Applicant has submitted a new application 2001/3, \"Application title\" to access resource(s) en title 11, en title 21.\n\nYou can review the application at http://example.com/application/7"
               (:body (email-to "assistant" mails))))))))
```
And for any tests where the event would send an email only to internal users, the additional test case could simply compare the entire return value of `emails`
```clojure
(deftest test-some-event
  (let [mails (emails base-events {#_...})]
    (testing "the original test case"
      (is (= {#_...}
             (email-to "handler" mails))))
    (testing "with external url"
      (with-redefs [rems.config/env (assoc rems.config/env :external-frontend-url "http://external-frontend-url.com/")]
        (is (= mails
               (emails created-events submit-event)))))))
```

The rest of the keys in the email are already covered by the original test cases so we can leave them out

The email feature as a whole is somewhat crucial to get right, so I wouldn't say having test cases for all events is too much. Even the obvious ones guard against regression later on and serve as documentation for maintainers and users.

pikariop:
Ok

Issue #3460, opened 2026-05-29
Handling errors when there is no application ID


meericsc:
If REMS encounters an error before the application has an ID, it will infinitely(?) try to load the page. For example in situation where user tries to apply access to two items that have different workflow via a link: https://rems-dev.2.rahtiapp.fi/application?items=16,19

We would need a logic to handle errors before an application is created. 
Redirect users to catalogue and show the error there?

Issue #3392, opened 2025-10-31
New catalogue items show with delay


meericsc:
When user creates a new catalogue item, it doesn't immediately show on the top of catalogue items list. After a page refresh or navigating to some other page and then back to catalogue items main page, the new catalogue item appears.

I couldn't reproduce this in our test instance, but I can reproduce in our own REMS instance.  Both are running the newest REMS version

PR #3461, opened 2026-06-10
Catalogue item hierarchy UI

### LINKED ISSUE #3411: Validate application before sending
Bigpicture would want to validate that there are specific datasets on the application before it is sent. 

We could do this with validate plugin and add that to the application. We would validate the application on send application. If the datasets are wrong, user would see an error saying why the sending of application failed

- Send application to plugin, plugin answers is application is OK
- If not ok, show error
- In application data model there is no catalogue item hierarchy as it does not exist yet (separate issue for creating this) 
- Plugin should get 1. application 2. catalogue item hierarchy either directly or it can also make an API call to get the info
- Tests + documentation

Consider if we need to add the same plugin also to other places, e.g. shopping cart (validate on apply) and change resources block (validate on save) on the application

### LINKED ISSUE #3412: Catalogue item hierarchy support (sub-dataset)
Currently, REMS does not support referencing one catalogue item to another. Bigpicture project would need to support the concept of datasets and additional "complementary" datasets. 

Add ability to mark a catalogue item as "part of" another catalogue item 
- Add "part of" to catalogue item in UI (json blob, postgres catalogue item data) 
- API support (optional field)
- In catalogue item editor add field where you can add and edit this info (e.g., Dataset A is part of Dataset B)
- Tests + documentation

- Related is the logic that allows only applying access to subdatasets if also the main dataset is added to application:  #3411

pikariop:
Close: https://github.com/CSCfi/rems/issues/3411
Fix: https://github.com/CSCfi/rems/issues/3412
Based on https://github.com/CSCfi/rems/pull/3424

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation
## Testing

## Follow-up
- [x] New tasks are created for pending or remaining tasks: https://github.com/CSCfi/rems/issues/3460

---

hier-1
(sending the application and approving it as a handler is redacted for brevity as they're unaffected)
hier-2
and finally,
hier-3

### DIFF
--- a/src/clj/rems/application/commands.clj
+++ b/src/clj/rems/application/commands.clj
@@ -7,6 +7,7 @@
             [rems.common.application-util :as application-util]
             [rems.common.form :as form]
             [rems.common.util :refer [build-index getx getx-in]]
+            [rems.db.entitlements]
             [rems.form-validation :as form-validation]
             [rems.permissions :as permissions]
             [rems.util :refer [assert-ex try-catch-ex]]
@@ -102,6 +103,22 @@
                  {:type :disabled-catalogue-item :catalogue-item-id (getx item :id)})]
     (when (seq errors)
       {:errors (vec errors)})))
+
+(defn- invalid-catalogue-item-hierarchy-error
+  "For such an item in `catalogue-item-ids` that has a parent (dependent), check that the parent item is also included in `catalogue-item-ids`, or `actor` has an entitlement to the parent item's resource."
+  [catalogue-item-ids actor {:keys [get-catalogue-item get-dependents get-entitlements]}]
+  (let [entitled-to-resids (into #{} (map :resourceid) (get-entitlements actor))
+        missing (into []
+                      (comp (mapcat (fn [id] (get-dependents {:catalogue-item/id id})))
+                            (map :catalogue-item/id)
+                            (remove (set catalogue-item-ids))
+                            (map get-catalogue-item)
+                            (filter (complement (comp entitled-to-resids :resource-id)))
+                            (map (fn [{:keys [id]}] {:catalogue-item/id id})))
+                      catalogue-item-ids)]
+    (when (seq missing)
+      {:errors [{:type :missing-top-level-item
+                 :top-level-item-ids missing}]})))
 
 (defn- licenses-not-accepted-error [application userid]
   (when-not (application-util/accepted-licenses? application userid)
@@ -240,6 +257,7 @@
   (or (must-not-be-empty cmd :catalogue-item-ids)
       (invalid-catalogue-items catalogue-item-ids injections)
       (unbundlable-catalogue-items catalogue-item-ids injections)
+      (invalid-catalogue-item-hierarchy-error catalogue-item-ids actor injections)
       (disabled-catalogue-items-error catalogue-item-ids injections)
       (let [workflow-id (-> (first catalogue-item-ids)
                             get-catalogue-item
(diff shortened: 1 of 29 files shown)


src/clj/rems/application/schema.clj
vuorioma, line 1: Was the intention here to separate schemas and functions from command related code?

I think its good practice. Looking at the namespace `ns rems.application.schema` i would expect for example, the application schema to be located in that namespace. 

Maybe we could create a new folder `src/clj/commands` so the namespace would become `rems.commands.schema`

vuorioma:
Do we want to enforce column length in new code? The clojure style guide advocates max 80.

src/clj/rems/application/commands.clj
```diff
@@ -291,6 +104,22 @@
     (when (seq errors)
       {:errors (vec errors)})))
 
+(defn- invalid-catalogue-item-hierarchy-error
+  "For such an item in `catalogue-item-ids` that has a parent (dependent), check that the parent item is also included in `catalogue-item-ids`, or `actor` has an entitlement to the parent item's resource."
+  [catalogue-item-ids actor {:keys [get-catalogue-item get-dependents get-entitlements]}]
```
vuorioma, line 108: Just rambling here, I dont understand the advantage of the injection pattern. I had to go through a chain of dozen references just to find the injected functions.

I believe it was implemented to enable mock testing, but what's wrong with, for example  `(with-redefs [db/get-catalogue-item mock_fn] ...) `to achieve that?

src/clj/rems/application/schema.clj
pikariop: Yes, it turned out that the schema definitions have a lot of references throughout the codebase, so otherwise I couldn't have added some of the new command injections without causing a circular dependency. I remember running into this earlier when trying to find out something, can't remember exactly what, but it traced into `rems.config/validate-config`, and because of that dependency we couldn't do whatever that thing was.

`rems.command.schema` would sound nice and I agree `rems.application.schema` isn't the best name for what actually just has the application command schema, when we also seem to have schemas in `rems.application.events` which I admit not having considered, but we also some command schemas in `rems.api.schema` and certain other API command schemas coupled with their respective endpoints, which makes sense (as the original implementation did, until the circular dependency thing) but it's another aspect to consider if we'd like to restructure things. Maybe we should file an issue for it?

I think we could rename this one `rems.application.command-schema` if that makes more sense?

src/clj/rems/application/commands.clj
pikariop: I suppose it also resolves the circular dependency problem to some extent? Having pure functions instead of mutating state directly is nice for sure, at least when developing them. I think one way to make it slightly more navigable in the IDE could be [protocols](https://blog.cofx.nl/dependency-injection-and-protocols-in-clojure.html), generally speaking.

Speaking of injections and having no direct requires to implementations, it seems I've added one here, `rems.db.entitlements`, possibly because clojure-lsp is sometimes a bit too helpful. It doesn't get used anywhere so I'll send in a commit to remove it shortly...

vuorioma:
I have done some UI testing. All good except for one case that may need review:
- Logged in as an owner and created a new catalogue item - default workflow,  no complementary items
- Attempted to create another catalogue item. Same as before but adding the previous item as complementary.  Could not find the previous item in the dropdown listing complementary items

pikariop:
> I have done some UI testing. All good except for one case that may need review:
>     * Logged in as an owner and created a new catalogue item - default workflow,  no complementary items
>     * Attempted to create another catalogue item. Same as before but adding the previous item as complementary.  Could not find the previous item in the dropdown listing complementary items

Couldn't replicate this. Possibly an instance of https://github.com/CSCfi/rems/issues/3392 ?

vuorioma (approved):

PR #3472, opened 2026-07-16
Add error handling in UI for when creating an application fails

### LINKED ISSUE #3460: Handling errors when there is no application ID

pikariop:
Fix #3460 

# Checklist for author

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/cljs/rems/new_application.cljs
+++ b/src/cljs/rems/new_application.cljs
@@ -12,14 +12,29 @@
 
 (rf/reg-event-fx
  ::enter-new-application-page
- (fn [{:keys [db]} [_ catalogue-item-ids]]
+ (fn [_ [_ catalogue-item-ids]]
    (post! "/api/applications/create"
           {:params {:catalogue-item-ids catalogue-item-ids}
-           :handler (fn [response]
-                      (remove-catalogue-items-from-cart! catalogue-item-ids)
-                      (replace-url! (str "/application/" (:application-id response))))
-           :error-handler (flash-message/default-error-handler
-                           :top [text :t.applications/application])})
+           :handler (fn [{:keys [application-id
+                                 success
+                                 errors]
+                          :as _response}]
+                      (cond
+                        success
+                        (do
+                          (remove-catalogue-items-from-cart! catalogue-item-ids)
+                          (replace-url! (str "/application/" application-id)))
+
+                        errors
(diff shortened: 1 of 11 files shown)


src/cljs/rems/new_application.cljs
```diff
@@ -12,14 +12,29 @@
...
+                          (remove-catalogue-items-from-cart! catalogue-item-ids)
+                          (replace-url! (str "/application/" application-id)))
+
+                        errors
+                        (do
+                          (replace-url! "/catalogue")
+                          (js/console.log errors)
+                          (flash-message/show-error! :top (->> errors
+                                                               (mapv (flash-message/argumentize-some-key :catalogue-item-id :catalogue-item-ids))
+                                                               flash-message/format-errors)))))
```
pikariop, line 34: Not exactly the way I'd like to fabricate an error handler but since it seems like `flash-message/format-errors` has previously been raised as a potential target for refactorings, and given that our API errors don't have an exact schema, this is what I figured to be a min-maxed improvement on things while staying within issue scope and leveraging existing functionality. 

Going forward, I'd suggest we design a more fine-grained schema than `s/Any` for the API errors, so consumers would have an easier time designing around the possible responses. I feel like the error should at minimum look like 
```clojure
{:type :t.x.errors/something-went-wrong ; as it currently is
 :domain-resource :catalogue-item       ; fixed key, qualified name of the object as the value 
 :ids [1]}                              ; fixed key, vector of values, id(s) of :domain-resource
```
So we could write a generic error formatter without having to guess at callsites what errors could be returned and what keys they might have

Additionally, we could look at the proposed [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html) for inspiration

vuorioma:
This solves #3460, approved.

The error message could be more user friendly going forward, for example "Something went wrong while creating your application. Please try again later", while hiding the technical detail under a dropdown if needed

vuorioma (approved):

Issue #3481, opened 2026-08-24
Limited support for newlines in labels


andrewpatto:
**Is your feature request related to a problem? Please describe.**
We have some labels in forms that would benefit from some blank lines

**Describe the solution you'd like**
It would be good if there was some simple way to put \n or \<br\> or some linefeed markers into
form text.

**Describe alternatives you've considered**
We could have a label for each and every line, though that feels pretty heavyweight.

**Additional context**

<img width="1123" height="437" alt="Image" src="https://github.com/user-attachments/assets/dac43600-ccb5-4304-8e49-2c1527b944f1" />

vs

<img width="1122" height="699" alt="Image" src="https://github.com/user-attachments/assets/16e2ebe5-390a-42a1-9726-3221a77e1e12" />

andrewpatto:
#2998 is the more complete solution to this - but possibly there is some simple parsing / regex magic that could be put in linkify() as an interim.

andrewpatto:
I can submit a PR for adding this to linkify if this is a path you'd consider (is only about 10 lines)

meericsc:
Hi, thanks for making an issue! This is something we've discussed before, but haven't had time to work on it.

We would definitely welcome a PR. Line break support would be a nice improvement for readability and for managing the label texts on the admin side. Thank you!

PR #3486, opened 2026-09-03
feat: Linebreak support in form labels

andrewpatto:
Converts \n characters found in form labels into actual HTML linebreaks - rather than rendering them as whitespace which is the current behaviour.

## Reviewability
- [x] Link to issue

## Backwards compatibility

## Documentation

## Testing

### DIFF
--- a/src/cljs/rems/util.cljs
+++ b/src/cljs/rems/util.cljs
@@ -131,23 +131,35 @@
 
 (def ^:private link-regex #"(?:http://|https://|www\.\w).*?(?=[^a-zA-Z0-9_/]*(?: |$))")
 
-(defn linkify
+(defn- linkify-links
   "Given a string, return a vector that, when concatenated, forms the
   original string, except that all substrings that resemble a link have
   been changed to hiccup links."
   [s]
+  (let [splitted (-> s
+                     (str/replace link-regex #(str "\t" %1 "\t"))
+                     (str/split "\t"))
+        link? (fn [s] (re-matches link-regex s))
+        text-to-url (fn [s] (if (re-matches #"^(http://|https://).*" s)
+                              s
+                              (str "http://" s)))]
+    (map #(if (link? %)
+            ^{:key %} [:a {:target :_blank :href (text-to-url %)} %]
+            %)
+         splitted)))
+
+(defn linkify
+  "Given a string, return a vector that, when concatenated, forms the
+  original string, except that all substrings that resemble a link have
+  been changed to hiccup links, and all newlines have been changed to
+  hiccup line breaks."
+  [s]
   (when s
-    (let [splitted (-> s
-                       (str/replace link-regex #(str "\t" %1 "\t"))
-                       (str/split "\t"))
-          link? (fn [s] (re-matches link-regex s))
-          text-to-url (fn [s] (if (re-matches #"^(http://|https://).*" s)
-                                s
-                                (str "http://" s)))]
-      (map #(if (link? %)
-              ^{:key %} [:a {:target :_blank :href (text-to-url %)} %]
-              %)
-           splitted))))
+    (let [lines (str/split s #"\n" -1)]
+      (->> lines
+           (map linkify-links)
+           (interpose (list ^{:key (gensym "br")} [:br]))
+           (apply concat)))))
 
 (defn visibility-ratio
   "Given a DOM node, return a number from 0.0 to 1.0 describing how much of an element is inside the viewport."
--- a/test/cljs/rems/test_util.cljs
+++ b/test/cljs/rems/test_util.cljs
@@ -50,4 +50,19 @@
              (linkify "(See http://www.abc.com?)"))))
     (testing "a link without http-prefix"
       (is (= ["(See www-page at " [:a {:target :_blank :href "http://www.abc.com"} "www.abc.com"] ".)"]
-             (linkify "(See www-page at www.abc.com.)"))))))
+             (linkify "(See www-page at www.abc.com.)"))))
+    (testing "change newlines to hiccup line breaks"
+      (is (= ["a" [:br] "b"]
+             (linkify "a\nb")))
+      (is (= ["a" [:br] "b" [:br] "c"]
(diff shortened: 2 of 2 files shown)


andrewpatto:
Closes #3481

andrewpatto:
This was the label text that I sent in via API

```
  about_text = <<EOT
A cancer of unknown primary (CUP) presents diagnostic uncertainty
and limited treatment options.

SUPER-NEXT (2021-2023) was an MRFF-funded national
clinical genomics study employing clinical whole genome and transcriptome
sequencing (WGTS) to help resolve diagnostic uncertainty and inform precision
treatments for CUP patients. 

In SUPER-NEXT, we implemented and compared WGTS, panel and cell-free DNA tests
reporting results through a molecular tumour board to evaluate diagnostic and
clinical impact these tests. A total of 299 patients were recruited to the study
from 12 Australian centres, with 251 patients receiving clinical reports. The
patient demographic of the CUP cohort was typical for this disease, with 55%
being male and a median age of 65 years.

Rebello, R.J., Posner, A., Dong, R. et al. 
Whole genome sequencing improves tissue-of-origin diagnosis and treatment options for cancer of unknown primary.
Nat Commun 16, 4422 (2025).

https://doi.org/10.1038/s41467-025-59661-x
EOT
```

So the representation is _actual_ newlines. I have not considered how newlines would be entered in the form builder UI.

Result

Screenshot 2026-09-03 at 11 55 46 am

andrewpatto:
In terms of backwards compatibility - I guess there could be existing labels that _have_ \n characters in labels that are currently just being displayed as spaces. Though I'm not quite sure how they would get into the data accidentally - but people who have deliberately put in \n into labels may be surprised when they start spontaneously working

andrewpatto:
Looking now at editing in the form UI - the "preview" seems to be accumulating extra \n for display as I am editing the label on the left. It is easy to enter newlines in the text area - so it is entirely possible existing users may have entered newlines in labels _not_ wanting them to actual be a line break. 

So we may have to rethink what we use as the triggers in linkify(). 

We could use the escape sequence "\n" (as opposed to the real line break single character). Or we could have a pseudo HTML <br/> or <br>? Thoughts? Or we could leave it as the real linebreak single character and make this more of a breaking change

pikariop:
Hello and thank you for your contribution!

I gave it a spin and it seems to be working alright in the applicant view. There's the preview thing you mentioned -- it seems like it does indeed accumulate extra line breaks when rendering the preview, but the actual data returned by the function is fine, and it seems to be recorded to the database and retrieved as you'd expect.

I had a look into what might be causing that and I'll add my findings in the review comments shortly.

src/cljs/rems/util.cljs
```diff
@@ -131,23 +131,35 @@
...
-                                (str "http://" s)))]
-      (map #(if (link? %)
-              ^{:key %} [:a {:target :_blank :href (text-to-url %)} %]
-              %)
-           splitted))))
+    (let [lines (str/split s #"\n" -1)]
+      (->> lines
+           (map linkify-links)
+           (interpose (list ^{:key (gensym "br")} [:br]))
+           (apply concat)))))
```
pikariop, line 162: Looking at the browser console, there's an error:

```
Warning: Encountered two children with the same key, `br658`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

That seems to happen because the `interpose` here evaluates the gensym expression only once and goes on to repeat it's result, while we would like to evaluate it each time for a new and unique symbol.

I would suggest we'd write that part as a function that makes us a `[:br]` with the appropriate metadata, and making a sequence of calling that function repeatedly, so we can then interleave it with the split lines. 

```suggestion
(let [lines (str/split s #"\n" -1)
      make-br #(list ^{:key (gensym "br")} [:br])]
      (->> lines
           (map linkify-links)
           (interleave (repeatedly make-br))
           (drop 1)
           (apply concat)))))
```

Another way to express it could be transducers which I think would fit this problem semantically quite well, as it seems like we've got a separate transformation step (`linkify-links`) and then it's just putting the lines back together. There's not an awful lot of transformation steps though, so whether it clarifies anything is a matter of taste, and I'm happy with whichever option you find more readable.
```suggestion
(let [make-br (fn [] ^{:key (gensym "br")} [:br])]
    (transduce
     (map linkify-links)
     (completing (fn [acc line]
                   (concat acc
                           (when-not (empty? acc)
                             [(make-br)])
                           line)))
     (str/split s #"\n" -1)))))
```
Apparently the React warning was right in that the elements might indeed get duplicated!

src/cljs/rems/util.cljs
```diff
@@ -131,23 +131,35 @@
 
 (def ^:private link-regex #"(?:http://|https://|www\.\w).*?(?=[^a-zA-Z0-9_/]*(?: |$))")
 
-(defn linkify
+(defn- linkify-links
```
pikariop, line 134: I think it would perhaps leave the commit history a bit tidier if we'd keep the original function as is and rename the new one. It would then look a bit more deliberate that existing functionality was unchanged, and the changes at the call sites make it apparent that the function now does something differently, but also that it didn't change the implied contract that there's going to be a vector of hiccup returned. 

I'm not sure what to call it, maybe just `label->hiccup` as we're making a couple of different transformations (who knows, maybe this will pave the way for more) and someone browsing the call sites would maybe be interested that the text content is going to be hiccup from there on.

pikariop (changes requested):

pikariop:
Re: whether to break functionality or introduce a formatting syntax, I think this is good as it is, it reflects the user experience and capability of the html text field in the form editor, and since the content is going to be mostly form field titles and help texts, it's perhaps a bit limited in how creative you can get with these things to begin with, at least in the general context of how websites implement these things and what the users might expect to see.

For backwards compatibility, I think we could write an administration script that looks for any newlines in form labels in the database and rewrites them as a single space. I can take care of that later once the pull request is merged.

